PR #1, opened 2019-04-10
RFC Process

hone:
* [rendered](https://github.com/hone/cnb-rfcs/blob/master/text/0001-rfc-process.md)

# Summary

The "RFC" (request for comments) process provide a consistent and controlled path for new features to enter the spec, libraries, or tools, so that all stakeholders can be confident about the direction Cloud Native Buildpacks is evolving in.

### DIFF
--- a/text/0000-rfc-process.md
+++ b/text/0000-rfc-process.md
@@ -54,17 +54,17 @@ To get an RFC into Cloud Native Buildpacks, first the RFC needs to be merged int
 
 - Fork the RFC repo: <https://github.com/buildpack/rfcs>
 - Copy `0000-template.md` to `text/0000-my-feature.md` (where 'my-feature' is descriptive. don't assign an RFC number yet).
-- Fill in RFC
+- Fill in RFC. Any section can be marked as "N/A" if not applicable.
 - Submit a pull request. The pull request is the time to get review of the proposal from the larger community.
 - Build consensus and integrate feedback. RFCs that have broad support are much more likely to make progress than those that don't receive any comments.
 
 Once a pull request is opened, the RFC is now in development and the following will happen:
 
-- It will be triaged and assigned to a team member in a future working group meeting.
-- That team member will be responsible for championing that RFC and ensuring it gets comments, feedback, or closed.
+- It will be triaged, labeled, and discussed in a future working group meeting. Working group meetings happen on a weekly cadence barring exceptions. Labeling marks a RFC as general or specific to a subteam.
 - The team will discuss as much as possible in the RFC pull request directly. Any outside discussion will be summarized in the comment thread.
 - When deemed "ready", a team member will propose a "motion for final comment period (FCP)" along with a disposition of the outcome (merge, close, or postpone). This is step taken when enough discussion of the tradeoffs have taken place and the team is in a position to make a decision. Before entering FCP, super majority of the team must sign off.
-- The FCP will lasts 10 days, so it's open for at least 5 business days.
+- The FCP will last 7 days. If there's unanimous agreement among the team the FCP can close early.
+- For voting, the binding votes are comprised of the core team (and subteam maintainers if labeled as a subteam RFC). Acceptance requires super majority of binding votes in favor. The voting options are the following: Affirmative, Negative, and Abstinence. Non-binding votes are of course welcome.
 - If no substantial new arguments or ideas are raised, the FCP will follow the outcome decided. If there are substantial new arguments, then the RFC will go back into development.
 
 Once an RFC has been accepted, the team member who merges the pull request should do the following:
--- a/text/0000-rfc-process.md
+++ b/text/0000-rfc-process.md
@@ -19,7 +19,7 @@ Now that we've entered the beta and people are trying out Cloud Native Buildpack
 # What it is
 [what-it-is]: #what-it-is
 
-RFCs record potential change and with the context and information at the given time. This process provides a defined process for anyone wishing to contribute to the Cloud Native Buildpacks project as well as collect a diverse set of inputs give opportunity for engagement. Anyone who chooses not to actively partcipate in any RFC is presumed to trust their colleagues on the matter. Once a RFC is accepted, they can be referenced as read-only documents in this repository until replaced or ammended by another RFC when context has significantly changed.
+RFCs record potential change and with the context and information at the given time. This process provides a defined process for anyone wishing to contribute to the Cloud Native Buildpacks project as well as collect a diverse set of inputs give opportunity for engagement. Anyone who chooses not to actively partcipate in any RFC is presumed to trust their colleagues on the matter. Once an RFC is accepted, they can be referenced as read-only documents in this repository until replaced or ammended by another RFC when context has significantly changed.
 
 # How it Works
 [how-it-works]: #how-it-works
@@ -46,11 +46,11 @@ What constitutes a "substantial" change may include the following:
 - process changes (project contributor)
 - governance (project contributor)
 
-If you submit a pull request the team deems warrants a RFC will politely closed with a request for one to be opened first.
+If you submit a pull request the team deems warrants an RFC, it will be politely closed with a request for an RFC to be opened first.
 
 ## Process
 
-To get a RFC into Cloud Native Buildpacks, first the RFC needs to be merged into the RFC repo. Once a RFC is merged, it's considered 'active' and may be implemented to be included in the project. These steps will get a RFC to be considered:
+To get an RFC into Cloud Native Buildpacks, first the RFC needs to be merged into the RFC repo. Once an RFC is merged, it's considered 'active' and may be implemented to be included in the project. These steps will get an RFC to be considered:
 
 - Fork the RFC repo: <https://github.com/buildpack/rfcs>
 - Copy `0000-template.md` to `text/0000-my-feature.md` (where 'my-feature' is descriptive. don't assign an RFC number yet).
@@ -67,7 +67,7 @@ Once a pull request is opened, the RFC is now in development and the following w
 - The FCP will lasts 10 days, so it's open for at least 5 business days.
 - If no substantial new arguments or ideas are raised, the FCP will follow the outcome decided. If there are substantial new arguments, then the RFC will go back into development.
 
-Once a RFC has been accepted, the team member who merges the pull request should do the following:
+Once an RFC has been accepted, the team member who merges the pull request should do the following:
 
 - Assign an id based off the pull request number.
 - Rename the file based off the id inside `text/`.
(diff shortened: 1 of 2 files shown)


jkutner:
Can we mark some of the sections or other bits as optional? Maybe Prior Art? I think that will help make the template more approachable.

hone:
@jkutner yeah, all of it is open change. It's common on other RFC forms which is why I put it there. It seems like a lot of the ones I ran across are derived from the Rust RFC. Ideally we hit a happy balance of people putting enough effort and it not being a burden to make one.

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+
+What constitutes a "substantial" change may include the following:
+
+- changes to the spec (buildpack author, platform operator)
+- adding/removing/changing pack commands (buildpack user, buildpack author)
+- lifecycle library interface (platform implementor)
+- process changes (project contributor)
+- governance (project contributor)
+
+If you submit a pull request the team deems warrants a RFC will politely closed with a request for one to be opened first.
```
nebhale:
```suggestion
If you submit a pull request the team deems warrants an RFC, it will be politely closed with a request for an RFC to be opened first.
```

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Now that we've entered the beta and people are trying out Cloud Native Buildpacks in the wild, we need to take more care in introducing change. As part of road of incubation, we need to demonstrate maturity and stability. This process documents a path for anyone to contribute and have input on features.
+
+# What it is
+[what-it-is]: #what-it-is
+
+RFCs record potential change and with the context and information at the given time. This process provides a defined process for anyone wishing to contribute to the Cloud Native Buildpacks project as well as collect a diverse set of inputs give opportunity for engagement. Anyone who chooses not to actively partcipate in any RFC is presumed to trust their colleagues on the matter. Once a RFC is accepted, they can be referenced as read-only documents in this repository until replaced or ammended by another RFC when context has significantly changed.
```
nebhale:
```suggestion
RFCs record potential change and with the context and information at the given time. This process provides a defined process for anyone wishing to contribute to the Cloud Native Buildpacks project as well as collect a diverse set of inputs give opportunity for engagement. Anyone who chooses not to actively partcipate in any RFC is presumed to trust their colleagues on the matter. Once an RFC is accepted, they can be referenced as read-only documents in this repository until replaced or ammended by another RFC when context has significantly changed.
```

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+- adding/removing/changing pack commands (buildpack user, buildpack author)
+- lifecycle library interface (platform implementor)
+- process changes (project contributor)
+- governance (project contributor)
+
+If you submit a pull request the team deems warrants a RFC will politely closed with a request for one to be opened first.
+
+## Process
+
+To get a RFC into Cloud Native Buildpacks, first the RFC needs to be merged into the RFC repo. Once a RFC is merged, it's considered 'active' and may be implemented to be included in the project. These steps will get a RFC to be considered:
```
nebhale:
```suggestion
To get an RFC into Cloud Native Buildpacks, first the RFC needs to be merged into the RFC repo. Once an RFC is merged, it's considered 'active' and may be implemented to be included in the project. These steps will get an RFC to be considered:
```

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+Once a pull request is opened, the RFC is now in development and the following will happen:
+
+- It will be triaged and assigned to a team member in a future working group meeting.
+- That team member will be responsible for championing that RFC and ensuring it gets comments, feedback, or closed.
+- The team will discuss as much as possible in the RFC pull request directly. Any outside discussion will be summarized in the comment thread.
+- When deemed "ready", a team member will propose a "motion for final comment period (FCP)" along with a disposition of the outcome (merge, close, or postpone). This is step taken when enough discussion of the tradeoffs have taken place and the team is in a position to make a decision. Before entering FCP, super majority of the team must sign off.
+- The FCP will lasts 10 days, so it's open for at least 5 business days.
+- If no substantial new arguments or ideas are raised, the FCP will follow the outcome decided. If there are substantial new arguments, then the RFC will go back into development.
+
+Once a RFC has been accepted, the team member who merges the pull request should do the following:
```
nebhale:
```suggestion
Once an RFC has been accepted, the team member who merges the pull request should do the following:
```

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Stick with the current process, which is opening issues on pull requests/issues in the respective repos or the roadmaps repo. This makes it hard for new people to know where, how to propose changes, and how things are decided. The current process makes it hard to track major changes to the project.
+
+# Prior Art
+[prior-art]: #prior-art
+
+The basic format of RFCs was [invented in the 1960s by Steve Crocker](https://en.wikipedia.org/wiki/Request_for_Comments#History). The current RFC process has been derived heavily from [Rust](https://rust-lang.github.io/rfcs/). Many other open source projects have adopted a RFC process:
```
nebhale:
```suggestion
The basic format of RFCs was [invented in the 1960s by Steve Crocker](https://en.wikipedia.org/wiki/Request_for_Comments#History). The current RFC process has been derived heavily from [Rust](https://rust-lang.github.io/rfcs/). Many other open source projects have adopted an RFC process:
```

nebhale:
I'm not sure if it should be defined here, but I was thinking about voting and the idea of a super-majority.  What form should voting take? 

* Affirmative Only
* Affirmative and Negative
* Affirmative, Negative, and Abstinance

nebhale:
> Can we mark some of the sections or other bits as optional? Maybe Prior Art? I think that will help make the template more approachable.

@jkutner Is optional required or could a user put in an N/A where appropriate?  And if someone put in N/A in a place we didn't feel comfortable with (I'm as lazy as the next person), we'd kick it back as part of discussion?

hone:
@nebhale as part of the multiweek FPC, I think it definitely is long, but just defaulted to 10 since I didn't know what number to pick. I'm fine with 3-7 days I'd think. The balance is giving everyone time to finalize comments if they haven't done so. Maybe the right answer here is to have it shorter now since we're still moving rapidly and as things stabilize more and the team gets bigger it'll make sense to slow down.

jkutner:
> I would have expected some change to the flags of pack to not require an RFC. I'd expect robust discussion in a GitHub issue... On the other hand, I could see a different change of flags to pack require an RFC. I don't have a good heuristic today...

@nebhale is the heuristic whether or not it's backwards compatible? Maybe not the only criteria, but it seems like something additive would be fine to discuss in GH issues. While something breaking (and its migration strategy) would require an RFC.

hone:
@jkutner @nebhale I'd argue that it'd be better to hopefully make the RFC light enough. Anything need would benefit as well from understanding context + motivation for adding another flag IMO as we saw today in our WG meeting.

sclevine:
Thoughts:
- Strongly agree that the template should be a suggested format rather that a strict format.
- Process wise, I think we could start by giving every committer the discretion to use the RFC process or not before making a change. We can tighten this up to require folks who aren't maintainers to submit RFCs in the future. When we get to that point, we can define the criteria that makes an RFC necessary on a subteam-by-subteam (or repo-by-repo) basis.
- I'd suggest that we use affirmative-only voting and only require an FCP when requested by the RFC author, maintainer, or core team member. This will speed up suggestions that everyone agrees on. We can change this process once we move out of beta.
- Who votes on RFCs? If the RFC is only applicable to one subteam, do only maintainers and core team members in the team have binding votes? Or are all RFCs project-wide with core team votes only (i.e., same rules as the spec)?
- I'm not sure if we need to start with a dedicated team member for each RFC. It may be enough to review each RFC at weekly WG meetings for now. Also, I think the RFC opener could be responsible for formatting their RFC appropriately before it is merged.
- Requiring each RFC to start as a PR seems heavy-weight. A PR could be required before it's accepted, but I think starting the RFC as a github issue might be sufficient.

nebhale:
I concur on the following points:
* suggested format
* committer discretion to start, and see how that works
* affirmative only voting
* No dedicated team member now (maybe add them in the future?)

Points of disagreement or clarification:
* I think FCPs should apply at all times, but if all binding votes are cast (not just the super-majority), can be closed early.  I'd like to leave the window open for "official" disagreement.
* RFCs should be Core Team + Project Maintainers (taking into account that some RFCs might not be project-specific).  That group has binding votes and needs a super-majority.
* I'm open to non-binding votes but if we want to exclude them I'm OK with that too.
* No dedicated team member now, but maybe add them in the future?
* I'm really attracted to the idea of starting as a GH issue, but concerned that we lose history

hone:
@nebhale @sclevine I'm curious when you think you/someone would want to deviate from the template? I think Ben's point about putting N/A for any section that doesn't seem necessary is a good one.

> Process wise, I think we could start by giving every committer the discretion to use the RFC process or not before making a change.
I like @nebhale's suggestion: Affirmative, Negative, and Abstinence

> Process wise, I think we could start by giving every committer the discretion to use the RFC process or not before making a change.
I disagree with this point. I'd much rather figure out a better scope of what should require a RFC where committers can make changes where needed and not open RFCs. Part of the point of the RFC is anything "substantial" should require a RFC from anyone. I also understand that this is going to be an evolving window. Ideally any of the high level features we talked about on Thursday would benefit from having a RFC written for them.

> I think FCPs should apply at all times, but if all binding votes are cast (not just the super-majority), can be closed early. I'd like to leave the window open for "official" disagreement.
I think this is a good one until we move into GA to revisit if we want to change.

> Who votes on RFCs?
> RFCs should be Core Team + Project Maintainers (taking into account that some RFCs might not be project-specific).
I think @nebhale's suggestion is good. I would imagine when we get bigger the Core Team will probably vote a lot less so subteams can have more autonomy?

> I'm not sure if we need to start with a dedicated team member for each RFC. It may be enough to review each RFC at weekly WG meetings for now.
I'm also fine with this. The dedicated member is probably premature, but it's mostly to ensure things don't become stale.

> Requiring each RFC to start as a PR seems heavy-weight. A PR could be required before it's accepted, but I think starting the RFC as a github issue might be sufficient.
I struggled with this too. I wanted to balance lightweight vs too much process. After seeing #2, I feel like it's probably ok to require it. It's just nice to have all the discussion in one place and inline comments. RFCs are substantial features and therefore should require some more work put behind it.

I'll update this PR with suggestions from this thread.

text/0000-rfc-process.md
```diff
@@ -0,0 +1,109 @@
...
+
+Once a pull request is opened, the RFC is now in development and the following will happen:
+
+- It will be triaged, labeled, and discussed in a future working group meeting. Working group meetings happen on a weekly cadence barring exceptions. Labeling marks a RFC as general or specific to a subteam.
+- The team will discuss as much as possible in the RFC pull request directly. Any outside discussion will be summarized in the comment thread.
+- When deemed "ready", a team member will propose a "motion for final comment period (FCP)" along with a disposition of the outcome (merge, close, or postpone). This is step taken when enough discussion of the tradeoffs have taken place and the team is in a position to make a decision. Before entering FCP, super majority of the team must sign off.
+- The FCP will last 7 days. If there's unanimous agreement among the team the FCP can close early.
+- For voting, the binding votes are comprised of the core team (and subteam maintainers if labeled as a subteam RFC). Acceptance requires super majority of binding votes in favor. The voting options are the following: Affirmative, Negative, and Abstinence. Non-binding votes are of course welcome.
```
ekcasey: What is our threshold for a "super majority"? Do we just mean "majority"?

nebhale: I believe that we'd aimed for super majority to mean 2/3 or greater and no single company can have more than 50% of countable votes.  This latter requirement might mean that a voter might be required to abstain when they otherwise have an opinion.

PR #4, opened 2019-04-26
RFC for new pack suggest-stacks command

ekcasey:

### DIFF
--- a/text/0004-pack-suggest-stacks.md
+++ b/text/0004-pack-suggest-stacks.md
@@ -1,0 +1,82 @@
+# Meta
+[meta]: #meta
+- Name: Suggest Stacks
+- Start Date: 2019-04-25
+- CNB Pull Request: (leave blank)
+- CNB Issue: #buildpack/pack#156
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+A `suggest-stacks` command will be added to the `pack` CLI help buildpack authors discover stacks. The output of this command will include the `io.buildpacks.stack.bionic` stack (maintained by the Cloud Native Buildpacks project) and other stacks that are popular within the community.
+
+# Motivation
+[motivation]: #motivation
+
+### Case 1: Buildpack author providing a buildpack to the community
+Given that buildpack authors must explicitly list supported stack in the buildpack's `buildpack.toml`, a buildpack author that hopes to provide a buildpack for comsuption by the community (an APM vendor for example) will want to know what stacks they should consider supporting.
+
+### Case 2: Buildpack author creating a builder image
+A buildpack author who wishes to create a builder but who does not want to create or maintain a stack should be able to easily find a maintained stack to use. `pack` and our documentation currently provides no guidance here. This is evidenced by questions from users in github and slack who are struggling to fill out the `stack` section of `builder.toml` with appropriate values. Making stacks discoverable in pack should significantly improve this experience.
+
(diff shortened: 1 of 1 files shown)


ameyer-pivotal:
👍

text/0000-pack-suggest-stacks.md
```diff
@@ -0,0 +1,81 @@
...
+- Name: Suggest Stacks
+- Start Date: 2019-04-25
+- CNB Pull Request: (leave blank)
+- CNB Issue: #buildpack/pack#156
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+A `suggest-stacks` command will be added to the `pack` CLI help buildpack authors discover stacks. The output of this command will include the `io.buildpacks.stack.bionic` stack (maintained by the Cloud Native Buildpacks project) and other stacks that are popular within the community.
```
jkutner: could this be `pack stacks`?

ekcasey: `pack suggest-stacks` would be consistent with the existing `pack suggest-builders` command. I think the idea behind the `suggest-*` command names is to convey to users that these lists are not comprehensive. If for example, I create a stack, users should not expect it to appear in this list.

text/0000-pack-suggest-stacks.md
```diff
@@ -0,0 +1,81 @@
...
+    Run Image: heroku/pack:18
+
+    Stack ID: org.cloudfoundry.stacks.cflinuxfs3
+    Maintainer: Cloud Foundry
+    Desctipion: (TO BE PROVIDED BY CLOUD FOUNDRY)
+    Build Image: cfbuildpacks/cflinuxfs3-cnb-experimental:build
+    Run Image: cfbuildpacks/cflinuxfs3-cnb-experimental:run
+```
+
+Other stacks can be add to the list of community stacks upon request from the maintainer, if it can be demonstrated that the images can be trusted by the community and the images will be actively patched with CVE fixes.
```
hone: Do we have other criteria of what marks something to be "trusted" besides patching CVEs? Do we have a metric of how fast CVEs should be patched that we're guaranteeing on our own bionic stack?

text/0000-pack-suggest-stacks.md
```diff
@@ -0,0 +1,81 @@
...
+[alternatives]: #alternatives
+
+As an alternative apporach, documentation could be used to improve stack discoverability. I hypothesize that building the discoverability into `pack` will make this information more easily accessible compared with documentation (many users will explore `pack --help` when they have a question, only resorting to documentation when necessary).
+
+If we do not solve this problem users will continue to struggle to use the `create-builder` command correctly, and therefore be unable to take advantage of one of the powerful features of `pack`.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Is providing run image mirrors valuable or unnecessary?
```
hone: I think it's definitely not necessary. If a buildpack author is just using this list to initially bootstrap, it's probably information that they don't need immediately. That being said, it is very valuable information.

ekcasey:
@hone I could see a compelling argument for only suggesting the bionic stack. It keeps things simple for users and eliminates the need to establish criteria for vetting entries to the community maintained stacks list. We could revisit the community maintained stacks list idea in the future if there is evidence after releasing this feature that it would be provide value.

PR #5, opened 2019-04-26
RFC for new pack inspect-image command

ekcasey:
[rendered](https://github.com/buildpack/rfcs/blob/inspect-image/text/0003-pack-inspect-image.md)


text/0000-pack-inspect-image.md
```diff
@@ -0,0 +1,96 @@
...
+[meta]: #meta
+- Name: Inspect Image
+- Start Date: 2019-04-25
+- CNB Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+An `inspect-image` command will be added to the `pack` CLI, allowing buildpack users to easily view metadata about an image generated by the Cloud Native Buildpack Lifecycle. The command will print the stack ID, stack metadata (run-image and run-image-mirrors), the buildpacks that contributed to the image, the bill of materials, and the processes types.
```
jkutner: This may be out of scope for this RFC, but as we're starting to get more and more of the hyphenated commands, I wonder if we should consider other command patterns, such as:

* `pack [topic]:[command]` (in this case `pack image:inspect`
* `pack [topic] [command]` (in this case `pack image inspect` (similar to `docker`)

I'm not necessarily against hyphenated commands, I just want to make sure we're choosing that pattern for good reasons.

ekcasey: I'm open to the idea of changing the pattern of all of our cli commands, but it seems like that should be a different RFC or issue. I wouldn't want the implementation of this RFC to be blocked on that discussion.

jkutner: yea, that's fair

text/0000-pack-inspect-image.md
```diff
@@ -0,0 +1,96 @@
...
+
+If we chose to use the layer metadata instead, I would propose removing the `[bom]` section from the image `metadata.toml` file, given that it currently serves no compelling purpose.
+
+The problem with either of these approaches is that buildpack authors are likely to view both the build plan entry claims and the layer metadata from a functional rather than an informative perspective. Data in these field may not be human friendly. 
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we add the buildpack id and version to the BOM entries?
+- Is there a better format for the BOM output than YAML? TOML would be in keeping with our ecosystem but it's harder to read.
```
ameyer-pivotal: It's my understanding that BOMs can become quite large. I wonder how we could mitigate this? Perhaps if a BOM is longer than a certain number of lines, we dump it to a file instead and give a link?

Otherwise, I think this new command makes a lot of sense.

text/0000-pack-inspect-image.md
```diff
@@ -0,0 +1,96 @@
...
+* using layer metadata would allow us to tell advanced users specifically which layers dependencies live in if we wished to at a later date
+
+If we chose to use the layer metadata instead, I would propose removing the `[bom]` section from the image `metadata.toml` file, given that it currently serves no compelling purpose.
+
+The problem with either of these approaches is that buildpack authors are likely to view both the build plan entry claims and the layer metadata from a functional rather than an informative perspective. Data in these field may not be human friendly. 
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we add the buildpack id and version to the BOM entries?
```
hone: For auditing purposes is it important to know where the buildpacks come from?

hone:
From the last specification WG, we've decided to split the bill of materials into a `--bom` command to display it.

sclevine:
@hone any additional questions? Would like to begin implementing this.

PR #6, opened 2019-05-06
RFC for pack logging interface

jkutner:
This is a proposal to add a generic logging interface in pack that can be used by consumers to inject custom loggers. In addition we propose a new implementation of the pack logger that uses a third-party library.

### DIFF
--- a/text/0005-pack-logging-refactor.md
+++ b/text/0005-pack-logging-refactor.md
@@ -1,0 +1,56 @@
+# Meta
+[meta]: #meta
+- Name: Pack Logging Interface
+- Start Date: 2019-05-06
+- CNB Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal to add a generic logging interface in pack that can be used by consumers to inject custom loggers. In addition we propose a new implementation of the pack logger that uses a third-party library.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


sclevine:
Strongly agree, although I'm not sure if the proposed interface is sufficient to cover the current functionality.

murphybytes:
@sclevine You can do something like the following code snippit that is from a proof of concept for this RFC that I'm working on.  So in places that take an io.Writer like [here](https://github.com/buildpack/pack/blob/master/build/phase.go#L118) you'd replace the `io.Writer` arguments with `logging.NewWriter(logger.Info).WithPrefix('foo')` for example.
```go 
// Writer will wrap a logging method in an io.Writer
type Writer struct {
	logMethod func(string)
	prefix    string
}

// NewWriter takes a log method and creates a Writer. For example
// w := NewWriter(log.Info)
func NewWriter(f func(string)) *Writer {
	return &Writer{
		logMethod: f,
	}
}

// Writes bytes to the embedded log function
func (w *Writer) Write(buf []byte) (int, error) {
	w.logMethod(w.prefix + string(buf))
	return len(buf), nil
}

// WithPrefix prepends prefix to log messages
func (w *Writer) WithPrefix(prefix string) *Writer {
	w.prefix = fmt.Sprintf("%s[%s] ", w.prefix, style.Prefix(prefix))
	return w
}
```

PR #7, opened 2019-05-15
RFC Process

hone:
I accidentally closed #1, This is the follow up pull request which should have incorporated all the feedback so far.

[rendered](https://github.com/hone/cnb-rfcs/blob/rfc/text/0000-rfc-process.md)


jkutner:
@sclevine @nebhale @ekcasey are there any more changes you’d like to see on this?

PR #8, opened 2019-05-30
RFC for renaming launch.toml

hone:
[rendered](https://github.com/hone/cnb-rfcs/blob/rename-app-toml/text/0000-rename-app-toml.md)

Let the bikeshed commence!

### DIFF
--- a/text/0000-rename-app-toml.md
+++ b/text/0000-rename-app-toml.md
@@ -1,0 +1,57 @@
+# Meta
+[meta]: #meta
+- Name: rename app.toml
+- Start Date: 2019-05-29
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


jkutner:
Is it possible that `project.toml` (with the same name and schema) becomes the app descriptor that was roughly described in #3? If yes, we should consider that when deciding on the name (even though it may not change anything).

hone:
@jkutner I'm fine with that. I'm not strongly opinionated on the filename itself, but we should move away from `app.toml` in the spec since there are no plans to make that change.

sclevine:
@nebhale Note that we voted in the WG meeting to revert the spec commit changing `launch.toml` to `app.toml`, given that there was never an RFC for it and it was never implemented in the lifecycle. This RFC is now about renaming `launch.toml` to something else.

@hone changing the title accordingly, hope you don't mind 😄

hone:
@sclevine yeah, no worries.
@nebhale I'm fine with that. Did you have new work in mind?

Should we just close/merge this in?

PR #9, opened 2019-06-10
RFC for additional metadata for app

iainsproat:
This is a proposal to allow additional metadata related to the app to be provided and exported to the image manifest.

### DIFF
--- a/text/0006-app-layer-metadata-source.md
+++ b/text/0006-app-layer-metadata-source.md
@@ -1,0 +1,102 @@
+# Meta
+[meta]: #meta
+- Name: Manifest - App metadata - add source url & hash
+- Start Date: 2019-06-10
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Adds Source URL and hash properties to app metadata in the manifest, `Config.Labels."io.buildpacks.lifecycle.metadata".app`.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+The provenance of the app should be transparent, and the image should be
+reproducible.  Providing metadata about the source URL and commit hash of the App assists
+in achieving this.
+
+- What use cases does it support?
+### Case 1
+This supports providing buildpack users, including users with quality assurance, legal or security roles, with data about the origin of packaged material.  This helps in their understanding of the provenance of the app.
+
+### Case 2
+This supports providing a buildpack user with information that would allow them to attempt to reproduce the build/packaging.  This may be done as part of a verification step related to the above use case.
+
+- What is the expected outcome?
+Additional metadata would be provided to allow for verification of the source and reproduction of the build.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This feature adds additional metadata to the image manifest.  It introduces two
+new properties related to the App layer.  These properties are the URL and hash
+of the source.
+
+It is proposed that these properties would be added as children of `Config.Labels."io.buildpacks.lifecycle.metadata".app`.
+
+We propose `source.url` and `source.hash` properties.
+
+Example:
+```
+{
+  Config: {
+    Labels: {
+      "io.buildpacks.lifecycle.metadata":"{\"app\":{\"source\":{\"url\":\"https://github.com/buildpack/rfcs.git\", \"hash\":\"a33a985597b04c36aeefd6b17c4ef593adb5dc01\"}}}"
+    }
+  }
+}
+```
+
+Unencoded, `io.buildpacks.lifecycle.metadata` is:
+```
+{
+  "app": {
+    "source": {
+      "url": "https://githubcom/buildpack/rfcs.git",
+      "hash": " a33a985597b04c36aeefd6b17c4ef593adb5dc01"
+    }
+  }
+}
+```
+
+# How it Works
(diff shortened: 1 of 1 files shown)


jkutner:
I want think about the finer points some more,  it in general I’m a fan of this idea. Thanks for writing it up!

iainsproat:
@sclevine - we've revised the RFC following your comments in our discussion on Monday.  It now proposes a new `io.buildpacks.app.metadata` property, instead of co-opting the existing `io.buildpacks.lifecycle.metadata`.  The `sha` property is renamed to `ref` to make it more generic and less tied to git.

cmccarthy101:
@sclevine  We have made changes to the RfC to reflect our chat about updating the How It Works section. This RfC should be good to vote on now.

nebhale:
Currently this proposal doesn't describe optionality or extensibility of this and probably needs to.  Consider the following cases:

1. Many projects build and test their application artifact using CI tools and only create the deployment artifact (sluglet before, image for CNB) towards the end of their CD pipeline.  This results in the input to the buildpack lifecycle not being obviously connected to a repo and commit.  _Is this metadata required to be populated?_
1.  Many enterprise applications use VCSs that not only not Git, but in many cases don't use hashes for ids (e.g. Subversion, CVS).  _Is this metadata extensible to allow identification schemes that don't match current DVCS conventions?_

sclevine:
How about something like this:

Label: `io.buildpacks.source.metadata` (optional)
Example values:

Git:
```js
{
   "id": "60d5fb7a7ad7c3b357a9d783b740f765d2a0d4d5", 
   "type": "git",
   "metadata": {
      "commit": "60d5fb7a7ad7c3b357a9d783b740f765d2a0d4d5",
      "refs": ["master", "v3.0"],
      "url": "https://github.com/example/source"
   }
}
```

Source upload:
```js
{
   "id": "146c4bce42545e6a4575283b32a7f01924ef86ce848273079693a42b52b27321", 
   "type": "image",
   "metadata": {
      "digest": "146c4bce42545e6a4575283b32a7f01924ef86ce848273079693a42b52b27321",
      "path": "/source",
      "repository": "hub.docker.io/example/image",
      "refs": ["example/image:mytag"]
   }
}
```
Where `id` is a unique identifier for an exact, immutable revision of the source code.

jchesterpivotal:
In Concourse the comparable concept to `id` is `version`, which can be an arbitrary JSON object. _Almost_ always it's a single-key object (eg, the git resource uses `{"ref":"...."}`. 

But occasionally it's necessary to use multiple fields to properly specify the version. The PR resource does this by including the PR number, repository and git commit as keys on each version.

iainsproat:
I like @sclevine's idea in principle, it feels nicely extendable.

But it feels as if this "wheel" should have been invented somewhere else before (wishful thinking maybe?!).

If we're to develop this metadata for source code identification for other (non-git) systems (present and/or future), then I think we should try to explore whether there's an existing standard ontology that we can adopt.

A quick search surfaced the following existing work or bodies working on software identification ontologies:
* https://github.com/codemeta/codemeta
* [OntoSoft](http://ontosoft.org/ontology/software/)
* [Research Data Alliance, Software Source Code Identification Working Group](https://www.rd-alliance.org/groups/software-source-code-identification-wg)
* [Software Citation Principles](https://www.force11.org/software-citation-principles)

However, at first glance I don't think they even particularly suit referencing commits within branches for git.  One of the recommendations in the Force11 principle is to use DOI's, which feel like a layer of indirection.

I'm not sure if there's something else out there; perhaps copying other software (as in @jchesterpivotal suggestion to follow Concourse's standards) to converge towards a de facto standard?

jchesterpivotal:
SPDX and CycloneDX also have relevant standards (and rely on other standards like Package URL, which I believe can refer to git commits).

In general, I agree that if there's a good standard we can adopt, we ought to.

sclevine:
Not sure if I like this better (as I think there are advantages to a single unique string for the version), but here's a concourse-like version:

Git:
```
{
   "type": "git",
   "version": { "commit": "60d5fb7a7ad7c3b357a9d783b740f765d2a0d4d5" }, 
   "metadata": {
      "refs": ["master", "v3.0"],
      "url": "https://github.com/example/source"
   }
}
```

Source:
```
{
   "type": "image",
   "version": { "digest": "146c4bce42545e6a4575283b32a7f01924ef86ce848273079693a42b52b27321" }, 
   "metadata": {
      "path": "/source",
      "repository": "hub.docker.io/example/image",
      "refs": ["example/image:mytag"]
   }
}
```

sclevine:
@iainsproat I think the above format addresses the concerns raised, as long as the metadata is clearly optional. Are you willing to update this RFC accordingly?

iainsproat:
Thanks Stephen.  It's in our backlog to update this RFC, we'll make the
changes soon (hopefully within the next couple of days).

cmccarthy101:
@sclevine 
We have updated the format of the source information in the RFC given the comments above. 

We have a question about the source information. Currently, buildpacks build images according to a single source repo. Is there any necessity to extend this to multiple sources? Eg having the label named `sources` and having an array of sources.

hone:
Overall, I like where this RFC is going.

@cmccarthy101 For multiple sources are these sources mirrors or multiple sources needed to construct the image?

iainsproat:
@hone - these would be multiple sources needed to construct the image.

PR #10, opened 2019-06-12
Build descriptor

jkutner:
I also tried to include:

```
[[build.packages]]
name = "rolldice"
```

But that's a bit complicated, and I didn't want to muddy the concept.

I also omitted any attempt to include `[launch]`, which might override launch-time config.


nebhale:
I'd like to see that spec say that if this file still exists in the source folder at the end of the `build` phase, it is removed by the lifecycle.  It's an incredibly dangerous file to leave hanging around in the case of an exploded JAR.

jkutner:
@nebhale this file shouldn't contain secrets (granted, people will put them in there). And there are cases where it will be useful to keep this file. Could we leave it up the Java buildpack to remove it manually? Or could we add a `self-desctruct = false` key (Stephen will hate that it defaults to true, so maybe another name)?

nebhale:
I'm still partial to removing as part of the spec.  Beyond the obvious problems with the Java classpath, there's also the problem of creating an image with secrets embedded within them.  From an enterprise perspective, this is one of the larger security problems they are grappling with when using Dockerfiles.  As this file _shouldn't_ have any usage at runtime, there's no reason it can't be removed to ensure fewer security vulnerabilities in created images.

PR #11, opened 2019-06-12
RFC: Distribution Specification

sclevine:
[Readable](https://github.com/buildpack/rfcs/blob/dist-spec/0000-spec-distribution.md).


sclevine:
@hone @jkutner @nebhale @ekcasey @matthewmcnew 
Addressed feedback from the WG meeting today. Now includes `package.toml`, `builder.toml`, and addresses usability feedback.

0000-spec-distribution.md
```diff
@@ -0,0 +1,302 @@
...
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
+mixins = ["build:git"]
+```
+
+`pack create-builder` will generate a builder image from buildpackages, buildpack blobs, and stack metadata.
+
+These properties will be specified in a `builder.toml` file.
+
+```toml
```
sclevine: @ekcasey @hone this UX gives builder creators maximum flexibility in defining a builder. Buildpackages are not even necessary.

sclevine:
@jkutner I added the buildpack ID to the environment as `$BP_ID`. @nebhale pointed out that otherwise, buildpacks would need to hardcode their ID to figure out which entry they correspond to in`buildpack.toml`.

sclevine:
Two things I realized need to be specified while playing with this API over the weekend:

1. The definition of `optional` on a order definition buildpack is that the lifecycle should repeat the group it is specified in without the order definition. If we naively assigned `optional` to all sub-buildpacks in the order definition, we could break nested order definitions that rely on some buildpacks not being optional.

   This is completely consistent with the current meaning of optional for a concrete buildpack implementation. The current behavior is really just a performance optimization on a “leaf node” that requires fewer re-tries.

2. We should define the expansion as depth-first in left-to-right over. This should work well with add-on type buildpacks that instrument a built app. It’s also easiest to understand (I think).

0000-spec-distribution.md
```diff
@@ -0,0 +1,307 @@
...
+A buildpack blob is a tgz containing a `buildpack.toml`. Example format:
+
+```toml
+[[buildpacks]]
+id = "io.buildpacks.nodejs"
+name = "Node.js Buildpack"
+version = "0.0.9"
+[[buildpacks.order]]
+group = [
+   { id = "io.buildpacks.node", version = "0.0.5" },
```
hone: can we make `version` optional if there's only 1 of that id specified?

0000-spec-distribution.md
```diff
@@ -0,0 +1,307 @@
...
+## Buildpack Blob Format
+
+A buildpack blob is a tgz containing a `buildpack.toml`. Example format:
+
+```toml
+[[buildpacks]]
+id = "io.buildpacks.nodejs"
+name = "Node.js Buildpack"
+version = "0.0.9"
+[[buildpacks.order]]
```
hone: every buildpack in this array must exist in the `[[buildpacks]]` key?

0000-spec-distribution.md
```diff
@@ -0,0 +1,307 @@
...
+
+`pack build` should accept a list of buildpacks via the `--buildpack` flag.
+The flag may be passed multiple times to construct a buildpack group.
+
+The value of each flag must be one of:
+- A buildpack ID of a buildpack on the builder, optionally followed by `@` and a version
+- A path to a buildpackage on the local filesystem
+- A reference to a buildpackage on a Docker registry  
+
+A version must be specified if the version is ambiguous.
```
hone: when would version be ambiguous?

0000-spec-distribution.md
```diff
@@ -0,0 +1,307 @@
...
+
+`pack create-package` will package a selection of
+
+- gzip compressed, tar archived buildpack blobs
+- other buildpackages
+- stack and buildpack metadata
+
+into a `.cnb` file, OCI image in a registry, or OCI image in a Docker daemon.
+
+These properties will be specified in a `package.toml` file.
```
hone: This file looks very similar to `buildpack.toml`. Is it crazy to think maybe they should be the same?

0000-spec-distribution.md
```diff
@@ -0,0 +1,307 @@
...
+id = "io.buildpacks.nodejs"
+version = "0.0.9"
+[[blobs.buildpacks]]
+id = "io.buildpacks.npm"
+version = "0.0.7"
+[[blobs.buildpacks]]
+id = "io.buildpacks.node"
+version = "0.0.5"
+
+[[packages]]
```
hone: Is this a new key? I couldn't find this in the existing spec.

0000-spec-distribution.md
sclevine: Nope, they are totally disjoint. E.g., you could fine a repo that just contains a single buildpack with this configuration.

0000-spec-distribution.md
sclevine: Only one of them where? A single version of the `order`-containing buildpack should represent an exact combination of versions of other buildpacks. Those buildpacks don't need to be defined in this file.

0000-spec-distribution.md
sclevine: If you have multiple versions of the same buildpack in the builder.

0000-spec-distribution.md
sclevine: `buildpack.toml` defines buildpacks -- it has their metadata, names, paths to their source, etc.

`package.toml` references buildpacks. It's the (potentially ephemeral) configuration file for `pack create-package`. It contains real-world URIs, etc. that are used to build buildpackages. This is information that can't live in buildpack.toml.

0000-spec-distribution.md
sclevine: This is a new key. There's no existing spec for any of these proposed files outside of this RFC.

0000-spec-distribution.md
```diff
@@ -0,0 +1,332 @@
...
+Example `package.toml` for Node.js with just blobs:
+
+```toml
+[default]
+id = "io.buildpacks.nodejs"
+version = "0.0.9"
+
+[[blobs]]
+uri = "https://example.com/nodejs.tgz"
+[[blobs.buildpacks]]
```
sclevine: @hone After discussing with @ekcasey, I think it's okay to make the `blobs.buildpacks` list optional. If unspecified, it means "all buildpacks contained in the blob." This should address the repetition between buildpack.toml and package.toml.

ameyer-pivotal:
I really like this. One non-trivial concern I have is with the naming "buildpackage". Since, "pack" is often short for "package", I think it will lead to confusion between "buildpack" and "buildpackage". Instead, I'd like to propose "buildpack bundle" (or something similar, if anyone has ideas).

sclevine:
It was called a bundle in the original proposal, but @ekcasey suggested buildpackage instead. I don't have a strong opinion.

0000-spec-distribution.md
```diff
@@ -0,0 +1,313 @@
...
+
+### App Developer
+
+`pack build` should accept a list of buildpacks via the `--buildpack` flag.
+The flag may be passed multiple times to construct a buildpack group.
+
+The value of each flag must be one of:
+- A buildpack ID of a buildpack on the builder, optionally followed by `@` and a version
+- A path to a buildpackage on the local filesystem
+- A reference to a buildpackage on a Docker registry
```
Zanadar: In our development workflow this seems non-ideal, because it looks like we lose the ability to directly use a buildpack, because we have to now go:
`buildpack-code -> [compiled buildpack] -> [buildpackage]. `
That makes development of buildpacks a bit more cumbersome, I think.

0000-spec-distribution.md
```diff
@@ -0,0 +1,313 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+## Overview of Changes
+
+- The `order.toml` and `buildpack.toml` files are merged into `buildpack.toml`.
+- `buildpack.toml` can contain multiple buildpacks.
+- A buildpack inside of `buildpack.toml` is either:
+  a. a single buildpack implementation specified by a `path` field or
+  b. an ordering of other buildpack IDs/versions
```
Zanadar: Can we compose buildpackages from each other?

Zanadar: It looks like yes (https://github.com/buildpack/rfcs/pull/11/files#diff-8e4e8df213b3904199aec818ffac0c54R249) unless I'm misunderstanding was "package"  here means.

sclevine: Yep, see the `[[packages]]` section in `package.toml`.

0000-spec-distribution.md
sclevine: That’s a good point. Ideally, the whole compilation process could be wrapped into `pack create-package`, but we would definitely need some kind of pre-package hook in `buildpack.toml`.

0000-spec-distribution.md
Zanadar: If this key refers to buildpackages, it should say that. If it doesn't I'm confused. I wonder if a more extensible concept like `bits`, where each bits says what it is, would allow future types of subpackage more easily. eg:
```
[[bits]]
uri = "https://example.com/rails.tgz"
[[bits]]
ref = "registry.example.com/nodejs:0.0.9"
[[bits]]
futurething = "some-other-packaging-format.fmt"
```

0000-spec-distribution.md
Zanadar: So basically have a tree where nodes are either buildpacks or trees of buildpacks...interesting.

0000-spec-distribution.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Buildpackage Format
+
+A buildpackage may exist as an OCI image on an image registry, an OCI image in a Docker daemon, or a `.cnb` file.
+
+A `.cnb` file is an uncompressed tar archive containing an OCI image. Its file name should end in `.cnb`.
+
+### Layer Blob
+
+Each FS layer blob in the image contains a single buildpack blob tgz and at least one symlink.
```
Zanadar: is the "single *buildpackage* blob" or "single tgz containing buildpacks"? Because the example below is confusing given that the checksums for each differant ID is the same. Maybe I'm just misreading here.

0000-spec-distribution.md
Zanadar: Similar feedback to above. This makes creating a builder more complicated (or maybe just confusing) given that buildpackages are now another possible component of the builder. Is the "best practice" to create a buildpackage per *the concept formerly known as a label* and then compose those or just mix and match buildpackages and buildpacks directly. As a buildpack distributor, I know have to decide that  I guess?

Zanadar:
Some feedback from the perspective of release engineering buildpacks for distribution. Overall in support, but I think this complicates our distribution and development story. Which is okay if we feel that this ultimately eases pains of consuming buildpacks. I missed the WG meeting, but is a registry somehow a better distribution point than a uri pointing to a tarball b/c of content adressability or something like that?

0000-spec-distribution.md
sclevine: A blob always means a tgz containing exactly one buildpack.toml and any number of buildpacks (either order definitions or concrete buildpacks).

0000-spec-distribution.md
sclevine: 👍

0000-spec-distribution.md
sclevine: I think I prefer separately lists, because it creates a static schema for the file. I'm not super attached to the names (`blobs` vs. `packages`) though.

0000-spec-distribution.md
Zanadar: ✔️ thanks.

0000-spec-distribution.md
sclevine: A buildpackage is a collection of buildpacks with a single entrypoint ID that references all of those buildpacks (via order definitions). It may work with any number of stacks.

A builder is a single stack image with any number of buildpacks or buildpackages on it.

The builder is useful for building on a workstation, where you need a single copy of the latest buildpacks. It's also useful for building on a platform that doesn't explicitly support CNB (like tekton or concourse).

A buildpackage is more useful for distributing a group of buildpacks that could be used on CF, Heroku, or another buildpack-native platform. It's also a useful way to extend builders that might not have all the buildpacks you need.

In the long term, once we have a centralized buildpack registry that maps buildpack IDs to their package locations, I expect that the builder and package.toml concepts might fade away.

sclevine:
Using a docker registry has a few key benefits:
1. Builders and buildpackages can be constructed dynamically on a registry without (significant) data transfer.
2. Only the changed blobs are downloaded when you download a new copy of a buildpackage or builder from a registry.

In the long term, a centralized buildpack registry might construct a buildpackage on-the-fly when you pull it by ID, e.g.,
```
docker pull bp.example.com/org.cloudfoundry.ruby
```
Could dynamically feed you a manifest that references the buildpack blobs contained by that ID.

0000-spec-distribution.md
jkutner: is `[[buildpacks.order]]` only allowed on any element in the list of `[[buildpacks]]`?

sclevine: Each buildpack in the top-level list may have a `path` or an `order`. The only other place `order` is valid is builder.toml (but that file is just a config for the pack CLI).

0000-spec-distribution.md
hone: I also do local buildpack development this way with `--buildpack` using `pack`. There's a huge ease of use factor when being able to use an exploded buildpack directory.

sclevine:
@jkutner I updated this to support the workflows you brought up today. Both `pack build --buildpack <blob-dir>` and `pack create-package <blob-dir>` are now supported.

I used the first buildpack in `buildpack.toml` for now, but happy to change as needed.

jkutner:
I was hoping we could arrive at something like this (I don't like having `buildpack` & `buildpacks` in the schema but ignore that):

```toml
[buildpack]
id = "io.buildpacks.nodejs"
name = "Node.js Buildpack"
version = "0.0.9"
[[buildpack.order]]
group = [
   { id = "io.buildpacks.node", version = "0.0.5" },
   { id = "io.buildpacks.npm", version = "0.0.7" },
]

[[buildpacks]]
id = "io.buildpacks.npm"
name = "NPM Buildpack"
version = "0.0.7"
path = "./npm-cnb/"
[buildpacks.metadata]
# ...
[[buildpacks.stacks]]
id = "io.buildpacks.stacks.bionic"

[[buildpacks]]
id = "io.buildpacks.node"
name = "Node Engine Buildpack"
version = "0.0.5"
path = "./node-cnb/"
[buildpacks.metadata]
# ...
[[buildpacks.stacks]]
id = "io.buildpacks.stacks.bionic"
```

The idea being that every `buildpack.toml` has a single `buildpack` that is akin to what you've defined as the "first buildpack".

But given that we didn't land on something like ^^, I think the file should be named `buildpacks.toml`. I'm happy to open that in its own RFC though.

I'm going to approve on this, but I still think that there may be a better way to capture the "first buildpack" as being special--in a way that is more obvious to the author/reader of the `buildpack.toml`.

sclevine:
Thanks @jkutner. Hopefully going with the first buildpack for now will make it easy to override that behavior with a different mechanism once we identify it. Happy to consider the current solution a starting point, and works towards functionality that makes dealing buildpack blobs directly an ideal experience.

Waiting for @hone to approve before merging.

sclevine:
@hone reminder about this RFC

sclevine:
> There were concerns brought up about mixing defining metadata and how to package, but from I can tell this is common in most language ecosystems in dependency managers like package.json or Cargo.toml that includes both metadata about the package and how to actually build them.

Just to clarify, I'm only concerned about putting a mandatory single URI for each buildpack reference in `buildpack.toml`. That's the only function I see `package.toml` providing right now. Once we have a registry, I'm very supportive of it going away, and of us adding a `registries` key to `buildpack.toml`.

PR #14, opened 2019-06-19
RFC: Auto-load User-provided Environment Variables

sclevine:
[Readable](https://github.com/buildpack/rfcs/blob/env-whitelist/0000-user-provided-env-var-whitelist.md).


jkutner:
I think we should note that this is a breaking change (an app that expects a provided JAVA_OPTS not to be loaded by default could break).

I'm lukewarm on the key name, but I understand your concern about flags that default to true. My feeling is that `secure-env` might not convey the intention well enough, especially because the primary use case isn't about security. Is there a better inverse of `load-env`? Maybe `serialize-env`?

hone:
@sclevine @nebhale are you doing this for most of the CF (non-CNB) buildpacks today? I wonder if this is expected behavior? Is it better to capture this kind of functionality into `libbuildpack`?

sclevine:
We already have this functionality in libbuildpack -- this RFC is intended to make it easier for buildpack authors to write simple buildpacks in the language of their choosing.

I've seen quite a few users surprised by the current behavior: you set environment variables, but they don't show up in the environment.

hone:
@sclevine thoughts on providing a blacklist as a buildpack author?

jkutner:
I think an blacklist would be a nice supplement to this. but the `libbuildpack` solution prevents "it just works" behavior for env vars, which stinks for new buildpack authors. 

Sidenote: I wonder if we should blacklist `PATH` by defafult.

sclevine:
@dzuelke suggests `clear-env`

hone:
Overall, I'm 👍 this change now. From the spec WG today, it looks like we're in agreement that we'd want to blacklist `PATH`, since there isn't a good way to deal with appending/prepending and knowing what the user wants and eval'ng the value they insert. Where do we handle this blacklist? I'm in favor of doing this at the lifecycle level. A concern that was raised was disjoint between what exists in `platform/env` vs what is autoloaded into the env. Would it be possible to have the default `PATH` blacklist as part of this RFC?

A separate platform level to blacklist env vars, could be a separate RFC and out of scope of this one imo.

sclevine:
> I'm in favor of doing this at the lifecycle level.

Does this mean that builds fail immediately if `PATH` is provided as a build time env var?

Or does it mean that without `clear-env = true`, `PATH` doesn't get set automatically? (Maybe the user can still suggest `PATH` values to the buildpack this way, as a mechanism to provide additional values?)

Should we blacklist the other lifecycle-managed env vars? (See: https://github.com/buildpack/spec/blob/master/buildpack.md#provided-by-the-lifecycle)

hone:
> Should we blacklist the other lifecycle-managed env vars? (See: https://github.com/buildpack/spec/blob/master/buildpack.md#provided-by-the-lifecycle)

I was wondering that too.

> Or does it mean that without clear-env = true, PATH doesn't get set automatically? (Maybe the user can still suggest PATH values to the buildpack this way, as a mechanism to provide additional values?)

I was leaning towards this. The thing that's unclear is how the buildpack would or should use `PATH`?

sclevine:
How about: for any of those five env vars, we only set them automatically (assuming no `clear-env = true`) if they are not already set in the real environment by the platform. 

Realistically, this just means blacklisting auto-set `PATH`, but it could prevent unexpected behavior if the platform sets the others.

dzuelke:
We could also have a list of "smart" env vars where we support automatic appending or prepending based on a `:` syntax without the existing variable being referenced.

Meaning that if you set `$PATH` to `/some/thing`, you're out of luck, but if you set it to `:/some/thing`, or `/some/thing:`, we append or prepend the existing value (and of course you could also do `/foo/bar::/bar/baz`; any empty segment essentially references the existing value).

That's how I've seen it done in other systems where an env var is used to append or prepend to a built-in path, where no existing env var is available to reference.

That's obviously then a slightly proprietary syntax, but OTOH, it's not like people will constantly shoot themselves in the foot by setting `$PATH` to a value on an existing high volume production app without prior testing or anything. Hopefully. :)

dzuelke:
But generally, I think we worry about the `$PATH` case too much. Yes, people will set it. But no, not all the time, and they can trivially use a `.profile` file if they want to customize it. It's a fairly advanced use case.

sclevine:
@hone Updated the RFC with @dzuelke's suggestion. The lifecycle-provided env vars (like `PATH`) are now prepended when they are user-provided. They always precede layer-provided values and platform-provided values.

sclevine:
@nebhale @ekcasey

PR #15, opened 2019-07-24
RFC: Buildpack configuration for distribution

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/buildpack-config-for-dist/text/0000-buildpack-config-for-dist.md)


sclevine:
After discussion at the WG meeting today, I would prefer the inner `buildpack.toml` file to have a flag or different top-level key (e.g., `[module]` instead of `[buildpack]`), so that inner buildpacks can't accidentally be packaged individually. Otherwise, it isn't possible for two or more buildpacks to reliably share files (such as large dependencies).

If we do intend for sub-buildpacks to be packaged individually, then I think we should drop monorepo support (including `path` in `order`) entirely from the spec, and then write our own individual tools to make buildpack monorepos easier. For example, you could write a packaging tool that translated `path` to an ID and version. I don't prefer this option, because I think it's useful for buildpacks to be able to share files.

That said, I'm approving this so that we can move forward. I don't think we can reach complete consensus on this.

text/0000-buildpack-config-for-dist.md
```diff
@@ -0,0 +1,298 @@
...
+[summary]: #summary
+
+This is a proposal to change the schema of `buildpack.toml` to separate the
+concerns of defining a buildpack and managing configuration of buildpacks.
+
+# Motivation
+[motivation]: #motivation
+
+In the approval of the [Distribution Spec](https://github.com/buildpack/rfcs/blob/master/text/0007-spec-distribution.md),
+we landed on a [compromise that no one was really happy with](https://github.com/buildpack/rfcs/pull/11#issuecomment-510483638).
```
sclevine: I disagree with this characterization. I liked the flexibility provided by the original proposal, and the drawbacks mentioned below are not significant for my personal use cases / preferences.

jkutner:
I've removed the `path` key from the `group` table, which means we're eschewing support for monorepo buildpacks for now.

sclevine:
Spec PR updated accordingly: https://github.com/buildpack/spec/pull/53

PR #16, opened 2019-07-24
RFC: Project descriptor

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/project-descriptor/text/0000-project-descriptor.md)

### DIFF
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -32,6 +32,7 @@ Terminology:
 
 * **project**: a repository containing source code for an app, service, function, buildpack or a monorepo containing any combination of those.
 * **image**: the output of a running a buildpack(s) against a project
+* **inline buildpack**: a buildpack that is part of a project's source code
 
 The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The new elements in  `buildpack.toml` support many different use cases including two new top level tables:
 
@@ -249,6 +250,8 @@ version = "0.0.1"
 id = "io.buildpacks.inline"
 version = "0.0.1"
 ```
+
+The `stack` key is optional for inline buildpacks, since they're meant to be used against the stack the project is using.
 
 There is an open question as to whether or not we should simplify the above example so that `[[project.buildpacks]]` can have a pointer to the buildpack defined in `[buildpack]` instead of repeating the `id` and `version`.
 
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -11,6 +11,8 @@
 
 This is a proposal to add new tables and keys to `buildpack.toml` that allow it to contain configuration for apps, services, functions, and buildpacks.
 
+**NOTE:** One of the proposed alternatives uses the filename `project.toml` with the same schema described in this RFC.
+
 # Motivation
 [motivation]: #motivation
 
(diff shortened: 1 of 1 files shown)


text/0000-project-descriptor.md
```diff
@@ -0,0 +1,353 @@
...
+## `[[project.buildpacks]]`
+
+The `[[project.buildpacks]]` array defines the buildpacks that will run against the app:
+
+```toml
+[[project.buildpacks]]
+id = "<buildpack ID (required)>"
+version = "<buildpack version (optional default=latest)>"
+optional = "<bool (optional default=false)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
```
sclevine: Should this be an image ref to a buildpackage? Otherwise, I think we'd need a centralized buildpack registry to resolve all the dependent IDs in the downloaded artifact.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,353 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This proposal merges to previously distinct concepts: projects/apps and buildpacks. However, by both concepts have many similar attributes and combining them enables the use of buildpacks to build buildpacks.
+- The use of a `buildpack.toml` to define an app may not be intuitive to a buildpack author who has already used it on a buildpack.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Name the file `project.toml` and keep the schema described herein.
```
sclevine: I like this option.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,353 @@
...
+- [`app.json`](https://devcenter.heroku.com/articles/app-json-schema)
+- [`.buildpacks`](https://github.com/heroku/heroku-buildpack-multi)
+- [heroku-buildpack-inline](https://github.com/kr/heroku-buildpack-inline)
+- [heroku-buildpack-multi-procfile](https://github.com/heroku/heroku-buildpack-multi-procfile)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Is `path` overriding `uri` and other elements confusing? Should these be the same key, and the execution environment figures out how to handle (i.e. `/user/local` versus `https://example.com`).
+- Do multiple `[[images]]`  result in multiple builds, or are the images derived from a single build?
```
sclevine: Thinking about this, it would be really easy to create multiple images in a single build by allowing buildpacks to add image names to layer metadata.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+- `[metadata]`: (optional) metadata about the repository
+
+Here is an overview of the complete schema:
+
+```toml
+[project]
+id = "<string>"
+name = "<string>"
+version = "<string>"
+authors = ["<string>"]
```
sclevine: Should this be more structured? E.g.,
```
[[authors]]
name = "Bob"
email = "bob@example.com"
```
Not suggesting it should, just wondering.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
+include = ["<string>"]
+exclude = ["<string>"]
+publish = "<boolean>"
```
sclevine: I'm assuming registry-specific fields aren't going to make it into the spec until a registry exists. Correct?

(Also, I'm assuming the bool is not a string.)

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+
+  [[project.buildpacks]]
+  id = "<string>" # may reference [buildpack.id] in this file
+  version = "<string>"
+  optional = "<boolean>" # question: is there actually a use case for this?
+  uri = "<url or path>"
+
+[buildpack] # used for inline buildpack
```
sclevine: Feels a bit weird that the inline buildpack can have a different version than the project, but I'm not sure if there's a nice solution to this.

sclevine: Also, how you do specify multiple inline buildpacks?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+# matches buildpack.toml
+
+[[order]] # used for inline buildpack
+# matches buildpack.toml
+
+[[stacks]] # allowed, but ignored for inline buildpack
```
sclevine: How does an inline buildpack specify stacks? Is it assumed to be compatible with all listed stacks? What if the inline buildpack is compatible with one of the `images` stacks but not another?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+# matches buildpack.toml
+
+[[images]]
+id = "<string>"
+path = "."
+stack = "<string>"
+
+  [images.launch.env]
+  key1 = "<string>"
```
sclevine: I'd prefer:
```
name = "<string>"
value = "<string>"
```
If only to match the k8s convention.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+license = "<string>"
+source = "<url>"
+include = ["<string>"]
+exclude = ["<string>"]
+publish = "<boolean>"
+
+  [[project.buildpacks]]
+  id = "<string>" # may reference [buildpack.id] in this file
+  version = "<string>"
+  optional = "<boolean>" # question: is there actually a use case for this?
```
sclevine: I think we should leave `optional` out. Unless we allow full `order` semantics, `optional` provides incomplete functionality anyways. And I think this list should be something that the user is explicitly deciding.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
+include = ["<string>"]
+exclude = ["<string>"]
+publish = "<boolean>"
+
+  [[project.buildpacks]]
+  id = "<string>" # may reference [buildpack.id] in this file
+  version = "<string>"
```
sclevine: Assuming `version` is optional here?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,401 @@
...
+version = "<string>"
+authors = ["<string>"]
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
+include = ["<string>"]
+exclude = ["<string>"]
+publish = "<boolean>"
+
+  [[project.buildpacks]]
```
sclevine: How would you feel about dropping this list and forcing users to use `[[images]]`?

text/0000-project-descriptor.md
jkutner: i wondered the same, but it could be `"Bob <bob@example.com>"`. i don't have strong feelings either way.

text/0000-project-descriptor.md
jkutner: I'm not sure we need to support that use case. What's the value of decomposing if their already inline/in-repo?

jkutner:
Closing in favor of #25. The inline buildpack (and probably scriptless buildpack) will be reborn in a forthcoming RFC

PR #17, opened 2019-07-24
RFC: Scriptless buildpacks

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/scriptless-buildpacks/text/0000-scriptless-buildpacks.md)


text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,203 @@
...
+cache = "<boolean>"
+launch = "<boolean>"
+build = "<boolean>"
+run = [ "<string>" ]
+
+  [buildpack.build.layers.metadata]
+  key = "<string>" # arbitrary elements that will be put directly into the layer metadata
+
+  [buildpack.build.layers.env]
+  KEY = "<string>" # env vars that will be put into `<layer_dir>/env`
```
sclevine: Any thoughts about this compared to the following?
```toml
name = "<string>"
version = "<string>"
```

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,203 @@
...
+
+  [[buildpack.build.layers.profile]]
+  # each table in this array will be turned into a <layer_dir>/profile.d script
+  name = "<string>"
```
sclevine: I wonder if these really need names. If we named them, e.g., `0.sh`,  `1.sh`, etc, then they would run in the order defined in the file.

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,203 @@
...
+- `[buildpack.build]`: (optional)  contains elements that are used to perform buildpack building.
+
+Both of these new tables are mutually exclusive with `[[buildpack.order]]`.
+
+## `[buildpack.detect]`
+
+```toml
+[buildpack.detect]
+run = ["<string>"] # bash commands
+requires = [ "<string>"] # defines required entries in the build plan
```
sclevine: Can we leave `requires` and `provides` in the original format? Otherwise, you can't request a particular version of a dependency from another buildpack.

sclevine:
I really like this idea!

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,204 @@
...
+  [buildpack.build.layers.metadata]
+  key = "<string>" # arbitrary elements that will be put directly into the layer metadata
+
+  [buildpack.build.layers.env]
+  KEY = "<string>" # env vars that will be put into `<layer_dir>/env`
+
+  [[buildpack.build.layers.profile]]
+  # each table in this array will be turned into a <layer_dir>/profile.d script
+  name = "<string>"
+  script = [ "<string>" ]
```
hone: I don't think these need to be an array. For implementation, wouldn't we just concat these with newlines and write to disk anyways? I propose just using a multline string.

```suggestion
  script = "<string>"
```

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,204 @@
...
+version = "0.0.9"
+
+[buildpack.detect]
+provides = [
+  "jdk"
+]
+
+[[buildpack.build.layers]]
+id = "jdk"
+cache = "true"
```
hone:
```suggestion
cache = true
```

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,204 @@
...
+launch = "true"
```
hone:
```suggestion
launch = true
```

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,204 @@
...
+build = "true"
```
hone:
```suggestion
build = true
```

text/0000-scriptless-buildpacks.md
hone: I assume `name = "<string>"` allows us to have a wider of valid names since it's a string. I do like the ergonomics of `KEY= "<string>"` slightly better, but don't feel too strongly either way.

hone:
I implemented this as an inline buildpack that you can use with `pack`: `pack --builder hone/scriptless`.

hone:
From the WG today, I'm going to implement a buildpack build env var, so people can use this as a non-inline buildpack.

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,204 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- `Dockerfile` is similar in the use case it addresses.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How do we merge the scriptless buildpack with scripts in `bin/` if they are eventually created?
+- What is the migrate path for a developer who has created a fairly complex scriptless buildpack, and now wants to port it to a `bin/build`.
```
josegonzalez: It would be cool to just have a way to `generate` the `bin/build`/`bin/detect` binaries from this setup.

text/0000-scriptless-buildpacks.md
jkutner: I think after some discussion in the WG meeting, we agreed that whatever this is it should match every other `env` section in our TOML files.

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,211 @@
...
+- `[buildpack.detect]`: (optional) contains elements that are used to perform buildpack detection.
+- `[buildpack.build]`: (optional)  contains elements that are used to perform buildpack building.
+
+Both of these new tables are mutually exclusive with `[[buildpack.order]]`.
+
+## `[buildpack.detect]`
+
+```toml
+[buildpack.detect]
+run = ["<string>"] # bash commands
```
sclevine: How about `inline` instead of `run`? Then we could support pointing to a script at a later date.

sclevine: Also curious if you feel like the list is better than:
```toml
inline = """
# bash script
"""
```

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,211 @@
...
+provides = [
+  "jdk"
+]
+
+[[buildpack.build.layers]]
+id = "jdk"
+cache = true
+launch = true
+build = true
+run = [
```
sclevine: How do you write metadata from here? How do you read the build plan?

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,211 @@
...
+
+  [buildpack.build.layers.metadata]
+  key = "<string>" # arbitrary elements that will be put directly into the layer metadata
+
+  [buildpack.build.layers.env]
+  KEY = "<string>" # env vars that will be put into `<layer_dir>/env`
+
+  [[buildpack.build.layers.profile]]
+  # each table in this array will be turned into a <layer_dir>/profile.d script
+  script = "<string>"
```
sclevine: Maybe the format could match the `layers` section? (i.e., `run`)

text/0000-scriptless-buildpacks.md
jkutner: i would support positional args like `$1` below this

text/0000-scriptless-buildpacks.md
```diff
@@ -0,0 +1,211 @@
...
+If `buildpack.detect.run` is not defined, and either `buildpack.detect` or `buildpack.build` are defined, then detection will always pass. Otherwise, the exit code of the last `run` command is used.
+
+## `[buildpack.build]`
+
+```toml
+[buildpack.build]
+run = [ "<string>" ]
+
+[[buildpack.build.layers]]
+id = "<string>"
```
sclevine: Recommend: `name` (layers don't have IDs)

jkutner:
@sclevine @hone @nebhale @ekcasey I've updated the RFC based on our discussion and the comments. I'm pretty keen on some variation of this: https://github.com/buildpack/rfcs/blob/scriptless-buildpacks/text/0000-scriptless-buildpacks.md#restricting-layer-access-in-scripts

jkutner:
There was a question along the lines:

> How do we handle logic that needs to access more than one layer?

If we use the mechanism proposed (where the layer script get `$1` as normal) and add a way to force layer regenerate, then I think this is solved.

thephw:
Related at least in terms of developer UX -- single file entry points really helps folks on-board and I like this idea a lot.

Story time. A former project that Deft worked on had about seven different services to coordinate with different lifecycles for each one. Each service was in a subdirectory of a monorepo and we had a single `Makefile` at the top directory with phony Makefile targets to orchestrate commands. Each service usually had some similar commands `web-test` `api-test` `web-up` `api-up` `web-logs` `api-logs` etc.

Even if everyone on the team didn't completely understand how the makefile targets worked most folks had no trouble consuming the services from the single entry point. That gave new folks on the team a leg up when trying to understand the workflow. We kind of tried it out expecting it to fall apart and by in large it held together better than I expected. Although the Makefile got pretty long.

Some of the positives from our experience:
- All common tasks were in a single file
- Targets in a Makefile behave very predictably and the behavior was well defined
- Possible to combine targets via target dependencies (`build: web-build api-build db-build`)
- Very easy to add a new target
- Possible to dependency chain make targets `make detect build`
- Generators can append to Makefiles
- Make was available nearly everywhere for everyone

Some downsides from our experience:
- Pipes can confuse exist codes and sometimes break in weird ways
- More complicated scripts or non-shell scripts typically needed to run as commands
- Make errors can be obtuse to read or understand
- Hard to test reliably

For buildpacks, I mostly think `make` offers some interesting corollaries to consider.

For funsies, I pseudo-coded a `Makefile` from our scaffolding on the prior project to show what it'd look like.

```
define USAGE
detect - determines whether buildpack should be applied
build - executes buildpack logic
inspect - shows metadata about your buildpack
usage - shows this prompt
endef
export USAGE

.DEFAULT_GOAL := usage

.SHELL: /bin/bash
.SHELLFLAGS = -o pipefail

# Everything below is a script target
.PHONY: $(MAKECMDGOALS)

detect:
	test -f Gemfile

build:
	./bin/build

inspect:
	./bin/explain_config_to_mere_mortals< buildpack.toml

usage:
	@echo "$$USAGE"
```

jkutner:
Closing because this has dropped in priority

PR #18, opened 2019-08-04
RFC: expose layer metadata during detect

sclevine:
[Readable](https://github.com/buildpack/rfcs/blob/e7504825dc8b2c4f77994e94f603c14ed100fc41/text/0000-expose-layer-metadata-for-detect.md).

### DIFF
--- a/text/0000-expose-layer-metadata-for-detect.md
+++ b/text/0000-expose-layer-metadata-for-detect.md
@@ -1,0 +1,40 @@
+# Meta
+[meta]: #meta
+- Name: Expose Layer Metadata During Detection Phase
+- Start Date: 2019-08-03
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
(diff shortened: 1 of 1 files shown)


text/0000-expose-layer-metadata-for-detect.md
```diff
@@ -0,0 +1,40 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+We extend `/bin/detect` to match `/bin/build`:
+```bash
+/bin/detect <layers> <platform> <plan>
+```
+
+We ensure that layer metadata is restored before detection and cache layers are restored after detection.
```
ekcasey: This would appear to conflict with https://github.com/buildpack/rfcs/pull/21. This RFC currentlty implies `analyze -> detect -> restore` instead of `detect -> analyze/restore`.

sclevine:
Closing this for now.

PR #19, opened 2019-08-05
RFC: API versioning

ameyer-pivotal:
[Readable](https://github.com/ameyer-pivotal/rfcs/blob/api-versions/0000-api-versions.md)

### DIFF
--- a/0000-api-versions.md
+++ b/0000-api-versions.md
@@ -21,23 +21,30 @@ To enable logic and/or validation based on `api`. For instance, we could keep us
 
 There would be two pieces to this:
 
-1. A reference to the current versions of all APIs placed somewhere in the spec (location within spec TBA). For instance, a Buildpack API and a Platform API.
+1. A reference to the current versions of all APIs placed in the spec. Currently, we would version the Buildpack API (buildpack-to-lifecycle contract) and the Platform API (lifecycle-to-platform contract).
 2. An `api` field in TOML files for entities that are dependent on a specific version of an API. For instance, `buildpack.toml` would have an `api` to represent the version of the buildpack API to which the buildpack conforms.
 
 Example `api` field for `buildpack.toml`:
 
 ```toml
-api = "<version>"
+api = "<major>.<minor>"
 
 [buildpack]
 id = "<string>"
 # ...
 ```
 
+The format of a version would be `<major>.<minor>` or `<major>` (with a `minor` of `0` implied). A change to the `major` version of an API would indicate a non-backwards-compatible change, while a change to the `minor` version would indicate availability of one or more opt-in features.
+
+A **missing or empty** `api` key would indicate `1.0` (the current version of the Buildpack API and Platform API, as of this writing). The next breaking change to either of the current APIs would bump that API version to `2.0`.
+
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Given the current state of the project, what would we deem to be the first "APIs?" Potentially there are two to begin with: a Buildpack API and a Platform API, but more may be warranted in the future.
-- What format should the version take? Integer? Semver?
-- Given the answer to the above, what version should we begin with for each API?
+- Where in the spec should we document current API versions?
 - What about files that are not spec'd, but rather are related to the `pack` implementation (e.g. `builder.toml`)?
+- Will it be confusing that the spec is already titled "Buildpack API v3 - Specification", while the Buildpack API version to be documented inside of it references only a subset of the spec (which begins at `1.0`, not `3.0`)?
+
+# Related RFCs
+[related-rfcs]: #related-rfcs
+[RFC: Lifecycle descriptor](https://github.com/buildpack/rfcs/pull/20)
--- a/0000-api-versions.md
+++ b/0000-api-versions.md
@@ -1,0 +1,43 @@
+# Meta
+[meta]: #meta
+- Name: API Versions
+- Start Date: 2019-08-05
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Begin versioning the various APIs that the spec describes. Then require these versions be specified in related TOML files, under an `api` field.
+
+# Motivation
+[motivation]: #motivation
+
+To enable logic and/or validation based on `api`. For instance, we could keep users from attempting to use a combination of incompatible buildpacks and lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+There would be two pieces to this:
+
+1. A reference to the current versions of all APIs placed somewhere in the spec (location within spec TBA). For instance, a Buildpack API and a Platform API.
+2. An `api` field in TOML files for entities that are dependent on a specific version of an API. For instance, `buildpack.toml` would have an `api` to represent the version of the buildpack API to which the buildpack conforms.
+
+Example `api` field for `buildpack.toml`:
+
+```toml
+api = "<version>"
+
+[buildpack]
+id = "<string>"
+# ...
+```
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Given the current state of the project, what would we deem to be the first "APIs?" Potentially there are two to begin with: a Buildpack API and a Platform API, but more may be warranted in the future.
+- What format should the version take? Integer? Semver?
+- Given the answer to the above, what version should we begin with for each API?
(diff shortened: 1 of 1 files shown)


0000-api-versions.md
```diff
@@ -0,0 +1,33 @@
...
+There would be two pieces to this:
+
+1. A reference to the current versions of all APIs placed somewhere in the spec (location within spec TBA). For instance, a Buildpack API and a Platform API.
+2. An `api` field in TOML files for entities that are dependent on a specific version of an API. For instance, `buildpack.toml` would have an `api` to represent the version of the buildpack API to which the buildpack conforms.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Given the current state of the project, what would we deem to be the first "APIs?" Potentially there are two to begin with: a Buildpack API and a Platform API, but more may be warranted in the future.
+- What format should the version take? Integer? Semver?
```
jkutner: Should the version match the spec version?

0000-api-versions.md
```diff
@@ -0,0 +1,33 @@
...
+
+To enable logic and/or validation based on `api`. For instance, we could keep users from attempting to use a combination of incompatible buildpacks and lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+There would be two pieces to this:
+
+1. A reference to the current versions of all APIs placed somewhere in the spec (location within spec TBA). For instance, a Buildpack API and a Platform API.
+2. An `api` field in TOML files for entities that are dependent on a specific version of an API. For instance, `buildpack.toml` would have an `api` to represent the version of the buildpack API to which the buildpack conforms.
```
jkutner: can you add an example? I think it would look like:

```
api = "<version>"

[buildpack]
id = "<string>"
...
```

0000-api-versions.md
ameyer-pivotal: Not necessarily, since they will be versioned independently of the spec

ameyer-pivotal: Or do you mean just the format?

0000-api-versions.md
```diff
@@ -0,0 +1,43 @@
...
+[motivation]: #motivation
+
+To enable logic and/or validation based on `api`. For instance, we could keep users from attempting to use a combination of incompatible buildpacks and lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+There would be two pieces to this:
+
+1. A reference to the current versions of all APIs placed somewhere in the spec (location within spec TBA). For instance, a Buildpack API and a Platform API.
```
hone: would this mean there's a distribution API (thinking that every file in the spec repo is a API version then?)

ameyer-pivotal: Potentially, though I'm not 100% sure how we want to break things down into APIs (by high-level sub-spec could be one option).

ekcasey:
I updated the lifecycle descriptor RFC to match the `<major>.<minor>` api version format that we decided upon for this RFC. 

https://github.com/buildpack/rfcs/pull/20

0000-api-versions.md
```diff
@@ -0,0 +1,50 @@
...
+
+The format of a version would be `<major>.<minor>` or `<major>` (with a `minor` of `0` implied). A change to the `major` version of an API would indicate a non-backwards-compatible change, while a change to the `minor` version would indicate availability of one or more opt-in features.
+
+A **missing or empty** `api` key would indicate `1.0` (the current version of the Buildpack API and Platform API, as of this writing). The next breaking change to either of the current APIs would bump that API version to `2.0`.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Where in the spec should we document current API versions?
+- What about files that are not spec'd, but rather are related to the `pack` implementation (e.g. `builder.toml`)?
```
hone: We should probably version them, but that'd be specific to `pack`. Dose anything depend on `pack` specific files?

0000-api-versions.md
```diff
@@ -0,0 +1,50 @@
...
+- Will it be confusing that the spec is already titled "Buildpack API v3 - Specification", while the Buildpack API version to be documented inside of it references only a subset of the spec (which begins at `1.0`, not `3.0`)?
```
hone: We could just use `3.0`.

0000-api-versions.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+The format of a version would be `<major>.<minor>` or `<major>` (with a `minor` of `0` implied). A change to the `major` version of an API would indicate a non-backwards-compatible change, while a change to the `minor` version would indicate availability of one or more opt-in features.
+
+A **missing or empty** `api` key would indicate `1.0` (the current version of the Buildpack API and Platform API, as of this writing). The next breaking change to either of the current APIs would bump that API version to `2.0`.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Where in the spec should we document current API versions?
```
hone: Does it make sense to just put it at the top of the respective `*.md` files?

0000-api-versions.md
```diff
@@ -0,0 +1,50 @@
...
+[motivation]: #motivation
+
+To enable logic and/or validation based on `api`. For instance, we could keep users from attempting to use a combination of incompatible buildpacks and lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+There would be two pieces to this:
+
+1. A reference to the current versions of all APIs placed in the spec. Currently, we would version the Buildpack API (buildpack-to-lifecycle contract) and the Platform API (lifecycle-to-platform contract).
```
hone: Does it make sense that there's a distribution version? If we're going to break it down does a version per markdown file make sense?

0000-api-versions.md
hone: Ah, I see this breaks the missing/empty case.

0000-api-versions.md
ekcasey: Nothing reads `builder.toml` other than `pack`. I would say we could add a schema version for these files if we wish later, but it isn't necessary yet.

0000-api-versions.md
ekcasey: What's interesting about this question is that the next API version we would support doesn't match the spec exactly at any commit. Example: we could put API version 2.0 at the top of the `buildpack.md` but that would seem to imply support for `store.toml` which isn't yet implemented. We could describe what 2.0 entails at the top of the spec and say `2.1` will add store.toml?

0000-api-versions.md
ekcasey: I could see that making sense eventually. Do we want to wait until/if we have a breaking change in the distribution spec?

hone:
I think there are some open questions that are still interesting, but overall the ideas in this RFC are great.

0000-api-versions.md
hone: yeah, I think that makes sense. In general, leaning on the side of more documentation for things is probably better.

0000-api-versions.md
hone: I wonder if we just need to version the spec in a branch (or tag) that's just accurate for that version. It's why @jkutner and I were talking about just 1 version of the API. It'd make that logistically easier.

hone:
An open question from me is, what warrants a version in the spec?

0000-api-versions.md
sclevine: Regardless of how we version the spec, there are two contracts that need separate versions. Otherwise, if we make a breaking change for platform maintainers that doesn't affect buildpacks, we would be forced to push a new major version on buildpack authors for no reason.

jkutner: @sclevine i wouldn't say it's for no reason; it's so only one version has to be managed. given how infrequently we will have breaking changes (especially after finalizing the spec) I suspect it won't be that big of a burden. That said, I'm fine if we decide on separate versions.

jkutner:
Going on my comment https://github.com/buildpack/rfcs/pull/20/files#r313872165 -- people may have to bump both platform buildpack api versions even if they don't change the contents of those files in order to use a particular lifecycle version (unless we support version ranges).

hone:
Looks like sem ver gets us limited (but maybe enough) version range support.

ameyer-pivotal:
@hone Good point on the implied range. I'll update the RFC to lay that out too.

0000-api-versions.md
nebhale: I think it should be mentioned as we expect to do this in the future.  However, with the no value implies 1.0, no one would be required to do anything about it today.

PR #20, opened 2019-08-05
RFC: Lifecycle descriptor

ekcasey:

[Readable](https://github.com/buildpack/rfcs/blob/lifecycle-descriptor/text/0000-lifecycle-descriptor.md)

NOTE: depends on #19

### DIFF
--- a/text/0000-lifecycle-descriptor.md
+++ b/text/0000-lifecycle-descriptor.md
@@ -25,12 +25,14 @@ Lifecycle can include a `lifecycle.toml` at it's root. Initially, the only suppo
 
 ```toml
 [api]
- platform = "1"
- buildpack = "1"
+ platform = "<major>.<minor>"
+ buildpack = "<major>.<minor>"
 
 [lifecycle]
-  version = "v0.4.0"
+  version = "<major>.<minor>.<patch>"
 ```
+
+The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more opt-in features.
 
 When `lifeycle.toml` exists but either of the `api.x` fields is omitted, `pack` will be assume that the lifecycle implements the latest API known to that version of `pack`. When `lifecycle.toml` is omitted, `pack` will initially warn users and assume the lifecycle implements the oldest known api version, for backwards compatibility. Eventually `pack` may require the presence of `lifecycle.toml`.
 
@@ -39,16 +41,17 @@ When `lifeycle.toml` exists but either of the `api.x` fields is omitted, `pack` 
 
 When `pack create-builder` creates a builder. The lifecycle API versions will be added to the `io.buildpacks.builder.metadata` label.
 
+Example:
 ```json
 {
 ...
-"lifecycle": {"version":  "v0.4.0", "api": {"platform":  "1", "buildpack":  "1"}}
+"lifecycle": {"version":  "0.4.0", "api": {"platform":  "1.0", "buildpack":  "1.1"}}
 }
 ```
 
 `pack` would ensure that the `api.buildpack` version matches the `api` version provided by all of the buildpacks that are added to the given builder.
 
-When `pack build` executes, it will check `lifecycle.api.platform` and it will either execute the correct commands for the given platform API, or fail if the platform API version is not supported. For example, if in the future we combine restorer/analyzer and cacher/exporter, then the platform API version of the next lifecycle release would be incremented. This will allow `pack` to know which binaries to execute with which flags.
+When `pack build` executes, it will check `api.platform` and it will either execute the correct commands for the given platform API, or fail if the platform API version is not supported. For example, if in the future we combine restorer/analyzer and cacher/exporter, then the platform API version of the next lifecycle release would be incremented. This will allow `pack` to know which binaries to execute with which flags.
 
 When `pack build` is run with the `--buildpack` flag, `pack` may use the buildpack api version from the builder metadata to determine whether the buildpack supplied with the `--buildpack` flag is compatible with the builder's lifecycle.
 
--- a/text/0000-lifecycle-descriptor.md
+++ b/text/0000-lifecycle-descriptor.md
@@ -34,7 +34,7 @@ Lifecycle can include a `lifecycle.toml` at it's root. Initially, the only suppo
 
 The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more additional features. For example, a lifecycle that specifies version 1.2 of the buildpack api supports buildpacks that specify versions 1.0, 1.1, 1.2, but not those that specify 1.3.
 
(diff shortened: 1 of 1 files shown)


text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,66 @@
...
+[what-it-is]: #what-it-is
+
+Lifecycle can include a `lifecycle.toml` at it's root. Initially, the only supported fields will be the `api.platform`, `api.buildpack` and `lifecycle.version`.
+
+```toml
+[api]
+ platform = "1"
+ buildpack = "1"
+
+[lifecycle]
```
ameyer-pivotal: Wonder if a top-level `lifecycle` field would be necessary. I guess we do the same for `buildpack.toml`, so it's probably fine.

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+When `pack create-builder` creates a builder. The lifecycle API versions will be added to the `io.buildpacks.builder.metadata` label.
+
+Example:
+```json
+{
+...
+"lifecycle": {"version":  "v0.4.0", "api": {"platform":  "1.0", "buildpack":  "1.1"}}
```
jromero: version in JSON format should be solely 0.4.0.

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,66 @@
...
+
+Lifecycle can include a `lifecycle.toml` at it's root. Initially, the only supported fields will be the `api.platform`, `api.buildpack` and `lifecycle.version`.
+
+```toml
+[api]
+ platform = "1"
+ buildpack = "1"
+
+[lifecycle]
+  version = "v0.4.0"
```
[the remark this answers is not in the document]
ekcasey: fixed

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+
+Example:
+```json
+{
+...
+"lifecycle": {"version":  "0.4.0", "api": {"platform":  "1.0", "buildpack":  "1.1"}}
+}
+```
+
+`pack` would ensure that the `api.buildpack` version matches the `api` version provided by all of the buildpacks that are added to the given builder.
```
jromero: Might be worth noting that this is dependent on #19 otherwise it's unclear how buildpacks provide an `api` version.

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+
+When `pack build` executes, it will check `lifecycle.api.platform` and it will either execute the correct commands for the given platform API, or fail if the platform API version is not supported. For example, if in the future we combine restorer/analyzer and cacher/exporter, then the platform API version of the next lifecycle release would be incremented. This will allow `pack` to know which binaries to execute with which flags.
```
jromero: What is `lifecycle.api.platform` referencing? Should it be "the lifecyle's `api.platform`"?

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Lifecycle can include a `lifecycle.toml` at it's root. Initially, the only supported fields will be the `api.platform`, `api.buildpack` and `lifecycle.version`.
+
+```toml
+[api]
+ platform = "<major>.<minor>"
+ buildpack = "<major>.<minor>"
```
jkutner: It does seem that having a single `api` version in https://github.com/buildpack/rfcs/pull/19 would make defining support in lifecycle a better experience. Is it possible that a user wants a combination of two versions (platform and buildpack) where there is not lifecycle version that matches?

I guess my point is: the purpose of two api versions is to prevent people from having to unnecessarily bump both when only one has changed. But unless we plan on supporting version ranges, people are going to have to bump both anyways.

ekcasey: The reason I think defining two version is important is that a breaking change in the platform/lifecycle interface doesn't necessarily imply a breaking change in the lifecycle/buildpack interface and vice versa. If we only had one `api` version, something like change a `detector` flag name would require every buildpack author to bump the `api` version in their `buildpack.toml` to indicate compatibility with the new lifecycle. That doesn't seem right to me.

text/0000-lifecycle-descriptor.md
ekcasey: I added a note to the PR description

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+
+[lifecycle]
+  version = "<major>.<minor>.<patch>"
+```
+
+The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more opt-in features.
```
hone: I think I understand where this is going, but can we just add another line from what we clarified in the working group today that says: supporting api version `1.2` means supporting `1.0`, `1.1`, `1.2`, but not `1.3`. This way we get built in version ranges based off of sem ver which handles most of the cases I think we'd want.

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+```toml
+[api]
+ platform = "<major>.<minor>"
+ buildpack = "<major>.<minor>"
+
+[lifecycle]
+  version = "<major>.<minor>.<patch>"
+```
+
+The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more additional features. For example, a lifecycle that specifies version 1.2 of the buildpack api supports buildpacks that specify versions 1.0, 1.1, 1.2, but not those the specify 1.3.
```
hone:
```suggestion
The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more additional features. For example, a lifecycle that specifies version 1.2 of the buildpack api supports buildpacks that specify versions 1.0, 1.1, 1.2, but not those that specify 1.3.
```

hone:
After the WG discussion today, I'm 👍 on the two version split, but I imagine for buildpack developers they'll mostly be concerned about the buildpack/lifecycle contract. We can figure out a way to market the versions that matter to the right audiences. As an operator, I imagine I'll need to care about both versions.

text/0000-lifecycle-descriptor.md
```diff
@@ -0,0 +1,69 @@
...
+ platform = "<major>.<minor>"
+ buildpack = "<major>.<minor>"
+
+[lifecycle]
+  version = "<major>.<minor>.<patch>"
+```
+
+The format of a api version would be <major>.<minor> or <major> (with a minor of 0 implied). A change to the major version of an API would indicate a non-backwards-compatible change, while a change to the minor version would indicate availability of one or more additional features. For example, a lifecycle that specifies version 1.2 of the buildpack api supports buildpacks that specify versions 1.0, 1.1, 1.2, but not those that specify 1.3.
+
+When `lifecycle.toml` exists, both of the`api.x` fields are required. When `lifecycle.toml` is omitted, `pack` will warn users and assume the lifecycle implements the oldest known api version, for backwards compatibility. Eventually `pack` may require the presence of `lifecycle.toml`.
```
nebhale: 👍

PR #21, opened 2019-08-06
Combined restorer/analyzer and cacher/exporter phases.

squee1945:
[Readable version](https://github.com/buildpack/rfcs/blob/ba7000a90b94a7a4c577d31e9a95d4e84bc4f7ba/text/0000-combined-restorer-analyzer-phases.md)


text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Combined `restorer` phase.
+
+The spec will formalize the (combined) `restorer` phase like this:
+
+```
+/lifecycle/restorer \
+  -group=[(required) group.toml file] \
+  -layers=[(required) layers directory] \
+  -image=[(optional) repository url for existing application image] \
```
jkutner: Is this equivalent to the positional argument, dubbed `repoName`, in `/lifecycle/analyzer`?

jkutner: I was wondering why this was optional, but I think maybe you're not trying to be precise here (which is fine).

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+## Combined `restorer` phase.
+
+The spec will formalize the (combined) `restorer` phase like this:
+
+```
+/lifecycle/restorer \
+  -group=[(required) group.toml file] \
```
ekcasey: Currently neither `-group` nor `-layers` are required. I think we should leave them as optional and keep the current defaulting behavior (first to an env var if it is present, then to a hardcoded value).

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+  -layers=[(required) layers directory] \
+  -image=[(optional) repository url for existing application image] \
+  -cache=[(optional) local directory, or repository url for cache image]
```
ekcasey: I think we should we add a flag for cache type or replace this with two optional flags `cache-volume` and `cache-image`. I worry that at some point a type of caching may be introduced that would prevent us from determining the type of caching from the format of cache destination value.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+The previous `/lifecycle/analyzer` phase will be removed from the spec.
+
+
+## Combined `exporter` phase.
+
+The spec will formalize the (combined) `exporter` phase like this:
+
+```
+/lifecycle/exporter \
+  -app=[(required) the application source directory] \
```
ekcasey: Same comment as for the restorer proposal. I think image tag reference(s) should by the only required input.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+  -base=[(required) repository url for the base image to build on top of] \
+  -group=[(required) group.toml file] \
+  -layers=[(required) layers directory] \
+  -image=[(required) repository url for new application image] \
+  -cache=[(optional) local directory, or repository url for cache image]
```
ekcasey: same comment as above regarding disambiguating cache types

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+
+The spec will formalize the (combined) `exporter` phase like this:
+
+```
+/lifecycle/exporter \
+  -app=[(required) the application source directory] \
+  -base=[(required) repository url for the base image to build on top of] \
+  -group=[(required) group.toml file] \
+  -layers=[(required) layers directory] \
+  -image=[(required) repository url for new application image] \
```
ekcasey: We should perhaps call out in the spec that this value must be a tag reference when it is passed to exporter, where the analogous field in the `restorer` can refer to a specific concrete image (either a digest reference or an image ID in the daemon case)

ekcasey: Also, the current reference lifecycle supports exporting to multiple tags provided they are on the same registry. We might want to spec that behavior.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Combined `exporter` phase.
+
+The spec will formalize the (combined) `exporter` phase like this:
+
+```
+/lifecycle/exporter \
+  -app=[(required) the application source directory] \
+  -base=[(required) repository url for the base image to build on top of] \
```
ekcasey: I think `run-image` might be a better name b/c it has a specific technical meaning in the CNB ecosystem and `run-image`s and `run-image-mirrors` are concepts referenced in several places by the lifecycle

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,130 @@
...
+
+A parallel optimization potential exists when combining the `cacher` and
+`exporter` phases.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Combined `restorer` phase.
+
+The spec will formalize the (combined) `restorer` phase like this:
```
ekcasey: I'm conflicted about going with `restorer` rather than `analyzer` given that `restorer` was previously the optional phase. `analyzer` makes more sense historically, although `restorer` might be more semantically meaningful.

text/0000-combined-restorer-analyzer-phases.md
jkutner: Alternatively, we could provide two `exporter` implementations (one that creates cache image and one for volumes). It might be useful to implement these to demonstrate who the exporter can be extended to support other caching mechanisms.

text/0000-combined-restorer-analyzer-phases.md
squee1945: I don't disagree, but I wanted to take the opportunity to suggest to a more applicable semantic.

I don't have a strong opinion here.

text/0000-combined-restorer-analyzer-phases.md
squee1945: Yes, this was the item named `repoName` in the lifecycle reference implementation.

I wasn't certain about the reasoning behind choosing a positional arg over a flag - do you know?

I was trying to be accurate with required/optional, so please call these out if they are incorrect.

I set it optional because I can imagine a case where it is not possible to reference a previous application image (but you still might be restoring cache layers). Thoughts?

text/0000-combined-restorer-analyzer-phases.md
squee1945: Acknowledged. Unless there are opposing opinions, I will change to mutually exclusive flags `-cache-volume` and `-cache-image`.

text/0000-combined-restorer-analyzer-phases.md
squee1945: I wasn't intending on removing the "env var overridable by flag" behavior of the current lifecycle implementation. I will add a section to formalize it.

But, even with env var, are `-group` and `-layers` _completely_ optional? E.g., is it valid to have no `CNB_LAYERS_DIR` and no `-layers` flag?

text/0000-combined-restorer-analyzer-phases.md
squee1945: Not opposed to different exporter binaries, though I'm not sure how one would structure that in the GitHub repo.

text/0000-combined-restorer-analyzer-phases.md
squee1945: Opinions on (1) multiple `-image` tags, (2) delimited (e.g., comma-separated) `-image` tag, or (3) a series of positional args?

text/0000-combined-restorer-analyzer-phases.md
ekcasey: That reasoning makes sense to me. Right now we will correctly handle an image tag reference that doesn't actually point at a previous image. I think it similarly makes sense to handle an omitted `-image` flag, although the cases where I would actually want to do that aren't as obvious to me.

Thinking about the inverse case, I wonder if we should also specify the `skip-layers` flag in the case that a user wants to `analyze` in order to restore [store.toml](https://github.com/buildpack/spec/blob/f08181f3e8e81e76c9d3758c0ed58303805ea8e0/buildpack.md#store.toml-toml) (when we support it) or take advantage of performance efficiencies during export, but doesn't want to restore cached layers or layer metadata (e.g `pack build --clear-cache`). I know we don't always specify all the flags but this one seems important to require, so the spec can be fully implemented.

text/0000-combined-restorer-analyzer-phases.md
ekcasey: When there is no env var the lifecycle falls back to a hardcoded default path https://github.com/buildpack/lifecycle/blob/master/cmd/cmd.go#L13

text/0000-combined-restorer-analyzer-phases.md
ekcasey: I vote for option 3, but I'm open to option 1 if you want to make a case for it.

In my mind `<image>` makes sense as the only positional argument because it feels different than the other parameters. The other flags provide optional ways to configure `export`. Image is "the thing you are exporting" so it takes a different/primary position in the usage.

But I suspect someone could make an equally valid argument for the opposite position.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,162 @@
...
+
+### `/lifecycle/analyzer`
+
+| Flag            | Required | Env var            | Default        | Description |
+| --------------- | -------- | ------------------ | -------------- | ------------|
+| `-group`        | optional | `CNB_GROUP_PATH`   | `./group.toml` | group.toml file |
+| `-layers`       | optional | `CNB_LAYERS_DIR`   | `/layers`      | layers directory |
+| `-platform`     | optional | `CNB_PLATFORM_DIR` | `/platform`    | platform directory |
+| `-image`        | optional | -                  | -              | repository url for existing application image |
+| `-cache-volume` | optional | -                  | -              | local directory, may not be used with `-cache-image` |
```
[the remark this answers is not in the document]
squee1945: I imagine the analyzer inspecting metadata in the manifest of both the previous application image and the cache image to determine which layers of each should be pulled by the restorer. E.g., in the current implementation, it seems like the full cache image is pulled down in restorer only to have some layers later invalidated by the analyzer (at least I think this is happening?) - this is exactly the sort of redundant work that I'm proposing a reversed analyzer/restorer ordering would avoid.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,162 @@
...
+The spec will formalize the `restorer` phase as follows:
+
+### `/lifecycle/restorer`
+
+| Flag            | Required | Env var            | Default        | Description |
+| --------------- | -------- | ------------------ | -------------- | ------------|
+| `-group`        | optional | `CNB_GROUP_PATH`   | `./group.toml` | group.toml file |
+| `-layers`       | optional | `CNB_LAYERS_DIR`   | `/layers`      | layers directory |
+| `-platform`     | optional | `CNB_PLATFORM_DIR` | `/platform`    | platform directory |
+| `-image`        | optional | -                  | -              | repository url for existing application image |
```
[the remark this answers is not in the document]
squee1945: Maybe we need a stronger definition of what these phases do. Based on naming alone, I would expect analyzer to work over image metadata and restorer to actually pull layers. Is that not correct?

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,162 @@
...
+| `-cache-image`  | optional | -                  | -              | cache image repository, may not be used with `-cache-volume` |
+
+The `restorer` phase is responsible to make available images and layers from
+previous builds to improve the efficiency of the current build. It will
+typically use information assembled by the `analyzer` phase to do so.
+
+### Why two phases, and not one?
+
+This proposal envisions a future extension to the spec where the buildpack
+itself can be involved in analysis, possible via an optional `/bin/analyze`
```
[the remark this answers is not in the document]
squee1945: Yes, @sclevine mentioned same. So it seems we'd need a new phase to pull off the `/bin/analyze` plan. Though `/bin/analyze` is a great name to use... any suggestions what this new step might be called? An example of what I (a buildpack dev) would like to do in this phase: hash my package-lock.json file and determine if a cache layer needed to be pulled at all.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,162 @@
...
+The spec will formalize the `analyzer` phase as follows:
+
+### `/lifecycle/analyzer`
+
+| Flag            | Required | Env var            | Default        | Description |
+| --------------- | -------- | ------------------ | -------------- | ------------|
+| `-group`        | optional | `CNB_GROUP_PATH`   | `./group.toml` | group.toml file |
+| `-layers`       | optional | `CNB_LAYERS_DIR`   | `/layers`      | layers directory |
+| `-platform`     | optional | `CNB_PLATFORM_DIR` | `/platform`    | platform directory |
+| `-image`        | optional | -                  | -              | repository url for existing application image |
```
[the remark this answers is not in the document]
squee1945: Changed throughout.

squee1945:
@sclevine @ekcasey I've removed the old file now; there is only a single file remaining. PTAL.

text/0000-combined-restorer-analyzer-phases.md
```diff
@@ -0,0 +1,162 @@
...
+phase called `exporter`.
+
+# Motivation
+[motivation]: #motivation
+
+In the current lifecycle reference implementation, the `restorer` phase
+fetches cached layers from a previous cache image. These layers can be
+arbitrarily large and, when using a remote repository, can represent a large
+amount of overhead. The `analyzer` phase runs after `restorer` and it
+may deterine that particular cache layers are no longer valid and remove them
```
hone:
```suggestion
may determine that particular cache layers are no longer valid and remove them
```

hone:
@squee1945 can you do the DCO sign-off?

squee1945:
@hone DCO sign-off is complete

PR #22, opened 2019-08-06
Service Bindings

nebhale:
A proposal for concretely defining how Service Binding information is exposed to buildpacks and applications.

### DIFF
--- a/text/0000-service-binding.md
+++ b/text/0000-service-binding.md
@@ -1,0 +1,155 @@
+# Meta
+[meta]: #meta
+- Name: Service Binding
+- Start Date: 2019-08-06
+- CNB Pull Request:
+- CNB Issue:
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+One of the common uses of buildpacks is to configure applications to run against remote "services" (a useful name, with perhaps too much baggage) such as APMs and RDBMSs.  In order to do this, the buildpacks need to know about the credentials (e.g. endpoint, username, password) required to connect to those services.  In order to make buildpacks portable and generically useful, these service credentials should be exposed in a standard way on all CNB-compliant platforms.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the specification describes a `CNB_SERVICES` environment variable, but does not describe what the payload should be.  This was always meant as a placeholder and as we approach 1.0, it should be completed with a more concrete description of where service bindings should be found and what their payloads should be.
+
+A complete service binding specification should enable buildpacks to determine what functionality an application developer would like to contribute (binding signals intent) and automatically do some or all of the following (not complete):
+
+  * Provision a binary agent at build time
+  * Configure an integration at runtime
+  * Scan incoming code for service vulnerabilities
+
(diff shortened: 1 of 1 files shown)


nebhale:
Summoning @markfisher and @scothis for their insights into Kubernetes idioms.

scothis:
Kubernetes secrets can be mounted as a single directory, with each key in the secret representing a file in that directory. Having the metadata for a single service be split between directories will make it harder to mange.

Likewise, structuring the service binding based on an attribute of the binding, other than name, will make it harder to manage as the tool creating the file system would need deep knowledge of the service.

I'd favor a form that puts all of the content for a single service into a single directory, and devolves the service.toml into individual files in that directory:

```plain
 platform
 └── services
     ├── primary-db
     │   ├── endpoint
     │   ├── kind
     │   ├── password
     │   ├── provider
     │   ├── tags
     │   └── username
     └── secondary-db
         ├── endpoint
         ├── kind
         ├── password
         ├── provider
         ├── tags
         └── username
 ```

nebhale:
@scothis The problem I see with that design (versus the `service.toml` metadata sitting next to the credentials) is that now tags, provider, kind, etc. all have to be keys within the secret rather than attributes (I might have the terminology wrong) on the resource itself.  Combine that with the fact that it's important to have the values of the metadata at build time and ideally _not_ have the values of the secrets then, I think there's an argument for keeping the metadata out of the secret payload.

nebhale:
@scothis Based on discussion in the WG meeting today, there was a preference for the metadata in files, but in a separate directory next to the secrets.  Does this compromise work for you?

nebhale:
Another concern was raised privately via the Heroku team

> the inconsistency between the proposed use cases stands out and if we decide to include service_kind and providers namespaces all services should follow this format with default namespaces for those without kind or provider attributes.

The kind/provider division was inspired by [Crossplane](https://crossplane.io/docs/v0.2/deploy.html) and serves as a mechanism for describing an abstract service type (e.g. MySQL, Redis) and a concrete implementation (e.g. CloudDB, Redis To Go).  I believe that all services, even the more esoteric ones, can be represented with the kind/provider division.  To take a couple of examples from the Heroku Add-ons catalog (kind/provider):

* MySQL/JawsDB MySQL
* MySQL/JawsDB Maria
* MySQL/ClearDB
* Redis/Redis To GO
* Redis/Redis Cloud
* Redis/Redis Green
* Redis/Heroku
* Redis/OpenRedis

Now those are the easy ones, multiple providers for an obvious single service kind.  But you could imagine something where there is a more 1:1 mapping like APMs.  In that case, most of the time you might want only a single kind/provider pair, but you might also want (as in the CF case) for a user to be able to manually create a servicing binding

* New Relic/New Relic
* New Relic/User Provided
* AppDynamics/AppDynamics SaaS
* AppDynamics/AppDynamics On Prem
* AppDynamics/User Provided

Finally, it'd be totally OK to have a `kind` that is hyper-specific to a single vendor:

* QRackajack/QRackajack
* Semaphore/Semaphore
* User Agent Identifier/User Agent Identifier

In short, I think it's totally OK to have a single kind/provider pair (with identical values for both) for some services as it leaves us the flexibility to have multiple providers per kind in other places.

PR #23, opened 2019-08-10
RFC: App Image Extensions (OS packages)

sclevine:
[Readable](https://github.com/buildpack/rfcs/tree/app-image-ext/text/0000-app-image-extensions.md).

### DIFF
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -9,55 +9,108 @@
 # Summary
 [summary]: #summary
 
-This RFC proposes an interface between a set of stack images and an app or platform in order to extend the images with app-specific packages.
+This RFC proposes:
+1. An interface between a set of stack images and a platform in order to extend the images
+2. A UX for the pack CLI for extending builders with OS packages
 
 # Motivation
 [motivation]: #motivation
 
-Allowing buildpacks to install OS packages dynamically during build would drastically increase CNB execution time, especially when CNB is used to build or rebase many apps with similar package requirements.
+Allowing buildpacks to install OS packages dynamically during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements.
 
-Mixins allow buildpack authors to create buildpacks that depend on an extended set of OS packages.
-
+Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time.
 However, it is not uncommon for application code to depend on OS packages.
-Given that app authors and platform maintainers experience increased CNB execution time directly, extending the stack at the app author's or platform maintainer's request may add flexibility without sacrificing performance.
+
+Given that app authors and platform maintainers experience increased build time directly, extending the stack at the app author's or platform maintainer's request may add flexibility without sacrificing performance.
 
 # What it is
 [what-it-is]: #what-it-is
 
+## Specification
+
 Stack image creators may add the following executables to their run and/or build images:
 
 ```
-/cnb/image/build/extend (pkg-cache-dir) | cwd: (app)
-/cnb/image/run/extend (pkg-cache-dir) | cwd: (app)
+/cnb/image/build/extend (pkg-cache-dir) (build-metadata-toml-file) | cwd: /
+/cnb/image/run/extend (pkg-cache-dir) (run-metadata-toml-file) | cwd: /
 ```
 * executes on base run or build images
-* both base image and extended image digests stored in app image metadata
-* on subsequent builds:
-  * if the chosen run/build base image digest is different than the one in app image metadata, the chosen digest is extended
-  * if the chosen run/build base image digest is the same, proceed to status
+* on subsequent builds, lookup latest upstream base image digest:
+  * if the digest is different than extended image's base, the new digest is extended
+  * if the digest is the same, proceed to status
 
 ```
-/cnb/image/build/status (pkg-cache-dir) | cwd: (app-dir)
-/cnb/image/run/status (pkg-cache-dir) | cwd: (app-dir)
+/cnb/image/build/status (pkg-cache-dir) (build-metadata-toml-file) | cwd: /
+/cnb/image/run/status (pkg-cache-dir) (run-metadata-toml-file) | cwd: /
 ```
 * executes on last-built extended run or build image (determined from last-built app image labels)
-* exit status 100 = needs extension
+* metadata guaranteed to be the same as extend
+* exit status 100 = existing base image digest is extended 
 * exit status 0 = does not need extension
 * exit status other = unknown
 
-The status executable must be invoked occasionally to poll for updates.
+The status executable may be invoked to determine whether updates are necessary.
+
+## UX
+
+### Creating an Extended Builder
+
+`pack create-builder sclevine/builder -b builder.toml`
+
+builder.toml:
+```toml
+[stack]
+id = "io.buildpacks.stacks.bionic"
+build-image = "example.com/build"
+run-image = "example.com/run"
+run-image-mirrors = ["example.org/run"]
+
+[extend]
+run-image = "sclevine/run"
+run-image-mirrors = ["scl.sh/run"]
+[extend.run]
+packages = ["git"] # run-metadata-toml-file (stored on run image metadata)
+[extend.build]
+packages = ["git"] # build-metadata-toml-file (stored on builder image metadata)
+```
+
+Behavior: creates builder extended with packages as well as `sclevine/run` extended with packages.
+
+### Upgrade an Extended Builder
+
+Both build and run images: `pack upgrade sclevine/builder`
+
+Run images (including mirrors): `pack upgrade sclevine/run`
+
+### Upgrade an App
+
+`pack upgrade sclevine/myapp`
+
+This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then run `pack rebase sclevine/run`.
 
 # How it Works
 [how-it-works]: #how-it-works
 
+## Specification
+
 Platforms may extend app images using the above interface by running the `extend` executable as root in a new container and creating an image from the result.
 Extending an image should generate a single layer.
 This should happen prior to the normal CNB build or rebase process.
-The `app` directory is used to determine what packages to install.
-The `pkg-cache` directory is used to cache package databases (e.g. to speed up `apt-get update`).
+The `*-metadata-toml-file` file is used to determine what packages to install.
+The `pkg-cache-dir` directory is used to cache package databases (e.g. to speed up `apt-get update`).
+
+## UX
+
+`pack upgrade` (on base image)
+1. See if new base image available, if so proceed to (3). Otherwise, (2).
+2. Run status, if updates available, proceed to (3). Otherwise, stop.
+3. Run extend and publish new image. Store extension metadata on images. 
+
+If running `pack upgrade` on a builder, perform operation against both builder and run images and publish both.
+If running `pack upgrade` on an app, perform operation against run image (with ephemeral output into app repo), then `pack rebase`.  
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 - Requires kaniko (or a similar tool) when a docker daemon is unavailable
-- Rebasing apps with package dependencies becomes many orders of magnitude slower
+- Fully rebasing apps with package dependencies becomes many orders of magnitude slower, due to upgrade needs
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -69,10 +69,10 @@ run-image-mirrors = ["example.org/run"]
 [extend]
 run-image = "sclevine/run"
 run-image-mirrors = ["scl.sh/run"]
-[extend.run]
-packages = ["git"] # run-metadata-toml-file (stored on run image metadata)
-[extend.build]
-packages = ["git"] # build-metadata-toml-file (stored on builder image metadata)
+[extend.run] # run-metadata-toml-file (stored on run image metadata)
+packages = ["git"]
+[extend.build] # build-metadata-toml-file (stored on builder image metadata)
+packages = ["git"]
 ```
 
 Behavior: creates builder extended with packages as well as `sclevine/run` extended with packages.
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -31,8 +31,8 @@ Given that app authors and platform maintainers experience increased build time 
 Stack image creators may add the following executables to their run and/or build images:
 
 ```
-/cnb/image/build/extend (pkg-cache-dir) (build-metadata-toml-file) | cwd: /
-/cnb/image/run/extend (pkg-cache-dir) (run-metadata-toml-file) | cwd: /
+/cnb/image/build/extend (build-metadata-toml-file) (pkg-cache-dir) | cwd: /
+/cnb/image/run/extend (run-metadata-toml-file) (pkg-cache-dir) | cwd: /
 ```
 * executes on base run or build images
 * writes metadata to `/cnb/image/build/metadata.toml` or `/cnb/image/run/metadata.toml`
@@ -41,8 +41,8 @@ Stack image creators may add the following executables to their run and/or build
   * if the digest is the same, proceed to status
 
 ```
-/cnb/image/build/status (pkg-cache-dir) (build-metadata-toml-file) | cwd: /
-/cnb/image/run/status (pkg-cache-dir) (run-metadata-toml-file) | cwd: /
+/cnb/image/build/status (build-metadata-toml-file) (pkg-cache-dir) | cwd: /
+/cnb/image/run/status (run-metadata-toml-file) (pkg-cache-dir) | cwd: /
 ```
 * executes on run or build base images
 * metadata arg is `/cnb/image/build/metadata.toml` or `/cnb/image/run/metadata.toml`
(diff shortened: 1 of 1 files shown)


josegonzalez:
So from a practical standpoint, where does that leave platforms? Where should extra os packages - and any extra knobs necessary to run them - be specified?

For the @dokku project, we currently have a `dokku-apt` community plugin that [supports a range of files](https://github.com/dokku-community/dokku-apt#usage) for modifying the "base image": Heroku might instead extend either their `app.json` or `heroku.yml` manifests. Other platforms may want this in some toml file, or filled in via an api of some sort.

The point of the above is that this spec doesn't really tell platforms what they should/should not do in regards to allowing image extensions, which makes it less portable between platforms. I also am not 100% sure we should be allowing/pushing blanket repository access for building image extensions.

sclevine:
I added more detail to clarify some of the questions above. I also pushed the package configuration entirely into the platform, which may address concerns about having a direct contract between the app directory and the stack.

Additionally, I suggested a pack CLI UX. 😄

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,117 @@
...
+id = "io.buildpacks.stacks.bionic"
+build-image = "example.com/build"
+run-image = "example.com/run"
+run-image-mirrors = ["example.org/run"]
+
+[extend]
+run-image = "sclevine/run"
+run-image-mirrors = ["scl.sh/run"]
+[extend.run]
+packages = ["git"] # run-metadata-toml-file (stored on run image metadata)
```
josegonzalez: So is this metadata meant to be the only thing that can/should be used by the platform?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,117 @@
...
+* exit status 0 = does not need extension
+* exit status other = unknown
+
+The status executable may be invoked to determine whether updates are necessary.
+
+## UX
+
+### Creating an Extended Builder
+
+`pack create-builder sclevine/builder -b builder.toml`
```
josegonzalez: I did a GitHub search, but there isn't a spec around what a `builder.toml` is. Is there a single place where I can read what defines that/how it's meant to be used?

josegonzalez: As well, would this be considered to be idempotent? As in, multiple invocations of the same `create-builder` command with the same args should be a noop?

text/0000-app-image-extensions.md
sclevine: Sorry, the comment was supposed to be on the line above. `extend.run` is arbitrary metadata and is the contents of `(run-metadata-toml-file)`. It's not just restricted to a list of packages.

text/0000-app-image-extensions.md
sclevine: The spec you're looking for is here: https://buildpacks.io/docs/using-pack/working-with-builders/#builder-configuration

The output image is always re-created from images specified in builder.toml, so I'm not sure what idempotent would mean in this context. Given the same inputs, the output should always be the same. It's not an operation that can be re-applied. (And `extend` is never run on an image that is already extended.)

josegonzalez: I mean like, if I wanted to put this in the context of my own platform, could I always run `create-builder` with that toml in place and expect it to return immediately because the input `builder.toml` was the same? That way I don't have to wait for an expensive rebuild step on each "deploy" of my app.

sclevine: `pack create-builder` could check to see if there's an existing builder at the specified tag, verify that its metadata matches builder.toml, swap out any buildpacks, and run `pack upgrade` on it.

Similarly, if the extended builder already exists, `pack extend-builder` could swap out buildpacks with those in the source image and run `pack upgrade`.

As long as the base image digests haven't changed and `status` passes, you wouldn't need to wait for a rebuild. `pack upgrade` will check this for you.

josegonzalez: If the answer turns from _could_ to _does_, I can see that being useful to many users, as otherwise there wouldn't be much of a difference from an image extension to a buildpack that vendors installed packages to a subdirectory in the workspace.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+If `extend.run-image` is not specified, `extend.run` is stored on the new builder metadata and used to dynamically extend the run image on `pack build`.
+
+### Building an App with Additional Packages
+
+`pack build sclevine/myapp`
+
+project.toml:
+```toml
+...
+[extend.run] # run-metadata-toml-file (stored on new run image metadata)
```
jkutner: `[extend]` is a little awkward as a top level key if we adopt an `[extensions]` key 😬

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+
+Behavior: creates a version of the current builder with additional packages and an ephemeral run image with additional packages, then does a normal `pack build`.
+
+Question: should we store the new builder image to make rebuild faster? If so, where? Should we generate a tag for it?
+
+### Upgrade an Extended Builder
+
+Both build and run images: `pack upgrade sclevine/builder`
+
+Run images (including mirrors): `pack upgrade sclevine/run`
```
jkutner: what happens here? it updates package indicies and reinstalls whatever package(s)?

text/0000-app-image-extensions.md
sclevine: I don't feel strongly about the naming, and placing the section under each image seems reasonable.

text/0000-app-image-extensions.md
sclevine: Re-installs the packages on top of the latest available upstream run image.

sclevine:
I made some minor simplifications to the proposed UX.

@hone @jkutner any thoughts?

text/0000-app-image-extensions.md
jromero: This also seemed unclear without this comment. It might be worth adding to the RFC.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+
+Attempting to rebase an app directly after it's been upgraded is not permitted and will fail, because the current base is now ephemeral.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Specification
+
+Platforms may extend app images using the above interface by running the `extend` executable as root in a new container and creating an image from the result.
+Extending an image should generate a single layer.
```
jromero: Is this solely a side-effect given the fact that extend is executed once and how it would be implemented?

Couldn't a build system execute `extend` potentially be able to yield multiple layers (for instance if it wants to leverage it's own caching)?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+
+What happens if you try to extend an extended builder?
+
+Three options:
+
+1. Fail
+
+2. Treat the extended builder as a normal builder, and override the metadata.
+   Advantages: Allows controlled, multi-level distribution model for changes. Easy to implement.
+   Disadvantages: Requires `pack upgrade`ing multiple levels of builders to fully patch the most extended builder.
```
jromero: This seems like the ideal and most clear choice IMO. How do you know, or why should you care if the builder you are extending is extending another builder?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+2. Run status, if updates available, proceed to (3). Otherwise, stop.
+3. Run extend and publish new image. Store extension metadata on images. 
+
+If running `pack upgrade` on a builder, perform operation against both builder and run images and publish both.
+If running `pack upgrade` on an app, perform operation against run image (with ephemeral output into app repo), then `pack rebase`.  
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Requires kaniko (or a similar tool) when a docker daemon is unavailable
```
jromero: An alternative can simply be that `extend-builder` requires a docker daemon.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+- Fully rebasing apps with package dependencies becomes many orders of magnitude slower, due to upgrade needs
```
jromero: Can this scenario be drawn out more? I'm trying to understand the exact workflow for an app maintainer to rebase their app images.

Wouldn't it be as simple as:

1. I need to update my run images: `pack upgrade my/run-image` - takes some time to install packages
2. I now rebase my app images: `pack rebase my/app` - near-instant

I don't know that the app image cares that it comes from this "extended" run image.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+[run.extend] # run-metadata-toml-file (stored on new run image metadata)
+packages = ["git"]
+[build.extend] # build-metadata-toml-file (stored on new builder image metadata)
+packages = ["git"]
+...
+```
+
+Behavior: creates a version of the current builder with additional packages and an ephemeral run image with additional packages, then does a normal `pack build`.
+
+Question: should we store the new builder image to make rebuild faster? If so, where? Should we generate a tag for it?
```
jromero: Since we're speaking about `pack build` I think it should. Would it being stored on the daemon with the image name based on the hash of the configuration (`hash('original image sha' + 'build.extend metadata')`) work?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+run-image-mirrors = ["scl.sh/run"]
+[extend.run] # run-metadata-toml-file (stored on new run image metadata)
+packages = ["git"]
+[extend.build] # build-metadata-toml-file (stored on new builder image metadata)
+packages = ["git"]
+```
+
+Behavior: creates a new version of an existing builder with additional packages as well as a new run image (`sclevine/run`) with additional packages.
+
+If `extend.run-image` is not specified, `extend.run` is stored on the new builder metadata and used to dynamically extend the run image on `pack build`.
```
jromero: Maybe this shouldn't be an option.

The thought that an app image with an extended (ephemeral) run image is no longer treated the same makes me think that these images shouldn't be ephemeral. A lot of complexity and restrictions are removed.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Specification
+
+Stack image creators may add the following executables to their run and/or build images:
+
+```
+/cnb/image/build/extend (build-metadata-toml-file) (pkg-cache-dir) | cwd: /
```
matthewmcnew: Is this cache-dir provided by the platform to share across multiple different extensions of the same stack?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+/cnb/image/run/extend (run-metadata-toml-file) (pkg-cache-dir) | cwd: /
+```
+* executes on base run or build images
+* writes metadata to `/cnb/image/build/metadata.toml` or `/cnb/image/run/metadata.toml`
+* on subsequent builds, lookup latest upstream base image digest:
```
matthewmcnew: What is the `upstream base image here`?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+...
+```
+
+Behavior: creates a version of the current builder with additional packages and an ephemeral run image with additional packages, then does a normal `pack build`.
+
+Question: should we store the new builder image to make rebuild faster? If so, where? Should we generate a tag for it?
+
+### Upgrade an Extended Builder
+
+Both build and run images: `pack upgrade sclevine/builder`
```
matthewmcnew: Would `sclevine/builder` be distributed with its extensions?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+
+## UX
+
+`pack upgrade` (on base image)
+1. See if new base image available, if so proceed to (3). Otherwise, (2).
+2. Run status, if updates available, proceed to (3). Otherwise, stop.
+3. Run extend and publish new image. Store extension metadata on images. 
+
+If running `pack upgrade` on a builder, perform operation against both builder and run images and publish both.
+If running `pack upgrade` on an app, perform operation against run image (with ephemeral output into app repo), then `pack rebase`.
```
matthewmcnew: Why not just  perform the extension on pack rebase?

text/0000-app-image-extensions.md
sclevine: Perhaps per user per stack ID?

text/0000-app-image-extensions.md
sclevine: The latest available version of the unextended run image.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+run-image = "example.com/run"
+run-image-mirrors = ["example.org/run"]
+
+[extend]
+run-image = "sclevine/run"
+run-image-mirrors = ["scl.sh/run"]
+[extend.run] # run-metadata-toml-file (stored on new run image metadata)
+packages = ["git"]
+[extend.build] # build-metadata-toml-file (stored on builder image metadata)
+packages = ["git"]
```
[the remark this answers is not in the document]
sclevine: Where would the modified run image be saved?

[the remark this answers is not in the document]
sclevine: Maybe `stack.run.extend.tag`?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,182 @@
...
+[builder]
+image = "example.com/builder"
+
+[extend]
+run-image = "sclevine/run"
+run-image-mirrors = ["scl.sh/run"]
+[extend.run] # run-metadata-toml-file (stored on new run image metadata)
+packages = ["git"]
+[extend.build] # build-metadata-toml-file (stored on new builder image metadata)
+packages = ["git"]
```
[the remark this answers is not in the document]
sclevine: Would it be more consistent with the new builder case above if we nested `build` and `run` under `builder` (or used top-level `build` and `run` keys for both)?

text/0000-app-image-extensions.md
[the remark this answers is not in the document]
sclevine: Ah, you suggested providing this via a CLI flag below. Strongly agree with that suggestion.

text/0000-app-image-extensions.md
sclevine: How often would we clean up old images?

text/0000-app-image-extensions.md
sclevine: Up to the author of the builder.

text/0000-app-image-extensions.md
sclevine: `pack upgrade` essentially becomes the new `pack rebase`.

I feel like `pack rebase` should always happen instantly without requiring a build container, and that it should break when rebasing isn't possible. This would discourage users from making unnecessary extensions that silently hurt performance.

text/0000-app-image-extensions.md
sclevine: This comment was about the purposed interface, not about the pack CLI specifically. Platforms like kpack would need kaniko or similar.

text/0000-app-image-extensions.md
sclevine: Just meant that the full rebase process (either with the two commands you specified, or with a single `pack upgrade my/app`) takes longer than having stack authors include more packages in stacks (or having a larger selection of dependency-providing buildpacks).

text/0000-app-image-extensions.md
sclevine: If we treat the extending an non-extended builder the same as extending an extended builder, it becomes possible to run `pack upgrade` and still have outdated packages with CVEs. We end up with the same problem that the Dockerfile model has (even though we have an easy fix with (3)).

text/0000-app-image-extensions.md
sclevine: It could, but it would probably need to look like multiple `extend` scripts (as the slices interface doesn't work when you already have layers). Do you have a suggestion for an interface that would result in multiple layers (considering that the generated layers depend on arbitrary input)?

jromero:
Work in progress alterations to the pack UX section of this RFC: https://hackmd.io/zuzsIAh5QGKcQt_EZAcXaw?view

dmikusa-pivotal:
I have a few questions:

1.) I've read over this a few times and the proposal seems to focus on providing metadata to indicate what is to be installed, but not how it's to be installed. For example, I want package "foo". It doesn't say anything about which package manager to use or if it even has to be a package manager. Am I understanding that correctly?

2.) Again, if I'm understanding correctly, it seems like `extend` is responsible for performing the actual installation. I wonder if the current metadata, just a list of `packages` is sufficient?

My concern is that you could potentially have different types of packages, different repos, repo keys that need to be specified. Heck, if you wanted to install some proprietary library you might just have a tgz archive. It seems like more metadata would be required for the `extend` binary to make decisions about how to properly install the requested packages & software.

Again, if I'm misunderstanding the proposal, please let me know.

My concerns are around installing proprietary software that doesn't come from standard OSS repos. PHP needs this in some cases (Oracle, MSSQL). It would be fantastic if this proposal covered that use case. Thanks

sclevine:
Likely deprecated in favor of #77.

sclevine:
Deprecated in favor of https://github.com/buildpacks/rfcs/pull/87.

PR #24, opened 2019-09-19
RFC: Buildpack Registry

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/buildpack-registry/text/0000-buildpack-registry.md)

### DIFF
--- a/text/0000-buildpack-registry.md
+++ b/text/0000-buildpack-registry.md
@@ -56,15 +56,15 @@ $ pack publish docker.io/foo/bar registry.buildpacks.io/myname/mycnb
 
 This will push the buildpackage to a Docker registry (`docker.io` in this example, but we will support others). Then it will register that image with the Buildpack Registry as a buildpack (`myname/mycnb` in this example). To ensure that the buildpack author has access to the account and the published image, the Buildpack Registry will do the following:
 
-1. Create a temporary public key
+1. Create a temporary private/public key pair
 1. Generate a random token
-1. Sign the token
-1. Add the signed token to the buildpackage metadata.
+1. Sign the token using the private key
+1. Add the token signature to the buildpackage metadata.
 1. Send the random token and the public key to the Buildpack Registry
-1. The Buildpack Registry will use the public key to sign the random token
-1. If the signed token matches the value stored in the image metadata we know the client is the one who published that image.
+1. The Buildpack Registry will use the public key and the signature in the metadata to verify the token
+1. If the signature is valid we know the client is the one who published that image.
 
-(I think this works, but if not, we could keep a private key in the Buildpack Registry, and the client could request the public key to use then encrypt the token. I'm fairly certain that works, but it requires an extra step that I think isn't necessary)
+This prevents the user from needing to login on two different systems (the buildpack registry and DockerHub or whatever backend registry they're using).
 
 To download a buildpack, any client may run:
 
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
Have you looked at https://github.com/opencontainers/artifacts?

jonjohnsonjr:
This is really interesting. I've love something like this to just front arbitrary registries with a vanity domain.

Aggregating across multiple registries is a cool idea, that could help with a lot of different problems. You could potentially use this to pull a buildpack/image by digest without having to specify a location cc @jchesterpivotal @dprotaso 

> * How will we index buildpacks to facilitate searching?
> * How will we index buildpacks to facilitate fast resolution (w/o spamming the registry)?

It's essentially impossible to do this right now in a portable way, see https://github.com/opencontainers/distribution-spec/issues/22

You should get involved in [steve's efforts to define search requirements](https://hackmd.io/mF80aTt8TC-GiGHOyPG-dw).

> The registry will not store any secrets. It will only store a mapping of Buildpack Registry account to third-party account.

Somewhat related is [signing artifacts](https://hackmd.io/Rs4G0I4lSAO-Fdrdchweqg).

Keeping in sync with the registries you're fronting will be tough. What happens if someone deletes a buildpack from the underlying registry?

I'll also reiterate that it's important to be careful in writing any software that assumes a central registry. A _lot_ of problems were caused by a dockerhub default, e.g.:
* Parsing references to images is *hard* (i.e. ambiguous) because omitting the registry defaults to dockerhub.
* For a long time, it was impossible to set up a mirror for anything other than docker. [Related](https://github.com/opencontainers/distribution-spec/pulls).

jkutner:
@jonjohnsonjr thanks for taking a look! The signing stuff in particular is very interesting, and I'll read up on the search requirements. 

The "keeping in sync" part is tricky, and I'll add that to *Risks*. I think deletion is the main concern (as opposed to pushing new tags in a repo w/o a corresponding buildpack).

With regard to it being a central registry, my hope is that it not. Heroku will run it's own instance (and may actually run several). I can image users wanting to run per-cluster instances of a buildpack registry too (to support private buildpacks).

jonjohnsonjr:
Steve opened some issues for signing and search if you're interested in participating in those conversations:
https://github.com/opencontainers/distribution-spec/issues/70
https://github.com/opencontainers/distribution-spec/issues/71

text/0000-buildpack-registry.md
```diff
@@ -56,15 +56,15 @@ $ pack publish docker.io/foo/bar registry.buildpacks.io/myname/mycnb
 
 This will push the buildpackage to a Docker registry (`docker.io` in this example, but we will support others). Then it will register that image with the Buildpack Registry as a buildpack (`myname/mycnb` in this example). To ensure that the buildpack author has access to the account and the published image, the Buildpack Registry will do the following:
 
-1. Create a temporary public key
+1. Create a temporary private/public key pair
 1. Generate a random token
-1. Sign the token
-1. Add the signed token to the buildpackage metadata.
+1. Sign the token using the private key
+1. Add the token signature to the buildpackage metadata.
```
wking: With a temporary signing keypair and only the signature in the buildpackage, how does the consumer find and trust the public key they use to verify the signature?

wking: Ah, this is just about verifying the push to registry, not about end users verifying anything.  The buildpack registry can just hold the public key in a local cache.  Sorry for the noise.

jkutner:
closing in favor of #35

PR #25, opened 2019-09-26
Simple Project descriptor

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/project-descriptor-simple/text/0000-project-descriptor.md)

### DIFF
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -32,6 +32,7 @@ Terminology:
 
 * **project**: a repository containing source code for an app, service, function, buildpack or a monorepo containing any combination of those.
 * **image**: the output of a running a buildpack(s) against a project
+* **inline buildpack**: a buildpack that is part of a project's source code
 
 The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The new elements in  `buildpack.toml` support many different use cases including two new top level tables:
 
@@ -249,6 +250,8 @@ version = "0.0.1"
 id = "io.buildpacks.inline"
 version = "0.0.1"
 ```
+
+The `stack` key is optional for inline buildpacks, since they're meant to be used against the stack the project is using.
 
 There is an open question as to whether or not we should simplify the above example so that `[[project.buildpacks]]` can have a pointer to the buildpack defined in `[buildpack]` instead of repeating the `id` and `version`.
 
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -46,11 +46,16 @@ Here is an overview of the complete schema:
 id = "<string>"
 name = "<string>"
 version = "<string>"
+authors = ["<string>"]
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
 include = ["<string>"]
 exclude = ["<string>"]
 
 [[images]]
-id = "<string>"
+ref = "<string>"
+suffix = "<string>"
 path = "."
 
   [[images.launch.env]]
@@ -73,10 +78,6 @@ path = "."
   uri = "<url or path>"
 
 [metadata]
-authors = ["<string>"]
-documentation = "<url>"
-license = "<string>"
-source = "<url>"
 # additional arbitrary keys allowed
 
 [extensions]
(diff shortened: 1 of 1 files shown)


text/0000-project-descriptor.md
```diff
@@ -0,0 +1,281 @@
...
+[[images]]
+id = "<string>"
+path = "."
+stack = "<string>"
+
+  [[images.launch.env]]
+  name = "<string>"
+
+  [[images.build.env]]
+  name = "<string>"
```
josegonzalez: Should we add `value = "<string>"` to these two?

jkutner: yes 🤦‍♂️

jkutner:
Updated based on discussion in working group. Changes include:

* added `value` for env vars
* changed `processes.cmd` to `processes.command` and included all other fields from `launch.toml`
* Removed `stack`
* Removed `publish`
* Removed `optional`
* Moved `license`, `author`, `source`, etc under `[metadata]`
* Added `[extensions]`

jkutner:
There is an unresolved question regarding how to tag multiple `[[images]]` w/o:
1. requiring developers to commit a registry URL/host to Git
1. requiring developers to manually retag the image after building

One of the options, which I'm proposing as a solution, is to have a platform use two value to construct the full image tag: `[[images]] id` and a new `--registry` flag on `pack build`. Thus, you might see a `project.toml` containing this:

```toml
[[images]]
id = "myorg/gateway"

[[images]]
id = "myorg/service1"
```

A user could then run:

```
$ pack buildpack myimage --registry gcr.io --publish
```

jkutner:
I've added [an example using image id prefix/suffix](https://github.com/buildpack/rfcs/blob/project-descriptor-simple/text/0000-project-descriptor.md#example-custom-image-prefix) as discussed.

jkutner:
@sclevine @nebhale @hone @ekcasey ready for re-review

jkutner:
I've moved the metadata keys in to `[project]` based on the precedent found in `pom.xml` and `package.json`.

I've added `ref` and `suffix` keys to `[[images]]` as described in WG. I chose `ref` because I was afraid `tag` too strongly referred to a complete tag. Happy to change that though.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,324 @@
...
+version = "<string>"
+authors = ["<string>"]
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
+include = ["<string>"]
+exclude = ["<string>"]
+
+[[images]]
+ref = "<string>"
```
sclevine: Small issue with `ref` vs `tag`: only tags are valid in this field, and other kinds of ref are not valid.
E.g., `gcr.io/example/myapp@sha256:2342ab...` is not valid, because it's not possible to know the digest ahead of time.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,324 @@
...
+- `[project]`: (optional) defines configuration for a project
+- `[[images]]`: (optional) defines configuration for an image
+- `[metadata]`: (optional) metadata about the repository
+- `[extensions]`: (optional) extensions to the spec
+
+Here is an overview of the complete schema:
+
+```toml
+[project]
+id = "<string>"
```
sclevine: Were we going to replace `id`/`version` with `ref` (or `tag`, noting my comment below), and then combine it with the individual `ref`s and `suffix`s?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,324 @@
...
+Given the elements described above, we seek to satisfy the following use cases:
+
+## Example: Basic app
+
+This is a complete `project.toml` for a simple app:
+
+```toml
+[project]
+id = "io.buildpacks.my-app"
+version = "0.1"
```
sclevine: Seems unusual to require a version to be checked into a git repo for project.toml to be valid. That breaks certain release strategies.

text/0000-project-descriptor.md
jkutner: i'm fine with `tag` then

text/0000-project-descriptor.md
jkutner: i think i'm fine with that. i'd probably move to make it optional though (which means there's no required id or tag, which might be a problem).

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+No other configuration is required.
+
+## Example: Monorepos
+
+The `[[images]]` array of tables allows the buildpack lifecycle to generate more than one image per build. Each table in the array may contain configuration that defines a different image for different parts of the code. For example:
```
sclevine: clarify per `pack build`, not lifecycle build

hone:
From @jonjohnsonjr, Julia also has a [Project.toml](https://github.com/JuliaLang/Pkg.jl/issues/805).

jkutner:
Closing in favor of #32. We'll open a new RFC for `[[images]]`, `[[extensions]]`, buildpacks and other things.

PR #26, opened 2019-10-09
Create 0000-allow-github-release-url-shorthand.md

jabrown85:
[Readable](https://github.com/buildpack/rfcs/blob/d750008a099ea3f8a3367fabbadd904cea42b8c4/text/0000-allow-github-release-url-shorthand.md)

### DIFF
--- a/text/0000-allow-github-release-url-shorthand.md
+++ b/text/0000-allow-github-release-url-shorthand.md
@@ -1,0 +1,65 @@
+# Meta
+[meta]: #meta
+- Name: Allow GitHub specific url parsing for remote buildpacks
+- Start Date: 2019-10-09
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Allow pack users to specify shorter GitHub urls when specifying remote buildpacks explicitly. These urls would contain the repo and potentially a tag name of the release to download.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


hone:
The convenience of ubiquity of GitHub support in bundler for Ruby was a well liked feature. GitHub is getting a package registry feature giving repos a docker registry.  should we be using that with buildpackages / distribution here?

With respect to vendor neutrality, I wonder if we need to add support for other vendors too and where does that stop? Are `pack` plugins eventually a solution that that problem?

nebhale:
I think I concur with @hone that the forthcoming build packages, which are image-based, likely mean that the URIs will start pointing at Docker Registry coordinates.  I'm not sure how this relates to the URIs one would then use though.

text/0000-allow-github-release-url-shorthand.md
```diff
@@ -0,0 +1,65 @@
...
+```
+pack build myimage --buildpack https://github.com/company/my-buildpack@v1.0.0
+pack build myimage --buildpack https://github.com/company/my-buildpack
+```
+
+The proposal would have pack use with the [GitHub Releases API](https://developer.github.com/v3/repos/releases/) when a url matches the above example patterns. A GitHub release would be found using the tag name specified in the url. If no tag is specified, the latest release is used.
+
+If no matching release is found, an error will be generated stating the buildpack could not be found. 
+
+If a matching release is found, the assets of that release will be searched for an entry with a `content_type` of `application/gzip`. This found asset's `browser_download_url` will become the target for the buildpack download process. If no matching asset is found, GitHub's default tar would become the target. This would fallback would allow repositories that are also a valid CNB to release without uploading an explicit tar.
```
ekcasey: I'm also not sure it is safe to assume that a buildpack release should have exactly one gzip artifact. I can imagine buildpack authors releasing two buildpack artifacts that are built from the same source but are configured/compiled/packaged to support different stacks.

jabrown85:
Closing this. We think the distribution and registry will be OK for now. We might revisit at a later time if the need comes back up.

PR #27, opened 2019-10-14
RFC: Change the lifecycle spec to support Windows.

micahyoung:
[Readable](https://github.com/buildpack/rfcs/blob/ac9b508de1c6469ee28d6ad44a0c2a80dc84806c/text/0000-windows-lifecycle.md)

Work-in-progress spec changes are here and subject to change
https://github.com/buildpack/spec/compare/master...greenhouse-org:windows-lifecycle

### DIFF
--- a/text/0000-windows-lifecycle.md
+++ b/text/0000-windows-lifecycle.md
@@ -1,0 +1,87 @@
+# Meta
+[meta]: #meta
+- Name: Support for Windows in the Lifecycle components
+- Start Date: 2019-10-07
+- CNB Pull Request: 
+- CNB Issue: 
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Implement all lifecycle phases for the Microsoft Windows Operating System and any OCI-compliant Windows container backend. Use Windows conventions as alternatives to Linux-specific conventions where necessary to achieve feature parity with current functionality.
+
+# Motivation
+[motivation]: #motivation
+
+Windows app developers already use containerization as a very important tool for easing deployment for headless Windows applications with tools like Docker for Windows, Cloud Foundry and Kubernetes with Windows Workers. However, the same Day 1 and 2 challenges that CNBs help solve - deployment conistency, dependency management, rebasing, etc - are equally prominent for these containerized Windows apps. Nearly every use-case and persona for Linux containers also exists for Windows, though there is generally less modern tooling available to ease app migration.
+
+A common use-case for CNBs with Windows apps would be migrating an existing deployment of a Windows .NET application. A traditional deployment consists of .NET framework apps authored in Visual Studio on a Windows desktop, deployed directly to Microsoft Internet Information Services on a Windows Server Virtual Machine with various Operating System-level, library-level and app-level interdependencies. Traditional app developers and operators often have to coordinate and manually perform app and dependency changes in these deployments. By migrating to Cloud Native Stacks, Buildpacks and app layers, these bespoke Windows deployments could be decomposed into OCI-image layers increasing automation, consistency and maintainability.
+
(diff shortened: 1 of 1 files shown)


text/0000-windows-lifecycle.md
```diff
@@ -0,0 +1,87 @@
...
+
+A Buildpacks User would expect to follow roughly the same steps and patterns for migrating apps for both Windows or Linux to Stacks and CNBs, though using Windows-specific artifacts, buildpacks and tooling where no cross-plaform standard exists. In the end, they would expect to generate Windows-specific OCI-images which they could host on any registry and which they would run on a Windows-based, OCI-compliant container server such as Docker for Windows, Windows Server containers or Kubernetes with Windows Workers.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC proposes adding official Windows support to the lifecycle which can be optionally used by all platforms and buildpacks. There will be a new Windows-specific, distributable lifecycle artifact containing cross-compiled versions of lifecycle commands. Buildpacks can be either Windows-specific or cross-platform by following Windows conventions. Windows-specific builder images can be created based on the lifecycle and buildpacks.  
+
+This introduces new functionality that is relevant to all personas that interact with buildpacks. This has no impact on existing users.
```
jkutner: despite this, i think any spec changes should probably enter as [extensions](https://github.com/buildpack/spec/tree/master/extensions), especially if there's a concern that the Docker for Windows is less than stable.

micahyoung: Thank Joe. I may have undersold the current stability of Docker for Windows - both the Desktop and Server container runtimes are far along - Docker Desktop got Windows containers in early 2018 and but Windows Server has had them since 2017. At this point, Pivotal PKS, Azure AKS and Google GKE have Windows Workers for Kubernetes in various Beta flavors but they'll be GA "any day now" and as far as I know, all of these soon-to-be-GA IaaS k8s are built against the same APIs that we'll depend on - the Docker client APIs and Windows OCI image format.

My caveat here in the RFC was mainly to point this historical context and how Windows Docker containers are only now where Linux Docker containers have been for some time and so might possibly see some last minute changes. The major APIs we need are well defined though and Windows and Linux should very soon be equally stable.

As for having Windows defined in extensions, I'll admit I'm new to the RFC concepts but I'm wondering if there's a risk of having Windows defined just as an extension may lead to more Linux-only design choices that might, when it's time to implement them for Windows, require less elegant solutions or just wouldn't possible without an upstream change. This might mean reworking already-implemented specs and features or needing to feature-flag them off on Windows.

I feel like we've seen some examples of this in the current spec. One that popped up was relying on Linux shebang kernel support to enable scripts and binaries with the same file names - it's a handy feature for Linux but doesn't work well Windows platforms which instead need explicit suffixes to determine whether they're executable or scripts. I'm not exactly sure what a cross-OS approach would have been but it certainly has lead us to need to come up with a new approach just for Windows, without the benefit of a single solution for both OSes.

I definitely wouldn't want Windows concerns to slow down RFC ideation but to me, it feels safer to resolve any issues well before spec and implementation happens or to explicitly agree to set aside Windows concerns when needed. This may even minimize the need for a Windows-specific spec concerns in the first place.

But I'd like hear other thoughts and concerns though.

micahyoung: Not trying to jump the gun but I opened a [spec PR](https://github.com/buildpack/spec/pull/62) with one potential approach to defining Windows within the current specs. I'm hoping this accounts for everything that needs to be defined but I'm still open to putting these wherever makes the most sense.

micahyoung: Sorry for the noise. Struggling the comment flow

PR #28, opened 2019-10-29
RFC: Pack Build Default Process Flag

hone:
[readable](https://github.com/hone/cnb-rfcs/blob/pack-build-process-flag/text/0000-pack-build-process-flag.md)

### DIFF
--- a/text/0000-pack-build-default-process-flag.md
+++ b/text/0000-pack-build-default-process-flag.md
@@ -1,6 +1,6 @@
 # Meta
 [meta]: #meta
-- Name: Pack Build Process Flag
+- Name: Pack Build Default Process Flag
 - Start Date: 2019-10-29
 - CNB Pull Request: [rfcs#28](https://github.com/buildpack/rfcs/pull/28)
 - CNB Issue: (leave blank)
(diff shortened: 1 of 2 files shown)


hone:
@sclevine that would just set the env var?

sclevine:
Yep, on the image so that `launcher` can read it.

jkutner:
I prefer setting `CMD` as it aligns better with standard OCI image stuff (which seems like something we should try to do when possible). For example, a consumer of the image would be able to easily understand why it's running the `worker` process instead of the `web` process without knowing anything about `CNB_` vars.

ekcasey:
I think `CNB_PROCESS_TYPE` is a better path forward. That means we are changing the default process type, but a user can still override by explicitly overriding `CNB_PROCESS_TYPE`. That override wouldn't work if we set `CMD` instead, given the current precedence. I don't see a compelling reason to change the precedence.

ekcasey:
I wonder if we could have two flags `--default-process <process-type>` which sets `CNB_PROCESS_TYPE` and another `--process <type>="<command>"` which creates an entry in `metadata.toml` that looks like

```
[[processes]]
type = "<type>"
command = "<command>"
args = []
direct = false
```

hone:
@nebhale @sclevine @jkutner updated the PR based off of @ekcasey's comments on our discussion. Can you please re-review.

hone:
With @jkutner's vote this has entered Final Comment Period. This will be open until next Thurs.

PR #31, opened 2019-12-05
Email RFC

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/email/text/0000-use-email-more.md)


hone:
By finalizing, I mean those entering the Final Comment Period. I'd like anyone in the community to have an opportunity to comment.

hone:
Should we use specific tags for emails we know we want to send, for instance for RFCs entering the FCP, we could use: [RFC FCP]?

PR #32, opened 2019-12-05
Minimal project descriptor

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/project-descriptor-minimal/text/0000-project-descriptor.md)

See also #25

### DIFF
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -32,6 +32,7 @@ Terminology:
 
 * **project**: a repository containing source code for an app, service, function, buildpack or a monorepo containing any combination of those.
 * **image**: the output of a running a buildpack(s) against a project
+* **inline buildpack**: a buildpack that is part of a project's source code
 
 The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The new elements in  `buildpack.toml` support many different use cases including two new top level tables:
 
@@ -249,6 +250,8 @@ version = "0.0.1"
 id = "io.buildpacks.inline"
 version = "0.0.1"
 ```
+
+The `stack` key is optional for inline buildpacks, since they're meant to be used against the stack the project is using.
 
 There is an open question as to whether or not we should simplify the above example so that `[[project.buildpacks]]` can have a pointer to the buildpack defined in `[buildpack]` instead of repeating the `id` and `version`.
 
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -17,8 +17,9 @@ This is a proposal for a new file, `project.toml`, containing configuration for 
 This proposal is meant to address all of the following issues:
 
 * Build metadata (i.e. information about the code being built)
-* [Application descriptor](https://github.com/buildpack/spec/issues/44)
 * [Ignoring files](https://github.com/buildpack/pack/issues/210)
+
+It also provides a foundation for solving the problems defined in the [Application descriptor issue](https://github.com/buildpack/spec/issues/44).
 
 # What it is
 [what-it-is]: #what-it-is
@@ -28,7 +29,7 @@ Terminology:
 * **project**: a repository containing source code for an app, service, function, buildpack or a monorepo containing any combination of those.
 * **image**: the output of a running a buildpack(s) against a project
 
-The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The elements in  `project.toml` support many different use cases including three top level tables:
+The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The elements in  `project.toml` support many different use cases including these top level tables:
 
 - `[project]`: (optional) defines configuration for a project
 - `[metadata]`: (optional) metadata about the repository
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -46,11 +46,16 @@ Here is an overview of the complete schema:
 id = "<string>"
 name = "<string>"
 version = "<string>"
+authors = ["<string>"]
+documentation = "<url>"
+license = "<string>"
+source = "<url>"
 include = ["<string>"]
 exclude = ["<string>"]
 
 [[images]]
-id = "<string>"
+ref = "<string>"
+suffix = "<string>"
 path = "."
 
   [[images.launch.env]]
@@ -73,10 +78,6 @@ path = "."
   uri = "<url or path>"
 
 [metadata]
-authors = ["<string>"]
-documentation = "<url>"
-license = "<string>"
-source = "<url>"
 # additional arbitrary keys allowed
 
 [extensions]
@@ -123,17 +124,21 @@ exclude = [
 
 The `.gitignore` pattern is used in both cases. The `exclude` and `include` keys are mutually exclusive, and if both are present the Lifecycle will error out. These lists apply to both buildpacks built with `pack create-package` and apps built with `pack build`.
 
+Any files that are excluded (either via `include` or `exclude`) will be excluded before the build (i.e. not only exluded from the final image).
+
 ## `[[images]]`
 
 Defines properties of the image(s) that are output from a build.
 
 ```toml
 [[images]]
-id = "<image name>"
+ref = "<string>"
+suffix = "<string>"
 path = "."
 ```
 
-* `id` (string, optional): by default inherits from `[project.name]`. If set will be used as a suffix for the tag of the OCI image produced
+* `ref` (string, optional): by default inherits from `[project.id]`. If set will be used as the tag of the OCI image produced
+* `suffix` (string, optional): If set will be used as the suffix to the tag of the OCI image produced
 * `path` (string, optional): by default uses the directory where this file lives
 
 ## `[image.launch.env]` and `[image.build.env]`
@@ -225,14 +230,14 @@ id = "io.buildpacks.monorepo-app"
 version = "0.1"
 
 [[images]]
-id = "my-service"
+ref = "my-service"
 path = "service/"
   [[images.processes]]
   name = "web"
   command = "java -jar target/service.jar"
 
 [[images]]
-id = "my-gateway"
+ref = "my-gateway"
 path = "gateway/"
   # array of processes used to run the image
   [[images.processes]]
@@ -261,24 +266,24 @@ These entries override any defaults in the builder image. If, for example, the p
 
 ## Example: Custom Image Prefix
 
-The `id` field of an `[[images]]` table is used as a suffix for the tag of the OCI image produced. If no prefex is provided, then it represents the entire tag. For example:
+The `ref` field of an `[[images]]` table is used as the tag of the OCI image produced. If no suffix is provided, then it represents the entire tag. For example:
 
 ```toml
 [[images]]
-id = "gcr.io/example/myimage:42"
+ref = "gcr.io/example/myimage:42"
 
 [[images]]
-id = "gcr.io/example/myimage:latest"
+ref = "gcr.io/example/myimage:latest"
 ```
 
-However, a given platform may support a prefix, in which the follow could be used:
+However, a given platform may override the ref, in which case the tag will be combined with the suffix:
 
 ```toml
 [[images]]
-id = ":42"
+suffix = ":42"
 
 [[images]]
-id = ":latest"
+suffix = ":latest"
 ```
 
 A platform such as `pack` may then allow a command where the provided image name is a prefix, like:
--- a/text/0000-project-descriptor.md
+++ b/text/0000-project-descriptor.md
@@ -44,12 +44,23 @@ version = "<string>"
 authors = ["<string>"]
 documentation-url = "<url>"
 source-url = "<url>"
-include = ["<string>"]
-exclude = ["<string>"]
 
 [[project.licenses]]
 type = "<string>"
 uri = "<uri>"
+
+[build]
+include = ["<string>"]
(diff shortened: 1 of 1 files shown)


jkutner:
Updated based on feedback from 12/5/19 WG meeting

josegonzalez:
Is the idea here that platforms would convert into the project descriptor from whatever they support for metadata files?

Would unimplemented attributes raise an error? So a specific platform might extend the descriptor internally, for instance for extending os package support.

jkutner:
> Is the idea here that platforms would convert into the project descriptor from whatever they support for metadata files?

We haven't defined how a platform will support this file. We hope that eventually it will make it's way into the spec. If it does, we generally believe that `lifecycle` should in some way honor it (when possible). How that manifests itself is undecided (a new `/lifecycle/describer` or inline with each binary maybe?).

>  Would unimplemented attributes raise an error?

No. The only error condition we define is when `exclude` and `include` are both present.

> So a specific platform might extend the descriptor internally, for instance for extending os package support.

Could this fit within the arbitrary `[metadata]` table?

iainsproat:
Is the `[project]` property `source-url` intended to duplicate the property (`"io.buildpacks.app.source".metadata.url`) in [RFC 0013](https://github.com/buildpack/rfcs/blob/master/text/0013-app-layer-metadata-source.md)?

Is it possible to include the metadata properties described in [RFC 0013](https://github.com/buildpack/rfcs/blob/master/text/0013-app-layer-metadata-source.md) within the project.toml file proposed in this RFC?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,150 @@
...
+
+- `[project]`: (optional) defines configuration for a project
+- `[metadata]`: (optional) metadata about the repository
+
+Here is an overview of the complete schema:
+
+```toml
+[project]
+id = "<string>"
+name = "<string>"
```
steren: It is unclear what is the difference between `id` and `name`

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,150 @@
...
+- Name: Minimal Project Descriptor
+- Start Date: 2019-06-11
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: https://github.com/buildpack/rfcs/pull/25
+
+# Summary
+[summary]: #summary
+
+This is a proposal for a new file, `project.toml`, containing configuration for apps, services, functions, and buildpacks.
```
steren: While `.toml` is a great readable format, I have concerns that picking `.toml` might add more friction to users coming from the container world, which seems to have converged towards using `.yaml`. In my opinion, this is a data point that should be factored in when picking a format.

As a middle ground, maybe this proposal could say that `.yaml` and `.json` format are also accepted, but `.toml` is preferred (in the end, they all represent structured data, and `.yaml` is even a superset of `.json`).

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,150 @@
...
+Here is an overview of the complete schema:
+
+```toml
+[project]
+id = "<string>"
+name = "<string>"
+version = "<string>"
+authors = ["<string>"]
+documentation-url = "<url>"
+source-url = "<url>"
```
briandealwis: `include` and `exclude` are used by the build process. How are the other fields here intended to be used?  (Are they turned into image annotations?) Should there be `builder` and `run-image` fields?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,150 @@
...
+
+This is a proposal for a new file, `project.toml`, containing configuration for apps, services, functions, and buildpacks.
+
+# Motivation
+[motivation]: #motivation
+
+This proposal is meant to address all of the following issues:
+
+* Build metadata (i.e. information about the code being built)
+* [Application descriptor](https://github.com/buildpack/spec/issues/44)
```
briandealwis: #44 includes environment variables which isn’t addressed in this proposal. These variables seemed intended to influence buildpack execution, but it could be useful to define variables to be propagated into the container environment too?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,150 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Terminology:
+
+* **project**: a repository containing source code for an app, service, function, buildpack or a monorepo containing any combination of those.
+* **image**: the output of a running a buildpack(s) against a project
+
+The target personas for this proposal is buildpack users who need to enrich or configure buildpack execution. The elements in  `project.toml` support many different use cases including three top level tables:
```
briandealwis: “three”?

text/0000-project-descriptor.md
jkutner: `name` is meant to be human readable, while `id` is machine readable. However, we're not specifying any behaviors associated with these fields.

text/0000-project-descriptor.md
jkutner: we're not specifying any behavior associated with the other fields.

I think any option on `pack` (including those you mentioned) is a good candidate for this descriptor. We batted around many different schemas, but are now trying to land on a very minimal subset of what was originally proposed.

jkutner:
@iainsproat i'm not sure if [RFC0013](https://github.com/buildpack/rfcs/blob/master/text/0013-app-layer-metadata-source.md) was ever implemented, but we should probably align this with it. However, I think the `app` should be replaced with `project` ("app" has become too overloaded)

text/0000-project-descriptor.md
josegonzalez: I think we should stick to `toml`. My reasoning is that all other parts of the buildpack spec use yaml, and as a platform maintainer, it would be quite frustrating to have to support other formats both in the platform itself and in support requests coming in from users.

The name `project.toml` also has fewer (no?) conflicts with other projects (I see that references to `project.yaml` in OpenStack and ArgoCD).

The `json` format doesn't allow for comments, which is why it wasn't picked.

iainsproat:
@jkutner It's true that [RFC0013](https://github.com/buildpack/rfcs/blob/master/text/0013-app-layer-metadata-source.md) has not been implemented yet. It has already been agreed upon and we have a pending [PR into the specification](https://github.com/buildpack/spec/pull/68). We could replace `app` with `project` though I'm not sure of the process required to update a previously accepted RFC. Does this require a new RFC?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,203 @@
...
+
+If both `exclude` and `include` are defined, the build process will error out.
+
+## `[[build.buildpacks]]`
+
+The build table may contain an array of buildpacks. The schema for this table is:
+
+```toml
+[[build.buildpacks]]
+id = "<buildpack ID (required)>"
```
ekcasey: I don't think `id` or `version` should be required if `uri` is provided. The buildpack at the `uri` will have an ID and version in it's `builpack.toml`. These fields then would add no additional information and would merely provide an opportunity for users to enter an incorrect value

text/0000-project-descriptor.md
hone: @jkutner can we actually put descriptions for each of these fields?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,205 @@
...
+id = "<string>"
+name = "<string>"
+version = "<string>"
+authors = ["<string>"]
+documentation-url = "<url>"
+source-url = "<url>"
+
+[[project.licenses]]
+type = "<string>"
+uri = "<uri>"
```
hone: @nebhale for the non SPDX use case, do people usually point to a URI or just include a file to the LICENSE on disk locally?

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,205 @@
...
+[project]
+id = "<string>"
+name = "<string>"
+version = "<string>"
+authors = ["<string>"]
+documentation-url = "<url>"
+source-url = "<url>"
+
+[[project.licenses]]
+type = "<string>"
```
hone: I believe, we should be following the [SPDX list](https://spdx.org/licenses/). The list is versioned as well, so we can point to a particular list we like.

hone: Also thoughts here on `id` or `name` over `type`?

text/0000-project-descriptor.md
jkutner: added description

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+* `id` - the machine readable identifier of the project (ex. "com.example.myservice")
+* `name` - the human readable name of the project (ex. "My Example Service")
+* `version` - and arbitrary string representing the version of the project
+
+## `[project.licenses]`
+
+* `type` - This uses the [SPDX 2.1 license expression](https://spdx.org/spdx-specification-21-web-version). This matches the identifier and must be included in the [SPDX Lincenes List](https://spdx.org/licenses/). As of this writing, it's on version 3.7.
+* `uri` - If this project is using a nonstandard license, then this key may be specified in lieu of `type` to point to the license.
```
elbandito: @jkutner does `in lieu of type` mean that `type` and `uri` should be mutually exclusive?

text/0000-project-descriptor.md
nebhale: I'd gone with type to preserve the possibility of using a non-SPDX value here.  For example, `type = "Pivotal Proprietary"` should be a reasonable value, and wouldn't be an "id" per se.

text/0000-project-descriptor.md
nebhale: The use of the URI is typically handled orthogonally from whether there's an SPDX value or not.  The customers that have asked us for this don't so much want a link to the boiler-plate of the license (the SPDX takes care of that in most systems), but rather proof that the project uses that license.  So [you'll see](https://github.com/cloudfoundry/openjdk-cnb/blob/master/buildpack.toml#L41-L42):

```
[[metadata.dependencies.licenses]]
type = "GPL-2.0 WITH Classpath-exception-2.0"
uri  = "https://openjdk.java.net/legal/gplv2+ce.html"
```

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,214 @@
...
+version = "<string>"
+```
+
+* `id` - the machine readable identifier of the project (ex. "com.example.myservice")
+* `name` - the human readable name of the project (ex. "My Example Service")
+* `version` - and arbitrary string representing the version of the project
+
+## `[project.licenses]`
+
+* `type` - This uses the [SPDX 2.1 license expression](https://spdx.org/spdx-specification-21-web-version). This matches the identifier and must be included in the [SPDX Lincenes List](https://spdx.org/licenses/). As of this writing, it's on version 3.7.
```
nebhale:
```suggestion
* `type` - This uses the [SPDX 2.1 license expression](https://spdx.org/spdx-specification-21-web-version). This matches the identifier and must be included in the [SPDX Licenses List](https://spdx.org/licenses/). As of this writing, it's on version 3.7.
```

nebhale: I don't think that there should be a hard requirement that **only** contain SPDX values.  It's a good suggestion, but I think it's reasonable to use a value that's broadly applicable but not open source (e.g. Oracle BCL, Pivotal Proprietary).  This is something that you can map back to an automated value, but doesn't have an SPDX id.

text/0000-project-descriptor.md
nebhale: They should not be mutually exclusive.  In our use of the same design, we've enforced the "at least one" requirement.

jkutner:
@nebhale i've addressed your feedback, and loosened the requirements on license type and uri.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,214 @@
...
+
+```toml
+[metadata.heroku]
+pipeline = "foobar"
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The `project.toml` contents will be read by the platform and/or the lifecycle. The launch image will contain labels for the data in all fields except `include` and `exclude`.
```
jromero: Should all of `[build]` be excluded from being added to the label? I don't see the value in replicating the buildpacks which are already stored in another label. In general, I would think that we would only add informational data to the label not so much configuration.

text/0000-project-descriptor.md
```diff
@@ -0,0 +1,214 @@
...
+This defines the buildpacks that a platform should use on the repo.
+
+Either an `id` or a `uri` are required, but not both. If `uri` is provided, `version` is not allowed.
+
+## `[[build.env]]`
+
+Used to set environment variables at build time, for example:
+
+```toml
+[[build.env]]
```
jromero: Would `launch.env` be viable to be added as well? Thinking about environment variables that you want when your application launches that are outside of buildpacks (or override those provided by the buildpacks). Can be added later if we still need to discuss it further.

jkutner: I think `launch.env` is a bit more contentious, and we should discuss it in a separate update RFC.

text/0000-project-descriptor.md
jkutner: agree, we should omit `build`

jkutner:
@hone @nebhale @sclevine Please review the very small change in f047502 and approve/reject. Otherwise the FCP will close on 1/16.

jkutner:
I see now that this was merged in be3761d

PR #33, opened 2019-12-06
Release process

jromero:
[readable](https://github.com/jromero/rfcs/blob/release-process/text/0000-release-process.md)

### DIFF
--- a/text/0000-release-process.md
+++ b/text/0000-release-process.md
@@ -1,0 +1,192 @@
+# Meta
+[meta]: #meta
+- Name: Release Process
+- Start Date: 2019-11-18
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The release process for the various components maintained by the Cloud Native Buildpacks organization.
+
+# Motivation
+[motivation]: #motivation
+
+Through the previous releases of pack and lifecycle we've come to learn that a time-based release cadence makes the
+most sense. Unfortunately, due to the fact that there is no formal process, making the release happen has been met by
+some inconsistencies in timing and confusion in processes. These are generally in the form of conversations revolving 
+the following questions _"when are we going to release?"_, _"what do we want to put in this release?"_, 
+_"should we wait for feature Y to release?"_, etc.
+
+# Consumer Impact
+
+#### Improvements
+
+At the moment, releases tend to get pushed back to ensure certain feature sets are released. With more
+definitive guidance we could plan around it and isolate features when necessary through the process of
+[CCB](#change-control-board) while maintaining the goal of releasing on time.
+
+#### Concerns
+
+If the [Release Cadence](#timetable) is too long, consumers would need to wait longer for features they request
+if they are not included in the currently scheduled release.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposed release process aims to be flexible while providing structure. It proposes a 3-phase release cycle for
+the purpose of:
+ 
+1. Enabling [scheduled releases](#timetable)
+2. Allowing for the [continuation of development](#concurrency)
+3. [Isolating releases](#release-branches) and their features
+4. Ensure [user acceptance](#user-acceptance-testing) takes place
+5. [Migration](#migration-guide) is considered
+
+### Phases
+Each release would consists of the following 3 phases:
+
+- development - standard development process _(issue -> PR -> merge)_
+- [feature complete](#feature-complete) - branch of code into a [release branch](#release-branches) for ongoing [UAT](#user-acceptance-testing)
+- [release execution](#release-execution) - the step of finalizing and publishing a release
+
+```text
+ release X
+
+ +---------------+      +------------------+      +---------+
+ |  development  | ---> | feature complete | ---> | release |
+ +---------------+      +------------------+      +---------+
+
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Timetable
+
+_The following schedules are suggestive and subject to change based on practicality._
+
+* Release Cadence: **Monthly** on the **1st Tuesday** of the month 
+* Feature Complete: **5 business days** prior to scheduled release
+* CCB Review: **daily**
+
+### Concurrency
+
+Due to the nature of [release branches](#release-branches), this enables development to continue on future releases
+while finalizing the current release. 
+
+```text
+ release X
+
+ +---------------+      +------------------+      +---------+
+ |  development  | ---> | feature complete | ---> | release |
+ +---------------+      +------------------+      +---------+
+
++---------------------------------------------------------------------------------+
+
+                        release Y
+
+                        +---------------+      +------------------+      +---------+
+                        |  development  | ---> | feature complete | ---> | release |
+                        +---------------+      +------------------+      +---------+
+```
+
+### Terminology
+
+#### User Acceptance Testing
+
+User acceptance testing (UAT) is the process of manually verifying that new features and fixes are **complete** and 
+**don't introduce issues**. Typically, there are two strategies:
+
+1. Rolling UAT - User acceptance is done per issue as the issue is completed.
+2. Feature-Complete UAT - User acceptance is done when all features are complete.
+
+More commonly Feature-Complete User Acceptance is used in practice because any additional changes could have unexpected
+side-effects invalidating prior completed UATs. In our case, Feature-Complete UAT fit perfectly during the
+[feature complete](#feature-complete) phase.
+
+#### Feature Complete
+
+Feature complete for a particular release is a period of time when changes are analyzed and vetted before being added to
+a release.
+
+The following take place:
+
+1. At the beginning, a release branch is created (`release/<version>`).
+2. At the beginning, a [change control board](#change-control-board) is organized on a daily basis to review additional
+requested changes.
+3. During, only changes approved by the CCB are merged into the release branch.
+
+#### Change Control Board
+
+A change control board would consist of core developers that, on a daily basis, organize changes that might
+be suitable for [release branches](#release-branches) during the feature complete.
+
+The board members typically decide on whether a change should be included on the release based on the following criteria:
+
+- Impact
+- Effort
+- Risk
+
+#### Release branches
+
+Release branches are protected branches that will be created at the beginning of feature complete. These branches are
+protected, similar to `master`, if not more so depending on the [change control board](#change-control-board) members. 
+
+The release branch will be merged into `master` upon release.
+
+```text
+                 master
+
+                   o
+                   |
+                   |    release/1.2.3
+                   |
+                   |          o commit: fix/critical-or-minor
+                   |          |
+                   |          |
+commit: feature/Y  o          |
+                   |          |
+                   |          |
+                   |          |
+                   |          |
+                   +----------+ <--+ start feature complete
+                   |
+                   |
+                   |
+                   |
+commit: feature/X  o
+                   |
+                   |
+                   +
+```
+
+#### Migration Guide
+
+A migration guide is a document which details breaking changes for migrating from prior versions.
+
+### Release Execution
+
+Upon execution of the release the following will take place:
+
+- A github release will be created containing the following:
+    - **Artifacts**
+    - **Release Notes**
+    - **Migration Guide**, when appropriate.
+- The release branch will be tagged as `v<version>`.
+- The release branch will be merged into `master`.
+
+# Prior Art
+[prior-art]: #prior-art
+
+### References
+
+* Release branches: https://docs.microsoft.com/en-us/azure/devops/repos/git/git-branching-guidance?view=azure-devops#manage-releases
+* Change control board: https://en.wikipedia.org/wiki/Change_control_board
+* Feature Complete and CCB: http://www.professionalqa.com/code-freeze
+
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
--- a/text/0000-release-process.md
+++ b/text/0000-release-process.md
@@ -9,7 +9,10 @@
 # Summary
 [summary]: #summary
 
-The release process for the various components maintained by the Cloud Native Buildpacks organization.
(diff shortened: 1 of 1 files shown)


text/0000-release-process.md
```diff
@@ -0,0 +1,194 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Timetable
+
+_The following schedules are suggestive and subject to change based on practicality._
+
+* Release Cadence: **Monthly** on the **1st Tuesday** of the month 
+* Code Freeze: **5 days** prior to scheduled release
```
Zanadar: Is this working days or calendar days? Calendar days has the code freeze starting on the last thursday of the month. Is that assumption correct? Seems like that might not be enough time to do much in the way of  UAT.

text/0000-release-process.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Migration Guide
+
+A migration guide is a document which details breaking changes for migrating from prior versions.
+
+### Release Execution
+
+Upon execution of the release the following will take place:
+
+- A github release will be created containing the following:
```
Zanadar: What about formalizing the cutting of an `-rc.1` release when we have a code freeze. This lets some of our consumers test out candidate changes. We could cut them as pre-releases and if the CCB lets in more changes, we could cut `-rc.2`, etc?

text/0000-release-process.md
```diff
@@ -0,0 +1,194 @@
...
+2. At the beginning, a [change control board](#change-control-board) is organized on a daily basis to review additional
+requested changes.
+3. During, only changes approved by the CCB are merged into the release branch.
+
+#### Change Control Board
+
+A change control board would consist of core developers that, on a daily basis, organize changes that might
+be suitable for [release branches](#release-branches) during the code freeze.
+
+The board members typically decide on whether a change should be included on the release based on the following criteria:
```
Zanadar: We should probably formalize how one joins the CCB or who comprises the CCB initially. Maybe in another RFC?

text/0000-release-process.md
jromero: That's a great idea! I hadn't thought about the possibility of providing pre-release builds. I've got a few initial thoughts:

- I wonder if it's something we want to take on right now or defer onto a separate RFC. 
- My initial concern would be that it might not be as stable as we'd like with such a short release cycle. 
- Given that we would have a short release cycle I'm not sure how much value pre-release builds would provide. We would need very tight feedback loops for reporting issues and addressing them.

text/0000-release-process.md
jromero: It should be business days.

text/0000-release-process.md
jromero: Good question. Given this is a community project rather than a team I could see some strategies that differ from common practices. 

FWIW, typically it would involve product owner, development lead, and QA lead.

text/0000-release-process.md
Zanadar: I think we could have a followup RFC on the process of release candidates, for sure. And yes I agree that sort turnaround makes this not super useful.

jromero:
Additional feedback:

- code freeze has a negative connotation - as to sound as if the developers are no longer doing anything. Really should rename to _code complete_.
- propose staggering pack/lifecycle releases as it would have additional positive properties.

jkutner:
I think an Unresolved Question is "who's on the CCB?"

Are there any drawbacks to this proposal?

Zanadar:
Moving to final comment period

Zanadar:
Disussion in WG about how to handle out of band release:
- Maybe announce on slack

Zanadar:
This to be moved into entry in `community.md`

text/0000-release-process.md
```diff
@@ -0,0 +1,192 @@
...
+### Release Execution
+
+Upon execution of the release the following will take place:
+
+- A github release will be created containing the following:
+    - **Artifacts**
+    - **Release Notes**
+    - **Migration Guide**, when appropriate.
+- The release branch will be tagged as `v<version>`.
+- The release branch will be merged into `master`.
```
Zanadar: An aspect of this should also be, "on thursday, a standing item will be discussing the upcoming releases".

What this disucssion entails, I'm not certan.

zmackie:
> Disussion in WG about how to handle out of band release:
> 
> * Maybe announce on slack

We've created a #release-planning channel in slack for these discussion

text/0000-release-process.md
hone: Is the CCB for now not just the maintainers for the subteam the project falls under?

text/0000-release-process.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Timetable
+
+_The following schedules are suggestive and subject to change based on practicality._
+
+* Release Cadence: **Monthly** on the **1st Tuesday** of the month
```
hone: Are all releases happening on the 1st Tues? Does this only apply to `pack`, to `lifecycle`? What about if we want to release one of them earlier b/c of a breaking change, do we need to wait a month then?

hone: I also wonder if a monthly release cadence is too much churn?

text/0000-release-process.md
hone: Depending on the lifecycle of the release I can imagine the following:
* At the start of a new release, rundown/celebration of previous release and discussion for the proposed features being worked on in the new one.
* During development, quick status check/any open questions on features
* Week before feature complete, a warning that it closes next Tues.

text/0000-release-process.md
```diff
@@ -0,0 +1,296 @@
...
+* Change control board: https://en.wikipedia.org/wiki/Change_control_board
+* Feature Complete and CCB: http://www.professionalqa.com/code-freeze
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Who makes up the Change Control Board, and how are meetings scheduled?
+- Will the release branch be maintained after the changes have been merged back in to master? Could, or should, this
+branch be used to create future patches?
+- Is the proposed release cadence too ambitious given historic feature delivery speed? Is spending roughly 25% of the
```
hone: * [Rust](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html#choo-choo-release-channels-and-riding-the-trains) does every 6 weeks, they also use release trains which is fairly similar here.
* [Firefox](https://hacks.mozilla.org/2019/09/moving-firefox-to-a-faster-4-week-release-cycle/) used to do 6-8 weeks, but has moved to 4 weeks.
* [Ember.js](https://github.com/emberjs/rfcs/blob/master/text/0056-improved-release-cycle.md) also does a 6 week release cadence.
* [Chrome](https://chromium.googlesource.com/chromium/src/+/master/docs/process/release_cycle.md) also releases every 6 weeks, but they take 7 weeks to stabilize any beta release.
* [Ubuntu](https://ubuntu.com/about/release-cycle) is every 6 months.

It seems like 6 weeks is the most popular cadence that I've seen.

text/0000-release-process.md
```diff
@@ -0,0 +1,296 @@
...
+time in a release window too much?
+- How is the person, or group, responsible for actually building and publishing a release determined?
```
hone: On the rails team, there's usually a release manager who's assigned and in charge of making that happen.

text/0000-release-process.md
```diff
@@ -0,0 +1,296 @@
...
+
+* Release branches: https://docs.microsoft.com/en-us/azure/devops/repos/git/git-branching-guidance?view=azure-devops#manage-releases
+* Change control board: https://en.wikipedia.org/wiki/Change_control_board
+* Feature Complete and CCB: http://www.professionalqa.com/code-freeze
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Who makes up the Change Control Board, and how are meetings scheduled?
+- Will the release branch be maintained after the changes have been merged back in to master? Could, or should, this
```
hone: Not sure if this is a separate RFC (but maybe not), there should be some policy that talks about once we release what is the support policy. The ruby core team has a release manager who's in charge of the branch and is the gatekeeper of what gets in. Do we differentiate between bug and security fixes?

hone:
Just because this RFC hasn't been accepted, I think we can trial this whole process. That being said, there are a decent number of open questions that I think that need to be answered before this moves forward.

text/0000-release-process.md
```diff
@@ -0,0 +1,296 @@
...
+                   |
+                   |
+                   +
+```
+
+#### Migration Guide
+
+A migration guide is a document which details breaking changes for migrating from prior versions.
+
+### Release Execution
```
hone: Announcements via e-mail, slack, social media(?) should be included as part of the release plan so we don't forget.

text/0000-release-process.md
jkutner: I think we should also add comms here (email, twitter, slack #announcements, etc)

jkutner:
I like @hone's suggestion to trial the process. But I think there are too many open questions to finalize as is. In particular I'd like to see more clarity on:

* The CCB (who's on it, how you get on it)
* If this policy applies to all repos (even imgutil, and libbuildpack?)
* How are security releases are handled (if we need to bypass the  regular cadence)?
* How lifecycle and pack interplay (do we stagger their release cycles to help w/ breaking changes?)

I think those points are captured well in the inline comments by others, but I wanted to reiterate.

nebhale:
Moving back out of FCP because we were a little over-eager.

text/0000-release-process.md
jromero: Yes, I think we agreed on that. I've updated the RFC to reflect that.

text/0000-release-process.md
jromero: @zmackie / @hone

After @zmackie and I tried to hash this out we incorporated the role of a release manager. (refer to updated doc for details). It does include the idea of relaying status during working group meetings but given the role is short lived (within feature complete) I do think the listed items lend themselves more to the PM role. Thoughts?

jromero: @jkutner added!

text/0000-release-process.md
jromero: Updated the summary to present this as a _template_ that only pack and lifecycle would use initially. The specifics on dates are agreed upon outside this document.

text/0000-release-process.md
jromero: Updated to 6 weeks as the initial proposal.

text/0000-release-process.md
jromero: We are opting to leave out-of-band release out of scope of this RFC, if that works for everyone. We have tried to make that clearer in the summary.

text/0000-release-process.md
jromero: Added

jromero:
@jkutner we've made some changes that should shine light on some of the questions you've gathered:

> The CCB (who's on it, how you get on it)

This is now answered. TL;DR; maintainers of component

> If this policy applies to all repos (even imgutil, and libbuildpack?)

Summary updated. This process is intended to be a "template" that will be applied to `pack` and `lifecycle`. Other components can choose to follow it.

> How are security releases are handled (if we need to bypass the regular cadence)?

Summary update. This RFC intentionally leaves out out-of-band releases to be hashed out later as they provide their own set of challenges.

> How lifecycle and pack interplay (do we stagger their release cycles to help w/ breaking changes?)

Summary update. As a template, the scheduling of interoperable components is left up to a higher order as well as future definitions in an "out-of-band" release process.

text/0000-release-process.md
```diff
@@ -0,0 +1,319 @@
...
+### References
+
+* Release branches: https://docs.microsoft.com/en-us/azure/devops/repos/git/git-branching-guidance?view=azure-devops#manage-releases
+* Change control board: https://en.wikipedia.org/wiki/Change_control_board
+* Feature Complete and CCB: http://www.professionalqa.com/code-freeze
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Could, or should, the release branch be used to create future patches once it's merged to master?
```
ekcasey: I don't see any reason to delete the branches. I think we should leave them, and use them for patches releases if necessary.

hone: I've seen this work really well in practice in both Ruby + Rails projects where needing to do bugfixes and timely security practices, this just keeps it pretty clean.

text/0000-release-process.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+#### Migration Guide
+
+A migration guide is a document which details breaking changes and provides actions that may be taken to migrate from
+prior versions.
+
+### Release Manager
+
+A release manager is a role assigned to an individual selected to own the release process. A release manager should be
```
hone: Is the release manager assigned per release or is this a more long standing role?

text/0000-release-process.md
```diff
@@ -0,0 +1,319 @@
...
+part of the component's maintainers sub-team. They may volunteer, or be selected by other members of the component's maintainers
+sub-team. Their role and responsibilities start as soon as a release going into [feature complete](#feature-complete).
+
+The responsibilities of the release manager include, but are not limited to:
+
+- Communicating status during working group meetings.
```
hone: Should we elaborate things covered in these meetings?

text/0000-release-process.md
hone: What is the distinction in your head of PM/Release Manager?

zmackie:
Outstanding Q: does this apply to lifecycle?

text/0000-release-process.md
hone: The downside of having these branches around is there's a ton of branches on the repo. The plus side is bugfixes or other contributions can be easily PR'd even outside of the subteams.

sclevine:
From WG 4/15: @jromero will update to reflect that just pack is affected for now

jromero:
> From WG 4/15: @jromero will update to reflect that just pack is affected for now

Updated

jromero:
@joe / @hone any additional clarifications necessary or concerns?

hone:
@jromero nope, just nitpick formality, that we should make sure to make it clear when unresolved questions are resolved/out of scope for future readings.

sclevine:
Moving to FCP as of 2020-05-06, will be merged 2020-05-13 assuming no other feedback is presented.

PR #34, opened 2019-12-06
Lifecycle Compatibility Verification

jromero:
[readable](https://github.com/jromero/rfcs/blob/lifecycle-compat-verification/text/0000-lifecycle-compat-verification.md)

_Continuation of #29_

### DIFF
--- a/text/0000-lifecycle-compat-verification.md
+++ b/text/0000-lifecycle-compat-verification.md
@@ -1,0 +1,331 @@
+# Meta
+[meta]: #meta
+- Name: Lifecycle Compatibility Verification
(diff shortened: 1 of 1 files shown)


jromero:
@hone  / @nebhale Any additional changes necessary for this to be approved?

nebhale:
I move for an FCP, ending 16 Jan, for merge.  If @hone approves before that day, we'll merge early.

hone:
I do like how flags feel less hidden since it can be viewed in the help of a binary. There's pretty massive benefits to the env var too. Potentially, we could do both?

PR #35, opened 2019-12-11
Client side registry

jkutner:
[Readable](https://github.com/buildpack/rfcs/blob/client-side-registry/text/0000-client-side-buildpack-registry.md)

### DIFF
--- a/text/0000-client-side-buildpack-registry.md
+++ b/text/0000-client-side-buildpack-registry.md
@@ -43,7 +43,6 @@ When a buildpack author would like to publish and share their buildpack on the r
 
 Sometimes a buildpack author may have pushed up a bad version that they wish to not be available in the index. In order to not break builds, it will not be possible to fully remove an entry from the index. Instead, the entry in the index will be marked as "yanked". This information can than be used when resolving which buildpacks to fetch.
 
-1. Log into the buildpack registry using `pack registry-login`.
 1. Use `pack yank-buildpack <namespace>/<name> <version>` command to yank a buildpack.
 
 If a buildpack author wants to undo the yank and make the buildpack version available in the index, they can use the `--undo` flag.
--- a/text/0000-client-side-buildpack-registry.md
+++ b/text/0000-client-side-buildpack-registry.md
@@ -208,16 +208,18 @@ For performance reasons, the git history will be periodically squashed. This str
 
 ### Namespace ownership
 
-Each `namespace` will be owned by a
+Each `namespace` will be owned by an entity. The mapping of this entity, called owner, to a namespace will have the following structure:
 
 ```
-{
-  "namespace" : "<string>",
-  "owner" : [
-    "id" : "<string>",
-    "type" : "<string>"
-  ]
-}
+[
+  {
+    "namespace" : "<string>",
+    "owner" : [
+      "id" : "<string>",
+      "type" : "<string>"
+    ]
+  }
+]
 ```
 
 * `namespace` - can represent a set or organization of buildpacks. matches `namespace` in the primary index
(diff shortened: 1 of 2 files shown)


jkutner:
There were a lot of questions in the WG about the directory structure, and how it relates to namespaces. There is a concern that a small set of namespace will cause the distribution of files in the repo to be poor (i.e. there will be a small number of orgs as compared to package names in crates.io).

The following alternatives were discussed:

## Digest

We could generate a digest for a namespace/name and use that to create the directories, such as:

```
a9
├── 25
│   └── example_foobar
└── 74
    └── example_foocal
```

## Namespace as top-level dirs

We could put the two-dir structure under some namespace dirs:

```
example
└──fo
    ├── ob
    │   ├── example_fooball
    │   └── example_foobar
    └── oc
        └── example_foocal
```

hone:
I was just thinking you could dir based off of the name of the buildpack but include the namespace in the file. For example:

```
fo
├── ob
│    ├── example_fooball
│    └── example_foobar
└── oc
     └── example_foocal
```

Even if it's not evenly split, is there a downside to that?

thephw:
Search feels like a good start for discoverability. What would be some other good services we could provide for discovery? Some indicators of maturity, health, and popularity of the individual buildpacks may provide additional data for folks to query what to use.

jonjohnsonjr:
Re: file layout -- consider using https://github.com/opencontainers/image-spec/blob/master/image-layout.md for some of this.

> Folders with be split by two nested folders. The reasoning for this is so the index doesn't run into a files in a folder limit.

The distribution of files under blobs/ is definitely a concern (it's come up before), but I wonder if we can resolve that upstream in the spec.

jkutner:
One bit of feedback we got from the Crates folks is that human readable index dirs/files can help significantly with debugging. For that reason, and the desire to support "handmade" PRs to the index, I want to move forward with @hone's proposal to base the dir on the name of the buildpack but include the namespace in the file. For example:

```
fo
├── ob
│    ├── example_fooball
│    └── example_foobar
└── oc
     └── example_foocal
```

text/0000-client-side-buildpack-registry.md
```diff
@@ -0,0 +1,302 @@
...
+
+If a buildpack author wants to undo the yank and make the buildpack version available in the index, they can use the `--undo` flag.
+
+```
+$ pack yank-buildpack --undo <namespace>/<name> <version>
+```
+
+## Using a Buildpack from the Registry
+
+A buildpack can be pulled from the regsitry by running `pack pull-package` or `docker pull`. However, most users will not consume the buildpack in such a raw format. The following sections describe how each persona will use the registry.
```
nebhale: This command should be symmetric, so probably `pack pull-buildpack`.

text/0000-client-side-buildpack-registry.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Manipulating the Index
+
+In general, the file will be append only for add. This will add new entries through `pack publish-buildpack`.
+
+When performing a `pack yank-buildpack`, it will rewrite the corresponding line by adjusting the `yanked` field to `true`.
+
+When performing a `pack yank-buildpack --undo`, it will rewrite the corresponding line by adjusting the `yanked` field to `false`.
+
+### Squashing the Index
```
nebhale: Why is this broadly necessary? It feels like shallow clones would be sufficient here?

text/0000-client-side-buildpack-registry.md
```diff
@@ -0,0 +1,302 @@
...
+    ]
+  }
+]
+```
+
+* `namespace` - can represent a set or organization of buildpacks. matches `ns` in the primary index
+* `owner.id` - the identifier of the user or group that owns a namespace. this will be specific to `owner.type`
+* `owner.type` - the type of owner (i.e. a Github account, a Google account, etc)
+
+Namespaces are allocated to owners on a first-come-first-serve basis. However, we will retain the right to retroactively change ownership.
```
nebhale: 👍

text/0000-client-side-buildpack-registry.md
jkutner: I don't have a strong opinion here. This was based on advise from the crates.io folks. We can dig into why they found this necessary.

nebhale: Trust in Rust.  No need to research.

hone: Indeed. I found this from an [internal rust post](https://internals.rust-lang.org/t/cargos-crate-index-upcoming-squash-into-one-commit/8440).

> Performing an incremental update of a shallow checkout is very expensive as a server operation. The CocoaPods project on GitHub [has historically run into problems](https://thenextweb.com/dd/2016/03/08/cocoapods-popular-basically-took-five-github-servers/) with this strategy. Effectively this isn’t a scalable solution on large scale.

hone:
@jkutner can you clean up this branch?

PR #36, opened 2019-12-11
RFC to remove pack run command

ekcasey:
[Readable](https://github.com/buildpack/rfcs/blob/remove-pack-run/text/0000-remove-pack-run.md)


ekcasey:
@scothis I believe (correct me if I am wrong) that you originally contributed this feature to `pack` with a `riff` use case in mind, but said use case didn't materialize in the end. Your input would be valuable here.

scothis:
That’s correct. It’s fine to remove if it’s causing confusion or a burden

Zanadar:
I think this is a great change.  Focussing on the **build** experience seems completely the right way to go. And you can alias `pack-run=docker run` for demos 😜

jromero:
@hone Are there any outstanding questions/concerns regarding the removal of `pack run`?

text/0000-remove-pack-run.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Other tools like `docker` do not provide a single command for build+run
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Do current users use the `pack run` command?
```
hone: Unrelated to this PR, it'd be great to get metrics on usage from our CLI.

text/0000-remove-pack-run.md
```diff
@@ -0,0 +1,46 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It was originally intended to support rapid build/run cycles during development, some users may find it useful for this reason.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep `pack run` the way it is
+- Invest in building out a better `pack run` experience
```
hone: I think this sounds great, but probably scoped for a different RFC.

PR #38, opened 2019-12-13
Fix urls to point to new github org

Zanadar:
- Also some whitespace cleanup


sclevine:
This just updates links due to the org rename, and doesn't propose an RFC. Merging with no votes.

PR #39, opened 2019-12-16
Add RFC for circle ci orb

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/circleci-orb/text/0000-circleci-orb.md)

### DIFF
--- a/text/0000-circleci-orb.md
+++ b/text/0000-circleci-orb.md
@@ -1,0 +1,67 @@
+# Meta
+[meta]: #meta
+- Name: Support a CircleCI Orb
+- Start Date: 2019-12-15
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal for officially supporting a [CircleCI Orb](https://circleci.com/orbs/) that runs buildpacks.
+
+# Motivation
+[motivation]: #motivation
+
+One of the questions we often get is "how do I use buildpacks in my production pipeline?" The `pack` CLI is great for local development, but integrating into a CI/CD pipeline requires a little imagination. By official supporting a CircleCI Orb, we'll make it dead simple for people to use buildpacks on CircleCI and we'll help users of other CI/CD platforms get an idea how to integrate buildpacks.
+
+# What it is
+[what-it-is]: #what-it-is
+
+We publish a `buildpacks/pack` Orb that can be used in a repo's `.circleci/config.yml` thusly:
+
+```yaml
+version: 2.1
+description: Cloud Native Buildpacks sample app
+
+orbs:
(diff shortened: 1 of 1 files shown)


jkutner:
@sclevine i've added a drawback, and an alternative section

thephw:
Do we know anyone at CircleCI or did we kind of pick it out of the air as a thing some of us use? I think we could mitigate some of the drawbacks by having those folks take ownership of it after a reference implementation.

jkutner:
@sclevine per your request, I've added a section in the *Alternatives* section on [running lifecycle directly](https://github.com/buildpacks/rfcs/blob/circleci-orb/text/0000-circleci-orb.md#run-lifecycle-instead-of-pack) and outlined its drawbacks. PTAL

hone:
@thephw I have a friend who works there and I'm trying to see if they can help give us pointers. From talking with him, I don't think the Community/Partner Engineering wants to take ownership over this. As for a reason we picked it, it's one of the more public cloud CI services. We're also using it internally here at Salesforce + Heroku for some things.

text/0000-circleci-orb.md
```diff
@@ -0,0 +1,122 @@
...
+    jobs:
+      - pack/build:
+          image-name: myimage
+          builder: heroku/buildpacks:18
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+We move the [jkutner/buildpacks-orb](https://github.com/jkutner/buildpacks-orb) repo into the `github.com/buildpacks` org, and publish it as `buildpacks/pack`
```
hone: We should clarify what subteam is responsible for owning this once it is moved over even if it's obvious. In this case, it'd be the platform team.

text/0000-circleci-orb.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+In the future, we should implement an orb/job that runs lifecycle independently from pack. Ideally if we have a `/lifecycle/prepare` or similar phase that handles `project.toml` and setup/download of buildpacks. But in order to support the full feature set of pack we will start with the proposal in this RFC.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [circleci/docker](https://circleci.com/orbs/registry/orb/circleci/docker)
+
+# Unresolved Questions
```
hone: Should we use this as the start for documentation on "How to move beyond `pack`"? What I mean by that, is the process of how this is implemented just as important as the working thing so others can learn how to use CNB in their own prod environment. Sample vs Implementation.

text/0000-circleci-orb.md
```diff
@@ -0,0 +1,124 @@
...
```
hone: @jromero brought up a good question in one of the WG meetings that we should make note of, what is the criteria for platforms we're going to support especially for 3rd party integrations like Tekton, CircleCI, Jenkins, etc?

PR #40, opened 2019-12-17
Add RFC to define mixins for the Bionic stack

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/bionic-mixins/text/0000-bionic-mixins.md)


sclevine:
@jonjonsonjr as discussed

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+```
+run stackID: "io.buildpacks.stacks.bionic"
+run mixins: ["git"]
+build stackID: "io.buildpacks.stacks.bionic"
+build mixins: ["build:build-essentials", "git"]
+```
+
+And buildpack authors would update their `buildpack.toml` files from:
+```toml
+[stack]
```
jkutner: `[[stacks]]`?

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+run mixins: []
+build stackID: "org.cloudfoundry.stacks.cfbionicstack"
+build mixins: []
+```
+To:
+```
+run stackID: "io.buildpacks.stacks.bionic"
+run mixins: ["git"]
+build stackID: "io.buildpacks.stacks.bionic"
+build mixins: ["build:build-essentials", "git"]
```
Zanadar: `build:build-essential` (I'm assuming per the above text this should match the package name

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+And buildpack authors would update their `buildpack.toml` files from:
+```toml
+[[stacks]]
+id = "org.cloudfoundry.stacks.cfbionicstack"
+```
+To:
+```toml
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
+mixins = ["build:build-essentials", "git"] # NOTE: only need to list mixins they actually require!
```
Zanadar: `build:build-essential`

Zanadar:
Think @sclevine you meant @jonjohnsonjr

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+
+- This could encourage buildpack authors and app developers to create a large number of unmaintained stack images that have different sets of Ubuntu packages.
+- Breaking change for buildpack authors once old stack images are removed.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Neither the spec nor this RFC suggest whether or not stack authors or app developers are allowed to make additional changes to stack images without using mixin metadata.
+- This RFC doesn't propose a way for the project to describe changes to `ubuntu:bionic` that are not official LTS Ubuntu 18.04 packages.
+- When package groups (like `build-essential`) or dependent packages are installed, must stack images always specify all the packages added? Or is it only necessary when stack image authors want buildpacks to be able to depend on the dependent packages?
```
nebhale: The latter.  In order to make a mixin as broadly compatible as possible you'll probably want to list both the package group and all the packages that it includes.

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+We would change the image metadata from:
+```
+run stackID: "org.cloudfoundry.stacks.cfbionicstack"
+run mixins: []
+build stackID: "org.cloudfoundry.stacks.cfbionicstack"
+build mixins: []
+```
+To:
+```
+run stackID: "io.buildpacks.stacks.bionic"
```
hone: I would imagine that you may keep the `cfbionicstack` as a stackID. A stack image may not just be the base image + some packages. I was originally expecting some linkage that `heroku-18` is `ubuntu:bionic` plus some extra mixins as far as a buildpack is concerned, but is also its own stackID.

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Binoic."
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Motivation
+[motivation]: #motivation
+
+This will allow:
+1. Buildpack authors to publish buildpacks that are compatible with more stack images without requiring those authors to explicitly specify vendor-specific stack IDs.
+2. Vendor-specific buildpacks to work together in the same build, as long as those buildpacks support `io.buildpacks.stacks.bionic`.
```
hone: Definitely interested in this use case on the Heroku side as we've working more with the Riff folks. Forking a buildpack just to change the `buildpack.toml` isn't great.

text/0000-bionic-mixins.md
ekcasey: Even if the image includes things beyond "base image + some packages", do the buildpacks depend on those differences to execute successfully?

If not, these images can still declare their stack ID to be `io.buildpacks.stacks.bionic` despite these differences. If so, these buildpacks won't be interoperable with other platforms.

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,75 @@
...
+This RFC proposes that we formalize the existing definition of `io.buildpacks.stacks.bionic` and define mixins on that stack to be Ubuntu 18.04 LTS package names (without version or architecture info) from official package mirrors. 
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Binoic."
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Motivation
+[motivation]: #motivation
+
+This will allow:
+1. Buildpack authors to publish buildpacks that are compatible with more stack images without requiring those authors to explicitly specify vendor-specific stack IDs.
```
briandealwis: How can a platform specify that their custom stack adheres to `io.buildpacks.stacks.bionic`?  (How would `heroku/buildpacks:18` transition, considering there are Buildpacks in the wild relying on the `heroku-18` stack ID?)

text/0000-bionic-mixins.md
jabrown85: I gave defining a buildpack properly with mixins a shot. It gets complicated. I would bet most buildpack authors don't even know what they depend on that is not in base bionic without some serious work.

I had to build a build/run image that was _just_ bionic to test my buildpack and what it depends on. Then, I had to run as root to install packages and basically play whack-a-mole as I ran into errors while building an example app. What's the incentive for buildpack authors to do this work? How do they know if they are relying on a package's dependency and it gets removed in next version?

Also, all the buildpacks I tried the process on needed `curl` and `ca-certificates`. If you've only ever built buildpacks against CF/HK images, I can imagine you would never even know those packages exist and just assume you don't need to list them. It seems likely that a great majority of the buildpacks would have a lot of the same packages.

Finally, what about packages that the buildpack doesn't rely on itself, but the majority of the target audience for the buildpack would require? This came up when discussing with @hone. He probably has a real concrete example package but say that most of our customers build ruby apps that require `libreadline`. The buildpack itself doesn't _require_ it and it's optional for building ruby itself but the customer's code it is expecting to build most likely would need it. We could list it as a `mixin` for the buildpack but that would essentially tie the buildpack to Heroku's bionic images that include it. Users who build ruby code that doesn't rely on ruby built with `libreadline` can't use the buildpack on another `bionic` image either.

jabrown85:
Here are some more thoughts after our chat @sclevine @hone @jkutner 

I think we all agreed that the `io.buildpacks.stacks.bionic` stack really needs to be `bionic` + a few essentials (e.g. `ca-certificates`, `libssl`).

I think we could enhance `stack.id` to give buildpack developers a better experience without them needing to understand the complexity.

An example might be a buildpack author for `company/ruby` would be encouraged to target `io.buildpacks.stacks.bionic#language-essentials`. `language-essentials` in this case could be `libxml`, `libreadline`, and `libuv`. Resolution of the stack still resolves to `io.buildpacks.stacks.bionic` but the definition of those _buildpack mixin packs?_ would be part of the platform to ensure the best compatibility.

A `#bash-essentials` pack of mixins maybe includes `jq` and `yj`. This could be the recommended stack for buildpack developers who aren't compiling code as a part of their buildpack.

I would limit buildpack authors to a single pack identifier when defining their stack id. I wouldn't want `io.buildpacks.stacks.bionic#language-essentials#bash-essentials` to be accepted. The limitation of a single pack of mixins could be combined with additional `mixins`. So you could require `imagemagick` in addition to the ones provided by the mixin pack. I also would expect a stack of `io.buildpacks.stacks.bionic` and mixins of `io.buildpacks.mixinpacks.bionic.language-essentials,io.buildpacks.mixinpacks.bionic.bash-essentials` to work just fine.

Later, if a buildpack is failing to build on a new stack that doesn't work because the buildpack author is targeting a pack that includes something they are not actually required to use, the author could then go learn what exactly is in that pack and start specifying the mixins in a more selective way. They could stop requiring `yj` but start detecting when it's available during execution to improve compatibility across stacks. This is the sort of work I would expect buildpack authors at Heroku/Pivotal would do in order to target tiny stacks but only because we are paid to work on them :) 

Anyway, that is what I was thinking while we were chatting earlier. I think having a few mixin packs targeted at the buildpack author's purpose could lower the barrier. Once they hit a limitation, the existing granularity of the mixinx would allow the buildpack author to take that dive into understanding stacks and mixins.

text/0000-bionic-mixins.md
sclevine: So your suggestion is that stack authors should add all packages to the mixin list, but are not required to? Would this imply that a given stack run/build image can have arbitrary contents that aren't specified by the ID or mixins, as long as those contents don't disappear when newer versions are published?

text/0000-bionic-mixins.md
sclevine: Platforms could publish new images with the `io.buildpacks.stacks.bionic` ID  to new image tags and deprecate their bespoke stack image tags once everyone has migrated.

Alternatively, platforms could give buildpack authors a certain amount of time to add both stacks to their buildpack.toml files before publishing images with the `io.buildpacks.stacks.bionic` ID over the existing image tags.

sclevine:
@hone @jkutner @jabrown85 updated based on the feedback above.

I recorded the `io.buildpacks.stacks.bionic#language-essentials` suggestion as an unresolved question for a future RFC because it would require significant changes to the spec, lifecycle, pack, etc. Let me know if my understanding of that suggestion is inaccurate.

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,85 @@
...
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Proposal
+[proposal]: #proposal
+
+This RFC proposes that we define `io.buildpacks.stacks.bionic` such that:
+1. The base set of packages is `ubuntu:bionic` with `ca-certificates`, `libssl1.1`, and `openssl` added.
+2. Mixins without an `=` are Ubuntu 18.04 LTS package names without version or architecture info from official package mirrors.
+3. Mixins that begin with `set=` are defined by the project as such:
+  a. `build:set=shell-utils` represents the change set: `build:curl`, `build:jq`, and yj CLI at build-time.
```
jabrown85:
```suggestion
  a. `build:set=shell-utils` represents the change set: `build:curl`, `build:jq`, and `build:yj` at build-time.
```

text/0000-bionic-mixins.md
```diff
@@ -9,13 +9,23 @@
...
 Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
 
+# Proposal
+[proposal]: #proposal
+
+This RFC proposes that we define `io.buildpacks.stacks.bionic` such that:
+1. The base set of packages is `ubuntu:bionic` with `ca-certificates`, `libssl1.1`, and `openssl` added.
+2. Mixins without an `=` are Ubuntu 18.04 LTS package names without version or architecture info from official package mirrors.
+3. Mixins that begin with `set=` are defined by the project as such:
+  a. `build:set=shell-utils` represents the change set: `build:curl`, `build:jq`, and yj CLI at build-time.
```
hone:
```suggestion
  a. `build:set=shell-utils` represents the change set: `build:curl`, `build:jq`, and `yj` CLI at build-time.
```

text/0000-bionic-mixins.md
sclevine: yj is not available as an Ubuntu package, so I went with "yj CLI at build-time" instead of "`build:yj`."

Maybe it's worth defining `build:set=yj` as well?

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,85 @@
...
+The Cloud Native Buildpacks project informally defines stack build and run images with ID `io.buildpacks.stacks.bionic` as containing exactly the set of Ubuntu packages in the `ubuntu:bionic` image on Docker Hub.
+This RFC proposes that we formalize the existing definition of `io.buildpacks.stacks.bionic` and define mixins on that stack to be Ubuntu 18.04 LTS packages. 
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Binoic."
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Proposal
+[proposal]: #proposal
+
+This RFC proposes that we define `io.buildpacks.stacks.bionic` such that:
```
hone: Does it expand the scope too much to set a framework for adding new stacks under the `io.buildpacks.stacks` namespace or is that just a separate proposal. For instance, I think setting the `set=` pattern is probably more valuable then just defining what all the sets are in this proposal. @sclevine

sclevine: We could define `set=` for `io.buildpacks.stacks.*`, but it might be safer to hold off until we have additional stacks to define it for. The `set=` convention may be applicable to most base images that use OS packages today, but container base images could change in unexpected ways in the future.

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,85 @@
...
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The Platform API specification allows stack maintainers to define stack mixins for a given stack.
+The Cloud Native Buildpacks project informally defines stack build and run images with ID `io.buildpacks.stacks.bionic` as containing exactly the set of Ubuntu packages in the `ubuntu:bionic` image on Docker Hub.
+This RFC proposes that we formalize the existing definition of `io.buildpacks.stacks.bionic` and define mixins on that stack to be Ubuntu 18.04 LTS packages. 
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Binoic."
```
briandealwis: I don't understand this requirement: equivalency seems to be key. For example, say I'm authoring a buildpack for  small niche language.  I *want* to be able to assert that my buildpack should work on anything claiming to be equivalent to Bionic.  I don't see the use of asserting that it will only work for `ubuntu:bionic`.

```suggestion
In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Bionic."
```

briandealwis:
```suggestion
In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Bionic."
```

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,85 @@
...
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
```
briandealwis: I find this language a bit confusing.  On first read, I thought it meant that the images *must be derived from* `ubuntu:bionic`, but I think it really means *provides equivalent functionality*  to `ubuntu:bionic`.  Otherwise images like `gcr.io/gcp-runtimes/ubuntu_18_0_4` won't qualify.   And it may be worth expanding _functionality_: does it mean that `/var/lib/dpkg` should be retained and that new packages can be installed with `apt-get`?

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,87 @@
...
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Bionic."
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Proposal
+[proposal]: #proposal
+
+This RFC proposes that apply the following set of restrictions to base images marked with stack ID `io.buildpacks.stacks.bionic`:
+1. The base set of packages is the set of packages distributed in `ubuntu:bionic` with `ca-certificates`, `libssl1.1`, and `openssl` added. The base image does not have to be based on `ubuntu:bionic`, but must contain packages the provide exactly the same functionality to processes running as the CNB-specified non-root user.
+2. When invoked as root, `apt-get` must be configured to use at least the set of package repositories as `ubuntu:bionic`.  `apt-get update` may be needed to download the package repository indices before `apt-get install` is functional.
```
sclevine: @briandealwis do these clarifications address your concerns?

text/0000-bionic-mixins.md
```diff
@@ -0,0 +1,87 @@
...
+
+In general, a stack is a contract that should be more like "Ubuntu Bionic" and less like "Platform XYZ's Re-distribution of Ubuntu Bionic."
+Stack images can still be provided by platforms/vendors, but they should use the standard CNB stack contracts where applicable.
+
+# Proposal
+[proposal]: #proposal
+
+This RFC proposes that apply the following set of restrictions to base images marked with stack ID `io.buildpacks.stacks.bionic`:
+1. The base set of packages is the set of packages distributed in `ubuntu:bionic` with `ca-certificates`, `libssl1.1`, and `openssl` added. The base image does not have to be derived from the `ubuntu:bionic` image, but must contain packages that provide exactly the same functionality to processes running as the CNB-specified, non-root user.
+2. When invoked as root, `apt-get` must be configured to use at minimum the set of package repositories that `ubuntu:bionic` is configured to use. `apt-get update` may be needed to download the package repository indices before `apt-get install` is functional.
```
sclevine: @briandealwis do these clarifications address your concerns?

sclevine:
@hone added the final location of the definition of `io.buildpacks.stacks.bionic` as an unresolved question, as requested.

text/0000-bionic-mixins.md
briandealwis: lgtm, thanks

PR #43, opened 2020-01-14
RFC to isolate registry creds from builder images

ekcasey:
[Readable](https://github.com/ekcasey/rfcs/blob/dont-trust-builders/text/0000-dont-trust-builders.md)

### DIFF
--- a/text/0000-dont-trust-builders.md
+++ b/text/0000-dont-trust-builders.md
@@ -24,11 +24,11 @@ Users may not realize that credentials are given to builder images and experimen
 # What it is
 [what-it-is]: #what-it-is
 
-### Ephemeral Exporter Images
+### Ephemeral Lifecycle Images
 
-When `pack build` executes and is provided the `--publish` flag, it will generate an ephemeral "exporter image" on the fly, by adding a single layer container the correct lifecycle to a scratch image. `pack` will then use this image to created the `analyze`, `restore`, and `export` containers. `pack` will clean up the image when the build completes.
+When `pack build` executes and is provided the `--publish` flag, it will generate an ephemeral "lifecycle image" on the fly, by adding a single layer container the correct lifecycle to a scratch image. `pack` will then use this image to created the `analyze`, `restore`, and `export` containers. `pack` will clean up the image when the build completes.
 
-When generating exporter images `pack` will read the lifecycle version from the builder metadata and fetch the matching lifecycle version from the lifecycle github release assets.
+When generating lifecycle images `pack` will read the lifecycle version from the builder metadata and fetch the matching lifecycle version from the lifecycle github release assets.
 
 If the lifecycle version specified in a builder image does not match any known lifecycle release, `pack build` will fail with a helpful message.
 
@@ -57,7 +57,7 @@ Therefore, `pack` will treat every builder as a "trusted builder" to [potentiall
 # Prior Art
 [prior-art]: #prior-art
 
-Creating ephemeral exporter images is similar to work `pack build` already does when it creates ephemeral builder images (e.g when it adds another layer because of the `--buildpack`).
+Creating ephemeral lifecycle images is similar to work `pack build` already does when it creates ephemeral builder images (e.g when it adds another layer because of the `--buildpack`).
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-dont-trust-builders.md
+++ b/text/0000-dont-trust-builders.md
@@ -2,7 +2,7 @@
 [meta]: #meta
 - Name: Isolate Registry Credentials from Builder Images
 - Start Date: 2020-01-14
-- CNB Pull Request:
+- CNB Pull Request: [rfcs#43](https://github.com/buildpacks/rfcs/pull/43)
 - CNB Issue:
 - Supersedes: N/A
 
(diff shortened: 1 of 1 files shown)


text/0000-dont-trust-builders.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+??
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+- Registry credential isolation in irrelevant when `--publish` is not supplied, should we prioritize performance or consistency in this case?
+- Is there a better name for this concept than `exporter` image
```
jromero: Could we tie it in with https://github.com/buildpacks/lifecycle/issues/191 and have a lifecycle image instead? One that could be multi-purpose for running exporter for this use case and also other phases for others.

ekcasey: The name "exporter image" might be misleading. Per the text of this RFC the entire lifecycle would be present in the image.

I am in favor of doing buildpacks/lifecycle#191, but I don't think the order of operations matters much here. The lifecycle/exporter image can exist with or without that optimization.

text/0000-dont-trust-builders.md
```diff
@@ -0,0 +1,70 @@
...
+### `--trust-builder` Flag
+If `pack build` is run with the `--publish` and `--trust-builder` flag, then all lifecycle steps will be run in a single container.
+The registry credentials will be provided only to `analyzer` and `exporter` processes.
+
+Users can run `pack trust-builder <build-image>` to permanently trust a builder. 
+
+### The Daemon Image Exception
+
+When `pack build` is run without the `--publish` flag, registry credentials are not necessary.
+Therefore, `pack` will treat every builder as a "trusted builder" to optimize performance.
```
hone: What's the optimization?

text/0000-dont-trust-builders.md
```diff
@@ -0,0 +1,70 @@
...
+- Keep trusting builders but plaster warnings everywhere
+
+# Prior Art
+[prior-art]: #prior-art
+
+??
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+- Registry credential isolation in irrelevant when `--publish` is not supplied, should we prioritize performance or consistency in this case?
```
hone: What does the consistency case look like?

text/0000-dont-trust-builders.md
jromero: Alternative considerations: I think we discussed this in passing at some point.

1. We could take only the `lifecycle` layer from the builder image and create an ephemeral container.
2. A lifecycle layer should have a recognizable digest per version/OS which we can "verify". If they don't match the user could still use it via `--trust-builder` or `--force` options.

Ultimately, I'd hate for users to have to think about another image to be honest.

text/0000-dont-trust-builders.md
```diff
@@ -0,0 +1,70 @@
...
+### The Daemon Image Exception
+
+When `pack build` is run without the `--publish` flag, registry credentials are not necessary.
+Therefore, `pack` will treat every builder as a "trusted builder" to optimize performance.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Adding another image (the `exporter` image) makes the mechanics of `pack build` harder for users to understand
+* CNB team must maintain & publish exporter images
```
hone: Are you thinking this belongs under the implementation team?

text/0000-dont-trust-builders.md
ekcasey: #46 details potential optimizations

text/0000-dont-trust-builders.md
ekcasey: > 1. We could take only the lifecycle layer from the builder image and create an ephemeral container.

If we don't trust the image we can't trust it's implementation of lifecycle

ekcasey: > 2. A lifecycle layer should have a recognizable digest per version/OS

Given that we don't ship the lifecycle as a layer, I don't think we should use a layer digest as an identifier. builders don't have to be created with `pack` in which case difference in timestamps history could change the digest. Compression algorithms are another potential difference

ekcasey: Even if we could verify the lifecycle implementation running in the builder container opens up other attack vectors.

text/0000-dont-trust-builders.md
ekcasey: It would involve running `analyzer` and `exporter` in trusted containers, even though there are no registry credentials to protect

text/0000-dont-trust-builders.md
ekcasey: After some thought, I am wondering if we can avoid publishing images and instead create ephemeral images on the fly. All of the logic we need to make that work can be repurposed from the implementation of `pack create-builder`.

ekcasey: I updated the RFC to reflect this change

text/0000-dont-trust-builders.md
```diff
@@ -0,0 +1,70 @@
...
+[what-it-is]: #what-it-is
+
+### Exporter Images
+
+For every release of the lifecycle we will publish an `exporter` image to `docker.io/cnbs/exporter:<version>`,
+where `<version>` is the version of the lifecycle and the image that contains only lifecycle in a scratch image.
+
+When `pack build` executes and is provided the `--publish` flag, it will use the `exporter` image to created the `analyze`, `restore`, and `export` containers.
+
+If the lifecycle version specified in a builder image does not match any known `exporter` image `pack build` will fail
```
hone: For the helpful error message, do we want to show people how to create their own lifecycle image or point people to trusting the builder?

text/0000-dont-trust-builders.md
hone: I think `exporter` is maybe not a great name b/c it's used for both analyze and export phases. Are you against just using "lifecycle" here?

text/0000-dont-trust-builders.md
hone: seems like the actual optimizations are out of scope of this RFC. Thoughts on rephrasing? We can replace the link if/when the other RFC gets merged.

```suggestion
Therefore, `pack` will treat every builder as a "trusted builder" to [potentially optimize performance](https://github.com/buildpacks/rfcs/pull/46).
```

text/0000-dont-trust-builders.md
ekcasey: It would point people towards trusting the builder.

text/0000-dont-trust-builders.md
ekcasey: If we build the image on the fly, users may not even need a name for this concept. But I agree lifecycle image is better than `exporter`. I'll update this.

text/0000-dont-trust-builders.md
hone: I'm :+1: optimizing performance here since that's probably the default getting started path for most people. It doesn't seem like there isn't any material difference to the end user by doing consistency.

text/0000-dont-trust-builders.md
hone: Does building an image on the fly help here?

hone: I wonder if we should do that and/or list valid versions of lifecycle

hone: From slack with @ekcasey:

> I think folks who contribute to lifecycle are the people who will realistically run into this error. If you create a builder with your dev version of the lifecycle, you haven’t made a mistake per se but it can’t run properly w/o being trusted
> we could instead lift valid lifecycles and assume that anyone making their own lifecycle should know what to do?

hone:
@ekcasey can you fix the DCO check on my suggestion commit :P

nebhale:
I move that this RFC be sent to FCP with a disposition of merge, closing 1700 PT, February 6, 2019.

PR #46, opened 2020-01-22
RFC: lifecycle/all

ekcasey:
[Readable](https://github.com/buildpacks/rfcs/blob/lifecycle-all/text/0000-lifecycle-all.md)

### DIFF
--- a/text/0000-lifecycle-all.md
+++ b/text/0000-lifecycle-all.md
@@ -54,7 +54,7 @@ The following flags optional flags can customize behavior:
 | `-cache-dir`    | optional | `CNB_CACHE_DIR`        | -                         | path to cache directory |
 | `-cache-image`  | optional | `CNB_CACHE_IMAGE`      | -                         | cache image tag reference |
 | `-order`        | optional | `CNB_ORDER_PATH`       | `/cnb/order.toml`         | path to order.toml |
-| `-stack`        | optional | `CNB_STACK_PATH`       | -                         | path to stack.toml |
+| `-stack`        | optional | `CNB_STACK_PATH`       | `/cnb/stack.toml`         | path to stack.toml |
 | `-launch-cache` | optional | `CNB_LAUNCH_CACHE_DIR` | -                         | path to launch cache directory |
 | `-launcher`     | optional | `CNB_LAUNCHER_PATH`    | `/cnb/lifecycle/launcher` | path to launcher binary |
 | `-buildpacks`   | optional | `CNB_BUILDPACKS_DIR`   | `/cnb/buildpacks`         | path to buildpacks directory |
--- a/text/0000-lifecycle-all.md
+++ b/text/0000-lifecycle-all.md
@@ -64,6 +64,7 @@ The following flags optional flags can customize behavior:
 | `-version`      | optional | -                      | false                     | show version |
 | `-skip-restore` | optional | -                      | false                     | do not restore metadata from previous image or data from cache
 | `-tag` (multiples allowed) | optional | -           |                           | additional tags to apply to exported image
+| `-previous-image`| optional | -                     | `<image-name>`              | image to analyze and reuse layers from
 
 Most of the these flags are existing lifecycle phase flags. `-skip-restore` is a a new addition that modifies the behavior of `analyzer` similarly to the existing `-skip-layers` flag, and, in addition, skips the `restorer` phase all together.
 
--- a/text/0000-lifecycle-all.md
+++ b/text/0000-lifecycle-all.md
@@ -28,7 +28,7 @@ The existence of `all` will constitute a non-breaking but substantive change to 
 
 This new binary supports two main goals.
 
-### Goal 1: faster `pack`
+### Goal 1: faster `pack` builds
 By creating a single container and invoking `/cnb/lifecycle/all` instead of creating a container per lifecycle phase, pack can shave approximate `6s` from the execution time of `pack build`.
 
 `pack` may not always want to do this (see https://github.com/buildpacks/rfcs/pull/43 for an explanation of why this is not always ideal). However, it could be very beneficial to users that are wiling to trust their builder images.
@@ -40,7 +40,6 @@ As members of the community experiment with incorporating CNBs into other platfo
 [what-it-is]: #what-it-is
 
 ### Usage
-USAGE:
 
 `/cnb/lifecycle/all <image-name>`
 
@@ -70,7 +69,7 @@ Most of the these flags are existing lifecycle phase flags. `-skip-restore` is a
 
 This RFC proposes a `-tag` flag as the interface to provide additional tags, rather than multiple positional arguments (as is currently accepted by `exporter` ) for clarity and it's similarity to `docker build`.
 
-Some existing lifecycle flags (e.g. `-group` on the `detector` `builder` and `exporter` phases) can are not necessary. They refer to files that were previously used to pass information between lifecycle phase processes. When the phases run in the same process, those values can be passed in memory.
+Some existing lifecycle flags (e.g. `-group` on the `detector` `builder` and `exporter` phases) are not necessary. They refer to files that were previously used to pass information between lifecycle phase processes. When the phases run in the same process, those values can be passed in memory.
 
 
 ### User
@@ -81,7 +80,7 @@ In the daemon case pack will run `all` as `root` but the buildpacks' `/bin/detec
 
 ### Credential Management
 
-Right now, when `pack` runs w/ the `--publish` flag `analyzer`,and `exporter` are provided with registry credentials via the `CNB_REGISTRY_AUTH` environment variable.
+Right now, when `pack` runs w/ the `--publish` flag `analyzer` and `exporter` are provided with registry credentials via the `CNB_REGISTRY_AUTH` environment variable.
 
 In the registry case pack will set `CNB_REGISTRY_AUTH` when invoking `all`. To prevents buildpacks from having read access to those credentials the build and detect implementations will ensure this variable is not present in the environment of the `/bin/detect` and `/bin/build` processes.
 
@@ -101,6 +100,8 @@ Require platforms/integrations to invoke all lifecycle phases in order in all si
 [prior-art]: #prior-art
 
 Most in container build tools do not spin up multiple containers.
+
+Regarding the introduction of the `-tag` flag, `docker build -t` is prior art.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-lifecycle-all.md
+++ b/text/0000-lifecycle-all.md
@@ -1,0 +1,99 @@
+# Meta
+[meta]: #meta
+- Name: Lifecycle All Binary
+- Start Date: 2020-01-21
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+### `/cnb/lifecycle/all`
+A new lifecycle binary shall be found at the path `/cnb/lifecycle/all` within a lifecycle archive or builder image. When invoked will run the following phases sequentially:
+
+* detector
+* analyzer
+* restorer
+* builder
+* exporter
+
+### `/cnb/lifecycle/*` are still available
+Each of these phases will continue to be available individually as provided by the current lifecycle.
+`launcher` will remain a separate binary
+
+### Platform API 0.3
+The existance of `all` will constitute a non-breaking but substantive change to the Platform API bring the subsequent lifecycle release to Platform API 0.3. Given that we do not have a mechanism for indicating breaking and nonbreaking platform API change pre 1.0, 
+
+# Motivation
+[motivation]: #motivation
+
+This new binary supports two main goals.
+
+### Goal 1: faster `pack`
+By creating a single container and invoking `/cnb/lifecycle/all` instead of creating a container per lifecycle phase, pack can shave approximate `6s` from the execution time of `pack build`.
+
+`pack` may not always want to do this (see https://github.com/buildpacks/rfcs/pull/43 for an explanation of why this is not always ideal). However, it could be very beneficial to users that are wiling to trust their builder images.
+
+### Goal 2: Easier integration of the lifecycle into other platforms and CI tools
+As members of the community experiment with incorporating CNBs into other platform and CI tools, the difficulty of invoking each lifecycle phase correctly and sequentially draws many to the easier solution of using `pack` to orchestrate the lifecycle. While `pack` can provide a simple way for these developers to implement these integrations, there may be performance drawbacks when compared with using the lifecycle directly. `/cnb/lifecycle/all` will provide an easier interface, lowering the barrier to using the lifecycle without pack, thus enabling platform developers choose between `pack` and `lifecycle` based solely upon the needs of their integration.
+
+# What it is
+[what-it-is]: #what-it-is
+
+### Usage
+USAGE:
+
+`/cnb/lifecycle/all <image-name>`
+
+The following flags optional flags can customize behavior:
+
+| Flag            | Required | Env var                | Default                   | Description |
+| --------------- | -------- | ---------------------- | ------------------------- | ------------|
+| `-layers`       | optional | `CNB_LAYERS_DIR`       | `/layers`                 | path to layers |
+| `-platform`     | optional | `CNB_PLATFORM_DIR`     | `/platform`               | path to platform directory |
+| `-run-image`    | optional | `CNB_RUN_IMAGE`        | -                         | reference to run image |
+| `-app`          | optional | `CNB_APP_DIR`          | `/workspace`              | path to app directory |
+| `-cache-dir`    | optional | `CNB_CACHE_DIR`        | -                         | path to cache directory |
+| `-cache-image`  | optional | `CNB_CACHE_IMAGE`      | -                         | cache image tag reference |
+| `-order`        | optional | `CNB_ORDER_PATH`       | `/cnb/order.toml`         | path to order.toml |
+| `-stack`        | optional | `CNB_STACK_PATH`       | -                         | path to stack.toml |
+| `-launch-cache` | optional | `CNB_LAUNCH_CACHE_DIR` | -                         | path to launch cache directory |
+| `-launcher`     | optional | `CNB_LAUNCHER_PATH`    | `/cnb/lifecycle/launcher` | path to launcher binary |
+| `-buildpacks`   | optional | `CNB_BUILDPACKS_DIR`   | `/cnb/buildpacks`         | path to buildpacks directory |
+| `-daemon`       | optional | `CNB_USE_DAEMON`       | false                     | export to docker daemon |
+| `-uid`          | optional | `CNB_USER_ID`          | -                         | UID of user in the stack's build and run images |
+| `-gid`          | optional | `CNB_GROUP_ID`         | -                         | GID of user's group in the stack's build and run images |
+| `-version`      | optional | -                      | false                     | show version |
+| `-skip-restore` | optional | -                      | false                     | do not restore metadata from previous image or data from cache
+
+Most of the these flags are existing lifecycle phase flags. `-skip-restore` is a a new addition that modifies the behavior of `analyzer` similarly to the existing `-skip-layers` flag, and skips the `restorer` phase all together.
+
+
+Some existing lifecycle flags (e.g. `-group` on the `detector` `builder` and `exporter` phases) can are not necessary. They refer to files that were previously used to pass information between lifecycle phase processes. When the phases run in the same process, those values can be passed in memory.
+
+ ### Credential Management
+
+// TODO
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+// TODO
+
+# Alternatives
+[alternatives]: #alternatives
+
+// TODO
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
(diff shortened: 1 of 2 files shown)


text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,108 @@
...
+| `-version`      | optional | -                      | false                     | show version |
+| `-skip-restore` | optional | -                      | false                     | do not restore metadata from previous image or data from cache
+| `-tag` (multiples allowed) | optional | -           |                           | additional tags to apply to exported image
+
+Most of the these flags are existing lifecycle phase flags. `-skip-restore` is a a new addition that modifies the behavior of `analyzer` similarly to the existing `-skip-layers` flag, and, in addition, skips the `restorer` phase all together.
+
+This RFC proposes a `-tag` flag as the interface to provide additional tags, rather than multiple positional arguments (as is currently accepted by `exporter` ) for clarity and it's similarity to `docker build`.
+
+Some existing lifecycle flags (e.g. `-group` on the `detector` `builder` and `exporter` phases) can are not necessary. They refer to files that were previously used to pass information between lifecycle phase processes. When the phases run in the same process, those values can be passed in memory.
```
ashwin-venkatesh:
```suggestion
Some existing lifecycle flags (e.g. `-group` on the `detector` `builder` and `exporter` phases) are not necessary. They refer to files that were previously used to pass information between lifecycle phase processes. When the phases run in the same process, those values can be passed in memory.
```

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,108 @@
...
+
+### User
+
+Right now, when `pack` runs w/o the `--publish` flag `analyzer` and `exporter` are run as root. This is required so that these phases can connect to the mounted docker daemon socket. Historically the `builder` and `detector` binaries are never run as root.
+
+In the daemon case pack will run `all` as `root` but the buildpacks' `/bin/detect` and `/bin/builder` scripts will be invoked as the provided user. This will require a change to build and detect implementations.
+
+### Credential Management
+
+Right now, when `pack` runs w/ the `--publish` flag `analyzer`,and `exporter` are provided with registry credentials via the `CNB_REGISTRY_AUTH` environment variable.
```
ashwin-venkatesh:
```suggestion
Right now, when `pack` runs w/ the `--publish` flag `analyzer` and `exporter` are provided with registry credentials via the `CNB_REGISTRY_AUTH` environment variable.
```

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,108 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+Require platforms/integrations to invoke all lifecycle phases in order in all situations, even when running in a single container.
+
+# Prior Art
+[prior-art]: #prior-art
+
+Most in container build tools do not spin up multiple containers.
+
```
ashwin-venkatesh: You could add the usage of the `tag` flag from docker as prior art.

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,108 @@
...
+### Platform API 0.3
+The existence of `all` will constitute a non-breaking but substantive change to the Platform API, bringing us to Platform API 0.3 . Although platform will interpret this as a breaking change, we do not have a mechanism for indicating non-breaking platform API change pre 1.0. Bumping the platform API number is necessary so that platforms like `pack` know whether or no they can use this feature.
+
+# Motivation
+[motivation]: #motivation
+
+This new binary supports two main goals.
+
+### Goal 1: faster `pack`
```
ashwin-venkatesh:
```suggestion
### Goal 1: faster `pack` builds
```

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,108 @@
...
+`pack` may not always want to do this (see https://github.com/buildpacks/rfcs/pull/43 for an explanation of why this is not always ideal). However, it could be very beneficial to users that are wiling to trust their builder images.
+
+### Goal 2: Easier integration of the lifecycle into other platforms and CI tools
+As members of the community experiment with incorporating CNBs into other platform and CI tools, the difficulty of invoking each lifecycle phase correctly and sequentially draws many to the easier solution of using `pack` to orchestrate the lifecycle. While `pack` can provide a simple way for these developers to implement these integrations, there may be performance drawbacks when compared with using the lifecycle directly. `/cnb/lifecycle/all` will provide an easier interface, lowering the barrier to using the lifecycle without pack, thus enabling platform developers choose between `pack` and `lifecycle` based solely upon the needs of their integration.
+
+# What it is
+[what-it-is]: #what-it-is
+
+### Usage
+USAGE:
```
ashwin-venkatesh:
```suggestion
```

jabrown85:
If [`bin/develop`](https://github.com/buildpacks/spec/blob/master/buildpack.md#development) gets built and get it's own binary at `/cnb/lifecycle/develop` should we consider `/cnb/lifecycle/build` or another name?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+
+* detector
+* analyzer
+* restorer
+* builder
+* exporter
+
+Each of these phases will continue to be available individually as provided by the current lifecycle.
+
+### Platform API 0.3
```
nebhale: Love this callout and I don't have a problem being "more" breaking than would be strictly necessary.

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+The existence of `all` will constitute a non-breaking but substantive change to the Platform API, bringing us to Platform API 0.3 . Although platform will interpret this as a breaking change, we do not have a mechanism for indicating non-breaking platform API change pre 1.0. Bumping the platform API number is necessary so that platforms like `pack` know whether or no they can use this feature.
+
+# Motivation
+[motivation]: #motivation
+
+This new binary supports two main goals.
+
+### Goal 1: faster `pack` builds
+By creating a single container and invoking `/cnb/lifecycle/all` instead of creating a container per lifecycle phase, pack can shave approximate `6s` from the execution time of `pack build`.
```
nebhale: Is there a reason why this is preferable to running the command `/cnb/lifecycle/detector && /cnb/lifecycle/analyzer && /cnb/lifecycle/restorer && /cnb/lifecycle/builder && /cnb/lifecycle/exporter`?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+
+`pack` may not always want to do this (see https://github.com/buildpacks/rfcs/pull/43 for an explanation of why this is not always ideal). However, it could be very beneficial to users that are wiling to trust their builder images.
+
+### Goal 2: Easier integration of the lifecycle into other platforms and CI tools
```
nebhale: This is a more compelling argument, when compared to pure performance which I think can be enhanced in other ways.

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+### Credential Management
+
+Right now, when `pack` runs w/ the `--publish` flag `analyzer` and `exporter` are provided with registry credentials via the `CNB_REGISTRY_AUTH` environment variable.
+
+In the registry case pack will set `CNB_REGISTRY_AUTH` when invoking `all`. To prevents buildpacks from having read access to those credentials the build and detect implementations will ensure this variable is not present in the environment of the `/bin/detect` and `/bin/build` processes.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Making it easy to run all of the lifecycle in a single container will encourage folks to do so. However, this requires consumers to be more thoughtful about security and credential isolation. We will need to harden `builder` and `detector` implementation to ensure buildpacks are not given root access or registry credentials.
```
nebhale: Do we believe that it won't be possible to harden these implementations such that they can be trusted?  Do we think that no matter what we'll always need a fallback to multiple containers?

/cc @sclevine

text/0000-lifecycle-all.md
ekcasey: The benefits of `/cnb/lifecycle/all` in the `pack` vs. stringing all of the phases together are mostly aesthetic. If we pass all of the required flags to each lifecycle phase, the command starts getting very long, complicated, and redundant looking.

text/0000-lifecycle-all.md
ekcasey: If we are running the lifecycle binaries that come packaged in the builder image, we can never harden them sufficiently to trust them. This is not because of potential problems with our implementation but rather, there is no guarantee that the files at those paths contain our implementation of the lifecycle.

We could append a layer to the image and essentially inject our lifecycle every build but this seems risky. There are numerous attack vectors if you don't trust the provenance of the builder image.

text/0000-lifecycle-all.md
ekcasey: Even if we want this feature mainly to accomplish `Goal 2` I think `pack` should use it. The ux and the reliability of `/cnb/lifecycle/all` will be improved if `pack` uses it in anger.

I can reverse the order of the goals if we think this is the more compelling one.

nebhale: Not necessary.

text/0000-lifecycle-all.md
matthewmcnew: Do we know what the primary cause of the `6s` increase in multi container builds? Is it purely the docker startup overhead?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Goal 2: Easier integration of the lifecycle into other platforms and CI tools
+As members of the community experiment with incorporating CNBs into other platform and CI tools, the difficulty of invoking each lifecycle phase correctly and sequentially draws many to the easier solution of using `pack` to orchestrate the lifecycle. While `pack` can provide a simple way for these developers to implement these integrations, there may be performance drawbacks when compared with using the lifecycle directly. `/cnb/lifecycle/all` will provide an easier interface, lowering the barrier to using the lifecycle without pack, thus enabling platform developers choose between `pack` and `lifecycle` based solely upon the needs of their integration.
+
+# What it is
+[what-it-is]: #what-it-is
+
+### Usage
+
+`/cnb/lifecycle/all <image-name>`
```
matthewmcnew: Currently, pack prepends lifecycle output with the step of the container builds. Such as: `===> DETECTING`. 

Are we expecting the `/lifecycle/all` binary prepend the output or will pack still detect the current step in the build?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,109 @@
...
+| `-launch-cache` | optional | `CNB_LAUNCH_CACHE_DIR` | -                         | path to launch cache directory |
+| `-launcher`     | optional | `CNB_LAUNCHER_PATH`    | `/cnb/lifecycle/launcher` | path to launcher binary |
+| `-buildpacks`   | optional | `CNB_BUILDPACKS_DIR`   | `/cnb/buildpacks`         | path to buildpacks directory |
+| `-daemon`       | optional | `CNB_USE_DAEMON`       | false                     | export to docker daemon |
+| `-uid`          | required | `CNB_USER_ID`          | -                         | UID of user in the stack's build and run images |
+| `-gid`          | required | `CNB_GROUP_ID`         | -                         | GID of user's group in the stack's build and run images |
+| `-version`      | optional | -                      | false                     | show version |
+| `-skip-restore` | optional | -                      | false                     | do not restore metadata from previous image or data from cache
+| `-tag` (multiples allowed) | optional | -           |                           | additional tags to apply to exported image
```
matthewmcnew: What image would the lifecycle analyze for previous layers? Could we add a `previous-image` to separate the tags that we right to from the resulting location?

text/0000-lifecycle-all.md
ashwin-venkatesh: In this case, pack would have to be more clever when invoking `/lifecycle/all` to know which phase is running. I suspect the `all` binary might have to be better about logging which phase is currently being run.

text/0000-lifecycle-all.md
ashwin-venkatesh: The argument provided as the `<image name>` would be the one that the lifecycle would analyze, no? `-tag` will be used to specify additional write locations.

text/0000-lifecycle-all.md
ashwin-venkatesh: It appeared to be from the overhead of both starting up as well as tearing down the container. We didn't do a complete teardown of where in the container lifecycle we were losing time. Comparing the execution time of the binaries within the container vs the execution time of the entire container had a constant ∆ of roughly `1.5s`. So running 1 container instead of 5 started should about a `6-7s` improvement in end to end pack build times.

text/0000-lifecycle-all.md
ekcasey: As a stop gap we could add the phase to the lifecycle logs. I think the better longterm solution is the structured logging RFC I have long been threatening to submit. This should be a good forcing function.

matthewmcnew: +1 to structured logging

text/0000-lifecycle-all.md
matthewmcnew: It seems a bit confusing to me that `tags` are not the complete list of locations that the lifecycle will export to especially because <image-name> doesn't have the `tag` name inside of it. 

Additionally, at times it can be useful to explicitly call out the image input to the `analyze` step. For example: kpack could pass in the previous image's digest to the analyzer. You obviously wouldn't export to an image name with the digest.

ekcasey: @matthewmcnew I am not opposed to adding an optional flag for the previous image image, so long as it default to the export tag when not supplied.

The rationale behind the single positional argument is that an image tag is the only **required** input. Whereas every other flag configures the build process, the `<image-name>` is the primary thing the process acts on.

Given that we already enforce that all exported tags are tags on the same registry, what if we changed the value passed to `-t` to be a "tag tag" (in the `<registry.com/repo/name:tag`> sense). I am not sure exporting to multiple different repositories in the same registry is really a valid use case. Maybe that would make the usage clearer - only the positional argument gets to define the repository.

ekcasey:
@jabrown85 

> should we consider /cnb/lifecycle/build or another name?

`build` seems too close to `builder` and therefore likely to be confusing.

`/cnb/lifecycle/combo`?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,110 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+### `/cnb/lifecycle/all`
+A new lifecycle binary with the name `lifecycle/all` will be included in each released lifecycle archive. Within a builder image, it will be found at the path `/cnb/lifecycle/all`. When invoked, it will run the following phases sequentially:
+
+* detector
```
jkutner: Is it possible that a poorly written buildpack could mutate the workingDir during `detect` and those mutations would stick around through the other phases? Then when the same buildpack is run in Tekton, for example, with each phase in it's own container the output/results of buildpack execution against the same code are different?

Maybe there are other scenarios that don't involve a poorly written buildpack.

If so, is there any way to prevent that, such as running in phase in a sandbox (like a temp dir that we copy into during each phase)?

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,110 @@
...
+- Start Date: 2020-01-21
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+### `/cnb/lifecycle/all`
+A new lifecycle binary with the name `lifecycle/all` will be included in each released lifecycle archive. Within a builder image, it will be found at the path `/cnb/lifecycle/all`. When invoked, it will run the following phases sequentially:
```
jromero: RE: name

Creating a thread to discuss the name of this new binary.

Previous comments:

> If bin/develop gets built and get it's own binary at /cnb/lifecycle/develop should we consider /cnb/lifecycle/build or another name? 
https://github.com/buildpacks/rfcs/pull/46#issuecomment-577376828 - @jabrown85 

> build seems too close to builder and therefore likely to be confusing.
> /cnb/lifecycle/combo?
> https://github.com/buildpacks/rfcs/pull/46#issuecomment-577857191 - @ekcasey

jromero: Not a fan of `combo`.... 

Also, FWIW, it already isn't all in that it doesn't include `rebaser`.

text/0000-lifecycle-all.md
ekcasey: Given that the working dir is `/layers` and both `/layers` and `/workspace` are volumes, changes made by buildpacks to these dirs are already persisted between phases. Changes made within `/buildpacks` wouldn't be persisted by the permissions should already be preventing buildpacks from making any such changes.

text/0000-lifecycle-all.md
ekcasey: @jromero gonna start a thread just to shoot down every idea without adding a suggestion? ;)

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,110 @@
...
+[prior-art]: #prior-art
+
+Most in container build tools do not spin up multiple containers.
+
+Regarding the introduction of the `-tag` flag, `docker build -t` is prior art.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Is there a better name than `lifecycle/all`? It doesn't include `launcher` or `rebaser` functionality so it isn't truly `all`.
```
jkutner: * `lifecycle/accio` 
* `lifecycle/alohomora` 
* `lifecycle/morsmordre` (dark magic)
* `lifecycle/sectumsempra` (very dark magic)

text/0000-lifecycle-all.md
jabrown85: `lifecycle/make` or `lifecycle/constructor`?

text/0000-lifecycle-all.md
ekcasey:
```
lifecycle/sectumsempra
```
lifecycle/silencio

text/0000-lifecycle-all.md
ekcasey: ooo, I think `/lifecycle/make` is my favorite so far (@ashwin-venkatesh agrees 💯 )

text/0000-lifecycle-all.md
matthewmcnew: Love the optional previous image flag.

matthewmcnew: I fear using "tag tag" as the value passed to -t would be confusing because it would be differ from the "tag" terminology in docker and kpack.

matthewmcnew: Could we just reuse the multiple positional arguments like what is in exporter today for multiple tags?

ekcasey: A single positional argument and additional `-tag`s seems more honest? The `exporter` currently treats the first of the positional arguments differently than the rest. In this proposed phase the positional argument will be used as the default value for the analyzer image. If one of the tags in special, 1 positional arg + flags seems to indicate that more clearly.

text/0000-lifecycle-all.md
jromero: @ekcasey :D Wanted to start the thread but didn't have any _good_ ideas myself.

Binary name suggestions:

- `/lifecycle/phases`

Maybe we don't need a new binary at all and only provide a subcommand to the [single lifecycle binary](https://github.com/ashwin-venkatesh/rfcs/blob/multicall-binary/text/0000-lifecycle-multicall-binary-for-build.md).

- `lifecycle/lifecycle build-app`
- `lifecycle/lifecycle run-phases`

jromero: I'm ok with `lifecycle/make` but it could be confused with [`make`](https://www.gnu.org/software/make/)

nebhale:
I move that this RFC be sent to FCP with a disposition of merge, closing 1700 PT, February 5, 2019.

text/0000-lifecycle-all.md
```diff
@@ -0,0 +1,110 @@
...
+- Name: Lifecycle All Binary
+- Start Date: 2020-01-21
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+### `/cnb/lifecycle/all`
```
ekcasey: Following suggestion in WG I am going to use this comment thread to conduct an informal poll to select the name for this binary. After this comment I will add comments each containing a single suggestion:

1. Please vote with a 👍
2. No more than 2 votes per person
3. Feel free to express dislike with a 👎 (full disclosure these may not factor into the counting)
4. To add a suggestion, add a comment containing only the suggestion

ekcasey: `lifecycle/all`

ekcasey: `lifecycle/make`

ekcasey: `lifecycle/combo`

ekcasey: `lifecycle/phases`

ekcasey: `lifecycle/constructor`

ekcasey: `lifecycle/creator`

ekcasey: `/lifecycle/build`

text/0000-lifecycle-all.md
dgageot: `lifecycle/Ash_nazg` to rule them all

zmackie:
@ekcasey Will make changes to codify `lifecycle/creator`

Issue #44, opened 2020-01-17
Default value of $PORT


Zanadar:
It's a common source of confusion that the containers we produce don't start, because they're waiting for a port to bind to. Should we have a way for there to be a default value of $PORT?

Basically , we always require folks to invoke their container with some version of `-e PORT=3000`. It would be nice if they would just turn on without this needing to be set, and for the default to be something sane. 

My vote is for `PORT=8080`

nebhale:
Can you please provide some examples of this confusion?

Zanadar:
>  i'm trying to get cnb for dotnet to build local docker container. so far what i'm getting doesn't run, i assume it binds to IPv6. is there a way to override listening interface, and what sets the launch command in for this buildpack - i'm unable to find it.

>  i actually just figured it out - it requires you set PORT env var

_____

Me:
> When we're booting the app in the httpd-cnb, where does $PORT come from?

Zanadar:
> wouldn't there be a default listening port if the app doesn't set it?

Zanadar:
> Rackup apps with the ruby buildpack assumes PORT be specified in runtime environment and it seems to not get set
- https://buildpacks.slack.com/archives/C94UJCNV6/p1576506128019900

Zanadar:
>  I was able to get farther, I’m having trouble getting it to map to a specific port. (which I suspect is more an error in my usage)

- https://buildpacks.slack.com/archives/C94UJCNV6/p1575401687206700?thread_ts=1573598283.079300&cid=C94UJCNV6

Zanadar:
(The above unattributed comments are from internal users on corp. slack)

nebhale:
Understood on the lack of attribution.  Do you have a read on how many of those are side-effects of the _buildpacks_ requiring `$PORT` (e.g. putting it into a command line or a `.profile.d`) versus the frameworks themselves expecting it?

My reason for this line of questioning is that there's a set of buildpacks for a specific language (😉) for which this hasn't come up and it was because we made an explicit effort to remove our dependency on `$PORT`.  If this can be solved via updates to buildpacks rather than more specification, I'd prefer to go that route.

Zanadar:
So the buildpacks have a default value for `$PORT`? That's a viable solution.

[...]

PR #47, opened 2020-01-22
Initial RFC for platform default ports

zmackie:
[readable](https://github.com/Zanadar/rfcs/blob/default-port/text/0000-platform-default-port.md)

### DIFF
--- a/text/0000-platform-default-port.md
+++ b/text/0000-platform-default-port.md
@@ -1,0 +1,50 @@
+# Meta
+[meta]: #meta
+- Name: Default ports
+- Start Date: 2020-01-22
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Proposes a mechanism whereby a platform can expose a default port for builds.
+
+# Motivation
+[motivation]: #motivation
+
+A number of users have complained that the containers they created with CNBs don't run "out of the box" because of an unset `$PORT` variable. This leads to a confusing initial experience and makes the path from running a simple `pack build` to starting a container unnecessarily rocky. If platforms had a default value of `$PORT` they could expose to the lifecycle, which could easily be overridden during a build, it would reduce some of that friction.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Platforms that build images with CNB tech create images with a default value for `PORT=8080`. Buildpacks can override that value if it makes sens for their individual case; for example a `postgres` buildpack - if there were such a thing - would override `$PORT` with `5432` which a common convention. In cases where there is no assumed value because the language community hasn't standardized (for example in ruby or nodejs) the images produced by, for example, pack, could still start without the added step of configuring this value during invocation.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The platform spec would dictate that `PORT` must be defined by the platform at the outset of a build, but that buildpacks can override this value.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I don't see any downsides to providing a default for this variable, but I'm sure they exist. The only downside would be added complexity of the spec.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Don't do anything. Continue to explain this to users of pack.
+- Always set `PORT=1234` because its a cool number.
+- Leave this out of the spec and just make the change in `pack`
+
+
+# Prior Art
+[prior-art]: #prior-art
+
+Not sure here?
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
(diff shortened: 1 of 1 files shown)


Zanadar:
https://github.com/buildpacks/rfcs/issues/44#issuecomment-575759322

Zanadar:
Porting comments

Zanadar:
From @nebhale :
https://github.com/buildpacks/rfcs/issues/44#issuecomment-575759322
> Understood on the lack of attribution. Do you have a read on how many of those are side-effects of the buildpacks requiring $PORT (e.g. putting it into a command line or a .profile.d) versus the frameworks themselves expecting it?
>
> My reason for this line of questioning is that there's a set of buildpacks for a specific language (😉) for which this hasn't come up and it was because we made an explicit effort to remove our dependency on $PORT. If this can be solved via updates to buildpacks rather than more specification, I'd prefer to go that route.

https://github.com/buildpacks/rfcs/issues/44#issuecomment-575916181
> Not that buildpacks have a default, but rather the buildpacks don't depend on `$PORT` being defined. If there's no expectation of `$PORT`, there's no need for a default.
> 
> In the Java world, none of the frameworks depend on `$PORT` and typically simply default to `8080`. When the frameworks default to something else, they _still_ don't depend on `$PORT`. It does require a user to know based on their application, they may need to know `-p 8080:8080` or `-p 8888:8888`, but at the very least there's no dependency on a set, valid `$PORT`.


https://github.com/buildpacks/rfcs/issues/44#issuecomment-576708622
> Anecdotally, there has historically been [pushback](https://github.com/ratpack/ratpack/issues/239) against frameworks understanding `$PORT` directly. I don't think this is as universal of a behavior as you'd expect.
> 
> I generally feel that the era of platforms pushing values _into_ containers (e.g. picking the port that will be exported) is waning. We're much more seeing platforms that allow arbitrary configuration or transparent mapping for these ports. For example, there's no `$PORT` in the `postgres` image, there's a straight assumption that Postgresql will always be exposed on `5432` and you configure your execution of the container to expose that properly. I think we should be adopting the same pattern; stick with the default ports that frameworks choose, and use configuration to map those out to the world.

Zanadar:
From @jkutner 
https://github.com/buildpacks/rfcs/issues/44#issuecomment-575959243
> `PORT` has become a common enough standard that it is jarring for users when it's not set by default. Because Spring doesn't pick it up by default, many apps will have `--server.port=$PORT` in their `java` command, which creates a rocky experience with CNB.

text/0000-platform-default-port.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+A number of users have complained that the containers they created with CNBs don't run "out of the box" because of an unset `$PORT` variable. This leads to a confusing initial experience and makes the path from running a simple `pack build` to starting a container unnecessarily rocky. If platforms had a default value of `$PORT` they could expose to the lifecycle, which could easily be overridden during a build, it would reduce some of that friction.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Platforms that build images with CNB tech create images with a default value for `PORT=8080`. Buildpacks can override that value if it makes sens for their individual case; for example a `postgres` buildpack - if there were such a thing - would override `$PORT` with `5432` which a common convention. In cases where there is no assumed value because the language community hasn't standardized (for example in ruby or nodejs) the images produced by, for example, pack, could still start without the added step of configuring this value during invocation.
```
jromero: As a developer I would expect the opposite behaviour. I would expect that there is a default set by the buildpack that I could then override via parameter at build time. In most cases I believe env vars is the way to communicate configuration to buildpacks.

Essentially what I picture is an order of precedence:
1. Environment variables
2. Buildpack defaults

For example:
1. `pack build ... -env POSGRES_PORT=6543 ...`
2. buildpack defaults to `5432` if `POSGRES_PORT` is not present

Maybe this is simply a question of buildpack best practices such as informing the user of the port being bound and what they can do to override it OR there could be a better mechanism for discovering configurability within buildpacks.

jromero: What if..., hear me out, we had a standard way to declare configuration within `buildpack.toml` that platforms could then use to display configuration elements.

For example a `buildpack.toml`:
```toml
[buildpack]
id = "..."
version = "..."

[config]
[[config.env]]
name = "POSTGRES_PORT"
description="Port to serve PostgreSQL on."
default="5432"
```

`pack` could then do various things with that information such as:

```shell script
$ pack get-buildpack-config postgres@1.2.3

Environment Variables:
  POSTGRES_PORT        Port to serve PostgreSQL on. (default 5432)

```

jromero: 🌈 This discussion will be moved to a seperate RFC.

zmackie:
One suggestion from WG is that this live in `pack` vs the spec proper. There would be definite tradeoffs to the current clean seperation between `pack` and the images it produces (because `pack` doesn't change the environment of the images it produces).

zmackie:
Will close and go back to the drawing board.

PR #49, opened 2020-01-27
RFC: spec API releases

ekcasey:
[Readable](https://github.com/buildpacks/rfcs/blob/spec-api-releases/text/0000-spec-api-branches.md)

This RFC proposes a branching process to allow us to stage changes for a new platform or buildpack API versions, while accommodating @sclevine's preference that master of the spec always describe a real set of API versions.

### DIFF
--- a/text/0000-spec-api-branches.md
+++ b/text/0000-spec-api-branches.md
@@ -1,0 +1,58 @@
+# Meta
+[meta]: #meta
+- Name: Spec API branches
+- Start Date: 2020-01-27
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The version of the specification is represented by two numbers, the Buildpack API and the Platform API. When there are major changes to these APIs the API number is bumped. Rather than requiring a bump to the API for each change that enters the spec, this RFC proposes adding protected branches for future API version to the spec repo. Breaking changes can be PR'd to these branches and the branches themselves will be merged to master when the core team decided that a group of changes together shall constitute the next API version.
+
+# Motivation
+[motivation]: #motivation
+
+Changes to the Platform or Buildpack API can be disruptive, especially right now when both are pre-stable (0.x) and every change is treated as breaking. 
+
+When we want to make several changes to a given API in short succession (i.e. in the time windows between subsequent releases of the reference lifecycle), it would be nice to use a single new API number to represent a set of breaking changes.
+
+Given that we have decided that master of the spec repo shall always describe a specific set of API versions, we should stage changes on a branch, and merge that branch to master when we want to assign a number to the set of changes.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The spec repo shall have protected branches representing the next API version (right now those would be `platform/0.3` and `buildpack/0.3`). PRs to the spec that change an API (all non-cosmetic changes) should be made to those branches.
+
+When they decide it is appropriate, the core team, in concsultation with the implementation team will merge an API branch to master, bump the API versions, in the spec README (https://github.com/buildpacks/spec#api-versions), and apply a [tag](https://github.com/buildpacks/spec/releases) to master. 
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Example - Platform API
+For example, if we were following this process right now, [buildpacks/spec#68](https://github.com/buildpacks/spec/pull/68) would be made against the `platform/0.3` branch. [rfcs#46](https://github.com/buildpacks/rfcs/pull/46) is accepted, a corresponded set of spec changes would also be PR'd to the spec repo.
+
+Once the complete set of changes that should represent Platform API 0.3 is fully staged on the branch it will be merged to master. A release of lifecycle that supports Platform API 0.3 can be shipped, and users will be able to easily lookup implemented API.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* PRing the spec becomes a more complicated process
+* Core team will have a new resposibility, deciding when to finalize a new API version
+
+# Alternatives
+[alternatives]: #alternatives
+
+- PR all spec changes to master, use release tags to indicate finalized API version
+- Bump the API version for every single change to the spec
+
+# Prior Art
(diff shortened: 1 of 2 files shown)


text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+When we want to make several changes to a given API in short succession (i.e. in the time windows between subsequent releases of the reference lifecycle), it would be nice to use a single new API number to represent a set of breaking changes.
+
+Given that we have decided that master of the spec repo shall always describe a specific set of API versions, we should stage changes on a branch, and merge that branch to master when we want to assign a number to the set of changes.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The spec repo shall have protected branches representing the next API version (right now those would be `platform/0.3` and `buildpack/0.3`). PRs to the spec that change an API (all non-cosmetic changes) should be made to those branches.
+
+When they decide it is appropriate, the core team, in concsultation with the implementation team will merge an API branch to master, bump the API versions, in the spec README (https://github.com/buildpacks/spec#api-versions), and apply a [tag](https://github.com/buildpacks/spec/releases) to master.
```
hone:
```suggestion
When they decide it is appropriate, the core team, in consultation with the implementation team will merge an API branch to master, bump the API versions, in the spec README (https://github.com/buildpacks/spec#api-versions), and apply a [tag](https://github.com/buildpacks/spec/releases) to master. 
```

text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+## Example - Platform API
+For example, if we were following this process right now, [buildpacks/spec#68](https://github.com/buildpacks/spec/pull/68) would be made against the `platform/0.3` branch. [rfcs#46](https://github.com/buildpacks/rfcs/pull/46) is accepted, a corresponded set of spec changes would also be PR'd to the spec repo.
+
+Once the complete set of changes that should represent Platform API 0.3 is fully staged on the branch it will be merged to master. A release of lifecycle that supports Platform API 0.3 can be shipped, and users will be able to easily lookup implemented API.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* PRing the spec becomes a more complicated process
+* Core team will have a new resposibility, deciding when to finalize a new API version
```
hone: Would we want to have this fall in line with #33, namely when `lifecycle` gets released? I can see us dealing with spec changes that wouldn't require a `lifecycle` release as well.

text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- PR all spec changes to master, use release tags to indicate finalized API version
```
hone: I think we can achieve the objectives here with something like this by having a moving stable branch and master is active development. We tell github to use the stable branch as default, so users always see a real API version.

text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Changes to the Platform or Buildpack API can be disruptive, especially right now when both are pre-stable (0.x) and every change is treated as breaking. 
+
+When we want to make several changes to a given API in short succession (i.e. in the time windows between subsequent releases of the reference lifecycle), it would be nice to use a single new API number to represent a set of breaking changes.
+
+Given that we have decided that master of the spec repo shall always describe a specific set of API versions, we should stage changes on a branch, and merge that branch to master when we want to assign a number to the set of changes.
```
hone: What do we do about changes that are in master today, but aren't implemented?

ekcasey: with `/bin/develop` leaving the spec (buildpacks/spec#71), only `store.toml` remains unimplemented. It's not a particular complicated feature, so I vote we prioritize catching up with the spec before we release the next spec version.

text/0000-spec-api-branches.md
ekcasey: imo every lifecycle should fully implement a specific documented set of API versions, but the lifecycle release and the spec release don't need to happen simultaneously.

In other words, if the spec defines buildpack API version 0.3, a subsequent lifecycle release can implement buildpack API 0.2 or 0.3, but it cannot implement half of 0.3 features. Obviously, the preference would be for catching up with the spec as quickly as possible.

text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+The spec repo shall have protected branches representing the next API version (right now those would be `platform/0.3` and `buildpack/0.3`). PRs to the spec that change an API (all non-cosmetic changes) should be made to those branches.
+
+When they decide it is appropriate, the core team, in consultation with the implementation team will merge an API branch to master, bump the API versions, in the spec README (https://github.com/buildpacks/spec#api-versions), and apply a [tag](https://github.com/buildpacks/spec/releases) to master.
```
hone: Since there are two API versions here, should we clarify/document how we name the tags?

text/0000-spec-api-branches.md
```diff
@@ -0,0 +1,58 @@
...
+- Name: Spec API branches
+- Start Date: 2020-01-27
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The version of the specification is represented by two numbers, the Buildpack API and the Platform API. When there are major changes to these APIs the API number is bumped. Rather than requiring a bump to the API for each change that enters the spec, this RFC proposes adding protected branches for future API version to the spec repo. Breaking changes can be PR'd to these branches and the branches themselves will be merged to master when the core team decided that a group of changes together shall constitute the next API version.
```
jromero: Can we expand this to say that any specification including extensions would follow this workflow?

ekcasey: I updated the wording, please let me know if it is clear.

text/0000-spec-api-branches.md
ekcasey: updated, have a look

nebhale:
I move that this RFC be sent to FCP with a disposition of merge, closing 1700 PT, February 19, 2019.

PR #50, opened 2020-01-28
RFC: source_date_epoch for build timestamps

zmackie:
[readable](https://github.com/zmackie/rfcs/blob/source_date_epoch/text/0000-source-date-epoch.md)

### DIFF
--- a/text/0000-source-date-epoch.md
+++ b/text/0000-source-date-epoch.md
@@ -40,7 +40,10 @@ More knobs to turn, more configurability.
 # Prior Art
 [prior-art]: #prior-art
 
-https://reproducible-builds.org/docs/source-date-epoch/
+- Spec: https://reproducible-builds.org/docs/source-date-epoch/
+
+- [`ko`](https://github.com/google/ko/tree/ca1b2a1dedfdb13b4955018631fd6566e5d1a13d#why-are-my-images-all-created-in-1970)
+- [`rules_docker`](https://github.com/bazelbuild/rules_docker/blob/371f328c4357e33b64bbdc4eaaf59b4544efa31e/container/image_test.py#L231-L236) (via https://github.com/bazelbuild/bazel/issues/2240)
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-source-date-epoch.md
+++ b/text/0000-source-date-epoch.md
@@ -1,0 +1,48 @@
+# Meta
+[meta]: #meta
+- Name: Source Date Epoch
+- Start Date: 2020-01-28
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Allow platforms to define an build epoch. This would improve the UX of images builds at the expense of reproducibility.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the lifecycle sets all build dates to the unix epoch for reproducibility. This can be confusing for newcomers because it looks like their builds are "40 years old". This should at the least be documented. Going one step further, the lifecycle could respect the value of `$SOURCE_DATE_EPOCH` if set by platforms, which would allow an optional override.
+
+# What it is
+[what-it-is]: #what-it-is
+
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
If you want some more prior art, [`ko`](https://github.com/google/ko/tree/ca1b2a1dedfdb13b4955018631fd6566e5d1a13d#why-are-my-images-all-created-in-1970) and [`rules_docker`](https://github.com/bazelbuild/rules_docker/blob/371f328c4357e33b64bbdc4eaaf59b4544efa31e/container/image_test.py#L231-L236) (via https://github.com/bazelbuild/bazel/issues/2240) both support this.

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the lifecycle sets all build dates to the unix epoch for reproducibility. This can be confusing for newcomers because it looks like their builds are "40 years old". This should at the least be documented. Going one step further, the lifecycle could respect the value of `$SOURCE_DATE_EPOCH` if set by platforms, which would allow an optional override.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Lifecycle respects `$SOURCE_DATE_EPOCH` if provided by platform by using that value as the timestamp for everything it creates. This could have a default value of zero for repro. and be set optionally to allow platforms a better UX. eg. `pack build MY_APP --build-epoch 12345` would set `SOURCE_DATE_EPOCH=12345` during build  time and all layers would have that timestamp.
```
ekcasey: > all layers would have that timestamp


Does "layers" refer to layer history entries or does it include layer file timestamps. If layer file timestamps are not standardized, lifecycle cannot reuse cached layers anymore. We currently use the layer diffID (uncompressed layer digest), to determine whether a layer has changed

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,51 @@
...
+
+Lifecycle respects `$SOURCE_DATE_EPOCH` if provided by platform by using that value as the timestamp for everything it creates. This could have a default value of zero for repro. and be set optionally to allow platforms a better UX.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More knobs to turn, more configurability.
+
+# Alternatives
+[alternatives]: #alternatives
```
ekcasey: Alternative: Document the current behavior and collect feedback to determine if the ux is a problem.

zmackie: We haven't documented it, but I've so far seen 4 questions about "why are my images being created in 1980?", FWIW. At least people are taking up our releases!

ekcasey: If `pack` still creates images in 1980 by default, and a flag is needed to supply the timestamp, I don't think that mitigates the initial surprise/confusion. It provides users with a mechanism to change the behavior once they understand it and if they have a reason to care about the consequences

zmackie:
Maybe https://github.com/buildpacks/docs/issues/112 is the way.

text/0000-source-date-epoch.md
zmackie: I'll look into why the alternatives provide this knob: the thesis is that the default for these other platforms is `timestamp == now` vs our default which is `timestamps == 1980` (set unchanging value), so the mechanism for setting it to a reproducible value was needed.

zmackie: Reading links above: in the case of `ko` they default to a fixed timestamp ([docs](https://github.com/google/ko/tree/ca1b2a1dedfdb13b4955018631fd6566e5d1a13d#why-are-my-images-all-created-in-1970)). In bazel build it seems as if the they timestamp to *now* and the have the variable for overrides.

That still points to at least *starting* with the documentation and then moving on to a knob from there.

hone: @zmackie should we close this RFC for now and see if we want to re-open a new one with your new data or do you want to keep this open until you get that data?

zmackie:
will add reasonings into rfc

mikegehard:
I love the idea of adding some documentation around this just so we can point new comers to this.

I am working on a training class that uses pack and I'm sure everyone will have questions.

zmackie:
Documentation PR: https://github.com/buildpacks/docs/pull/123

jromero:
@zmackie given that we've documented this and there's not been much activity or further requests for this specific feature should this be closed or moved to draft?

zmackie:
@jromero Happy to close this PR.

edudar:
Is it possible to bring this back for consideration?

PR #52, opened 2020-02-03
RFC: optional RFC section for Spec changes

zmackie:
[readable](https://github.com/zmackie/rfcs/blob/rfc-spec-reminder/text/0000-template-changes.md)

Signed-off-by: Zander Mackie <zmackie@gmail.com>

### DIFF
--- a/text/0000-template-changes.md
+++ b/text/0000-template-changes.md
@@ -32,7 +32,7 @@ Discuss prior art, both the good and bad.
 + # Spec. Changes (OPTIONAL)
 + [spec-changes]: #spec-changes
 +
-+ Does this RFC entail any proposed changes to the specifications (either buildpack or platform)?
++ Does this RFC entail any proposed changes to the specifications (either buildpack or platform) or extensions?
 + This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
 +
 ```
--- a/text/0000-template-changes.md
+++ b/text/0000-template-changes.md
@@ -1,0 +1,61 @@
+# Meta
+[meta]: #meta
+- Name: RFC template changes
+- Start Date: 2020-02-03
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+As a response to the process of approving [RFC 0013](0013-app-layer-metadata-source.md) which ended up requiring spec changes, this RFC proposes amending the template to include an optional (non-binding) section to document spec changes that fall out of an RFC.
+
+# Motivation
+[motivation]: #motivation
+
+Improve the process of moving RFCs into spec and implementation.
+
+# What it is
+[what-it-is]: #what-it-is
+
+[0000-template.md](../0000-template.md) gets a new section:
+
+```diff
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
++
++ # Spec. Changes (OPTIONAL)
++ [spec-changes]: #spec-changes
++
++ Does this RFC entail any proposed changes to the specifications (either buildpack or platform)?
++ This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
++
+```
+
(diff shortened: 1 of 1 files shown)


text/0000-template-changes.md
```diff
@@ -0,0 +1,61 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
++
++ # Spec. Changes (OPTIONAL)
++ [spec-changes]: #spec-changes
++
++ Does this RFC entail any proposed changes to the specifications (either buildpack or platform)?
```
nebhale: Or extension.

text/0000-template-changes.md
```diff
@@ -0,0 +1,61 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
++
++ # Spec. Changes (OPTIONAL)
++ [spec-changes]: #spec-changes
++
++ Does this RFC entail any proposed changes to the specifications (either buildpack or platform) or extensions?
```
ekcasey:
```suggestion
+ Does this RFC entail any proposed changes to the specifications or extensions?
```

Distribution too. Might as not provide a comprehensive list for the sake of future proofing

text/0000-template-changes.md
```diff
@@ -0,0 +1,61 @@
...
++ This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
++
```
ekcasey: It would be nice to provide some examples here to help folks who are less familiar with the spec start to consider this question: new lifecycle flags, new buildpack.toml fields, new fields in the buildpackage label etc.

text/0000-template-changes.md
zmackie: Agree being more general here would be good to future-proof.

text/0000-template-changes.md
zmackie: > It would be nice to provide some examples here to help folks who are less familiar with the spec start to consider this question: new lifecycle flags, new buildpack.toml fields, new fields in the buildpackage label etc.

Agree that examples would be helpful, will add. However, I don't think me message here is that the onus is necessarily on the RFC opener, at least initially, to add to this section. My thinking was that if an RFC evolved to a place where spec changes where necessary, a reviewer should call that out and the author would add it on when they received that feedback.

text/0000-template-changes.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Summarize the discussion in a comment.
+
+# Prior Art
+[prior-art]: #prior-art
+
+Not sure
```
hone: I believe the alternative (Do Nothing) would be for RFC authors to debate spec changes after a RFC gets accepted inside the spec PR which is where we are today.

text/0000-template-changes.md
jromero: What about something like a checklist? This would help people easily think through all the potential candidates. Some changes apply to multiple specs.

```
If this RFC proposes changes to a specification, what specification?

- [ ] platform
- [ ] buildpack
- [ ] distribution
- [ ] extension: ___________
```

text/0000-template-changes.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More work for RFC authors, although the contents of this section should likely be supplied by the project maintainers leading the discussion.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Summarize the discussion in a comment.
```
jkutner: I think we could also put the spec changes (as described herein) under "How It Works" without requiring a new section. I feel like this might make the template more approachable for those who don't know the spec well enough, but have a good RFC otherwise.

nebhale: The problem this aims to solve is that we had that before and not one of us twigged that there should have been a spec change paragraph.  That seems to indicate that the status quo has an issue.

nebhale:
I move that this RFC be sent to FCP with a disposition of merge, closing 1700 PT, February 27, 2019.

text/0000-template-changes.md
hone: Can we get the examples added here?

PR #54, opened 2020-02-04
RFC: self-documenting buildpacks

zmackie:
[readable](https://github.com/zmackie/rfcs/blob/self-documenting-buildpacks/text/0000-self-documenting-buildpacks.md)

### DIFF
--- a/text/0000-self-documenting-buildpacks.md
+++ b/text/0000-self-documenting-buildpacks.md
@@ -1,0 +1,46 @@
+# Meta
+[meta]: #meta
+- Name: Self Documenting Buildpacks
+- Start Date: 1-23-2020
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC suggest a mechanism where buildpacks can provide documentation for the benefit of their consumers.
+
+# Motivation
+[motivation]: #motivation
+
+At the moment the origin and provinence of a buildpack are are sort of a black box. Users have asked where to find information about the repos for buildpacks so they can file issues, but we would also like a method to in-situ allow buildpacks to surface their configuration options or anything else their authors think pertinent
+
+# What it is
+[what-it-is]: #what-it-is
+
+Would add an additional, optional, field to `buildpack.toml` -  `docs`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The `docs` field would be of `TOML` string type. It would be expected that this field contains markdown. The rendering the contents of this field would be up to the platform implementers, but I could imagine an interface like `pack inspect-builder BUILDER --docs BUILDPACK` that would render this output to a terminal. This would allow buildpack authors to fill out arbitary information about their buildpacks such as repo, author, build and run time env vars, and other sorts of miscellaneoous documentation and configuration.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Rendering markdown is hard. Yet another thing in the spec. Too generic.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- A `link` field that just points somewhere else so buildpack authors would have to host their own docs.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- Rdoc
+- Golang method docs
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions


zmackie:
Not sure if this is platform or implementation? Probably the latter (or maybe learning team :) )

text/0000-self-documenting-buildpacks.md
```diff
@@ -0,0 +1,46 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Would add an additional, optional, field to `buildpack.toml` -  `docs`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The `docs` field would be of `TOML` string type. It would be expected that this field contains markdown. The rendering the contents of this field would be up to the platform implementers, but I could imagine an interface like `pack inspect-builder BUILDER --docs BUILDPACK` that would render this output to a terminal. This would allow buildpack authors to fill out arbitary information about their buildpacks such as repo, author, build and run time env vars, and other sorts of miscellaneoous documentation and configuration.
```
ekcasey: If we believe the primary mode of viewing the docs would be in a terminal with pack, I don't think Markdown, as a text-to-html tool, is the best option.

If we used specific fields for known types of configuration (build and run time env vars, plan entries, etc.) we could print that data in a tabular format.

text/0000-self-documenting-buildpacks.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Rendering markdown is hard. Yet another thing in the spec. Too generic.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- A `link` field that just points somewhere else so buildpack authors would have to host their own docs.
```
ekcasey: I think I prefer this alternative:
1. It's relatively easy for us to implement
1. It's low effort for buildpack authors to take advantage of
1. Buildpack authors have the option to update the docs without shipping a new version of the buildpack
1. Given that many buildpack docs are github READMEs, the link approach has the incidental benefits of taking people to the repo where they can track progress and file issues.

text/0000-self-documenting-buildpacks.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+At the moment the origin and provinence of a buildpack are are sort of a black box. Users have asked where to find information about the repos for buildpacks so they can file issues, but we would also like a method to in-situ allow buildpacks to surface their configuration options or anything else their authors think pertinent
+
+# What it is
+[what-it-is]: #what-it-is
+
+Would add an additional, optional, field to `buildpack.toml` -  `docs`.
```
ekcasey: In order for information to be available to `pack inspect-builder`, it will need to be added to builder/bpkg labels. This RFC should include those labels, especially if they necessitate a change to the distribution spec. We could use a `buildpack.toml` fields to populate the label(s).

text/0000-self-documenting-buildpacks.md
jromero: I'm stuck in the middle. I agree with @ekcasey in regards to having a list of know properties for parsing purposes.

eg. This is a very rough swag:

```toml
[buildpack]
name = "A cool buildpack for Java."

[docs]
details = """
a lot of details
...
...
"""
url = "http://learnmore.com/README.md" # url to "learn more"

# Environment variables that could be provided by the user
[[docs.env]]
name = "BP_DEBUG"
description = "Enables debug logging."
default = "false"
scope = ["build"]

[[docs.env]]
name = "PORT"
description = "Port to bind to."
default = "8080"
scope = ["build", "launch"]

[[docs.plan]]
# Plan entries (need to research)
```

I like the idea of markdown for purposes outside of CLI but that feature could be added later. FWIW, there are libraries (and sure we can roll our own) to strip markdown: https://github.com/writeas/go-strip-markdown

text/0000-self-documenting-buildpacks.md
jromero: If it is a link, which I'm not opposed to in the short-term. I don't see why we wouldn't just display it in `pack inspect-builder` WITHOUT having to call out that I want to see this one other attribute.

text/0000-self-documenting-buildpacks.md
zmackie: I think *unknown* entries are probably the issue here. I worry that if we go down the path of having fields for *known* entries, we'll have to update that more frequently than would be desired or just create a situation where buildpacks cant be adequately documented because we're missing their config keys.

jromero: If we don't want to go with the full spec'ing of individual properties at this point in time then I vote we opt for the `link` option. The UX for `docs` being arbitrary content seems odd and would wager that wouldn't be as useful as a buildpack being able to specify a uri to where that information can be hosted (and subsequently updated without releasing).

nebhale:
Count me in with the crowd that only wants a link out to documentation.  I don't think we have a great understanding of what things are useful in documentation (should you document how application developers use and configure the buildpack, should you document how other buildpacks `provides/requires` the buildpack, what information is useful to each of those audiences), and so any attempt to standardize is almost certainly incorrect for the long term.  `pack inspect-[buildpackage|builder]` should put a link in the table of buildpacks and be done with it.

zmackie:
Closing in favor of https://github.com/buildpacks/rfcs/pull/55

PR #56, opened 2020-02-11
RFC: Buildpack URIs

jromero:
[readable](https://github.com/buildpacks/rfcs/blob/buildpack-uris/text/0000-buildpack-uris.md)

### DIFF
--- a/text/0000-buildpack-uris.md
+++ b/text/0000-buildpack-uris.md
@@ -41,11 +41,16 @@ directory and image.
 
 | Short Name | Format | Examples |
 |--- |--- |--- |
-| Filesystem | `file://[<host>]/<path>` | `file:///./my/buildpack.tgz`<br>`file:///home/user/my/buildpack.tgz`
+| Relative | `<path>` | `./my/buildpack.tgz`<br>`/home/user/my/buildpack.tgz`
+| Filesystem | `file://[<host>]/<path>` | `file:///my/buildpack.tgz`<br>`file:///home/user/my/buildpack.tgz`
 | URL | `http[s]://<host>/<path>` | `http://example.com/my/buildpack.tgz`<br>`https://example.com/my/buildpack.tgz`  
-| Docker | `docker://[<host>]/<path>[:<tag>]` | `docker:///ubuntu:latest`<br>`docker://gcr.io/distroless/nodejs` 
-| CNB Builder | `cnb:builder://[<id>[@<version>]]` | `cnb:builder://`<br>`cnb:builder://bp.id`<br>`cnb:builder://bp.id@bp.version`
+| Docker | `docker://[<host>]/<path>[:<tag>⏐@<digest>]` | `docker://gcr.io/distroless/nodejs`<br>`docker:///ubuntu:latest`<br>`docker:///ubuntu@sha256:45b23dee08...`
+| CNB Builder | `cnb:builder://[<host>]/[<id>[:<version>]]` | `cnb:builder://`<br>`cnb:builder:///bp.id`<br>`cnb:builder:///bp.id:bp.version`
 
+
+### Relative
+
+A URI without a scheme is treated as a relative reference as per [RFC 3986 section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1).
 
 ### Filesystem
 
@@ -66,14 +71,16 @@ eg. `docker:/ubuntu:latest`
 ### CNB Builder
 
 - `cnb:builder://` - A reference to ALL the buildpacks in the builder.
-- `cnb:builder://bp.id` - A reference to a buildpack with id `bp.id` in the builder. If there is more than one version
+- `cnb:builder:///bp.id` - A reference to a buildpack with id `bp.id` in the builder. If there is more than one version
 available, a error should be thrown.
-- `cnb:builder://bp.id@bp.version` - A reference to a buildpack with id `bp.id` at the specific version `bp.version`.
+- `cnb:builder:///bp.id:bp.version` - A reference to a buildpack with id `bp.id` at the specific version `bp.version`.
+- Similar to the `file` scheme, there is a minimal declaration for omitting host by using a simple slash (`/`). 
+eg. `cnb:builder:/bp.id:bp.version`
 
 # How it Works
 [how-it-works]: #how-it-works
 
-### Usage
+### pack
 
 Certain `pack` CLI arguments would take URIs supported instead of the current inconsistent format.
 
@@ -82,21 +89,58 @@ Some examples:
 ```shell script
 pack build my-app \
   --buildpack cnb:builder://bp.id@bp.version \                # buildpack from builder
-  --buildpack file:///../path/to/buildpack/ \                 # relative file
+  --buildpack file:///home/user/path/to/buildpack/ \          # absolute via file
+  --buildpack /home/user/path/to/buildpack/ \                 # absolute via schemeless
+  --buildpack ../path/to/buildpack/ \                         # relative file
   --buildpack docker:/cnbs/some-package \                     # Docker Hub image
   --buildpack docker://gcr.io/cnbs/sample-package-2:bionic \  # GCR image (with tag)
   --buildpack cnb:builder://                                  # All buildpacks in builder
 ```   
 
-#### Additional considerations
+### Config Files
 
-`pack` could default a certain scheme to provide additional ease-of-use. For example,
-if  `cnb:builder` is the default scheme, users can simply use `pack build ... --buildpack bp.id@bp.version`. We might
-want to wait until the registry is available though to make that the default.
+Config files would replace `image` with `uri`, using the `docker://` scheme, where applicable.
+
+`package.toml`:
+```toml
+[buildpack]
+uri = "."
+
+[[dependencies]]
+uri = "docker:/my/image:latest"
+
+# ...
+``` 
+
+`builder.toml`:
+```toml
+# ...
+
+[[buildpacks]]
+uri = "./my/buildpacks/"
+
+[[buildpacks]]
+uri = "docker:/my/image:latest"
+
+# ...
+```
+
+# Additional considerations
+
+#### `pack` User Expirience
+
+`pack` could provide additional logic in order for users to have a nicer user experience when providing references
+via CLI. For example, `pack` could have an order of precedence to make a best guess at which type of reference
+they are referring to or provide additional feedback if it's ambiguous.
+
+#### Registry vs Daemon
+
+
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- The user would need to type an additional scheme prefix.
+- The user would need to type an additional scheme prefix for any URI which is not a relative reference.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-buildpack-uris.md
+++ b/text/0000-buildpack-uris.md
@@ -1,0 +1,124 @@
+# Meta
+[meta]: #meta
+- Name: Buildpack URIs
+- Start Date: 02/07/2020
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+A set of predefined URIs for referring to buildpacks from various sources (protocols). 
+
+_Please note that the scheme denotes protocol and not type._
+
+# Motivation
+[motivation]: #motivation
+
+Given that we support a multitude of sources for buildpacks, a certain level of ambiguity has naturally presented
+itself. We currently _try_ to resolve the various sources through a complex set of rules yet the ambiguity to users
+still exists. In config files, this ambiguity is at times resolved by introducing a separate key name to signify
+the source (protocol).
+
+The aim of this RFC is to propose a set of URI schemes that could be used in both CLI arguments and config files
+while solving the ambiguity problem.
+
+A few references to further express the problem:
+
+- [A poll](https://buildpacks.slack.com/archives/C94UJCNV6/p1580920669081100) to understand what users expect when
+providing buildpacks and there is overlap of sources.
+- [Issue](https://github.com/buildpacks/pack/issues/489) to provide additional clarity in logs due to ambiguity. 
+- [A conversation](https://buildpacks.slack.com/archives/CJ6B92ZSB/p1578502454017000) when trying to determine how to
+provide reference to all buildpacks in builder.
+- [Issue](https://github.com/buildpacks/pack/issues/376) where an `image` key was added to differentiate between
+directory and image. 
+ 
+# What it is
+[what-it-is]: #what-it-is
+
+### Scheme
+
+| Short Name | Format | Examples |
+|--- |--- |--- |
+| Filesystem | `file://[<host>]/<path>` | `file:///./my/buildpack.tgz`<br>`file:///home/user/my/buildpack.tgz`
+| URL | `http[s]://<host>/<path>` | `http://example.com/my/buildpack.tgz`<br>`https://example.com/my/buildpack.tgz`  
+| Docker | `docker://[<host>]/<path>[:<tag>]` | `docker:///ubuntu:latest`<br>`docker://gcr.io/distroless/nodejs` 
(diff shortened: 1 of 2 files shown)


text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,128 @@
...
+provide reference to all buildpacks in builder.
+- [Issue](https://github.com/buildpacks/pack/issues/376) where an `image` key was added to differentiate between
+directory and image. 
+ 
+# What it is
+[what-it-is]: #what-it-is
+
+### Scheme
+
+| Short Name | Format | Examples |
```
nebhale: There appears to be an inconsistency here in that only one of the URI schemes refers to a "type" of the thing being referenced.  For example, what if I want to refer to a builder at a URL or a buildpackage in the registry?  I think there are two ways of handling this.  Either the URI **never** has a reference to the type and the type is inferred contextually (`--buildpackage cnb://cloudfoundry/openjdk:1.2.0 --builder https://localhost/builder.tgz`) or the URI **always** contains the type to reduce ambiguity (`--buildpackage cnb:package:registry://cloudfoundry/openjdk:1.2.0 --builder cnb:builder:https://localhost/builder.tgz`).

I prefer the latter due to its extensibility and inability to be ambiguous:

```
cnb:[builder|package|dependency|...]:<any valid URI>
```
```
cnb:package:file:///./my/buildpack.tgz
cnb:builder:file:///home/user/my/builder.tgz
```
```
cnb:package:http://example.com/my/buildpack.tgz
cnb:builder:https://example.com/my/builder.tgz
```
```
cnb:package:docker:///mybuildpack:latest
cnb:builder:docker://gcr.io/mybuilder:latest
```
```
cnb:package:registry:///mybuildpack:latest
cnb:builder:registry://gcr.io/mybuilder:latest
```

jromero: I tried to make it explicit in the summary but maybe haven't justified the why.

> Please note that the scheme denotes protocol and not type.

The "type" IMO should be handled via other methods and scheme should only detail protocol. Things like [media-type](https://github.com/opencontainers/image-spec/blob/master/media-types.md) or in our simplistic case for OCI images labels could be used.

nebhale: Then `cnb:builder://` has type information and shouldn't.

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,128 @@
...
+itself. We currently _try_ to resolve the various sources through a complex set of rules yet the ambiguity to users
+still exists. In config files, this ambiguity is at times resolved by introducing a separate key name to signify
+the source (protocol).
+
+The aim of this RFC is to propose a set of URI schemes that could be used in both CLI arguments and config files
+while solving the ambiguity problem.
+
+A few references to further express the problem:
+
+- [A poll](https://buildpacks.slack.com/archives/C94UJCNV6/p1580920669081100) to understand what users expect when
```
jkutner: This is great! can we put the results somewhere durable?

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,128 @@
...
+- `cnb:builder://bp.id` - A reference to a buildpack with id `bp.id` in the builder. If there is more than one version
+available, a error should be thrown.
+- `cnb:builder://bp.id@bp.version` - A reference to a buildpack with id `bp.id` at the specific version `bp.version`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Usage
+
+Certain `pack` CLI arguments would take URIs supported instead of the current inconsistent format.
```
jkutner: Just to be clear, you mean that the naked `foo/bar` format would be disallowed?

text/0000-buildpack-uris.md
zmackie: We *could* pay for our slack instance :) somehow. But for now, here's an image of the poll: 
Screen Shot 2020-02-12 at 1 39 10 PM

text/0000-buildpack-uris.md
jromero: It's not really a type. It's the "protocol" of how to search for buildpacks within a builder. That's at least my interpretation/justification.

text/0000-buildpack-uris.md
jromero: Right unless we want to make a very explicit default but I thought we would want to wait until the registry is available.

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,128 @@
...
+- `<host>` is optional and defaults to `index.docker.io`.
+- Similar to the `file` scheme, there is a minimal declaration for omitting host by using a simple slash (`/`). 
+eg. `docker:/ubuntu:latest`
+
+### CNB Builder
+
+- `cnb:builder://` - A reference to ALL the buildpacks in the builder.
+- `cnb:builder://bp.id` - A reference to a buildpack with id `bp.id` in the builder. If there is more than one version
+available, a error should be thrown.
+- `cnb:builder://bp.id@bp.version` - A reference to a buildpack with id `bp.id` at the specific version `bp.version`.
```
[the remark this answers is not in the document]
ekcasey: @nebhale in that case we might have a problem with the `docker://` scheme as well. Presumably we would want to be able to refer to images by digest:  `<registry>/<repo>@<digest>`

text/0000-buildpack-uris.md
jkutner: Is there anything preventing us from accepting the naked format, and *always* having it mean one thing (i.e. we always put `cnb:builder://` or whatever in front of it)? That means you only need to add a scheme if you want a local `file://` or whatever.

jromero: Yes, that's listed under additional considerations: https://github.com/buildpacks/rfcs/blob/buildpack-uris/text/0000-buildpack-uris.md#additional-considerations

Also, it seems like we might need to have the "naked" format be relative URIs. See @ekcasey comment below.

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,128 @@
...
+directory and image. 
+ 
+# What it is
+[what-it-is]: #what-it-is
+
+### Scheme
+
+| Short Name | Format | Examples |
+|--- |--- |--- |
+| Filesystem | `file://[<host>]/<path>` | `file:///./my/buildpack.tgz`<br>`file:///home/user/my/buildpack.tgz`
```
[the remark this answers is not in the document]
jromero: Some additional context:

image

text/0000-buildpack-uris.md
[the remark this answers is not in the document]
jromero: @ekcasey this doesn't look like it applies to the `docker://` scheme as it's [only reserved in the authority segment](https://tools.ietf.org/html/rfc2396#section-3.2), it is [allowed in the path segment](https://tools.ietf.org/html/rfc2396#section-3.3).

zmackie: > A non-empty scheme component followed by a colon (:), consisting of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus (+), period (.), or hyphen (-). Although schemes are case-insensitive, the canonical form is lowercase and documents that specify schemes must do so with lowercase letters. Examples of popular schemes include http, https, ftp, mailto, file, data, and irc. URI schemes should be registered with the Internet Assigned Numbers Authority (IANA), although non-registered schemes are used in practice.[b]

This suggests that our scheme should be `cnb-builder`.

jromero: Wikipedia is a loose form of a source. :D

Here's an example of a valid URI that has been spec'd containing multiple colons:

[URN](https://tools.ietf.org/html/rfc8141#section-2):
> A Uniform Resource Name (URN) is a Uniform Resource Identifier (URI)
   that is assigned under the "urn" URI scheme and a particular URN
   namespace, with the intent that the URN will be a persistent,
   location-independent resource identifier.

```
urn:example:1/406/47452/2
```

jromero: This section might also help: https://tools.ietf.org/html/rfc3986#section-1.1.2

jromero: And another: https://tools.ietf.org/html/rfc3986#section-3

image

zmackie: https://tools.ietf.org/html/rfc3986#section-3.1
> scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

>  Scheme names consist of a sequence of characters beginning with a
   letter and followed by any combination of letters, digits, plus
   ("+"), period ("."), or hyphen ("-").  Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for

zmackie: BTW, I may be wrong here, but I think we're discussing a URI (vs a URN which does not indicate how to access a resources); if that's true the scheme is a bit tighter than the UR_N_ scheme.

zmackie: Maybe I'm being pedantic, but I just find it hard to viusally differentiate between the scheme and the authority if the scheme has colons in it. To my eyes, that first colon indicates a seperation between the scheme and the rest of the URI, but that's just me.

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,172 @@
...
+# ...
+```
+
+# Additional considerations
+
+#### `pack` User Expirience
+
+`pack` could provide additional logic in order for users to have a nicer user experience when providing references
+via CLI. For example, `pack` could have an order of precedence to make a best guess at which type of reference
+they are referring to or provide additional feedback if it's ambiguous.
```
jkutner: I think I'd recommend this order:

1. builder image
2. registry
3. file system path

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,172 @@
...
+#### Registry vs Daemon
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- The user would need to type an additional scheme prefix for any URI which is not a relative reference.
+
+# Alternatives
```
jkutner: As an alternative, what if we defined `cnb` URN patterns. Unlike URLs, which identify resources by specifying their locations, URNs are location-independent identifiers assigned within defined namespaces.

 For example:

```
# reference a buildpack in the builder
urn:cnb:builder:[<id>[:<version>]]

# reference a buildpack on the default cnb-registry
urn:cnb:registry:[<id>[:<version>]]
```

In this way, the Builder image or the Registry location can be defined independently. I think this will be pretty common (`pack` already has a default builder, and will have a default registry).

If you need to specify buildpacks in more than one registry, then use the following URI:

```
# always points to a buildpack on a specific cnb-registry
cnb://[<host>]/[<id>[:<version>]]
```

Because there's no concept of pointing to buildpacks in more than one builder, we don't need to define a URI for buildpacks-in-a-builder (i.e. there's no need to define the location, so the URN should do).

Then we can honor all the other `file://`, `docker://`, etc schemes defined in the RFC.

zmackie: Oooh I like this!

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,179 @@
...
+
+- Predefined set of URI schemes for OCI images created by [Skopeo](https://github.com/containers/skopeo).
+- [from=builder][from-builder] syntax.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Where would this be documented long-term?
+    - Answer: Would solely live in the docs site.
+- How does `urn:cnb:registry:bp.id:bp.version` translate to a `namespace` + `name` as detailed in [RFC #22](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)?
```
jromero: There is an unresolved question on the original RFC:

- [Should we split id into namespace and id?](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md#unresolved-questions)
    - @jkutner, was there any insight there that might inform this unresolved question?

text/0000-buildpack-uris.md
jromero: Thank you! I've included this suggestion into the latest changes.

text/0000-buildpack-uris.md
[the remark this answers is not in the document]
jromero: Added `relative` URI syntax.

text/0000-buildpack-uris.md
jkutner: As of now, a `namespace/name` is an acceptable form of buildpack ID (like ID is a superset). So think defining the URN as `urn:cnb:registry:id:version` is sufficient for this proposal. In a future proposal we might define more strict rule about ID, but this definition would stay the same. Does that make sense?

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,187 @@
...
+
+| Name | Format | Examples |
+|--- |--- |--- |
+| Relative | `<path>` | `./my/buildpack.tgz`<br>`/home/user/my/buildpack.tgz`
+| Filesystem | `file://[<host>]/<path>` | `file:///my/buildpack.tgz`<br>`file:///home/user/my/buildpack.tgz`
+| URL | `http[s]://<host>/<path>` | `http://example.com/my/buildpack.tgz`<br>`https://example.com/my/buildpack.tgz`  
+| Docker | `docker://[<host>]/<path>[:<tag>⏐@<digest>]` | `docker://gcr.io/distroless/nodejs`<br>`docker:///ubuntu:latest`<br>`docker:///ubuntu@sha256:45b23dee08...`
+| CNB Registry | `cnb://[<host>]/[<id>[:<version>]]` |  `cnb:///my-org/my-bp`<br>`cnb://index.buildpack.io/my-org/my-bp`
+| CNB Builder Resource | `urn:cnb:builder[:<id>[:<version>]]` | `urn:cnb:builder`<br>`urn:cnb:builder:bp.id`<br>`urn:cnb:builder:bp.id:bp.version`
+| CNB Registry Resource | `urn:cnb:registry[:<id>[:<version>]]` | `urn:cnb:registry:bp.id`<br>`urn:cnb:registry:bp.id:bp.version`
```
jkutner: In pack we already use `<id>@<version>`. Should we make this `urn:cnb:registry:<id>@<version>` instead of the `:`?

text/0000-buildpack-uris.md
jromero: The RFC has been update to incorporate URNs to reference buildpacks in certain scenarios and additional considerations for registry resolved buildpacks.

text/0000-buildpack-uris.md
jromero: I have no preference but do like the simplicity of using colon `:` as a general segment  separator.

jkutner: agree. but i'd hate to have two different way's of handling it:

`pack build --buildpack example/foo@1.0`

and

`pack build --buildpack urn:cnb:registry:example/foo:1.0`

jromero: Two thoughts.

1. We could migrate pack to always use colon `:`
2. `example/foo:1.0` does have the undertone of being "mutable" where `@` is "immutable" thanks to docker and node. Is there anything we want to carry forward from that?

Does that help in any way?

jromero: From WG, let's go with `@` for URNs and CNB Registry. It implies immutability, is inline current implementation, and treats `id` + `version` as a unit vs separate segments.

text/0000-buildpack-uris.md
```diff
@@ -0,0 +1,189 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- Predefined set of URI schemes for OCI images created by [Skopeo](https://github.com/containers/skopeo).
+- [from=builder][from-builder] syntax.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Where would this be documented long-term?
```
hone, line 183: I wonder if we should move this to an extension spec or part of the [project.toml one](https://github.com/buildpacks/spec/pull/70).

jromero: Update to detail extension spec.

nebhale:
I move for an FCP, ending 29 April, for merge.

PR #59, opened 2020-03-10
Implements Optional spec field

zmackie:
Implements #52 

Signed-off-by: Zander Mackie <zmackie@gmail.com>


nebhale:
Resolved with 47270e3

PR #60, opened 2020-03-11
RFC for buildpack version formats

danielleadams:
[readable](https://github.com/buildpacks/rfcs/blob/da/buildpack-versions/text/0000-buildpack-versions.md)

### DIFF
--- a/text/0000-buildpack-versions.md
+++ b/text/0000-buildpack-versions.md
@@ -1,0 +1,62 @@
+# Meta
+[meta]: #meta
+- **Name:** Semantic Version Formats for Buildpack Versions
+- **Start Date:** 2020-03-11
+- **Supersedes:** N/A
+
+# Summary
+[summary]: #summary
+
+The Buildpacks team wants to enforce format requirements for buildpack versions but will not require version semantics. This RFC explains that buildpack authors are expected to use [semantic version](https://semver.org/) format without enforcing any semantic versioning rules.
+
+# Motivation
+[motivation]: #motivation
+
+There aren't specific guidelines in the buildpack documentation or spec for buildpack authors that explain how versioning works for buildpacks. It should be explicitly stated that the API expects a semantic version format and that authors are not required to follow semantic version semantics.
+
+Additionally, there are benefits of having a defined format because tools, such as `pack`, will be able to sort and filter buildpacks as needed. For example, with an expected semantic version format, a `latest` version is defined.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors are required to use the semantic versioning format, such as `0.10.4`. Currently, the Buildpack API already works with any version format. This requires no additional code to the current code bases, but instead clarification around what the Buildpacks projects expect.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Buildpacks team is making a commitment to the future of the project that buildpack developers will have the freedom to use any rules for versioning while using the semantic version format, and this decisions will not break the Buildpack API.
+
+### Tasks to complete:
+- explicitly state that semantic version format is expected in the buildpack spec, as well as the docs
+- change any current messaging (ie. variable names or code comments) that states otherwise
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This decision may lead to inconsistency across buildpacks for developers that are using buildpacks or multiple buildpacks.
+- Buildpack authors may request more guidance on buildpack versioning.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Do nothing/Don't explicitly specify expected version format.
+  - Ideally, we don't want to make future versions of buildpacks incompatible with what the Buildpacks team expects the future Buildpack API to look like, such as expecting a semantic version format for sorting.
+- Enforce semantic versioning rules for buildpack versions.
+  - There isn't an immediate benefit to this since versioning can tie into a development process and workflow (and should not be dictated by the tool).
+
+# Prior Art
+[prior-art]: #prior-art
+
+- Semantic Versioning docs: https://semver.org/
+- Issue opened to specify version requirements in the spec: https://github.com/buildpacks/spec/issues/72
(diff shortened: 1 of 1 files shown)


text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,56 @@
...
+[summary]: #summary
+
+The buildpacks project **does not** want to enforce rules or format requirements for buildpack versions. This RFC explains that buildpack authors can use any pattern (ie. semantic versioning) when deciding how to release versions of buildpacks.
+
+# Motivation
+[motivation]: #motivation
+
+There aren't specific guidelines in the buildpack documentation for buildpack authors that explain how versioning works for buildpacks. It should be explicitly stated that the API supports any version format.
+
+Therefore, buildpack authors have clear documentation that they can use any versioning system that works for their development workflow.
```
jkutner: "should have"? actually, i think this sentence could be cut.

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,56 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors can version buildpacks how they want. This requires no additional code, but instead clarification around how versions are referred to.
```
jkutner: it might be worth calling out that this is the status quo, and what you are proposing is to codify that in the spec.

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,56 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Buildpack team is making a commitment to the future of the project that buildpack developers will have the freedom to format buildpacks anyway, and these format decisions will not break the Buildpack API.
```
jkutner: "format buildpack versions"?

jromero:
Was there not a discussion that we had in a working group meeting that was leaning more towards semver format without the semantics? This would allow for ranges and potentially dependency conflict resolution.

danielleadams:
@jromero At the Buildpack WG meeting, there was a discussion to not enforce semantic versioning (iirc). What would "semver format without the semantics" mean?

hone:
From the WG: we want to at least enforce a sem ver format, so we can tell easily what is latest + sorting.

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,55 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors are required to use the semantic versioning format, such as `0.10.4`. Currently, the Buildpack API already works with any version format. This requires no additional code to the current code bases, but instead clarification around what the Buildpacks projects expect.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Buildpacks team is making a commitment to the future of the project that buildpack developers will have the freedom to use any rules for versioning while using the semantic version format, and this decisions will not break the Buildpack API.
```
hone, line 25: I think this technically breaks the spec since currently, we [don't specify versioning for buildpacks](https://github.com/buildpacks/spec/blob/master/buildpack.md#buildpacktoml-toml).

As part of this, we should add a [spec change section](https://github.com/buildpacks/rfcs/blob/master/text/0029-template-changes.md).

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,55 @@
...
+- **Name:** Semantic Version Formats for Buildpack Versions
+- **Start Date:** 2020-03-11
+- **Supersedes:** N/A
+
+# Summary
+[summary]: #summary
+
+The Buildpacks team wants to enforce format requirements for buildpack versions but will not require version semantics. This RFC explains that buildpack authors are expected to use [semantic version](https://semver.org/) format without enforcing any semantic versioning rules.
+
+# Motivation
```
hone, line 12: @ekcasey is it worth addding some of the benefits we get for having this format for things in `pack`, i.e. latest.

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,60 @@
...
+[prior-art]: #prior-art
+
+- Semantic Versioning docs: https://semver.org/
+- Issue opened to specify version requirements in the spec: https://github.com/buildpacks/spec/issues/72
+- Code comment that makes it seem sem-ver is expected: https://github.com/buildpacks/libbuildpack/blob/b2e4a9ef07aff26c430dfac2fbc928ee4164af35/buildpack/info.go#L27
+
+# Spec. Changes
+[spec-changes]: #spec-changes
+
+The RFC proposes a breaking change to the spec because it will enforce a buildpack version format that is not specified. It will need to be defined here: https://github.com/buildpacks/spec/blob/master/buildpack.md#buildpacktoml-toml
```
danielleadams, line 55: @hone I added a spec changes section that specifies a breaking change

text/0000-buildpack-versions.md
hone: @danielleadams is it worth mentioning here that one of the benefits of versioning in that platforms can actually make sense of version numbers based of the semantic versioning format number. i.e. "latest" is defined now.

danielleadams: yeah, I can add that.

text/0000-buildpack-versions.md
```diff
@@ -0,0 +1,62 @@
...
+[summary]: #summary
+
+The Buildpacks team wants to enforce format requirements for buildpack versions but will not require version semantics. This RFC explains that buildpack authors are expected to use [semantic version](https://semver.org/) format without enforcing any semantic versioning rules.
+
+# Motivation
+[motivation]: #motivation
+
+There aren't specific guidelines in the buildpack documentation or spec for buildpack authors that explain how versioning works for buildpacks. It should be explicitly stated that the API expects a semantic version format and that authors are not required to follow semantic version semantics.
+
+Additionally, there are benefits of having a defined format because tools, such as `pack`, will be able to sort and filter buildpacks as needed. For example, with an expected semantic version format, a `latest` version is defined.
```
danielleadams, line 17: @hone I added an additional part about being able to sort and filter buildpacks in the future.

PR #63, opened 2020-03-12
Image exposes metadata for all layers that participated in an image's build

pclalv:


text/0031-expose-all-layer-metadata.md
```diff
@@ -0,0 +1,138 @@
...
+6. I build `some-app @ 000000` without a cache (to "seed" the cache).
+7. I build `some-app @ 111111` with the cache from `some-app @ 000000`
+   and reproduce the issue.
+
+This example conflates the related issue of knowing what version of an
+application a particular layer arose from; however, the first step
+towards surfacing that information could be including all layer
+metadata in an image. Given that the layer metadata will be preserved,
+the buildpack author user or the platform could add the application's
+version (e.g. a git SHA) to the layer metadata.
```
ekcasey: The lifecycle, as of Platform API 0.3, does currently provide a (poorly document) mechanism for platforms to set source metadata on the image (https://github.com/buildpacks/rfcs/blob/master/text/0013-app-layer-metadata-source.md, buildpacks/lifecycle#230)

text/0031-expose-all-layer-metadata.md
```diff
@@ -0,0 +1,138 @@
...
+where launch-layer metadata is stored. Doing so would be mostly
+backwards compatible; the layer metadata for all layers is a superset
+of the layer metadata for launch layers.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* It's not perfectly backwards compatible; users who expect to be able
+  to use the value of the `io.buildpacks.lifecycle.metadata` label to
+  enumerate an image's launch layers will get different results;
```
ekcasey: I wonder if a different label or a different key in the structured label could help here:

Option 1: `io.buildpacks.cache.metadata` contains all of the metadata that we currently put in the cache. A user could compare the cache metadata from an image to a specific cache's metadata. This could include all `cache=true` layers but we still aren't capturing `cache=false && launch=false` layer metadata.

Option 2: right now, in the label, buildpacks have a `layers` key, we could change to `exportedLayers` and `unexportedLayers` or something like that

I would like to think more about the best way to organize this but I think keeping it separate somehow is probably a good idea.

text/0031-expose-all-layer-metadata.md
pclalv: I'm sorry, but it's not entirely clear to me how you mean to relate
that lifecycle feature to this RFC. To me, that lifecycle feature
seems to answer a related question, "What version of the app did I use
as input when I made this image?" But in general, the layers used in
that image may have been made with different versions of the same app.

For example, consider an app foo-api and a git repo with the same
name, and say I've already built `registry.example.com/foo-api:000000`
with the the foo-api repo @ `000000`.
1. Now say I build `registry.example.com/foo-api:111111` with the the
   foo-api repo @ `111111`, and all of my cached layers from the
   previous build (for the image
   `registry.example.com/foo-api:000000`) are valid.
1. As a result of that build, I've got an
   `registry.example.com/foo-api:111111` image whose app source code
   comes from the sha `111111`, but whose layers were all generated
   using the app as of the sha `000000`.

In this case, I believe that `registry.example.com/foo-api:111111`
would have a label `Config.Labels."io.buildpacks.app.source"` that
identifies the app source code as coming from the sha `111111`, but
there's nothing to tell me that the launch layers (let alone
non-launch layers) were generated with the sha `000000`. So, that's
the issue that I'm trying to relate.

This example actually does come my team's use of buildpacks tools.

To try to relate this to your concept of 'source metadata', I might
say that there are actually many sources involved in image builds,
perhaps the most obvious two being the application source code and the
cached layers. In addition to having 'source metadata' at the image
level, to say what version of the application source code the image
contains, it's also valuable to have 'source metadata' at the layer
level to say what version of the application source code was on-disk
when a buildpack generated that layer.

pclalv:
If you're still uncomfortable with the application version example
after reading [my comment][my-comment], or even if you'd like another
example, how about something like this?

In general, it's not possible to identify the non-launch layers that
participated in an image build, but it would be useful to do
so. Consider the case of an app build that uses a NodeJS buildpack;
the app source code does not vendor a `node_modules` directory, and
while `node_modules` will be used at build time, `node_modules` do not
need to be present in the launch image (i.e. because we only used
`node_modules` to compile the JS that lives in the launch image); so,
`node_modules` are only present in a cache layer. In order to debug a
problem with the compiled JS in some image image `foo-js:some-tag`, it
would be useful, at worst _necessary_, to identify exactly the
`node_modules` cache layer that participated in the build for the
image `foo-js:some-tag`, ideally by inspecting a label on
`foo-js:some-tag`.

[my-comment]: https://github.com/buildpacks/rfcs/pull/63#discussion_r394618456

text/0031-expose-all-layer-metadata.md
ekcasey: One reason I am resisting the idea of adding source metadata to the layer metadata, is that the source metadata isn't necessarily available to the buildpack at build time. One example: once `pack` supports `include` and `exclude` in `project.toml`, the `.git` directory is a likely candidate to be excluded.

I see that point you  are making, the source may influence what was created in a given layer. I think there is almost always a more direct way to express the relevant info. To take your `node_modules` example, the hash of the `package-lock.json` can be used to determine whether an existing cache layer containing `node_modules` is valid or if needs to be updated. The buildpack can already use the layer metadata to prevent a layer withoutdated models from being added to a new image. You could take the visibility a step further and add the actual contents of the resolved tree to the layer metadata (assuming this RFC get's accepted) or to the bill-of-materials (right now). If I am debugging an issue, I want to know what state at build time was relevant to the layer generation, without having to trace that back to the git sha of the last commit that changed the `package-lock.json`, especially since many other unrelated but important changes could have happened since then. If each of my cache-only layers came from a different sha, it wouldn't be easy to re-seed the cache using that information.

I really like the idea of adding all layer metadata to the image, but I think source metadata in particular, as an example, introduces distracting complications. Nothing would stop your buildpack from adding source metadata if you want it (and it is always available in your use cases) but I don't think it's the cleanest example for the RFC.

pclalv: My team's nodejs buildpack does use a content digest of
`package-lock.json` to validate its cache. That value has its place,
but it's no good for doing a reverse look-up; rather than require any
sort of reverse look up, I think that the layer should tell a user
exactly where it came from. For me, a checksum of `package-lock.json`
is not a more direct expression; perhaps I haven't fully under your
case.

I'm sure that I don't understand what you mean here:

>You could take the visibility a step further and add the actual
>contents of the resolved tree to the layer metadata (assuming this
>RFC get's accepted) or to the bill-of-materials (right now).

I do agree with you on this point:

>If I am debugging an issue, I want to know what state at build time
>was relevant to the layer generation, without having to trace that
>back to the git sha [...]

However, what if it was a bug in the buildpack that lead to some bad
cache? In that case, it would be useful to know the git SHA that was
used to generate the bad cache, and the version of the buildpack, to
reproduce the issue.

The use of git SHA that I'm trying to describe isn't about validating
the cache. It's about knowing how to reproduce the state of the cache
exactly in order to debug issues that arise when buildpacks, the
cache, and the application come together.

It's not ideal to have to checkout an old SHA and build it, but
sometimes it will be necessary to do so, particularly if the problem
exists in the buildpack code. I'm not recommending that anyone use
this as their go-to technique for debugging buildpack issues, but
sometimes there's nothing else that can be done. In general, I think
that there will be many debugging cases that would benefit from
knowing the version of the application in addition to the version of
the buildpack.

Layer source metadata is indeed distracting and complicating; if I
could add one last complication, it'd be that I don't want to have to
add source metadata for each of my layers - I just want to get that
for free, without adding boilerplate to each of my buildpacks.

That said, even now I have some doubts about making this reply,
because it's not clear to me that we're understanding one another.

sclevine: > the source metadata isn't necessarily available to the buildpack at build time. One example: once pack supports include and exclude in project.toml, the .git directory is a likely candidate to be excluded.

To me, this is definitely a problem with the use case. But I don't have a strong opinion about what example is used in the RFC, even though it seems like a more simple example would suffice.

nebhale:
Waiting for Working Group discussion

hone:
Paul came to the Working Group session today talking about his use case. We may not want to expose build concerns on the image labels since it may affect reproduceability. Instead, Paul is going to look at the report.toml RFC #70.

hone:
@pclalv are you ok with me moving this into draft?

jkutner:
@pclalv I accidentally moved this to "Draft" because I thought it was superseded by #94, but now I see that's in draft too. Feel free to reopen this if you need/want to. Sorry!

PR #64, opened 2020-03-15
Add Author to RFC Metadata

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/author/text/0000-add-author.md)

### DIFF
--- a/text/0000-add-author.md
+++ b/text/0000-add-author.md
@@ -28,6 +28,8 @@ You're not going to believe this, but you add an `Author(s):` field to the Meta 
 
 We will also rename the `CNB Pull Request` to `RFC Pull Request`.
 
+Because `CNB Issue` is rarely used, we will also renamed it to `Related Issues`.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/text/0000-add-author.md
+++ b/text/0000-add-author.md
@@ -1,0 +1,55 @@
+# Meta
+[meta]: #meta
+- Name: Add Author to RFC Metadata
+- Start Date: 2020-03-14
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This is a proposal to add an `Author(s):` field to the `Meta` section of the RFC template.
+
+# Motivation
+[motivation]: #motivation
+
+When an RFC is merged, and then renamed, the commit history is lost. This makes it difficult to remember who wrote the RFC.
+
+# What it is
+[what-it-is]: #what-it-is
+
+A new `Author(s):` field in the Meta section of the RFC template.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+You're not going to believe this, but you add an `Author(s):` field to the Meta section of the RFC template.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Even though the author will be know, the commit history is still hard to find.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## RFC PR Link
+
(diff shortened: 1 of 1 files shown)


text/0000-add-author.md
```diff
@@ -0,0 +1,55 @@
...
+[drawbacks]: #drawbacks
+
+Even though the author will be know, the commit history is still hard to find.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## RFC PR Link
+
+Instead of an author field, we could enforce that a link to the original RFC PR be added to the RFC after it's merged. This would make it easier to find the author and the history.
```
hone: We should be doing this already like [here](https://github.com/buildpacks/rfcs/blob/master/text/0003-pack-inspect-image.md).

ekcasey: Maybe the name could be improved. `RFC PR` would disambiguate between the rfc PR itself and a draft PR implementing the RFC.

And while we are at it, `CNB Issue` appears blank in most merged RFCs. Maybe `Related Issues` instead, to optionally link to any relevant issues?

text/0000-add-author.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Summary
+[summary]: #summary
+
+This is a proposal to add an `Author(s):` field to the `Meta` section of the RFC template.
+
+# Motivation
+[motivation]: #motivation
+
+When an RFC is merged, and then renamed, the commit history is lost. This makes it difficult to remember who wrote the RFC.
```
hone: Can you add the bit about multiple authors here from the WG?

text/0000-add-author.md
```diff
@@ -0,0 +1,59 @@
...
+[what-it-is]: #what-it-is
+
+A new `Author(s):` field in the Meta section of the RFC template.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+You're not going to believe this, but you add an `Author(s):` field to the Meta section of the RFC template.
+
+We will also rename the `CNB Pull Request` to `RFC Pull Request`.
```
hone, line 29: I added both the RFC Pull Request and the CNB Pull Request to this item: https://github.com/buildpacks/rfcs/blob/master/text/0005-contractual-build-plan.md

Should we drop one, keep it the same, or just make it more clear. I agree that having the discussion linked is important.

text/0000-add-author.md
```diff
@@ -0,0 +1,59 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+A new `Author(s):` field in the Meta section of the RFC template.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+You're not going to believe this, but you add an `Author(s):` field to the Meta section of the RFC template.
```
hone, line 27: Is this just name, github handle, email?

text/0000-add-author.md
jkutner: I've changed this to leave the `CNB ...` fields alone. we can address those in another RFC.

text/0000-add-author.md
jkutner: updated to `Github username`

text/0000-add-author.md
```diff
@@ -0,0 +1,60 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- Rust uses an ["RFC PR" link](https://github.com/rust-lang/rfcs/blob/master/0000-template.md)
+- TensorFlow has an ["Author(s)" field](https://github.com/tensorflow/community/blob/master/rfcs/yyyymmdd-rfc-template.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Does this make it look like only a few person contributed to the RFC (when in reality they are a collaborative process)?
```
hone, line 55: Maybe, but with the easy access to the discussion you can always check the history. We could also encourage / welcome more folks to the RFC process. :)

PR #65, opened 2020-03-15
Update JSON structure in RFC0022

jkutner:
This fixes a mistake that was made in the original RFC.


hone:
From the WG, we shouldn't be updating existing RFCs if it changes the original intent. @jkutner will submit addendum RFCs.

PR #67, opened 2020-03-24
Build Plan Merging

dwillist:
RFC for adding `build` and `launch` flags to the BuildPlan and Buildpack Plan

[Readable](https://github.com/dwillist/rfcs/blob/build-launch-flags/text/0000-build-plan-merging.md)


nebhale:
I'd also like to propose that we demote `version` as part of this RFC.  After a year of messing around with the declarative build plan, I don't think that other buildpacks can responsibly ask for a version of a dependency.  For example, what are the odds that a buildpack could ask for a version of `node-engine` that differed from the one in `package.json` or a version of Ruby that didn't match the application's `Gemfile.lock`.  Don't even get me started about Java applications expecting Java 8 and getting Java 11 🙄.

I think that the only actor that can properly request a version of a dependency is the application itself (or a manual override done via env-var, etc.).

hone:
From discussion on slack in #spec, the original authors will update this RFC.

sclevine:
Note from chat on 5/20: would be good to note that build-time-only dependencies should be excluded from the BOM. An interface for indicating that (e.g., `build` being a special capability or a separate `exclude` flag) may be desirable

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+This proposal adds a new field to `provides`:
+- `provides.strategy.version` field holds a `collect` key which indicates if the version fields from all of the requires should be merged together or if they should collected into an array.
+  - This field is structured to be extensible to other fields in `requires`.
+
+This proposal also removes both the `requires.version` and `requires.metadata` fields and replaces them with three new fields each with their own purpose:
+- `requires.version` field holds all of the version requirement information.
+  - `constraint` can be any semver constraint used to decide which dependency the providing buildpack will install.
+  - `source` is the origin of the constraint (i.e. it came from `package.json`).
```
hone, line 91: Is this field required? Also, what do you use it for, the BOM at the end?

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+  - Collects all  of the data from various `requires` entries into a unique entry for a given dependency.
+  - Allows for the implementation of information merging strategies for the `version`, `capabilities`, and `modifiers`.
+    - All version constraints are merged together into the `version` field.
+      - If the merge would result in a valid semver constraint, then image build will continue.
+      - If the merge fails to produce a valid semver constraint (i.e. a constraint the is unreachable such as "1.10.\*, 1.12.\*"), then it would cause the image build to fail.
+    - All `capabilities.build` and `capabilities.launch` for a given dependency will be `or` merged together.
+    - All keys in `modifiers` will be merged with a last in wins strategy.
+- The `collection` list will only contain elements if the `provides` strategy was set to a `collect = true`.
+  - If the `collection` field is present the `version` field will be empty.
+  - The `collection` field has been structured to be extensible to other fields in `requires`, this will allow for the extension of `provides.strategy`.
```
hone, line 129: If there's only 1 field here, do we need a separate table?

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+We arrived at this format for several reasons:
+- Moves some common buildpack operations into the lifecycle.
+  - Collects all  of the data from various `requires` entries into a unique entry for a given dependency.
+  - Allows for the implementation of information merging strategies for the `version`, `capabilities`, and `modifiers`.
+    - All version constraints are merged together into the `version` field.
+      - If the merge would result in a valid semver constraint, then image build will continue.
+      - If the merge fails to produce a valid semver constraint (i.e. a constraint the is unreachable such as "1.10.\*, 1.12.\*"), then it would cause the image build to fail.
+    - All `capabilities.build` and `capabilities.launch` for a given dependency will be `or` merged together.
+    - All keys in `modifiers` will be merged with a last in wins strategy.
+- The `collection` list will only contain elements if the `provides` strategy was set to a `collect = true`.
```
hone, line 127: Is the default value `false` or is this a required field?

In addition, what's the use case for when this is `false`?

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+
+We arrived at this format for several reasons:
+- Moves some common buildpack operations into the lifecycle.
+  - Collects all  of the data from various `requires` entries into a unique entry for a given dependency.
+  - Allows for the implementation of information merging strategies for the `version`, `capabilities`, and `modifiers`.
+    - All version constraints are merged together into the `version` field.
+      - If the merge would result in a valid semver constraint, then image build will continue.
+      - If the merge fails to produce a valid semver constraint (i.e. a constraint the is unreachable such as "1.10.\*, 1.12.\*"), then it would cause the image build to fail.
+    - All `capabilities.build` and `capabilities.launch` for a given dependency will be `or` merged together.
+    - All keys in `modifiers` will be merged with a last in wins strategy.
```
hone, line 126: Do modifiers end up in the BOM?

text/0000-build-plan-merging.md
ekcasey: +1 to @hone 's question

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+  some-modifer = "modifier"
+[[entries.collection]]
+[entries.collection.version]
+  constraint = "<dependency constraint>"
+  source = "<dependency constraint source>"
+```
+In each entry the `name` field is **unique**.
+
+We arrived at this format for several reasons:
+- Moves some common buildpack operations into the lifecycle.
```
ekcasey, line 119: Why is the lifecycle the right place for this vs. the language bindings or helper libraries? If the lifecycle merely provides _all_ of the data, a helper library can give present it in a merged on unmerged format on demand, without requiring that preference to move into the spec (as `provides.strategy`).

nebhale:
I find this update to the RFC to be fundamentally at odds to my objections from it's previous incarnation and not only doesn't account for my recommendations [here](https://github.com/buildpacks/rfcs/pull/67#issuecomment-604094608), but strengthens the contract that I think should have been removed.

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+  - Collects all  of the data from various `requires` entries into a unique entry for a given dependency.
+  - Allows for the implementation of information merging strategies for the `version`, `capabilities`, and `modifiers`.
+    - All version constraints are merged together into the `version` field.
+      - If the merge would result in a valid semver constraint, then image build will continue.
```
hone, line 123: so this semver check only happens if you choose to merge?

text/0000-build-plan-merging.md
```diff
@@ -0,0 +1,260 @@
...
+# Motivation
+[motivation]: #motivation
+
+## State of Paketo Buildpacks
+- The current method that the Paketo Buildpacks uses to communicate information to an earlier buildpack is to set an arbitrary string field in the Build Plan (TOML) `requires.metadata` section.
+- Putting this information in the `requires.metadata` obfuscates the flow of information between buildpacks.
+- This implicit contract between dependent buildpacks should be formalized.
+
+### Example:
+ We consider the interactions between the Paketo `node-engine` and `npm`:
```
natalieparellano, line 24: This is really just for my understanding, but would it be possible to provide an example of what the build plan and buildpack plan look like with today's implementation? That might help to illustrate the changes introduced with this RFC.

text/0000-build-plan-merging.md
hone: From the WG, yes they would.

nebhale:
Final Comment Period with close disposition, closing on 10 June, 2020.

PR #68, opened 2020-03-24
Image Labels

nebhale:
A proposal for adding arbitrary labels to images using buildpacks.

### DIFF
--- a/text/0000-image-labels.md
+++ b/text/0000-image-labels.md
@@ -1,0 +1,69 @@
+# Meta
+[meta]: #meta
+- Name: Image Labels
+- Start Date: 2020-03-24
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Buildpacks should be able to expose arbitrary label metadata on images.  This content should be exposed by enhancing buildpack artifacts.
+
+# Motivation
+[motivation]: #motivation
+
+While buildpacks create OCI images, the standard does not expose all elements that OCI images can support to buildpack authors.  One of these elements, [image labels][l], is regularly requested by users in order to contribute both [standard][s] and arbitrary metadata to their generated images.  Once delivered, this functionality would let buildpack authors contribute labels to the images their buildpacks create.
+
+[l]: https://docs.docker.com/engine/reference/builder/#label
+[s]: https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpacks must be able to contribute arbitrary key-value pairs that will be placed as labels on images.  This can be added as a non-breaking change to `launch.toml` as a peer to `processes` and `slices`. It should be comprised of a collection of items, each with `key` and `value` entries.
+
+```toml
+[[labels]]
+key = "<label key 1>"
+value = "<label value 1>"
+
+[[labels]]
+key = "<label key 2>"
+value = "<label value 2>"
+```
+
+Buildpack users will likely want to be able to contribute labels directly as well, but exposure of that functionality should be a platform-concern and implemented in a UX-native way by each platform rather than being standardized.  A buildpack that allows contribution via user-configuration could be used as a stop-gap solution until a platform supports UX-native contribution and as the only solution for platforms that never do.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The collected contents of the `labels` entries contributed by each buildpack should be contributed to the image's `.Config.Labels` collection.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There's no reason not to do this.
+
+# Alternatives
+[alternatives]: #alternatives
+
+This is a straightforward community request and no other designs have been considered.
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [Docker `LABEL` Directive][d]
+
+[d]: https://docs.docker.com/engine/reference/builder/#label
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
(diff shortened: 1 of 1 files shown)


text/0000-image-labels.md
```diff
@@ -0,0 +1,69 @@
...
+Buildpacks must be able to contribute arbitrary key-value pairs that will be placed as labels on images.  This can be added as a non-breaking change to `launch.toml` as a peer to `processes` and `slices`. It should be comprised of a collection of items, each with `key` and `value` entries.
+
+```toml
+[[labels]]
+key = "<label key 1>"
+value = "<label value 1>"
+
+[[labels]]
+key = "<label key 2>"
+value = "<label value 2>"
```
jromero: Is there a specific distinction as to why these would be added to the `launch.toml` vs a directory structure similar to that of environment variables?

nebhale:
> I assume that for labels with the same key, later labels in [[labels]] override earlier labels...

@sclevine Is this something to be described in an RFC, or should it just make it into the spec PR?

text/0000-image-labels.md
nebhale: Creating a directory structure assumes that the relationship will only ever be key-value and can never be extended beyond that.  The `processes` collection, with it's late addition of `direct` is a good example where the flexibility of a collection of TOML objects is useful.

jromero: With that in mind, isn't labels key/value. Are we anticipating that `labels` will get additions?

From my limited-knowledge perspective, it seems odd that what I consider global attributes (env and labels) are set via two _very_ different mechanism. Would we consider moving _env_ to `launch.toml` otherwise?

nebhale: Labels are key-value today, but it's easy to envision some kind of conditional or late reconciliation, etc. possibility in the future and there's no need to close that functionality off today by choosing an inflexible representation.

`/env` is a different case since one of the most important target platforms (Kubernetes) already has an opinion on how to represent environment variables (value-per-file).

text/0000-image-labels.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Buildpack users will likely want to be able to contribute labels directly as well, but exposure of that functionality should be a platform-concern and implemented in a UX-native way by each platform rather than being standardized.  A buildpack that allows contribution via user-configuration could be used as a stop-gap solution until a platform supports UX-native contribution and as the only solution for platforms that never do.
```
matthewmcnew: Is this suggesting that platform should also provide a mechanism to add arbitrary labels to the resulting image?

nebhale: @matthewmcnew Platforms could choose to do so if they wish.  This RFC would not mandate it.

matthewmcnew: In that case there would still be the space for an rfc, to instruct the lifecycle to add arbitrary labels. So, that the platform could enable that.

text/0000-image-labels.md
```diff
@@ -0,0 +1,69 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Buildpacks should be able to expose arbitrary label metadata on images.  This content should be exposed by enhancing buildpack artifacts.
+
+# Motivation
```
hone: can you add a concrete example of why people would use this feature in the buildpack?

text/0000-image-labels.md
jabrown85: I suddenly don’t like the name `launch.toml` if it now contains image labels, processes, and slices. But maybe that is just me.

jkutner: @jabrown85 we've talked about changing the name of `launch.toml` before. I think we should consider changing it at the same time we consider changing the keys in the `<layer>.toml` (`build`, `cache`, `launch`)

ekcasey: To @jromero 's suggestion about adding labels with `<layer>/labels/KEY` files instead of in `launch.toml`, I think there a practical dealbreaker to consider here in addition to the format concern:

* Adding labels in a directory structure would embed the information in the layer itself, so a reused launch layer that is not restored to disk on a rebuild would lose its labels. The fact that `exporter` needs to read the labels makes it materially different from launch time env vars, which are read by the launcher.

Another tradeoff to consider:
* Putting labels in `launch.toml` makes them a per-buildpack concern rather than a per-layer concern. This means that buildpacks must regenerate labels on every build. **I feel like this is the correct call**. If for any reason we wanted a reused layer to bring its labels with it, they would need to be in `<layer.toml>`.

text/0000-image-labels.md
ekcasey: additional exporter flags and/or fields in `project-metadata.toml` might be appropriate for platform contributed layers

text/0000-image-labels.md
```diff
@@ -0,0 +1,71 @@
...
+value = "<label value 2>"
+```
+
+Buildpack users will likely want to be able to contribute labels directly as well, but exposure of that functionality should be a platform-concern and implemented in a UX-native way by each platform rather than being standardized.  A buildpack that allows contribution via user-configuration could be used as a stop-gap solution until a platform supports UX-native contribution and as the only solution for platforms that never do.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The collected contents of the `labels` entries contributed by each buildpack should be contributed to the image's `.Config.Labels` collection.
+
```
ekcasey: Should we call out the merge strategy for duplicate labels? Given the label naming guidelines this hopefully won't be a problem, but we should be explicit just in case. I propose "last wins", using buildpack execution order.

Another option would be to fail to build.

ekcasey: realizing now that @sclevine already raised this point https://github.com/buildpacks/rfcs/pull/68#pullrequestreview-380578707

hone:
Moving this to FCP with a disposition to accept by April 9th.

PR #69, opened 2020-03-31
RFC for custom CA certs

zmackie:
[Viewable](https://github.com/zmackie/rfcs/blob/rfc/custom-ca-certs/text/0000-custom-ca-certs.md)

### DIFF
--- a/text/0000-custom-ca-certs.md
+++ b/text/0000-custom-ca-certs.md
@@ -13,23 +13,37 @@ In order to allow developers to use custom CA certs during build, this RFC:
 1. Suggests an interface between stacks and the platform to allow adding ca-certs to builds and for creating and extending builders with CA certs
 2. Suggest a `pack` UX for using certs during build and for creating and extending builders with CA certs
 
-This RFC builds on some of the ideas proposed in https://github.com/buildpacks/rfcs/pull/23
+**Note**: This RFC builds on some of the ideas proposed in https://github.com/buildpacks/rfcs/pull/23 to begin producing a generic interface for extending build/run images.
 
 # Motivation
 [motivation]: #motivation
 
 Use of custom CA certs is pervasive, especially for developers in enterprise environments. These developers need to fetch assets during build from corp artifact stores, publish to registries with custom certs, and make certs available during runtime for container images. Furthermore, developers might want to more granularly apply certificates, for example leaving certs out of a runtime container that were available during build. In order to account for the possible diversity of methods for adding CA certs in various stacks, the stack itself must embed this knowledge. But we should avoid the possibility of ad-hoc methods and stack forking, by coming up with a community solution.
 
+This solution should further be extensible for other ways of extending builders, thus should have a relatively generic interface. This RFC proposes using that interface for adding CA certs, but leaves the details of other forms of extension up to subsequent RFCs.
+
 # What it is
 [what-it-is]: #what-it-is
 
 ### Specification
-Stack authors may include an `add-cert` executable to the build and/or run images of their stack:
+Stack authors may include an `extend` executable to the build and/or run images of their stack:
 ```
-/cnb/image/{build,run}/add-cert <cert.pem>
+/cnb/image/{build,run}/extend ({build,run}-extend-toml-file) (pkg-cache-dir) | cwd: /
 ```
 
-The platform will invoke this executable for each cert used during an build. This executable should take a single argument, and serve as a wrapper for adding certs to the particular stack in question. Platforms use the certs available on the build image for publishing to registries. All lifecycle phases that could do egress will have build certs.
+For the purpose of CA certs, extend should examine the `[certs] = []` key of the file supplied (by the platform) in the first argument. The file should follow this schema:
+
+```toml
+# extend.toml
+[certs] = [
+    """-----BEGIN CERTIFICATE-----
+AASDASDASDASD
+ASDASDASDSADASD
+ASDSADSADsda
+etc
+-----END CERTIFICATE-----""",
+]
+```
 
 ### UX
 
@@ -104,7 +118,7 @@ certs = ["./cert3.pem"] # Relative to location of `builder.toml`
 # How it Works
 [how-it-works]: #how-it-works
 
-Platforms may extend app images with custom CA certs using the above interface by running the `add-cert` executable as root in a new container and creating an image from the result. Extending an image in this fashion should generate a single layer. This should happen prior to the normal CNB build or rebase process.
+Platforms may extend app images with custom CA certs using the above interface by running the `extend` executable as root in a new container and creating an image from the result. Prior to this execution, the platform will be responsible for populating the contents of `[certs]` in `*-extend-toml`. Extending an image in this fashion should generate a single layer. This should happen prior to the normal CNB build or rebase process.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -126,6 +140,7 @@ This RFC builds on some of the ideas proposed in https://github.com/buildpacks/r
 
 - Build vs Runtime CA-certs? Do we need to split them or should we just collapse this into one?
 - Dynamically extended run image vs just adding to launch image (comes to the same thing?)
+- How does the platform communicate to buildpacks that certs are available?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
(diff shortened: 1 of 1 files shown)


text/0000-custom-ca-certs.md
```diff
@@ -0,0 +1,133 @@
...
+/cnb/image/{build,run}/add-cert <cert.pem>
+```
+
+The platform will invoke this executable for each cert used during an build. This executable should take a single argument, and serve as a wrapper for adding certs to the particular stack in question. Platforms use the certs available on the build image for publishing to registries. All lifecycle phases that could do egress will have build certs.
+
+### UX
+
+##### pack
+Given the following command:
+`pack build --cert /cert1.pem --cert:build /cert2.pem --cert:run /cert3.pem`
```
matthewmcnew, line 38: I am not sure if it makes sense to add certs to the resulting run image especially as the default option. 

The certs will obvious be needed at build time. However, it is a common practice for the platform that runs the resulting image to provide the necessary certs and rotating them when necessary.

matthewmcnew:
This will be complicated for platforms that cannot easily create ephemeral images such as kubernetes based platforms like kpack. The platform will need to schedule a build for the image and write that image to a registry accessible by the kubernetes container runtime. 

If `/cnb/image/{build,run}/add-cert` updated a single path specific to the stack such as `/etc/ssl/certs`. The platform would be able to mount that path in all the subsequent steps to avoid needing to create a new image. This would also generalize well for the run image as a platforms could use the same approach to add ca certs to built image at runtime.

text/0000-custom-ca-certs.md
zmackie: Thank makes sense, but it would be nice to have some mechanism for baking them in too.

zmackie:
> This will be complicated for platforms that cannot easily create ephemeral images such as kubernetes based platforms like kpack. The platform will need to schedule a build for the image and write that image to a registry accessible by the kubernetes container runtime.
> 
> If `/cnb/image/{build,run}/add-cert` updated a single path specific to the stack such as `/etc/ssl/certs`. The platform would be able to mount that path in all the subsequent steps to avoid needing to create a new image. This would also generalize well for the run image as a platforms could use the same approach to add ca certs to built image at runtime.

Interesting, but how do you handle ephemeral builders (my understanding was that there was a way to use arbitary buildpacks on a builder; how does it get constructed in that case?

That said, the idea of use a volume and sharing it is an interesting and good idea (probably better than using an image), and potentially more performant.

frayer:
As a developer in an enterprise setting with custom CA certs, this is of interest.  Custom CA's show up in our build and run phases.

Any thoughts on how the metadata of extended certificates could be visible to buildpacks that may perform extra actions with them?  A JDK buildpack for example may wish to contribute `[extend.run]` certificates to the Java `cacerts` file for runtime cert trust.

zmackie:
@frayer I'm not too aware of how you'd normally add certs the the Java `cacerts` file, but I'm guessing that buildpacks should just assume that the system certs are available for use. @nebhale @jkutner know more about Java CNBs so can probably weigh in here..

matthewmcnew:
> Interesting, but how do you handle ephemeral builders (my understanding was that there was a way to use arbitary buildpacks on a builder; how does it get constructed in that case?

At this point in kpack, we don't support ephemeral builders and encourage users to create new builders in the kpack api. At some point we may allow builds to be scheduled with an order that contains a subset of the buildpacks in a builder by writing a custom order definition.

text/0000-custom-ca-certs.md
```diff
@@ -0,0 +1,148 @@
...
+
+This solution should further be extensible for other ways of extending builders, thus should have a relatively generic interface. This RFC proposes using that interface for adding CA certs, but leaves the details of other forms of extension up to subsequent RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+### Specification
+Stack authors may include an `extend` executable to the build and/or run images of their stack:
+```
+/cnb/image/{build,run}/extend ({build,run}-extend-toml-file) (pkg-cache-dir) | cwd: /
```
sclevine, line 31: Maybe: `(pkg-cache-dir)` -> `(cache-dir)`
Might be worth removing it though, or at least explaining why it's not used in this RFC.

text/0000-custom-ca-certs.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+For the purpose of CA certs, extend should examine the `[certs] = []` key of the file supplied (by the platform) in the first argument. The file should follow this schema:
+
+```toml
+# extend.toml
+[certs] = [
```
sclevine, line 38: TOML syntax here should be `certs =`, and we may want to use a separate top-level table name

text/0000-custom-ca-certs.md
```diff
@@ -0,0 +1,148 @@
...
+[motivation]: #motivation
+
+Use of custom CA certs is pervasive, especially for developers in enterprise environments. These developers need to fetch assets during build from corp artifact stores, publish to registries with custom certs, and make certs available during runtime for container images. Furthermore, developers might want to more granularly apply certificates, for example leaving certs out of a runtime container that were available during build. In order to account for the possible diversity of methods for adding CA certs in various stacks, the stack itself must embed this knowledge. But we should avoid the possibility of ad-hoc methods and stack forking, by coming up with a community solution.
+
+This solution should further be extensible for other ways of extending builders, thus should have a relatively generic interface. This RFC proposes using that interface for adding CA certs, but leaves the details of other forms of extension up to subsequent RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+### Specification
```
sclevine, line 28: Need an explanation of how rebasing will work (and probably need to introduce `pack upgrade` from #23)

text/0000-custom-ca-certs.md
natalieparellano: If we leave the run image out of scope for now, does that obviate the need to do rebase along with this piece of work?

nebhale:
@matthewmcnew @zmackie Ephemeral builders should be achievable simply by pushing a new manifest containing the appropriate layers to a registry, no build necessary.  The same should be achievable here, right?  Just synthesize a new manifest into a registry?

text/0000-custom-ca-certs.md
matthewmcnew: Yes 🙂

matthewmcnew:
@nebhale Yes but, that is still complicated because we would still need to build the layers with some form of a "docker build" type mechanism. 

Today we:
dynamically create a builder -> schedule build pod with init container that injects certs

Creating ephermal builders with ca certs in them would require us to:
schedule build pod to create ca cert layers -> upload to registry -> dynamically create builder -> schedule build pod without init containers.

nebhale:
But that's only done once per stack+certs combo and effectively creates a new stack that is then used with manifest-based synthesis.  So yes it does have to be scheduled once, but then can be re-used infinitely amortizing that time across an entire enterprise, no?

matthewmcnew:
Yes, that's possible but, not without insignificant complexity.

I think it probably makes sense to do as part of the larger app extensions work. 

However, it seems unfortunate that we need to execute some form of in cluster docker build with heightened privileges just to provide ca certs to builds.

nebhale:
Ah the niceties of ca-certs being included in different flavors of linux, in different ways and places.

text/0000-custom-ca-certs.md
zmackie: don't know what a full separate table gets us.

hone:
> I think it probably makes sense to do as part of the larger app extensions work.

Are we waiting on the app extensions RFC for this RFC to get in? These RFCs look super closely tied together. I agree with the motivations behind this RFC.

jkutner:
Same as @hone, I agree with the motivations. However, I'd like to both this and https://github.com/buildpacks/rfcs/pull/23 use buildpacks as an interface for providing the cacerts/extension. I'm working through that with Stephen, and we can probably upstream whatever we land on into this RFC.

sclevine:
Converting to draft pending root buildpack RFC.

sclevine:
Closing as abandoned and superseded by other RFCs.

wilkinsona:
@sclevine Can you please link to the superseding RCFs?

sclevine:
Stack buildpacks are intended as a robust solution to this problem: https://github.com/buildpacks/rfcs/pull/111

However, an approach using regular buildpacks has removed some of the pressure to deliver #111 for this use case:
https://github.com/paketo-buildpacks/ca-certificates

PR #70, opened 2020-04-01
RFC for the export report: report.toml

matthewmcnew:
[Readable](https://github.com/matthewmcnew/rfcs/blob/export-report/text/0000-export-report.md)

### DIFF
--- a/text/0000-export-report.md
+++ b/text/0000-export-report.md
@@ -70,11 +70,13 @@ This will require an additional flag which might complicate the lifecycle interf
 - [The kbld tool parses the pack output with a regex to find the built image digest](https://github.com/k14s/kbld/blob/5597786d8369e966f2e7217b24fd058f4a910675/pkg/kbld/image/pack.go#L17). 
 - Tekton allows tasks to surface a built image digest with an [OCI image layout index.json](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#surfacing-the-image-digest-built-in-a-task).
 
-# Unresolved Questions
-[unresolved-questions]: #unresolved-questions
+# Out of Scope
+[out-of-scope]: #out-of-scope
+
+Some decisions regarding `report.toml` are explicitly out of scope for this RFC:  
 
 - How/if `pack` should expose this information to users in a programmatic way?
-- Is there any additional metadata that should be provided in report.toml?
+- Should report.toml include additional metadata regarding layer usage?
 
 # Spec. Changes
 [spec-changes]: #spec-changes
--- a/text/0000-export-report.md
+++ b/text/0000-export-report.md
@@ -1,0 +1,76 @@
+# Meta
+[meta]: #meta
+- Name: Export Report
+- Start Date: (fill in today's date: 2020-04-01)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+The lifecycle export stage should optionally write a machine parseable metadata "report" about the resulting exported image(s).
+
+# Motivation
+[motivation]: #motivation
+
+The logging output of the lifecycle export stage displays the built image's digest or image id. However, this information is not easily available for a platform that needs to utilize the resulting image from a build. 
+
+Without this information, a platform will need to resort to suboptimal mechanisms such as parsing the exported output or reading the most recently tagged image on the registry. 
+
+
+# What it is
+[what-it-is]: #what-it-is
+
+The lifecycle exporter/creator binary will accept an optional flag `report` which provides a path to `report.toml`. When the flag is supplied the export stage will write a toml representation of the built image for platform consumption.
+
+Here is an overview of the suggested schema for report.toml:
+
+```
+[image]
+tags = ["index.docker.io/image/name:latest", "index.docker.io/image/name:other-tag"]
+identifier = "sha256:c8be1b8f4d60d99c281fc2db75e0f56df42a83ad2f0b091621ce19357e19d853"
+```
+
+By representing this in toml, we can extend this information with additional metadata when needed. 
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The export stage will write report.toml after exporting and caching is complete. There is no default path for report.toml as its location will be platform-specific. If the export steps fails to complete no report.toml is expected to be written.
+
+This addition will require a bump in the platform api.
+
+Docker daemon based platforms such as `pack` can read the report.toml by utilizing [docker cp](https://docs.docker.com/engine/reference/commandline/cp/).
+
+Kubernetes based platforms can utilize the [pod termination message](https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#customizing-the-termination-message) to retrieve the reports on build completion.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This will require an additional flag which might complicate the lifecycle interface.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Not doing anything and continuing to require platforms to retrieve this information via alternative means.
+- Writing only the built image identifier to an unstructured file. This would be simpiler for some platforms to parse but, would limit the options for extensibility.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [The kbld tool parses the pack output with a regex to find the built image digest](https://github.com/k14s/kbld/blob/5597786d8369e966f2e7217b24fd058f4a910675/pkg/kbld/image/pack.go#L17). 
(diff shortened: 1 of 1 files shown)


text/0000-export-report.md
```diff
@@ -0,0 +1,76 @@
...
+tags = ["index.docker.io/image/name:latest", "index.docker.io/image/name:other-tag"]
+identifier = "sha256:c8be1b8f4d60d99c281fc2db75e0f56df42a83ad2f0b091621ce19357e19d853"
+```
+
+By representing this in toml, we can extend this information with additional metadata when needed. 
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The export stage will write report.toml after exporting and caching is complete. There is no default path for report.toml as its location will be platform-specific. If the export steps fails to complete no report.toml is expected to be written.
```
hone, line 40: I'm inclined to think this is generally useful for most platforms. I'm a big fan of setting a default path that could be overrideable instead.

text/0000-export-report.md
```diff
@@ -0,0 +1,76 @@
...
+
+- [The kbld tool parses the pack output with a regex to find the built image digest](https://github.com/k14s/kbld/blob/5597786d8369e966f2e7217b24fd058f4a910675/pkg/kbld/image/pack.go#L17). 
+- Tekton allows tasks to surface a built image digest with an [OCI image layout index.json](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#surfacing-the-image-digest-built-in-a-task).
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How/if `pack` should expose this information to users in a programmatic way? 
+- Is there any additional metadata that should be provided in report.toml?
+- Is report the best name for this file and should it be marshaled into toml?
```
hone, line 70: `export.toml`?

If this is an export report, could it include the layer usage/size/reuse(?) without needing to dig into the docker image itself.

hone:
Based on the [Spec Changes RFC](https://github.com/buildpacks/rfcs/blob/master/text/0029-template-changes.md), do we need to add a `[spec-changes]` section?

text/0000-export-report.md
ekcasey: I would suggest `./report.toml` (or ./<new-name>.toml) (similar to `./group.toml` and `./analyzed.toml` default paths) which generally resolves to `/layers/report.toml` given that the working dir is usually the layers dir.

We should call out that the default can be changes with an env var or overrode with an exporter flag.

text/0000-export-report.md
ekcasey: Given we have an `analyzed.toml`, maybe `exported.toml` is the most consistent

matthewmcnew: I like `exported.toml`. Updated the rfc.

text/0000-export-report.md
matthewmcnew: Updated to default to the working directory.

text/0000-export-report.md
```diff
@@ -0,0 +1,76 @@
...
+[what-it-is]: #what-it-is
+
+The lifecycle exporter/creator binary will accept an optional flag `exported` which will provides a path to `exported.toml`. When the flag is supplied the export stage will write a toml representation of the built image for platform consumption. When the flag is not provided it will default to writing exported.toml in the working directory which by convention is `/layers`.
+
+Here is an overview of the suggested schema for exported.toml:
+
+```
+[image]
+tags = ["index.docker.io/image/name:latest", "index.docker.io/image/name:other-tag"]
+identifier = "sha256:c8be1b8f4d60d99c281fc2db75e0f56df42a83ad2f0b091621ce19357e19d853"
```
jromero, line 32: `identifier` appears to be generic. It would be nice to know if it's an Image ID (in the daemon case) or Digest (in the case of a registry).

hone:
Adding this as a separate comment. This might be scope creep, but does it make sense to include other image/layer information like the layer usage/size/reuse(?) without needing to dig into the docker image itself.

matthewmcnew:
Working group discussion on April 16th:

- `report.toml` might be a more appropriate and extensible name than the export specific name: `exported.toml`. There is likely a need for another RFC to make the files produced by the lifecycle consistent and easy to aggregate into `report.toml`.

-  Layer usage/size/reuse would be a good candidate for `report.toml` as it is useful information that is not currently accessible elsewhere. 

- A general philosophy for `report.toml` is that it should contain information that is not accessible elsewhere. For example, the bill of materials should not be included as it is available as an image label and would be unnecessarily duplicated in `report.toml`. 

cc @hone @ekcasey

 Please let me know if I missed anything or misrepresented any part of our discussion.

matthewmcnew:
@hone @sclevine 

Do you have thoughts on what layer info might look like in report.toml?

Perhaps, something like this?

```toml
[[layers]]
buildpack = "some_buildpack/ruby"
name = "mri"
on_export = "reused"
cached = true
launch = true
build = true
size = 17813

[[layers]]
buildpack = "some_buildpack/ruby"
name = "cache"
on_export = "cached"
cached = true
launch = false
build = false
size = 64811

File: text/0000-export-report.md [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [The kbld tool parses the pack output with a regex to find the built image digest](https://github.com/k14s/kbld/blob/5597786d8369e966f2e7217b24fd058f4a910675/pkg/kbld/image/pack.go#L17). 
+- Tekton allows tasks to surface a built image digest with an [OCI image layout index.json](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#surfacing-the-image-digest-built-in-a-task).
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How/if `pack` should expose this information to users in a programmatic way?
```
hone, line 76: Do we want to resolve this before accepting the RFC or is that just a separate RFC?

text/0000-export-report.md
```diff
@@ -0,0 +1,82 @@
...
+- Is there any additional metadata that should be provided in report.toml?
```
hone, line 77: I'd still like to see the layer size info and like the toml structure you laid out.

text/0000-export-report.md
matthewmcnew: As this is primarily requested to enable workflows in kubernetes like platforms that invoke the lifecycle directly, I would like to push that to a separate rfc if possible.

hone: @jromero thoughts on inclusion in `pack`?

jromero: The only case in which we'd be likely to use this would be for something like the [`quiet` option proposed](https://github.com/buildpacks/pack/issues/285) where we'd silence the output of the lifecycle and only use the information provided by this file to output only the desired information. _(This could have also been achieved via structured logging. 😄)_

I would agree with @matthewmcnew that pack is not the primary target for this feature and hence fleshing that out isn't necessary.

hone: @matthewmcnew can you put in here that this question is out of scope.

text/0000-export-report.md
hone: If you don't feel strongly about adding the layer sizing stuff in here, feel free to mark this as out of scope too.

ekcasey:
@matthewmcnew Some of the proposed metadata additions here may over with #63, which solves part of the problem by adding all layer metadata to the exported image even if `launch=false`. It seems like we should move forward with the original proposal and follow up with another RFC to address other desirable build data.

text/0000-export-report.md
matthewmcnew: Updated to indicate that this is out of scope for this RFC

text/0000-export-report.md
matthewmcnew: Updated to indicate that this is out of scope for this RFC

nebhale:
Entering Final Comment Period for merge disposition, closing 23 May, 2020.

text/0000-export-report.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+On a build exported to the docker daemon:
+```
+[image]
+tags = ["index.docker.io/image/name:latest", "index.docker.io/image/name:other-tag"]
+image_id = "9c0e1895b90f"
+```
+
+By representing this in toml, we can extend this information with additional metadata when needed. A general philosophy for report.toml is that it should contain information that is not accessible elsewhere. For example, the bill of materials should not be included as it is available as an image label and would be unnecessarily duplicated in report.toml.
```
mattmoor, line 43: Is there any decent tooling for extracting `toml` keys?  Something like `jq` for `toml`?

Suppose what I wanted in a particular result file (e.g. in Tekton) was the image digest, what is the guidance for extracting particular keys?

matthewmcnew: You could use [yj]() with jq. `yj -t | jq -r .image.digest`

hone: This is what we've been doing in the heroku buildpacks.

mattmoor: Cool, TIL.

PR #71, opened 2020-04-02
CNB_BUILDPACK_DIR environment variable

Malax:
[Readable](https://github.com/malax/rfcs/blob/cnb-buildpack-directory-env-var/text/0000-cnb-buildpack-directory-env-var.md)

### DIFF
--- a/text/0000-cnb-buildpack-directory-env-var.md
+++ b/text/0000-cnb-buildpack-directory-env-var.md
@@ -1,0 +1,67 @@
+# Meta
+[meta]: #meta
+- Name: CNB\_BUILDPACK\_DIRECTORY environment variable
+- Start Date: 2020-04-02
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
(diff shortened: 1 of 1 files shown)


sclevine:
This is also common on the CF side. My only request before approving would be that the env var be `CNB_BUILDPACK_DIR` to match project conventions.

We also may want to consider why an env var might be appropriate here vs. a positional arg like the other input directories.

Malax:
@sclevine I renamed the variable, thanks! :)

> We also may want to consider why an env var might be appropriate here vs. a positional arg like the other input directories.

I touched on that a little bit in the RFC, assuming we want to keep the amount of positional arguments low. It's harder to remember the order of argument than an actual name. That would be true for all other arguments though. What was the deciding factor to make `CNB_STACK` an env var and not an argument?

hone:
We should make the RFC decision here independent of the positional argument/env var discussion IMO. Cleaning that up should be a separate RFC.

sclevine:
In FCP as of 4/15

PR #72, opened 2020-04-03
RFC: Process specific environment variables

dsyer:


nebhale:
@dsyer Can you please put some concrete examples (existing buildpacks or example commands) of situations where the current system isn't sufficient.

dsyer:
The example I was working on was a Spring Boot app. In the default "web" profile, it just runs the app. In the "foo" profile it exposes the Prometheus endpoint for collecting metrics. Another profile might run the database migration scripts. Each one requires different environment variables, and env vars from one profile might pollute or interfere with another. E.g. if I'm not running db migrations I don't need the db admin password.

hone:
From last week, @sclevine will create a new RFC based on this work.

sclevine:
Converting this to a draft for now. I plan to pick it up eventually.

hone:
@jabrown85 identified a great use case for this and is going to pick this up.

jabrown85:
I opened a new PR [here](https://github.com/buildpacks/rfcs/pull/83) to continue this discussion further.

hone:
With #83, I motion for this to be closed.

PR #74, opened 2020-04-13
Add author and RFC link to template per RFC-0033

jkutner:
Per [RFC-0033](https://github.com/buildpacks/rfcs/blob/master/text/0033-add-author.md)


jkutner:
I can't figure out why the DCO bot thinks this isn't signed off.

PR #75, opened 2020-04-28
RFC: Register a Buildpack with Pack

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/pack-publish-buildpack/text/0000-pack-publish-buildpack.md)


text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,125 @@
...
+    - The issue will be opened against the `https://github.com/buildpacks/registry` repo, which is *not* the index. The `buildpacks/registry` defines the [namespace ownership as described in RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md#namespace-ownership).
+    - The issue body will contain structured data that defines the buildpack id, version, digest, and url.
+1. A Github action will detect the new issue and do the following:
+    - If this is the first time the `namespace` is used, it will add the namespace to the `buildpacks/registry` repo with the Github user who opened the issue as the owner.
+    - If this is *not* the first time the `namespace` is used, it will confirm that the Github user who opened the issue is an owner of the buildpack.
+    - Create and merge a PR against the `https://github.com/buildpacks/registry-index` repo using a Gitub token (i.e. all commits in that repo will be made by the same "user"). The PR will add the buildpack version described in the issue.
```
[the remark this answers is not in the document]
jkutner: Yea, I wasn't sure if there was a reason to keep a PR history or something. But we could definitely go straight to master.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,125 @@
...
+
+* `--undo` - this option will execute the same flow, but set `yanked=false`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Github is a single-point-of-failure
+* The user must have a Github account
+* The user must publish via a browser or provide a Github API token to the CLI
+* "publish" is ambigious (we use it in `pack package-buildpack --publish` to mean pushing an OCI image to a Docker registry).
```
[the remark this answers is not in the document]
jkutner: i thought about this too, but in some sense `--push` for those commands isn't accurate, since it's not actually pushing a local image (since it's writing directly to the registry). I also worry about the breaking change.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,125 @@
...
+Steps:
+
+1. Pulls the image manifest for the buildpackage at `url` if it does not already exist locally.
+1. Reads the following information from the image manifest (if any of these are missing, the command fails):
+    - `id` (must include a `namespace` and `name` in the form `<namespace>/<name>`)
+    - `version`
+1. Opens a link to a new Github Issue in the user's browser
+    - The user must be logged into Github
+    - The link will use [Github query parameters](https://help.github.com/en/github/managing-your-work-on-github/about-automation-for-issues-and-pull-requests-with-query-parameters) to pre-populate the title, body, and labels.
+    - The issue will be opened against the `https://github.com/buildpacks/registry` repo, which is *not* the index. The `buildpacks/registry` defines the [namespace ownership as described in RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md#namespace-ownership).
```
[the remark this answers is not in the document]
jkutner: Yea, I'm starting to think the same. In part because if `publish-buildpack` doesn't work against the registry-index, there's this weird asymmetry between `build -R` and `publish-buildpack` (talking to two different places).

This would also mean that if the default registry or `-R` value is a Github URL, we could still perform these steps (allowing people to more easily fork the index).

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,176 @@
...
+    - `id` (must include a `namespace` and `name` in the form `<namespace>/<name>`)
+    - `version`
+1. Makes a Git commit to the local registry index with a JSON payload represeting the buildpack version as defined in [RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)
+1. Uses Git to push the commit to the registry index's `origin`.
+
+Option:
+
+* `--native` - indicates that the native Git mechanism should be used
+
+## `pack yank-buildpack <buildpack-id-and-version>`
```
natalieparellano, line 117: Will there be a corresponding `--native` version of this command?

jkutner: yea, it probably should. we're going to revise this a bit, so I'll be sure to include that

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+    - The issue body will contain structured data that defines the buildpack id and version to yank.
+1. A Github action will detect the new issue and do the following:
+    - Confirm that the Github user who opened the issue is an owner of the buildpack.
+    - Create and merge a PR against the `https://github.com/buildpacks/registry-index` repo using a Gitub token (i.e. all commits in that repo will be made by the same "user"). The PR will set `yanked=true` for the buildpack version described in the issue.
+    - Close the Github issue.
+
+The following behavior will execute when the configured registry is of type `git`.
+
+1. Creates a `local` branch on the local registry index cache
+1. Makes a Git commit to the local registry index with an update setting `yank=true` in the JSON payload represeting the buildpack version as defined in [RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)
```
nebhale, line 211:
```suggestion
1. Makes a Git commit to the local registry index with an update setting `yank=true` in the JSON payload representing the buildpack version as defined in [RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)
```

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+This also has the following drawbacks:
+
+* The user *must* fork the `buildpacks/registry-index`
+* The contents of a PR to the `buildpacks/registry-index` repo are not intended to be human readable, which would make it difficult for users to audit before submitting.
+* This would result in commits against the index from the actual user, which requires that they DCO sign-off the commit (it's unclear if automating the sign-off is legally acceptable).
+* This differs from [rust-lang/crates.io-index](https://github.com/rust-lang/crates.io-index), which only has commits from a single user.
+
+## Using `push-buildpack` instead of `publish-buildpack`
+
+The command to release a new version could be `pack push-buildpack` to avoid ambiguity with other commands that use a `--publish` flag.
```
nebhale, line 254: Don’t think this is the way to go here.  I’d go with changing the flag in the other command first, but I also don’t think it’s enough of a problem to break compatibility.  My choice is to use `publish-buildpack` and leave the other commands with `—publish` until we have some concrete data that there’s a large amount of confusion.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Infer registry type
+
+Instead of configuring the registry `type`, Pack could infer the type from the URL. For example, any URL with the host `github.com` would be interpreted as type `github`. Any local file path would be interpreted as type `git`.
+
+## Direct commits to the remote index
+
+Instead of handling Github Issues on the `buildpacks/registry` repo, the `publish-buildpack` command could open a PR against the `buildpacks/registry-index`. This has the following advantages:
+
+* The operation could be performed headlessly without storing or transporting the user's Github token.
```
nebhale, line 243: Rather than storing a user token, why not just create a `pack` OAuth application, navigate the authentication flow, and keep that refresh token in `pack`’s config file.  Then the token never really leaves the users physical machine, and if `pack` is compromised it’s easily remedied.

text/0000-pack-publish-buildpack.md
jromero: I agree with your first remark: see https://github.com/buildpacks/rfcs/pull/75#discussion_r417480933

Speaking to the second I don't see how it wouldn't immediately be confusing. Our guide/docs will likely have these two commands side by side and we'd have to explain why the `--publish` just puts it in the OCI registry but then you have to "publish" it to the Buildpacks registry.

> One the OCI registry via `pack package-buildpack --publish` and the other to the CNB registry via `pack publish-buildpack`.

jromero: We can make it non-breaking by providing support for the prior flag as deprecated for some period of time.

nebhale: It wouldn’t be immediately confusing because, as a first approximation, no one reads the documentation.  In practice, while sitting at a command line and typing away (likely copying and pasting from somewhere else), the context of the command lends clarity to the flag.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+## Adding a Buildpack to the Registry
+
+When a buildpack author would like to publish a version of a buildpack on the registry, they will use the `pack publish-buildpack <url>` command. This command will create a Github Issue on the `https://github.com/buildpack/registry` (or similarly named) Github repo requesting the addition the new buildpack version to the index.
+
+## Yanking a Buildpack from the Registry
+
+Sometimes a buildpack author may have pushed up a bad version that they wish to not be available in the index. In order to not break builds, it will not be possible to fully remove an entry from the index. Instead, the entry in the index will be marked as "yanked". This information can then be used when resolving which buildpacks to fetch.
```
hone, line 32: From "How it Works" it looks like this also creates an issue. Can we add that to here as well?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+##### Type: Git Schema
+```toml
+[registries.<name>]
+type="git"
+url="<path to git repo>"
+```
+
+#### Default
+
+The following could be an implied default when no registries are present, or pre-seeded by Pack when the file is initalized.
```
hone, line 74: The benefit of generating the template is people can easily edit it. The downside is it's harder to update it.

text/0000-pack-publish-buildpack.md
jkutner: added below this line

text/0000-pack-publish-buildpack.md
jkutner: I think that's what we want to do longer term. I can capture this in the alternatives though.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,270 @@
...
+    - `id` (must include a `namespace` and `name` in the form `<namespace>/<name>`)
+    - `version`
+1. Opens a link to a new Github Issue in the user's browser
+    - The user must be logged into Github
+    - The link will use [Github query parameters](https://help.github.com/en/github/managing-your-work-on-github/about-automation-for-issues-and-pull-requests-with-query-parameters) to pre-populate the title, body, and labels.
+    - The issue will be opened against the `https://github.com/buildpacks/registry-index` repo.
+    - The issue body will contain structured data that defines the buildpack id, version, digest, and url.
+1. A Github action will detect the new issue and do the following:
+    - If this is the first time the `namespace` is used, it will add the namespace to the `buildpacks/registry-owners` repo with the Github user who opened the issue as the owner. The `buildpacks/registry-owners` repo defines the [namespace ownership as described in RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md#namespace-ownership).
```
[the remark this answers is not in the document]
jkutner: @ekcasey no, I'd rather create the namespace with the current user as the owner by default, and then provide a mechanism to change it a team/org later on. It's one less step for 90% of people to think about.

text/0000-pack-publish-buildpack.md
jkutner: addressed in 79933a43e0199e7dbebdce1026efd02e7e568dc0

text/0000-pack-publish-buildpack.md
nebhale: @jkutner Having heard my argument, can you confirm your choice here and I'll resolve.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,274 @@
...
+## Configuration
+
+A default registry will be configured in `.pack/config.toml` with the following schema.
+
+##### Base Schema
+```toml
+[registries]
+default="<name>"
+
+[registries.<name>]
```
sclevine, line 50: Could we move to something more fixed/extensible:
```
default-registry = "github"
[[registries]]
name = "github"
```

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+    - `version`
+1. Creates a `local` branch on the local registry index cache
+1. Makes a Git commit to the local registry index with a JSON payload represeting the buildpack version as defined in [RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)
+1. Uses Git to push the commit to the registry index's `origin`.
+1. Updates the `master` branch of the local registry cache from `origin`.
+
+Options:
+
+* `--buildpack-registry, -R` - the id of a registry configured in `.pack/config.toml`
+* `--headless` - If the registry type would default to opening a web browser, it will instead perform the action using an API. In the case of Github, it will use OAuth. Optionally, it will store a refresh token locally, but it may require re-auth on every invokation.
```
jkutner, line 181: @jromero could a headless version of this command shell out to a pre-authed Github CLI? https://cli.github.com/

nebhale: The GH CLI isn't popular enough (heck it's not even the most popular way of integrating `git` with GitHub) to require it as an additional dependency.

text/0000-pack-publish-buildpack.md
jkutner: fixed 5579706

text/0000-pack-publish-buildpack.md
jkutner: addressed 5579706

text/0000-pack-publish-buildpack.md
jkutner: Is reasonable to tell users who want to publish from CI that they must create an Oauth token, and provide it as an env var (and deal with any expiration themselves)? 

Alternatively, we could recommend that publishing on CI requires creating a forked registry, and then provide a Github action that they can setup to publish from their fork to the official repo. In this way, they can use the `git` type of repo (and `git push` to publish). The rest will happen within Github.

jkutner: addressed in ea88f91

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+[motivation]: #motivation
+
+The [Client-Side Registry RFC (0022)](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md) established that two CLI commands would allow users to add/remove buildpacks in the registry, but the implementation of those commands was vauge. In this document we describe the details of how they will work.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Adding a Buildpack to the Registry
+
+When a buildpack author would like to publish a version of a buildpack on the registry, they will use the `pack publish-buildpack <url>` command. This command will create a Github Issue on the `https://github.com/buildpack/registry` (or similarly named) Github repo requesting the addition the new buildpack version to the index.
```
hone, line 28: For future proofing and clarity, we should use the repo name "registry-index". "registry" makes me think of the component or service that serves up the index.

```suggestion
When a buildpack author would like to publish a version of a buildpack on the registry, they will use the `pack publish-buildpack <url>` command. This command will create a Github Issue on the `https://github.com/buildpack/registry-index` (or similarly named) Github repo requesting the addition the new buildpack version to the index.
```

text/0000-pack-publish-buildpack.md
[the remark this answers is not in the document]
hone: I also think we add `create-namespace` later on if we need it.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+
+1. Fetches the image config blob for the buildpackage at `url`.
+1. Reads the following information from the image config (if any of these are missing, the command fails):
+    - `id` (must include a `namespace` and `name` in the form `<namespace>/<name>`)
+    - `version`
+1. Opens a link to a new Github Issue in the user's browser
+    - The user must be logged into Github
+    - The link will use [Github query parameters](https://help.github.com/en/github/managing-your-work-on-github/about-automation-for-issues-and-pull-requests-with-query-parameters) to pre-populate the title, body, and labels.
+    - The issue will be opened against the `https://github.com/buildpacks/registry-index` repo.
+    - The issue body will contain structured data that defines the buildpack id, version, digest, and url.
```
hone, line 160: How do we handle error cases if the data doesn't match?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+
+Arguments:
+
+* `url` - the location of a buildpackage _in a Docker registry_ to be released in the buildpack registry
+
+Steps:
+
+The following behavior will execute when the configured registry is of type `github`.
+
+1. Fetches the image config blob for the buildpackage at `url`.
```
hone, line 152: Do we need to also do this check on the GitHub action side?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+1. Reads the following information from the image config (if any of these are missing, the command fails):
+    - `id` (must include a `namespace` and `name` in the form `<namespace>/<name>`)
+    - `version`
+1. Opens a link to a new Github Issue in the user's browser
```
hone, line 156: The original RFC suggested opening a PR with the changes in them. Since we already need to code this into `pack` for generating the JSON payload, does it make sense to do this for the GitHub pathway as well?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+1. Makes a Git commit to the local registry index with a JSON payload represeting the buildpack version as defined in [RFC-0022](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md)
+1. Uses Git to push the commit to the registry index's `origin`.
+1. Updates the `master` branch of the local registry cache from `origin`.
+
+Options:
+
+* `--buildpack-registry, -R` - the id of a registry configured in `.pack/config.toml`
+* `--auth <token>` - If the registry type would default to opening a web browser, it will instead perform the action using an API. In the case of Github, it will use the provided token to invoke the [Issues API](https://developer.github.com/v3/issues/)
+
+**NOTE:** Github Issues that are not recognized as a request to add/yank a buildpack version will be automatically closed. All PRs will also be automatically closed.
```
hone, line 183: If we choose to use PRs instead, we can just close the Issues tab.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+        ```toml
+        id = samples/hello-universe
+        version = 0.0.1
+        ```
+
+## `pack publish-buildpack <url>`
+
+Arguments:
+
+* `url` - the location of a buildpackage _in a Docker registry_ to be released in the buildpack registry
```
hone, line 146: Do we assume most people will host things on DockerHub? If so, should we allow people to just use `buildpacks/sample:latest`?

text/0000-pack-publish-buildpack.md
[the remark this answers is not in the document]
hone: release?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+$ pack yank-buildpack buildpacks/sample@0.0.1
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* `type` of registry isn't accounted for in the [Buildpack URI](https://github.com/buildpacks/rfcs/blob/master/text/0037-buildpack-uris.md) definition
+* Github is a single-point-of-failure for the official registry
+* The user must have a Github account to publish to the official registry
+* The user may only publish to the official registry via a browser
```
hone, line 235: This is definitely a big downside for the CI/automated workflow case.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+Instead of handling Github Issues on the `buildpacks/registry` repo, the `publish-buildpack` command could open a PR against the `buildpacks/registry-index`. This has the following advantages:
+
+* The operation could be performed headlessly without storing or transporting the user's Github token.
+
+This also has the following drawbacks:
+
+* The user *must* fork the `buildpacks/registry-index`
+* The contents of a PR to the `buildpacks/registry-index` repo are not intended to be human readable, which would make it difficult for users to audit before submitting.
+* This would result in commits against the index from the actual user, which requires that they DCO sign-off the commit (it's unclear if automating the sign-off is legally acceptable).
+* This differs from [rust-lang/crates.io-index](https://github.com/rust-lang/crates.io-index), which only has commits from a single user.
```
hone, line 256: This is only b/c they use a service here.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+
+Instead of handling Github Issues on the `buildpacks/registry` repo, the `publish-buildpack` command could open a PR against the `buildpacks/registry-index`. This has the following advantages:
+
+* The operation could be performed headlessly without storing or transporting the user's Github token.
+
+This also has the following drawbacks:
+
+* The user *must* fork the `buildpacks/registry-index`
+* The contents of a PR to the `buildpacks/registry-index` repo are not intended to be human readable, which would make it difficult for users to audit before submitting.
+* This would result in commits against the index from the actual user, which requires that they DCO sign-off the commit (it's unclear if automating the sign-off is legally acceptable).
```
hone, line 255: https://gitlab.cncf.ci/envoyproxy/envoy/commit/a38facb323373b33361570f12c2507a47e191545 :D

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Using `push-buildpack` instead of `publish-buildpack`
+
+The command to release a new version could be `pack push-buildpack` to avoid ambiguity with other commands that use a `--publish` flag.
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [rust-lang/crates.io-index](https://github.com/rust-lang/crates.io-index)
+
```
hone, line 266:
```suggestion
* [CocoaPods Specs](https://github.com/CocoaPods/Specs)
* [brew taps](https://docs.brew.sh/Taps)
```

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+## Direct commits to the remote index
+
+Instead of handling Github Issues on the `buildpacks/registry` repo, the `publish-buildpack` command could open a PR against the `buildpacks/registry-index`. This has the following advantages:
+
+* The operation could be performed headlessly without storing or transporting the user's Github token.
+
+This also has the following drawbacks:
+
+* The user *must* fork the `buildpacks/registry-index`
+* The contents of a PR to the `buildpacks/registry-index` repo are not intended to be human readable, which would make it difficult for users to audit before submitting.
```
hone, line 254: What would the user need to audit? Shouldn't the autofilled text be what should be submitted? If it's wrong, shouldn't the user have to fix the content it is generated from?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,276 @@
...
+## Using `push-buildpack` instead of `publish-buildpack`
+
+The command to release a new version could be `pack push-buildpack` to avoid ambiguity with other commands that use a `--publish` flag.
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [rust-lang/crates.io-index](https://github.com/rust-lang/crates.io-index)
+
+# Unresolved Questions
```
hone, line 267: Since the `--headless` option got removed, the biggest unresolved question is how to deal with CI. This has to be a large use case like at Heroku.

jkutner:
@ekcasey @sclevine @hone @nebhale I've made several updates based on the discussion from yesterday. PTAL before today's WG meeting if you can.

One thing I failed to address is the "in a Docker registry" restriction on the `url-or-tag` argument to `release-buildpack`. I'm very interested in supporting buildpackages in `.cnb` format, but it introduces some challenges that I'm not ready to handle yet. Are we comfortable punting on this for now?

text/0000-pack-publish-buildpack.md
jkutner: addressed in 4d8f5b7

text/0000-pack-publish-buildpack.md
jkutner: addressed in 4d8f5b7

text/0000-pack-publish-buildpack.md
jkutner: addressed in 8053b46

text/0000-pack-publish-buildpack.md
jkutner: They would not _need_ to audit, but if they attempted to verify that their PR was correct, it could be difficult. In addition, they might find things like the invalid JSON (i.e. no `,` at the end of each line) to be confusing.

hone:
> One thing I failed to address is the "in a Docker registry" restriction on the `url-or-tag` argument to `release-buildpack`. I'm very interested in supporting buildpackages in `.cnb` format, but it introduces some challenges that I'm not ready to handle yet. Are we comfortable punting on this for now?

@jkutner yeah, I thought we decided it wouldn't impact that at least from the lower level primitive.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,287 @@
...
+
+[[registries]]
+name="<string>"
+type="<string>"
+# <additional per "type" config entries>
+```
+
+The `default-registry` value must match one of the tables in the `[[registries]]` array with the same `name`. The `name` value can be any string.
+
+The `type` key in the `[[registries]]` tabl can be set to one of two values, described below:
```
hone, line 57:
```suggestion
The `type` key in the `[[registries]]` table can be set to one of two values, described below:
```

jkutner:
@hone probably `add-buildpack` and `yank-buildpack` as labels?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,287 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The [Client-Side Registry RFC (0022)](https://github.com/buildpacks/rfcs/blob/master/text/0022-client-side-buildpack-registry.md) established that two CLI commands would allow users to add/remove buildpacks to the registry, but the implementation of those commands was vague. In this document, we describe the details of how they will work.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Adding a Buildpack to the Registry
```
hone, line 26:
```suggestion
## Registering a Buildpack with the Registry
```

hone:
> @hone probably `add-buildpack` and `yank-buildpack` as labels?

I like the direction that goes. Should we stick to the commands here. `register-buildpack`?

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,287 @@
...
+        ```toml
+        id = samples/hello-universe
+        version = 0.0.1
+        ```
+
+## `pack register-buildpack <url>`
+
+Arguments:
+
+* `url-or-tag` - the location of a buildpackage _in a Docker registry_ to be released in the buildpack registry
```
jromero, line 146: In other documents we refer to this as the the `image-name`. Is the something special, or additional, when we say "url OR tag"?

text/0000-pack-publish-buildpack.md
jromero: FWIW, IIRC, PR's weren't possible due to the lack of API support from GH for creating PRs from patches.

text/0000-pack-publish-buildpack.md
```diff
@@ -0,0 +1,287 @@
...
+* `pack add-buildpack-registry` - a new command that will add to the `[[registries]]` config in `.pack/config.toml`.
+* `pack set-buildpack-registry` - a new command that will mutate the `default-registry` config in `.pack/config.toml`.
+* `pack publish-buildpack` - a new command that will perform both `package-buildpack` and `register-buildpack` in one step.
+* *Headless release* - a new option on `register-buildpack` would allow for a non-interactive (i.e. no browser) registration
+* *REST API* - as an alternative to `github` and `git` types of registries, we intend to support an `api` type that will use a registry service to broker registration requests (instead of using Github Issues).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* `type` of registry isn't accounted for in the [Buildpack URI](https://github.com/buildpacks/rfcs/blob/master/text/0037-buildpack-uris.md) definition
```
jromero, line 243: I'm failing to see where this is relevant? In what context would `type` be necessary in a UR(L|I)?

jkutner: @jromero what i mean here is that I'm not sure what those `cnb://` URLs really look like. Is it `cnb://github.com/buildpacks/registry-index/<ns>/<name>`? How do I reference a CNB on other Git servers?

text/0000-pack-publish-buildpack.md
jkutner: addressed in cc49f2f

text/0000-pack-publish-buildpack.md
jromero: Ah got it. This makes sense. Sounds like a superseding RFC to come for CNB URIs.

nebhale:
Final Comment Period with merge disposition, closing on 24 June, 2020.

elbandito:
How does the `message` get passed to the `git` commit and `github` issue during publish?  Is there a flag e.g. `-m` to pass the message value?

jkutner:
@elbandito yea, i think a `-m` would be appropriate

PR #76, opened 2020-05-08
Suggestions for: Publish a Buildpack with Pack #75

jromero:
- Adds a possible solution for types that allow for more publishing mechanism in the future
- Adds proposed formats for git messages and issues

### DIFF
--- a/text/0000-pack-publish-buildpack.md
+++ b/text/0000-pack-publish-buildpack.md
@@ -42,6 +42,104 @@ There are four mechanisms for publishing a buildpack:
 * Direct, which commits directly to the local Git repository for the index and makes a Git push to the origin
 * Yank, which removes a buildpack from the registry
 
+## Configuration
+
+##### Base Schema
+```toml
+[registries]
+default="<name>"
+
+[registries.<name>]
+type="<registry type>"
+# <additional per "type" config entries>
+```
+
+##### Type: GitHub Schema
+```toml
+[registries.<name>]
+type="github"
+url="<url to git repo>"
+issues-url="<url to issues for publishing>" # optional, default: "<url>/issues"
+```
+
+##### Type: GitHub Schema
+```toml
+[registries.<name>]
+type="git"
+url="<path to git repo>"
+```
+
+
+#### Default
+
+The following could be an implied default when no registries are present
+
+```
+[registries]
+default="official"
+
+[registries.official]
+type="github"
+url="https://github.com/buildpack/registry"
+
+[registries.official.github]
+issues-url="https://github.com/buildpack/registry/issues"
+```
+
+## Format
+
+### Type: Git
+
+A commit would be created with the following message:
+
+```
+[<ADD|YANK>] <namespace>/<name>@<version>
+
+<user message (optional)>
+```
+
+##### Example
+
+```
+[YANK] buildpacks-io-samples/hello-universe@0.0.1
+
+Major bug issues found. See https://github.com/buildpacks/samples/issues.
+```
+
+### Type: GitHub issue
+
+An issue would be created with the following content:
+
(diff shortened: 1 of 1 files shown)


text/0000-pack-publish-buildpack.md
```diff
@@ -42,6 +42,102 @@ There are four mechanisms for publishing a buildpack:
 * Direct, which commits directly to the local Git repository for the index and makes a Git push to the origin
 * Yank, which removes a buildpack from the registry
 
+## Configuration
+
+##### Base Schema
```
jkutner, line 47: is this for `.pack/config.toml`?

jkutner:
this reminds me of how Maven works with `settings.xml`. I like it, and it means we can make the `pack publish-buildpack` interface simpler.

text/0000-pack-publish-buildpack.md
```diff
@@ -42,6 +42,102 @@ There are four mechanisms for publishing a buildpack:
...
+
+- Title: `[<ADD|YANK>] <namespace>/<name>@<version>`
+- Body:
+
+        <user message (optional)>
+
+        ### Diff
+
+        ```patch
+        <patch contents>
```
jkutner, line 121: i'm not sure if i want to put the patch contents in the diff. It will be harder to validate than just some raw json (which can then be converted into the proper commit without the risk of the diff containing anything "bad")

text/0000-pack-publish-buildpack.md
jromero: Yes.

jromero: One of the downsides to this proposed change is that it would work more like `git remotes` where you name your registries instead of simply `--buildpacks-registry <URL>` it would be `--buildpacks-registry <name>` and we could have commands like `pack registry add <name> ...`

jkutner: given how few people with change this, i think that's ok.

I think it'll be more common to use the `cnb://<registry-url>/<ns>/<name>` URI in the `--buildpack` or `project.toml` fields.

text/0000-pack-publish-buildpack.md
jromero: "bad" like an invalid operation or ill formatted?

The main reasoning for this approach was to minimize the logic layer for applying the change. In this scenario the logic of applying is done at the client (`pack`) versus any automation that would need to be replicated elsewhere you wanted to host a registry. 

Given that the initial goal was to strive for something where the logic mostly lived on the client. The set of responsibilities could have looked like this:

client responsibilities:
- validate operation
- create operation diff
- sign diff
- create issue

registry responsibilities:
- check integrity
- check permissions
- apply diff
- commit

Thinking about it further there is still a big risk of ill behaved clients (or manual alterations) that would yield in a bad operation. A more secure approach, and maybe something you were already thinking about, would be to have the client create an operation (as per your suggestion JSON) and add a mechanism to `pack` to read this format and apply it via the "git/native" mechanism.

client responsibilities:
- validate operation
- create operation payload
- create issue

registry responsibilities:
- check permissions
- use something like `pack publish-buildpack --native --input <file>` to apply
    - validate the operation (again)
    - make changes
    - commit
    - push

jkutner: I don't think we can trust the diff--it could be malicious or malformed--so we'd need to have logic to process it anyways. It also makes it harder to manually submit the PR (which we may need in the future if we run a service).

I'd propose a similar structure like:

- Title: `[ADD/YANK] <ns>/<name>@<version>`
- Body:

      <user message (optional)>

      ### Data
      
      id = <ns>/<name>
      version = <version>
      addr = <address-to-image>
      digest = <digest-of-image>

PR #77, opened 2020-05-13
Add root buildpack as interface for app image extension RFC

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/app-image-ext-buildpacks/text/0000-app-image-extensions.md)

### DIFF
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -108,7 +108,7 @@ When the base image has not changed
 When there is an update to either the build or run base images
 1. Pull the new base image(s)
 1. Run the root buildpacks against the new image(s) without loading previous layers, and create an ephemeral image.
-1. Rebase the app on to the new emphemeral image.
+1. Rebase the app on to the new ephemeral image.
 
 ## Example: Apt Buildpack
 
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -11,9 +11,8 @@
 [summary]: #summary
 
 This RFC proposes:
-1. An interface between a set of stack images and a platform in order to extend the images
-2. A UX for the pack CLI for extending builders with OS packages
-3. A type of buildpack that runs as the root user
+1. A UX for the pack CLI for extending builders with OS packages
+1. A type of buildpack that runs as the root user
 
 # Motivation
 [motivation]: #motivation
@@ -23,7 +22,7 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
 Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time.
 However, it is not uncommon for application code to depend on OS packages.
 
-Given that app authors and platform maintainers experience increased build time directly, extending the stack at the app author's or platform maintainer's request may add flexibility without sacrificing performance. At the same time, we want the CNB interface to be highly orthogonal, which is why we've strived to "make everything a buildpack". Allowing some buildpacks to run with privileges would give users the flexibility they expect based on their experience with other tools like `Dockerfile`.
+Given that app authors and platform maintainers experience increased build time directly, extending the stack at the app author's request may add flexibility without sacrificing performance. At the same time, we want the CNB interface to be highly orthogonal, which is why we've strived to "make everything a buildpack". Allowing some buildpacks to run with privileges would give users the flexibility they expect based on their experience with other tools like `Dockerfile`.
 
 The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reuseable, and safe.
 
@@ -32,11 +31,11 @@ As an example, consider the use case of Acme.com. They have three teams that wan
 # What it is
 [what-it-is]: #what-it-is
 
-Stack image creators and buildpack users can use buildpacks to extend their build and/or run images.
+- *root buildpack* - a special case of buildpack that is run with privileges
 
-## Specification
+Application developers can use root buildpacks to extend their build and/or run images.
 
-Introduce a boolean `privileged` key in the `[buildpack]` table of `buildpack.toml`, which is defined as follows:
+We introduce a boolean `privileged` key in the `[buildpack]` table of `buildpack.toml`, which is defined as follows:
 
 ```
 [buildpack]
@@ -45,24 +44,117 @@ privileged = <boolean (default=false)>
 
 When `privileged` is set to `true`, the lifecycle will run this buildpack as the `root` user.
 
-For each Root Buildpack, the lifecycle will mount an overlay filesystem before the buildpack's build phase executes (excluding `/tmp`, `/cnb`, and `/layers`). The buildpack can then create layers normally by writing to the `<layers>/` directory the same way an unprivileged buildpack would. However, a new `paths` key would be added to the `<layers>/<layer>.toml` schema:
+For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
 
+The buildpack can then exclude directories from the layer by writing to the `launch.toml` using a new `[[excludes]]` table, for example:
+
+```toml
+[[excludes]]
+paths = [ "/var" ]
+cache = true
 ```
-paths = ["<path glob>"]
-launch = <boolean>
-build = <boolean>
-cache = <boolean>
+
+* `paths` - a list of path globs to exclude from the run image
+* `cache` - if set to `true`, the paths will be stored in the cache
+
+The `[process]` and `[[slices]]` tables can be used as normal.
+
+The following constraints will be enforced:
+* If a root buildpack is included in a `[[build.buildpacks]]` list in an app's `project.toml`, the build will fail.
+* If a user attempts to create a buildpackage including both root buildpacks and non-root buildpacks, the process will fail.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Building an App with Additional Packages
+
+Given a root buildpack with ID `example/git`, the following configuration can be used:
+
+`project.toml`:
+```toml
+[[stack.build.buildpacks]]
+id = "example/git"
+version = "1.0"
+
+[[stack.run.buildpacks]]
+id = "example/git"
+version = "1.0"
 ```
 
-The `paths` array defines the directories and files from the overlay filesystem that should be included in the layer. If an unprivileged buildpack uses the `paths` key, it will be ignored.
+When the following command is run, the app source code will be loaded into the image for access by the root buildpack.
 
-The `launch` and `build` keys can be used as normal, but how they are interpreted will differ based on where they are run (i.e. a launch-only layer will be discarded when the buildpack is run on a build-image, and a build-only layer will be discarded when the buildpack is run on a run-image).
+`pack build sclevine/myapp`
 
-If a Root Buildpack (i.e. a buildpack with `privileged = true`) is included in a `[[build.buildpacks]]` list in an app's `buildpack.toml`, the build will fail.
+This creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal `pack build`.
 
-## UX
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
+* The build image layers will be persisted as part of the cache, and can be reused on subsequent builds.
+* Any paths that are excluded using `launch.toml` and set to be cached will be stored in the cache, and can be reused on subsequent builds.
 
-### Creating an Extended Builder
+## Upgrade an App
+
+`pack upgrade sclevine/myapp`
+
+This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then do the equivalent of `pack rebase sclevine/myapp`.
+
+Attempting to rebase an app directly after it's been upgraded is not permitted and will fail, because the current base is now ephemeral.
+
+When the base image has not changed
+1. If the buildpacks are configure as idempotent (the default) load the previous layers onto the base images.
+1. Run the buildpacks, creating an emphemeral image.
+1. Rebase the app on to the new emphemeral image.
+
+When there is an update to either the build or run base images
+1. Pull the new base image(s)
+1. Run the root buildpacks against the new image(s) without loading previous layers, and create an emphemeral image.
+1. Rebase the app on to the new emphemeral image.
+
+## Example: Apt Buildpack
+
+A buildpack that reads an `apt.toml` file to install an arbitrary list of packages would have a `buildpack.toml` like this:
+
+```toml
+[buildpack]
+id = "example/apt"
+privileged = true
+```
+
+An end-user's `apt.toml` might look like this:
+
+```toml
+[build]
+packages = [
+  "libpq-dev"
+]
+
+[run]
+packages = [
+  "libpq",
+  "ffmpeg"
+]
+```
+
+The `bin/build` would look like:
+
+```bash
+#!/usr/bin/env bash
+
+apt update
+
+for package in $(cat apt.toml | yj -t | jq -r ".$CNB_STACK_TYPE.packages | .[]"); do
+    apt install $package
+done
+
+cat << EOF > launch.toml
+[[excludes]]
+paths = [ "/var" ]
+cache = true
+EOF
+```
+
+# Alternatives
+
+## Creating an Extended Builder
 
 `pack create-builder sclevine/builder --config builder.toml --run-image sclevine/run --run-image-mirror scl.sh/run`
 
@@ -87,7 +179,14 @@ Behavior: creates a new builder with layers applied from the `example/git` build
 
 **Note:** If `--run-image` is not specified, the builder metadata could be used to dynamically extend the run image on `pack build`. This would make every initial `pack build` slower, so I consider this optimization out-of-scope for this RFC.
 
-### Extended an Existing Builder
+Open Questions:
+- What happens if you try to extend an extended builder?
+   - There are three options:
+      1. Fail
+      1. Treat the extended builder as a normal builder, and override the metadata. Advantages: Allows controlled, multi-level distribution model for changes. Easy to implement. Disadvantages: Requires pack upgradeing multiple levels of builders to fully patch the most extended builder.
+      1. Always save the existing metadata and the run image references each time. On pack upgrade, replay each extension. Advantages: Easy to upgrade builders with confidence. Disadvantages: Consumers of extended builders may upgrade them unintentionally (by upgrading their own nested extensions).
+
+## Extended an Existing Builder
 
 `pack extend-builder sclevine/builder --extension-config extend.toml --run-image sclevine/run --run-image-mirror scl.sh/run`
 
@@ -106,127 +205,21 @@ Behavior: creates a new version of an existing builder with layers applied from 
 
 **Note:** If `--run-image` is not specified, the builder metadata could be used to dynamically extend the run image on `pack build`. This would make every initial `pack build` slower, so I consider this optimization out-of-scope for this RFC.
 
-### Building an App with Additional Packages
-
-`pack build sclevine/myapp`
-
-project.toml:
-```toml
-[[stack.build.buildpacks]]
-id = "example/git"
-version = "1.0"
-
-[[stack.run.buildpacks]]
-id = "example/git"
-version = "1.0"
-```
-
-Behavior: creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal `pack build`. The run image is persistented as part of the app image, and can be reused on subsequent builds.
-
-Questions:
-- should we store the new builder image to make rebuild faster? If so, where? Should we generate a tag for it? Can we use the buildpack cache for this?
-- should the buildpack have access to the app source code?
-
-### Upgrade an Extended Builder
-
-Both build and run images: `pack upgrade sclevine/builder`
-
-Run images (including mirrors): `pack upgrade sclevine/run`
-
-### Upgrade an App
-
-`pack upgrade sclevine/myapp`
-
-This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then do the equivalent of `pack rebase sclevine/myapp`.
-
-Attempting to rebase an app directly after it's been upgraded is not permitted and will fail, because the current base is now ephemeral.
-
-# How it Works
-[how-it-works]: #how-it-works
-
-## Specification
-
-Platforms may extend app images using the above interface by running the root buildpacks and creating an image from the result. This should happen prior to the normal CNB build or rebase process.
-
-## UX
-
-`pack upgrade` (on base image)
-1. See if new base image available, if so pull the new base image(s)
-1. Run the extension buildpack(s) against each base image.
-
-If running `pack upgrade` on a builder, perform operation against both builder and run images and publish both.
-If running `pack upgrade` on an app, perform operation against run image (with ephemeral output into app repo), then `pack rebase`.
-
-## Example: Imagemagick Buildpack
-
-For a buildpack that installs the Imagemagick package, the `buildpack.toml` would look like:
-
-```toml
-[buildpack]
-id = "example/imagemagick"
-privileged = true
-```
-
-The `bin/build` would look like:
-
-```bash
-#!/usr/bin/env bash
-
-apt update
-apt install imagemagick
-
-cat << EOF > ${1}/usr.toml
-paths = [ "/usr" ]
-launch = true
-build = true
-cache = true
-EOF
-
-cat << EOF > ${1}/var.toml
-paths = [ "/var" ]
-launch = false
-build = false
-cache = true
-EOF
-```
-
-# Alternatives
-
-- Instead of adding `paths` to the `<layer>.toml`, the lifecycle could save the entire overlay as a layer. This mimics the behavior of the `RUN` directive of `Dockerfile`.
-- Make the overlay filesystem optional by adding a boolean `overlay` key in the `[buildpack]` table in `buildpack.toml`. When disabled, the Root Buildpack would still have root access, but it's changes to the filesystem would be discarded.
+Open Questions:
+- What happens if you try to extend an extended builder?
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Requires kaniko (or a similar tool) when a docker daemon is unavailable
+- Requires kaniko (or a similar tool) for snapshotting
 - Fully rebasing apps with package dependencies becomes many orders of magnitude slower, due to upgrade needs
 - Reliability of rebasing apps with package dependencies degrades and network and other external factors may break the process.
-- Some buildpack users will not want to give root access to any-and-all buildpacks. Instead they may want to selectively whitelist certain root buildpacks.
-- The overlayfs could get messy
-   - is it portable (windows?)
-   - will there be unexpected edge cases?
-   - requires privileges?
-   - works for docker...
-
+- Some buildpack users will not want to give root access to arbitrary buildpacks. Instead they may want to selectively whitelist certain root buildpacks.
 
 # Questions
 [questions]: #questions
 
-What happens if you try to extend an extended builder?
-
-Three options:
-
-1. Fail
-
-2. Treat the extended builder as a normal builder, and override the metadata.
-   Advantages: Allows controlled, multi-level distribution model for changes. Easy to implement.
-   Disadvantages: Requires `pack upgrade`ing multiple levels of builders to fully patch the most extended builder.
-
-3. Always save the existing metadata and the run image references each time. On `pack upgrade`, replay each extension.
-   Advantages: Easy to upgrade builders with confidence.
-   Disadvantages: Consumers of extended builders may upgrade them unintentionally (by upgrading their own nested extensions).
-
-My preference is (3), but (1) would be a good place to start.
+* Should we allow non-root buildpacks to run in combination with root buildpacks?
 
 # References
 
@@ -244,20 +237,56 @@ This proposal adds a new key to the `[buildpack]` table in `buildpack.toml`:
 ```
 [buildpack]
 privileged = <boolean (default=false)>
+? = <boolean (default=false)>
 ```
 
 * `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user
+* `?` - when set to `true`, indicates that the buildpack is not idempotent. The lifecycle will provide a clean filesystem from the base image before each run (i.e. no cache). Non-idempotent buildpacks cannot be combined.
 
-## Layer Content Metadata (TOML)
+## `launch.toml`
 
-This proposal adds a new key to the Layer Content Metadata schema:
+This proposal adds a new top level table in `launch.toml`:
 
-```toml
-launch = false
-build = false
-cache = false
-paths = [ "<path glob>" ]
-
-[metadata]
-# buildpack-specific data
 ```
+[excludes]
+paths = [ "<path glob>" ]
+cache = <boolean (default=false)>
+```
+
+Path globs MUST:
+
+* Follow the pattern syntax defined in the Go standard library.
+* Match zero or more files or directories.
+* Be absolute
+
+Path globs are not restricted to the app dir as with [slices](https://github.com/buildpacks/spec/blob/master/buildpack.md#launchtoml-toml).
+
+## `project.toml`
+
+```
+[[stack.build.buildpacks]]
+id = "<buildpack ID (optional)>"
+version = "<buildpack version (optional default=latest)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+
+[[stack.run.buildpacks]]
+id = "<buildpack ID (optional)>"
+version = "<buildpack version (optional default=latest)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+```
+
+### `[[stack.build.buildpacks]]`
+
+This table MAY include a list of root buildpacks to execute against the builder image.
+
+### `[[stack.run.buildpacks]]`
+
+This table MAY include a list of root buildpacks to execute against the run image.
+
+## Environment
+
+## Provided by the Lifecycle
+
+| Env Variable      | Description         | Detect | Build | Launch
+|-------------------|---------------------|--------|-------|--------
+| `CNB_STACK_TYPE`  | 'build' or 'run'    | [x]    | [x]   |
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -24,7 +24,7 @@ However, it is not uncommon for application code to depend on OS packages.
 
 Given that app authors and platform maintainers experience increased build time directly, extending the stack at the app author's request may add flexibility without sacrificing performance. At the same time, we want the CNB interface to be highly orthogonal, which is why we've strived to "make everything a buildpack". Allowing some buildpacks to run with privileges would give users the flexibility they expect based on their experience with other tools like `Dockerfile`.
 
-The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reuseable, and safe.
+The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reusable, and safe.
 
 As an example, consider the use case of Acme.com. They have three teams that want to contibute privileged layers to every image built at the company. Each of these teams wants to manage their own artifacts and the mechanisms for installing them, but none of these teams are in control of the base image. The teams have special logic for when their layers should be added to an image; in some cases it's the precense of a Node.js app, and in others it's the use of Tomcat, etc. In the past, these teams were trying to contribute their layers by selectively adding them to every `Dockerfile` for every base image in the company, but this doesn't scale well. With Root Buildpacks, each team can manage their own artifacts, release cadence, and add logic to `bin/detect` to ensure the layers are added when needed.
 
--- a/text/0000-app-image-extensions.md
+++ b/text/0000-app-image-extensions.md
@@ -174,7 +174,7 @@ EOF
 
 # References
 
-Encorporated suggested UX tweaks from Javier Romero's doc: https://hackmd.io/zuzsIAh5QGKcQt_EZAcXaw?view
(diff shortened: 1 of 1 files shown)


text/0000-app-image-extensions.md
```diff
@@ -21,44 +23,42 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
-/cnb/image/build/extend (build-metadata-toml-file) (cache-dir) | cwd: /
-/cnb/image/run/extend (run-metadata-toml-file) (cache-dir) | cwd: /
+[buildpack]
+privileged = <boolean (default=false)>
 ```
-Status:
+
+When `privileged` is set to `true`, the lifecycle will run this buildpack as the `root` user.
+
+For each Root Buildpack, the lifecycle will mount an overlay filesystem before the buildpack's build phase executes (excluding `/tmp`, `/cnb`, and `/layers`). The buildpack can then create layers normally by writing to the `<layers>/` directory the same way an unprivileged buildpack would. However, a new `paths` key would be added to the `<layers>/<layer>.toml` schema:
```
sclevine, line 48: This means that some layers would be owned by CNB_UID and some layers would be owned by root. This will create complexity and complicate the security posture (especially if, e.g., SUID is allowed). Can we avoid these special buildpacks having two responsibilities?

text/0000-app-image-extensions.md
```diff
@@ -140,27 +146,68 @@ Attempting to rebase an app directly after it's been upgraded is not permitted a
...
+The `bin/build` would look like:
+
+```bash
+#!/usr/bin/env bash
+
+apt update
+apt install imagemagick
+
+cat << EOF > ${1}/usr.toml
+paths = [ "/usr" ]
```
sclevine, line 179: The puts the onus on the buildpack developer to understand exactly what installing the imagemagick does to the file system. Also, it's not safe to install imagemagick on the build image and rebase it on to the run image (same problem as rebasing with updated versions of the same image).

jkutner: > The puts the onus on the buildpack developer to understand exactly what installing the imagemagick does to the file system.

I think this would be mitigated by the availability of community maintained `example/apt` and `example/yum` buildpacks. In some cases you might have an `example/imagemagick` buildpack if the package required some specific logic.

> Also, it's not safe to install imagemagick on the build image and rebase it on to the run image

I think running the buildpack on both the build and run images covers this right?

text/0000-app-image-extensions.md
jkutner: What are the "two responsibilities"?

jkutner:
@nebhale @hone @ekcasey @sclevine while this is a PR to Stephen's [OS extensions RFC PR](https://github.com/buildpacks/rfcs/pull/23), it's probably easier to digest it by reading the [Readable](https://github.com/buildpacks/rfcs/blob/app-image-ext-buildpacks/text/0000-app-image-extensions.md) version.

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
-**TODO: `extend` should have a mechanism for outputting mixins that get added to the extended image stack label metadata.**
+cat << EOF > launch.toml
+[[excludes]]
+paths = [ "/var" ]
+cache = true
+EOF
+```
 
-## UX
+# Alternatives
```
matthewmcnew, line 155: Are these alternatives proposed as alternative workflows the user can choose or alternative implementations the rfc reviewers should consider?

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
+* If a user attempts to create a buildpackage including both root buildpacks and non-root buildpacks, the process will fail.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Building an App with Additional Packages
+
+Given a root buildpack with ID `example/git`, the following configuration can be used:
+
+`project.toml`:
```
matthewmcnew, line 73: How will buildpacks indicate support for being a run image buildpack vs a build image buildpack?

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
+
+```
+[buildpack]
+privileged = <boolean (default=false)>
+```
+
+When `privileged` is set to `true`, the lifecycle will run this buildpack as the `root` user.
 
-Stack image creators may add the following executables to their run and/or build images:
+For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
matthewmcnew, line 47: Will pack and platforms be expected to add the lifecycle to the run image to execute the run image buildpacks?

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
+1. Rebase the app on to the new emphemeral image.
+
+When there is an update to either the build or run base images
+1. Pull the new base image(s)
+1. Run the root buildpacks against the new image(s) without loading previous layers, and create an emphemeral image.
+1. Rebase the app on to the new emphemeral image.
+
+## Example: Apt Buildpack
+
+A buildpack that reads an `apt.toml` file to install an arbitrary list of packages would have a `buildpack.toml` like this:
```
matthewmcnew, line 114: Is it necessary to support running "root buildpacks" with access to application source like this example suggests? 

Requiring access to source feels confusing for run image buildpacks which presumably would not have access to the application source. Additionally, it precludes the option of pre-building builders with the root buildpacks for distribution.

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
+
+The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reuseable, and safe.
+
+As an example, consider the use case of Acme.com. They have three teams that want to contibute privileged layers to every image built at the company. Each of these teams wants to manage their own artifacts and the mechanisms for installing them, but none of these teams are in control of the base image. The teams have special logic for when their layers should be added to an image; in some cases it's the precense of a Node.js app, and in others it's the use of Tomcat, etc. In the past, these teams were trying to contribute their layers by selectively adding them to every `Dockerfile` for every base image in the company, but this doesn't scale well. With Root Buildpacks, each team can manage their own artifacts, release cadence, and add logic to `bin/detect` to ensure the layers are added when needed.
 
 # What it is
 [what-it-is]: #what-it-is
 
-## Specification
+- *root buildpack* - a special case of buildpack that is run with privileges
```
matthewmcnew, line 34: Could a better name for these buildpacks be stack buildpacks?

matthewmcnew:
I like the idea of buildpacks being the extensibility mechanism for stack extensions!

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
+
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
+* The build image layers will be persisted as part of the cache, and can be reused on subsequent builds.
+* Any paths that are excluded using `launch.toml` and set to be cached will be stored in the cache, and can be reused on subsequent builds.
+
+## Upgrade an App
+
+`pack upgrade sclevine/myapp`
+
+This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then do the equivalent of `pack rebase sclevine/myapp`.
```
matthewmcnew, line 98: Doesn't executing arbitrary "root buildpacks" before a rebase no longer allow rebase to be a safe operation?

text/0000-app-image-extensions.md
matthewmcnew: Also, will snapshotting be necessary on the build image? Could the "root buildpacks" run before the regular buildpacks in the same container without needing to create an ephermeral build image?

text/0000-app-image-extensions.md
jkutner: alternative workflows. but after discussion with @sclevine I think we're going to remove them entirely

jabrown85:
Do root buildpacks participate in `detect`? Could I write a ruby buildpack to require `imagemagick` from the`apt-get` root buildpack if I see a particular gem in use that is known to need it?

sclevine:
Note from chat on 5/20: would be good to ensure that the app developer doesn't need to pick whether the root buildpack runs on the run image vs. build image vs. both.

text/0000-app-image-extensions.md
```diff
@@ -21,48 +22,139 @@ Allowing buildpacks to install OS packages dynamically during build would drasti
...
 
-Stack image creators may add the following executables to their run and/or build images:
+For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
 
-Extension:
+The buildpack can then exclude directories from the layer by writing to the `launch.toml` using a new `[[excludes]]` table, for example:
+
+```toml
+[[excludes]]
+paths = [ "/var" ]
```
josegonzalez, line 53: Does this mean that root buildpack authors need to know what paths the packages the buildpack installs write to for layer exclusion?

text/0000-app-image-extensions.md
jkutner: we've decided to consolidate this this. However, a buildpack can use `CNB_STACK_TYPE to determine where it's running (mentioned in spec changes)

text/0000-app-image-extensions.md
jkutner: we should discuss this in WG. I think it's required for Detect, but idk about Build

text/0000-app-image-extensions.md
jkutner: i'm fine with either. I think to most end-users they will just be "buildpacks", so it's probably more a question for implementors of those buildpacks

text/0000-app-image-extensions.md
jkutner: I believe it makes it possible that rebase is unsafe. paging @sclevine

text/0000-app-image-extensions.md
jkutner: If they want to slim down the run image, then yes they do. Otherwise they can just ignore this file. This is a good thing to point out because this proposal is tilting toward giving the buildpack authors more responsibility. But with that, I think we keep a highly orthogonal interface (i.e. everything is a buildpack).

jkutner:
@jabrown85 yes, root buildpacks can `bin/detect` as normal

text/0000-app-image-extensions.md
josegonzalez: Yeah just thinking allowed. I think that might work well for cases where a root buildpack is specific to a particular package - say, installing curl - but not so well if they support arbitrary packages based on files in the application repo?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+cache = true
+```
+
+* `paths` - a list of path globs to exclude from the run image
+* `cache` - if set to `true`, the paths will be stored in the cache
```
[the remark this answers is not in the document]
jkutner: Yes, but as @matthewmcnew pointed out, we could choose to store it as an intermediate build image (no need for snapshot in that case I think).

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+
+The `[process]` and `[[slices]]` tables can be used as normal.
```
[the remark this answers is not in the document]
jkutner: yes, the app dir should be excluded from the snapshot layer. I'll call that out.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+version = "1.0"
+```
+
+When the following command is run, the app source code will be loaded into the image for access by the root buildpack.
+
+`pack build sclevine/myapp`
+
+All root buildpacks will be sliced out of the list of buildpacks, and run before non-root buildpacks. This creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal `pack build`.
+
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
```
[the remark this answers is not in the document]
jkutner: Yes, the root buildpacks must run twice: once on the build image and once on the run image. I'm not sure about reusing the run-image layer. It might not be possible.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+* The build image layers will be persisted as part of the cache, and can be reused on subsequent builds.
+* Any paths that are excluded using `launch.toml` and set to be cached will be stored in the cache, and can be reused on subsequent builds.
```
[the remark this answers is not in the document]
jkutner: If the buildpack cannot rebuild with the snapshot layer applied, then it is not idempotent and the yet-to-be-named key in the `buildpack.toml` must specify this.

text/0000-app-image-extensions.md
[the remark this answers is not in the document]
jkutner: After more thought, I'm unsure if we can get away from the snapshot layer entirely. We may need it to restrict what changes are preserved (for example, we do not want to preserve changes to the `/layers` or `/workspace` dirs)

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+## Building an App with Additional Packages
+
+Given a root buildpack with ID `example/git`, the following configuration can be used with a Golang app:
+
+`project.toml`:
+```toml
+[[build.buildpacks]]
```
natalieparellano, line 75: Should this be `detect.buildpacks`?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+When the following command is run, the app source code will be loaded into the image for access by the root buildpack.
+
+`pack build sclevine/myapp`
+
+All root buildpacks will be sliced out of the list of buildpacks, and run before non-root buildpacks. This creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal `pack build`.
+
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
+* The build image layers will be persisted as part of the cache, and can be reused on subsequent builds.
```
natalieparellano, line 91: Where would information about these layers (such as their digest) be stored? Is that the responsibility of the platform?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+* Any paths that are excluded using `launch.toml` and set to be cached will be stored in the cache, and can be reused on subsequent builds.
+
+## Upgrade an App
+
+`pack upgrade sclevine/myapp`
+
+This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then do the equivalent of `pack rebase sclevine/myapp`.
+
+Attempting to rebase an app directly after it's been upgraded is not permitted and will fail, because the current base is now ephemeral.
```
natalieparellano, line 100: Is it possible you could elaborate on this?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+We introduce a boolean `privileged` key in the `[buildpack]` table of `buildpack.toml`, which is defined as follows:
+
+```
+[buildpack]
+privileged = <boolean (default=false)>
+```
+
+When `privileged` is set to `true`, the lifecycle will run this buildpack as the `root` user.
+
+For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively to snapshotting, a platform may store a new stack iamge to cache the changes. All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
sclevine, line 47: Eventually, we could support `silces` to split up kaniko's userspace snapshots (and improve performance).

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+paths = [ "/var" ]
+cache = true
+```
+
+* `paths` - a list of path globs to exclude from the run image
+* `cache` - if set to `true`, the paths will be stored in the cache
+
+The `[process]` and `[[slices]]` tables can be used as normal.
+
+The following constraints will be enforced:
```
sclevine, line 62: Do these constraints exist to ensure that buildpack authors can't easily introduce OS package installation into many end-user workflows (and thus hurt performance across the ecosystem)?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,242 @@
...
+* If a user attempts to create a buildpackage including both root buildpacks and non-root buildpacks, the process will fail.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Building an App with Additional Packages
+
+Given a root buildpack with ID `example/git`, the following configuration can be used with a Golang app:
+
+`project.toml`:
```
sclevine, line 73: Isn't this contradicted by the above statement:
> If a root buildpack is included in a `[[build.buildpacks]]` list in an app's `project.toml`, the build will fail.

sclevine:
I'm essentially ready to approve this, but I want to chat more about how easy we should make it for app devs, platforms, and buildpack authors to apply root buildpacks.

On one hand, I don't want to apply arbitrary restrictions to app developers (so I think root buildpacks should be allowed in project.toml). But I want to avoid every meta-buildpack in the world including an apt root buildpack (so maybe root buildpacks should be disallowed in buildpackages).

text/0000-app-image-extensions.md
jkutner: this line should be removed, its left over from an older proposal

jkutner: fixed in 92c7a77

sclevine:
After thinking about the discussion during today's working group meetings, I think I'm okay with us providing a contract via the build plan that allows buildpacks to specify mixin requirements that may be satisfied by either the stack or a root buildpack.

However,

1. It should still be possible for a buildpack to statically require mixins in buildpack.toml, such that it can't be placed on a stack that does not have them. (This will allow buildpack authors to enforce performant builds.)

2. I still feel strongly that we shouldn't allow root buildpacks to be combined with regular buildpacks in a buildpackage.

This would allow buildpacks to depend on packages (and account for the python+sqlite case), but also allow platforms and buildpack authors to prevent performance degradation by providing/enforcing larger stack images.

text/0000-app-image-extensions.md
jkutner: No, that's not a table in the latest definition of project.toml. So I'm proposing that the root buildpacks be listed in the same list as non-root buildpacks.

text/0000-app-image-extensions.md
jkutner: I think the lifecycle would manage these layers (the same way as if you are storing the cache in an image). But I do need to address how they are stored when the cache is a volume. I'll fix this up

text/0000-app-image-extensions.md
jkutner: I added "In effect, we are saying that the use of root buildpacks breaks rebase." Does that help?

text/0000-app-image-extensions.md
jkutner: I actually proposed that `[[slices]]` would work as normal a few lines after this.

sclevine:
@hone @nebhale @ekcasey @jkutner curious if you think the above would be a good compromise? This uses @ekcasey’s idea to meet the requirements I heard, and gives proactive platform maintainers and buildpack authors a degree of control over performance (although it still feels somewhat risky)

nebhale:
Sounds good to me.  I have no skin in this game (I'm not planning on using this feature), and it doesn't seem any more dangerous than any other proposal.

jkutner:
@sclevine I'm still a little unsettled with:

> I still feel strongly that we shouldn't allow root buildpacks to be combined with regular buildpacks in a buildpackage.

I don't think it does a good job of accounting for the python+sqlite case (it pushes the burden onto the end-user, which is the opposite of the buildpacks value-prop). There are many other cases like this (such as Rails ActiveStorage with ffmpeg). 

What about this: if we allow root buildpacks to satisfy mixin requirements AND opt-out of running when those packages are already provided by the stack (such as by finding them in the build plan), then it eliminates some of the performance concerns.

With regard to an ecosystem of buildpackages that rampantly include root buildpacks, I'm just not worried about it. Most end-users will use one of the following:
* a single buildpackage 
* a buildpack maintained by an organization that knows not to abuse root buildpacks
* Their own custom buildpack

sclevine:
> it pushes the burden onto the end-user

I definitely agree that end-user developers should not have to deal with this. My intent isn't to force developers to specify the root buildpacks they want. I want builder authors / platform maintainers to have control over this instead. E.g., the heroku builder could supply an apt-buildpack in order.toml for every group.

> What about this: if we allow root buildpacks to satisfy mixin requirements AND opt-out of running when those packages are already provided by the stack (such as by finding them in the build plan), then it eliminates some of the performance concerns.

How would a platform maintainer prevent developers from installing OS packages that they don't want them to use? They would end up needing to fork meta-buildpacks to remove/replace root buildpacks.

> Most end-users will use one of the following:
> 
> - a single buildpackage
> - a buildpack maintained by an organization that knows not to abuse root buildpacks
> - Their own custom buildpack

Aren't we trying to build a large community registry of composable buildpacks in the long run?

matthewmcnew:
How will root buildpacks contribute mixins in the build plan? Will all mixin validation occur in detect?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+[[build.buildpacks]]
+id = "example/golang"
+version = "1.0"
+```
+
+When the following command is run, the app source code will be loaded into the image for access by the root buildpack.
+
+`pack build sclevine/myapp`
+
+All root buildpacks will be sliced out of the list of buildpacks, and run before non-root buildpacks. This creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal build phase without running the root buildpacks.
```
matthewmcnew, line 87: Could you clarify what the "ephemeral run image" is here? How does it relate to the result of extend phase?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+After the build phase, *a new extend phase* will run the root buildpacks against the run-image to create an ephemeral base image that will be used by the export phase instead of the configured run-image.
+
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
+* The build image layers will be persisted as part of the cache, and can be reused on subsequent builds.
+* Any paths that are excluded using `launch.toml` and set to be cached will be stored in the cache, and can be reused on subsequent builds.
+
+## Upgrade an App
```
matthewmcnew, line 95: While I understand the allure of a pack upgrade flow, I feel pretty strongly that this should just be accomplished by a complete rebuild. An "upgrade" that re-executes some but not all buildpacks seems arbitrary and confusing.

text/0000-app-image-extensions.md
sclevine: Confused about this also. Maybe
>  and an ephemeral run image with additional layers
was supposed to be replaced by the "new extend phase" below?

text/0000-app-image-extensions.md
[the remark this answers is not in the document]
sclevine: I think we should cache the previous run image layers locally if they can be reused. They need to be built locally anyways.

text/0000-app-image-extensions.md
sclevine: The `pack upgrade` flow is just supposed to run the root buildpacks, and only on the run image. This does bring up a serious issue though: the app isn't accessible during a rebase. 

@jkutner I think root buildpacks will need something like a `/bin/rebase` executable and/or some way of storing metadata about the app. Otherwise rebase would require the original source code.

Alternatively, we could only give root buildpacks access to metadata in project.toml and not the app directory.

text/0000-app-image-extensions.md
sclevine: Sorry I missed this. This is probably clear by now, but the idea is that we would re-play the root buildpacks on the new run image before rebasing.

sclevine:
I'm starting to worry about @matthewmcnew's point about rebasing being potentially unsafe after root buildpacks are used. Unless we guarantee that root buildpacks only install packages that remain ABI compatible on updates, rebase could change app behavior. This could break apps in production and cause users to lose faith the rebase process is safe.

If we decide that rebase isn't safe in those cases, then any use of root buildpacks will drastically hurt performance for at-scale security patches. Full rebuilds would be necessary for all images.

We can keep rebase safe with root buildpacks, but only if operators can whitelist which root buildpacks are available. If arbitrary meta-buildpacks can contain root buildpacks, then this becomes much harder.

More generally, I feel like root buildpacks make it too easy for buildpack authors to degrade the end-user experience (for both developers and operators) and that we need some sort of control to prevent that from happening. That control shouldn't hurt interoperability (e.g., prevent heroku buildpacks from being used with kpack + many apps).

I'm not strongly opinionated about how we achieve this, but disallowing root buildpacks and regular buildpacks to mix in a buildpackage would be a start. Would you be open to that option if we allowed root buildpacks to be installed as enforced pre-build hooks on builders? Then developers would **never** need to think about them in the OS package case, even when manually specifying buildpacks.

text/0000-app-image-extensions.md
matthewmcnew: Providing a `/bin/rebase` or restricting access to the app source seem to be workarounds to enable the "half a build" execution of pack upgrade, it seems much simpler to just do a complete rebuild. .

jkutner: > Alternatively, we could only give root buildpacks access to metadata in project.toml and not the app directory.

I think this is the way to go

josegonzalez:
Is it the case that a buildpack needs a package, or that an app needs a package? As an example, I might write an app that calls out - via subprocess - to headless chrome (installed via some deb package) to do Things™. The buildpack didn't link to chrome for any installation purpose, but it's a runtime requirement.

Asking as I'm not sure if that would impact the end user interface here.

jkutner:
@josegonzalez i think it's both. Some buildpacks will depend on a package, and some apps will depend on a package. Ideally the former is completely hidden from the end-user (only the buildpack author is aware of it). But I think we're afraid that any solution will force the end-user (app developer) to be aware of it.

jkutner:
Based on a breakout discussion with @hone @sclevine and @ekcasey, I'm going to try and tweak this RFC such that the following are true:

1. Root buildpacks must be _provided_ by the stack
1. Stack maintainers will include any root buildpacks they support at the time the Builder image is create.
1. End users can only use buildpacks that are _provided_ by the stack in the `[[build.buildpacks]]` table in `project.toml`.
1. The run-image will contain any root buildpacks provided by the stack, and those buildpacks will be removed from the final app image (but retained in the run-image so that they can be re-run on upgrade)
1. Buildpackages can only contain root buildpacks or non-root buildpacks (not both)

What is missing, IMO, is a way for an end-user to extend the stack (i.e. to provide their own root buildpacks). A few options I'd like to consider for this are:

* a `pack add-buildpack-to-builder` (or similarly named command). This would extend the builder with either a root buildpack or a non-root buildpack. For root buildpacks, they would be added to the stack.
* A mechanism that would extend the stack on-the-fly if a root buildpack is used in `[[build.buildpacks]]` but is not provided by the stack.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+`pack upgrade sclevine/myapp`
+
+This will run `pack upgrade sclevine/run`, generate an ephemeral image, and then do the equivalent of `pack rebase sclevine/myapp`.
+
+Attempting to rebase an app directly after it's been upgraded is not permitted and will fail, because the current base is now ephemeral. In effect, we are saying that the use of root buildpacks breaks rebase.
+
+When the base image has not changed
+1. If the buildpacks are configure as idempotent (the default) load the previous layers onto the base images.
+1. Run the buildpacks, creating an emphemeral image.
```
dfreilich, line 105:
```suggestion
1. Run the buildpacks, creating an ephemeral image.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+1. Rebase the app on to the new emphemeral image.
```
dfreilich, line 106:
```suggestion
1. Rebase the app on to the new ephemeral image.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+When there is an update to either the build or run base images
+1. Pull the new base image(s)
+1. Run the root buildpacks against the new image(s) without loading previous layers, and create an emphemeral image.
```
dfreilich, line 110:
```suggestion
1. Run the root buildpacks against the new image(s) without loading previous layers, and create an ephemeral image.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+1. Rebase the app on to the new emphemeral image.
```
dfreilich, line 111:
```suggestion
1. Rebase the app on to the new ephemeral image.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+[motivation]: #motivation
+
+Allowing buildpacks to install OS packages dynamically during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements.
+
+Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time.
+However, it is not uncommon for application code to depend on OS packages.
+
+Given that app authors and platform maintainers experience increased build time directly, extending the stack at the app author's request may add flexibility without sacrificing performance. At the same time, we want the CNB interface to be highly orthogonal, which is why we've strived to "make everything a buildpack". Allowing some buildpacks to run with privileges would give users the flexibility they expect based on their experience with other tools like `Dockerfile`.
+
+The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reuseable, and safe.
```
dfreilich, line 27:
```suggestion
The advantages of using a buildpack for privileged operations are the same as for unprivileged operations: they are composable, fast (caching), modular, reusable, and safe.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+As an example, consider the use case of Acme.com. They have three teams that want to contibute privileged layers to every image built at the company. Each of these teams wants to manage their own artifacts and the mechanisms for installing them, but none of these teams are in control of the base image. The teams have special logic for when their layers should be added to an image; in some cases it's the precense of a Node.js app, and in others it's the use of Tomcat, etc. In the past, these teams were trying to contribute their layers by selectively adding them to every `Dockerfile` for every base image in the company, but this doesn't scale well. With Root Buildpacks, each team can manage their own artifacts, release cadence, and add logic to `bin/detect` to ensure the layers are added when needed.
```
dfreilich, line 29:
```suggestion
As an example, consider the use case of Acme.com. They have three teams that want to contribute privileged layers to every image built at the company. Each of these teams wants to manage their own artifacts and the mechanisms for installing them, but none of these teams are in control of the base image. The teams have special logic for when their layers should be added to an image; in some cases it's the presence of a Node.js app, and in others it's the use of Tomcat, etc. In the past, these teams were trying to contribute their layers by selectively adding them to every `Dockerfile` for every base image in the company, but this doesn't scale well. With Root Buildpacks, each team can manage their own artifacts, release cadence, and add logic to `bin/detect` to ensure the layers are added when needed.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+When the following command is run, the app source code will be loaded into the image for access by the root buildpack.
+
+`pack build sclevine/myapp`
+
+All root buildpacks will be sliced out of the list of buildpacks, and run before non-root buildpacks. This creates a version of the current builder with additional layers and an ephemeral run image with additional layers, then does a normal build phase without running the root buildpacks.
+
+After the build phase, *a new extend phase* will run the root buildpacks against the run-image to create an ephemeral base image that will be used by the export phase instead of the configured run-image.
+
+* The run image layers are persistented as part of the app image, and can be reused on subsequent builds.
```
dfreilich, line 91:
```suggestion
* The run image layers are persisted as part of the app image, and can be reused on subsequent builds.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+# Questions
+[questions]: #questions
+
+* Should we allow non-root buildpacks to run in combination with root buildpacks?
+* How do the packages installed by a root buildpack related (or not related) to mixins
+    * That is, can you use a root buildpack to install a package that satisfies the requirements of a buildpacks' required mixins?
+
+# References
+
+Encorporated suggested UX tweaks from Javier Romero's doc: https://hackmd.io/zuzsIAh5QGKcQt_EZAcXaw?view
```
dfreilich, line 177:
```suggestion
Incorporated suggested UX tweaks from Javier Romero's doc: https://hackmd.io/zuzsIAh5QGKcQt_EZAcXaw?view
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+We introduce a boolean `privileged` key in the `[buildpack]` table of `buildpack.toml`, which is defined as follows:
+
+```
+[buildpack]
+privileged = <boolean (default=false)>
+```
+
+When `privileged` is set to `true`, the lifecycle will run this buildpack as the `root` user.
+
+For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively to snapshotting, a platform may store a new stack iamge to cache the changes. All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
dfreilich, line 47:
```suggestion
For each root buildpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the buildpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). As an alternative to snapshotting, the platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the buildpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+cache = true
+```
+
+* `paths` - a list of path globs to exclude from the run image
+* `cache` - if set to `true`, the paths will be stored in the cache
+
+The `[process]` and `[[slices]]` tables can be used as normal.
+
+The following constraint(s) will be enforced:
+* If a user attempts to create a buildpackage including both root buildpacks and non-root buildpacks, the process will fail.
```
dfreilich, line 63: What's the rationale behind this constraint? What is the concern it's addressing?

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Building an App with Additional Packages
+
+Given a root buildpack with ID `example/git`, the following configuration can be used with a Golang app:
```
dfreilich, line 70:
```suggestion
Given a root buildpack with ID `example/git`, the following configuration can be used in combination with a `example/golang` buildpack, for a Go app:
```

text/0000-app-image-extensions.md
dfreilich: I found the wording a bit confusing also. Is this what you meant?
```suggestion
All root buildpacks will be sliced out of the list of buildpacks, and run before non-root buildpacks. This creates a version of the current builder with additional layers and an ephemeral run image with additional layers. `Pack` then does a normal build phase, without running the root buildpacks.
```

text/0000-app-image-extensions.md
dfreilich: +1 to @jromero and @matthewmcnew . I think adding an `upgrade` would be a confusion for new consumers, who know they need to do something for their app, but aren't sure what. If there's an explicit need at a later point, great, but adding it preemptively to me sounds like the confusion risks outweigh the benefits.

text/0000-app-image-extensions.md
```diff
@@ -0,0 +1,243 @@
...
+- Fully rebasing apps with package dependencies becomes many orders of magnitude slower, due to upgrade needs
+- Reliability of rebasing apps with package dependencies degrades and network and other external factors may break the process.
+- Some buildpack users will not want to give root access to arbitrary buildpacks. Instead they may want to selectively whitelist certain root buildpacks.
+
+# Questions
+[questions]: #questions
+
+* Should we allow non-root buildpacks to run in combination with root buildpacks?
+* How do the packages installed by a root buildpack related (or not related) to mixins
+    * That is, can you use a root buildpack to install a package that satisfies the requirements of a buildpacks' required mixins?
```
dfreilich, line 173: Why wouldn't this be a positive outcome?

text/0000-app-image-extensions.md
jkutner: This addresses a concern that @sclevine had about root buildpacks proliferating through the ecosystem. Like, if we give this tool to people, they will use it too frequently. If we instead restrict its use, it forces buildpacks authors to explore alternative solutions.

I don't necessarily agree with that concern, but I'm willing to accept it in compromise (in part because it's easy to reverse).

text/0000-app-image-extensions.md
jkutner: I think it would be positive, but it's an open question because we aren't defining how it would work.

text/0000-app-image-extensions.md
dfreilich: This was correcting the spelling of presence (currently precense).

jkutner:
Closing in favor of #111 and #112

PR #78, opened 2020-05-13
Add RFC for Stack metadata

kvedurmu:
[Readable](https://github.com/kvedurmu/rfcs/blob/stack-metadata/text/0000-stack-metadata.md)

### DIFF
--- a/text/0000-stack-metadata.md
+++ b/text/0000-stack-metadata.md
@@ -103,7 +103,7 @@ LABEL io.buildpacks.stack.release_date="2020-05-12T05:17:02.390472"
 LABEL io.buildpacks.stack.description="Paketo Buildpacks base stack build image"
 LABEL io.buildpacks.stack.metadata="Some optionally defined metadata"
 ```
-
+During rebase, the `rebaser` will copy the `io.buildpacks.stack.*` labels from the new run image to the app image.
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/text/0000-stack-metadata.md
+++ b/text/0000-stack-metadata.md
@@ -21,6 +21,44 @@ Exposing additional metadata in stack images will make it easier for end users t
 - What is the expected outcome?  
 We converge on metadata we want to expose in stack images via labels, and all stack authors eventually support these.
 
+Here's the current output of the `pack inspect-image` command on a sample app:
+```
+Inspecting image: go-mod-simple-app
+...................
+LOCAL:
+
+Stack: io.buildpacks.stacks.bionic
+
+Base Image:
+  Reference: 610caf01e85e5da52c6a99b42fed80858c153de6d25b120940266846bec0872a
+  Top Layer: sha256:b6aadae5b30af0345f0ab82add7ed5b951d78230e86693c08f5c1ae8470a0ffb
+
+Run Images:
+  gcr.io/paketo-buildpacks/run:base-cnb
+...................
+```
+
+By adding the proposed metadata, `pack inspect-image` output of this app could eventually look like:
+```
+Inspecting image: go-mod-simple-app
+...................
+LOCAL:
+
+Stack: io.buildpacks.stacks.bionic
+Stack Metadata: 
+  Maintainer: "Paketo Buildpacks"
+  Homepage: "https://github.com/paketo-buildpacks/stacks"
+  OS Distro: "unbuntu 18.04"
+
+Base Image:
+  Reference: 610caf01e85e5da52c6a99b42fed80858c153de6d25b120940266846bec0872a
+  Top Layer: sha256:b6aadae5b30af0345f0ab82add7ed5b951d78230e86693c08f5c1ae8470a0ffb
+
+Run Images:
+  gcr.io/paketo-buildpacks/run:base-cnb
+...................
+```
+
 # What it is
 [what-it-is]: #what-it-is
 
@@ -28,14 +66,15 @@ This provides a high level overview of the feature.
 
 - Define the target persona: stack author  
 - Explaining the feature largely in terms of examples.  
-The following labels will be added to stack images:  
+The following labels can **optionally** be added by stack maintainers to stack images:  
 `io.buildpacks.stack.maintainer`: Name of Stack Maintainer  
 `io.buildpacks.stack.homepage`: URL for the stack (i.e repo link)  
 `io.buildpacks.stack.distro.name`: Name of OS Distribution  
 `io.buildpacks.stack.distro.version`: Version of OS Distribution  
-`io.buildpacks.stack.version`: Release Number
-`io.buildpacks.stack.release_date`: Release date of image
+`io.buildpacks.stack.version`: Release Number  
+`io.buildpacks.stack.release_date`: Release date of image  
 `io.buildpacks.stack.description`: Description  
+`io.buildpacks.stack.metadata`: Generic stack metadata
 
 
 # How it Works  
@@ -62,6 +101,7 @@ LABEL io.buildpacks.stack.distro.version="18.04"
 LABEL io.buildpacks.stack.version="25"
 LABEL io.buildpacks.stack.release_date="2020-05-12T05:17:02.390472"
 LABEL io.buildpacks.stack.description="Paketo Buildpacks base stack build image"
+LABEL io.buildpacks.stack.metadata="Some optionally defined metadata"
 ```
 
 # Drawbacks
@@ -92,4 +132,7 @@ Stack authors will define their own custom labels to surface the metadata above.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
-None.
+
+The platform spec should be updated to indicate that these keys can **optionally** be added on the build/run stack image.
+
+PR - https://github.com/buildpacks/spec/pull/89
--- a/text/0000-stack-metadata.md
+++ b/text/0000-stack-metadata.md
@@ -1,0 +1,95 @@
+# Meta
+[meta]: #meta
+- Name: Additional Stack Metadata
+- Start Date: 2020-05-12
+- Author(s): kvedurmu
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+We currently expose the Stack ID and Mixins via Labels on the build/run images. It would be great if we could leverage Labels to allow stack authors to specify additional image metadata (e.g OS Distro, Maintainer, Stack Version, etc.). The purpose of this RFC is to enumerate on additional metadata we want to expose in our stack images and standardize on these labels.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?  
+Exposing additional metadata in stack images will make it easier for end users to debug stack related issues. Also, by exposing this metadata, CNB platforms such as kpack can leverage this information to improve their UX. 
+- What is the expected outcome?  
+We converge on metadata we want to expose in stack images via labels, and all stack authors eventually support these.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.  
+
+- Define the target persona: stack author  
+- Explaining the feature largely in terms of examples.  
+The following labels will be added to stack images:  
+`io.buildpacks.stack.maintainer`: Name of Stack Maintainer  
+`io.buildpacks.stack.url`: URL for the stack (i.e repo link)  
+`io.buildpacks.stack.distro.name`: Name of OS Distribution  
+`io.buildpacks.stack.distro.version`: Version of OS Distribution  
+`io.buildpacks.stack.release`: Release Number
+`io.buildpacks.stack.release_date`: Release date of image
+`io.buildpacks.stack.description`: Description  
+
+
+# How it Works  
+[how-it-works]: #how-it-works  
+
+This is the technical portion of the RFC, where you explain the design in sufficient detail.  
+All labels listed above will be added as `Labels` in the stack's Dockerfile.   
+
+Example Dockerfile:
+```
+ARG cnb_uid=1000
+ARG cnb_gid=1000
+ARG stack_id="io.buildpacks.stacks.bionic"
+ENV CNB_USER_ID=${cnb_uid}
+ENV CNB_GROUP_ID=${cnb_gid}
+ENV CNB_STACK_ID=${stack_id}
+USER ${cnb_uid}:${cnb_gid}
+LABEL io.buildpacks.stack.id=${stack_id}
+LABEL io.buildpacks.stack.mixins="[\"ca-certificates\"............]"
+LABEL io.buildpacks.stack.maintainer="Paketo Buildpacks"
+LABEL io.buildpacks.stack.url="https://github.com/paketo-buildpacks/stacks/"
+LABEL io.buildpacks.stack.distro.name="ubuntu"
+LABEL io.buildpacks.stack.distro.version="18.04"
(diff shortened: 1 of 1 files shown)


text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+# Summary
+[summary]: #summary
+
+We currently expose the Stack ID and Mixins via Labels on the build/run images. It would be great if we could leverage Labels to allow stack authors to specify additional image metadata (e.g OS Distro, Maintainer, Stack Version, etc.). The purpose of this RFC is to enumerate on additional metadata we want to expose in our stack images and standardize on these labels.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?  
+Exposing additional metadata in stack images will make it easier for end users to debug stack related issues. Also, by exposing this metadata, CNB platforms such as kpack can leverage this information to improve their UX.
```
hone, line 20: What would `kpack` or a platform do with this data?

hone:
I assume this metadata is optional?

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Upstream OS distros have an existing pattern of doing this. (e.g `registry.access.redhat.com/ubi8/ubi`)
+- Buildpack Homepage field - https://github.com/buildpacks/rfcs/blob/master/text/0030-links-for-buildpacks.md  
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Any additional labels that we'd want to expose?
+
+# Spec. Changes (OPTIONAL)
```
hone, line 93: Would we want to define this into a spec or extension spec?

text/0000-stack-metadata.md
matthewmcnew: `kpack` provides a `stack` resource that currently only reports the digest of the provided run/build image. It would be useful to report more detailed metadata status of that stack. That way users could better understand where the stack comes form. Something like:

```
Id:             io.buildpacks.stacks.alpine
Run Image:      alpine/run@sha256:d70bf0fe11d84277997c4a7da94b2867a90d6c0f55add4e19b7c565d5087206f
Build Image:    alpine/build@sha256:663a1781cb60a3feb43b9a2a2eae6681ca0124509b47719b5a6bf770c7638712
Mixins:         ca-certificates, libssl1.1
Url:            https://github.com/alpinelinux/cnb-alpine
ReleaseDate:    2020-05-12T05:17:02.390472
....
```

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+- Explaining the feature largely in terms of examples.  
+The following labels will be added to stack images:  
+`io.buildpacks.stack.maintainer`: Name of Stack Maintainer  
+`io.buildpacks.stack.url`: URL for the stack (i.e repo link)  
+`io.buildpacks.stack.distro.name`: Name of OS Distribution  
+`io.buildpacks.stack.distro.version`: Version of OS Distribution  
+`io.buildpacks.stack.release`: Release Number
+`io.buildpacks.stack.release_date`: Release date of image
+`io.buildpacks.stack.description`: Description  
+
```
sclevine, line 39: Could we add a generic metadata field that is defined by the stack author?

E.g., `io.buildpacks.stack.metadata` could be defined by another out-of-spec RFC when the stack ID is io.buildpacks.stacks.bionic.

text/0000-stack-metadata.md
kvedurmu: We could even expose some of this output in existing `pack` commands such as `rebase` if we wanted to.

text/0000-stack-metadata.md
kvedurmu: It probably makes sense to define this in the core platform spec as optional fields here - https://github.com/buildpacks/spec/blob/master/platform.md#build-image

text/0000-stack-metadata.md
kvedurmu: Yeah that makes sense. I think that we should keep the other labels I proposed since we already use this pattern for the `stack.id` and `stack.mixins` and add this generic metadata label. How does that sound?

kvedurmu:
> I assume this metadata is optional?
@hone yup, that's what I was thinking.

text/0000-stack-metadata.md
ekcasey: It would be nice to include this info in the `pack inspect-image` output as well

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+This provides a high level overview of the feature.  
+
+- Define the target persona: stack author  
+- Explaining the feature largely in terms of examples.  
+The following labels will be added to stack images:  
+`io.buildpacks.stack.maintainer`: Name of Stack Maintainer  
+`io.buildpacks.stack.url`: URL for the stack (i.e repo link)  
+`io.buildpacks.stack.distro.name`: Name of OS Distribution  
+`io.buildpacks.stack.distro.version`: Version of OS Distribution  
+`io.buildpacks.stack.release`: Release Number
```
ekcasey, line 36: Maybe `version` here instead of `release`? It seems more consistent/generic

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.  
+
+- Define the target persona: stack author  
+- Explaining the feature largely in terms of examples.  
+The following labels will be added to stack images:  
+`io.buildpacks.stack.maintainer`: Name of Stack Maintainer  
+`io.buildpacks.stack.url`: URL for the stack (i.e repo link)
```
ekcasey, line 33: `homepage` would make this consistent with the buildpack metadata (assuming you mean git repo in the example here)

text/0000-stack-metadata.md
kvedurmu: That seems reasonable. Updated

text/0000-stack-metadata.md
kvedurmu: This makes sense to me. Updated.

text/0000-stack-metadata.md
nebhale: Agreed on this.  I quite like the pattern of top-level keys being wholly reserved for the spec (i.e. `io.buildpacks.stack.*`) and having a key that is wholly reserved for users that the spec will never use (`io.buildpacks.stack.metadata`).  Top-level keys as they are described here look good.

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+ARG cnb_uid=1000
+ARG cnb_gid=1000
+ARG stack_id="io.buildpacks.stacks.bionic"
+ENV CNB_USER_ID=${cnb_uid}
+ENV CNB_GROUP_ID=${cnb_gid}
+ENV CNB_STACK_ID=${stack_id}
+USER ${cnb_uid}:${cnb_gid}
+LABEL io.buildpacks.stack.id=${stack_id}
+LABEL io.buildpacks.stack.mixins="[\"ca-certificates\"............]"
+LABEL io.buildpacks.stack.maintainer="Paketo Buildpacks"
```
nebhale, line 58: Thoughts on this being a Git-style string, e.g. `Paketo Buildpacks <stack@paketo.io>`?

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+We converge on metadata we want to expose in stack images via labels, and all stack authors eventually support these.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.  
+
+- Define the target persona: stack author  
+- Explaining the feature largely in terms of examples.  
+The following labels will be added to stack images:
```
nebhale, line 31: Somewhere in the document, it should be called out that these are either optional or required.  I’m of the opinion that they should _all_ be optional, but don’t feel strongly about that.

text/0000-stack-metadata.md
nebhale: @hone I think this makes sense to go straight to core spec without a trial period in `extension`.  This is an augmentation to an already existent spec (labels on stack images) and we’re pretty clear on how it would be implemented and used.  I contrast this with the binding spec which is created from whole cloth with no understanding that it will work properly, be popular, and that all platforms will support it.

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+- Buildpack Homepage field - https://github.com/buildpacks/rfcs/blob/master/text/0030-links-for-buildpacks.md  
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Any additional labels that we'd want to expose?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+None.
```
nebhale, line 95: Given that this is a PR for a spec change, some sort of documentation needs to be added here, even it duplicates the discussion above.

text/0000-stack-metadata.md
kvedurmu: I'm not sure I understand the purpose of a Git-style string for this. I feel like a combination of `io.buildpacks.stack.maintainer` plus `io.buildpacks.stack.homepage` tells the user who maintains the image and where to go for more information.

nebhale: I'm not sure that `.homepage` implies contact information.  It points you to a place to learn about the stack, but not necessarily how to make contact with the maintainers.  I don't feel strongly on it though.

text/0000-stack-metadata.md
hone: @nebhale I'm :+1: that. @kvedurmu do you mind adding updating this section with the changes you put in the platform spec?

text/0000-stack-metadata.md
hone: > E.g., `io.buildpacks.stack.metadata` could be defined by another out-of-spec RFC when the stack ID is io.buildpacks.stacks.bionic.

@sclevine can you elaborate on what you hope to have in there?

text/0000-stack-metadata.md
hone: > Thoughts on this being a Git-style string, e.g. `Paketo Buildpacks <stack@paketo.io>`?

@nebhale I wonder if we should standardize this if we want to do that. We don't do it for [project.toml](https://github.com/buildpacks/spec/blob/master/extensions/project-descriptor.md#project).

text/0000-stack-metadata.md
hone: Can we include the kpack or pack examples in the RFC? :)

text/0000-stack-metadata.md
hone: optional but strongly encouraged?

text/0000-stack-metadata.md
hone: Yeah, I'm :+1: of making it straight to the spec, I just wanted to make sure it was put in the RFC.

text/0000-stack-metadata.md
ekcasey: > I quite like the pattern of top-level keys being wholly reserved for the spec (i.e. io.buildpacks.stack.*) and having a key that is wholly reserved for users that the spec will never use (io.buildpacks.stack.metadata).

While I agree in theory, we currently use a lot of `io.buildpacks.*.metadata` for our own purposes, rather than reserving them for users. (e.g `io.buildpacks.{lifecycle,build, builder}.metadata`)

nebhale: @ekcasey Well that's a kick in the shorts.  We can pick another key (`.user`) that represents the same thing, but alas is a naming inconsistency.  My priority stack is as follows:

1.  Designated area for all users data that we both agree to never spec inside of and that a user will never create outside of
2. Naming consistency

Even if that means we have some exceptions (e.g. `io.buildpacks{lifecycle,build,builder}.metadata`), I don't think we should introduce _new_ inconsistencies.

text/0000-stack-metadata.md
nebhale: Consistency first!  If we don't want to do it everywhere, let's not do it anywhere.

text/0000-stack-metadata.md
nebhale: My policy is MAY means WON'T and SHOULD means PROBABLY WON'T.  But  SHOULD is probably the right strength :)

nebhale:
@kvedurmu We'd like to get this moving again, can you please address the comments with new content?

kvedurmu:
Apologies @nebhale lost track of this. I believe I've addressed all the comments and [PRed](https://github.com/buildpacks/spec/pull/89) in a spec change for this. Let me know if I'm missing anything.

text/0000-stack-metadata.md
nebhale: @hone Do the latest changes resolve this comment?

text/0000-stack-metadata.md
nebhale: @kvedurmu I would written those into this RFC, but hey-ho a spec PR to start with will work too 😆

text/0000-stack-metadata.md
nebhale: @kvedurmu I don't think this one has been addressed yet.

kvedurmu: @nebhale oops just updated it.

text/0000-stack-metadata.md
hone: yep!

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,138 @@
...
+LABEL io.buildpacks.stack.version="25"
+LABEL io.buildpacks.stack.release_date="2020-05-12T05:17:02.390472"
+LABEL io.buildpacks.stack.description="Paketo Buildpacks base stack build image"
+LABEL io.buildpacks.stack.metadata="Some optionally defined metadata"
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This could make image rebasing slightly more complicated as labels like stack version will need to be updated.
```
hone, line 110: How would rebase work with this RFC?

matthewmcnew: I believe the `io.buildpacks.stack.*` labels would need to be updated on the app image.

hone: That sounds good to me. Can get how this works into the RFC itself?

hone:
I'm overall :+1: this RFC, but would like to see the rebase case addressed that was listed.

text/0000-stack-metadata.md
```diff
@@ -0,0 +1,138 @@
...
+LABEL io.buildpacks.stack.maintainer="Paketo Buildpacks"
+LABEL io.buildpacks.stack.homepage="https://github.com/paketo-buildpacks/stacks/"
+LABEL io.buildpacks.stack.distro.name="ubuntu"
+LABEL io.buildpacks.stack.distro.version="18.04"
+LABEL io.buildpacks.stack.version="25"
+LABEL io.buildpacks.stack.release_date="2020-05-12T05:17:02.390472"
+LABEL io.buildpacks.stack.description="Paketo Buildpacks base stack build image"
+LABEL io.buildpacks.stack.metadata="Some optionally defined metadata"
+```
+
```
hone, line 106:
```suggestion
During rebase, the `rebaser` will copy the `io.buildpacks.stack.*` labels from the new run image to the app image.
```

nebhale:
Final Comment Period with merge disposition, closing on 22 July, 2020.

nebhale:
@kvedurmu Please do a rebase where you sign off on all the commits.

PR #79, opened 2020-05-14
Non-breaking 0.x minor API versions

ekcasey:


text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
+# Meta
+[meta]: #meta
+- Name: Lifecycle API version support changes
+- Start Date: 2020-04-14
```
hone, line 4:
```suggestion
- Start Date: 2020-05-14
```

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+- Author(s): [Emily Casey](https://github.com/ekcasey)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+
+# Summary
+[summary]: #summary
+
+In order enable smoother migrations and upgrades this RFC proposes the following changes to the API version compatibility rules:
```
twoseat, line 13:
```suggestion
To enable smoother migrations and upgrades this RFC proposes the following changes to the API version compatibility rules:
```

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. When the lifecycle increments an API versions in the `0.x` version line, this shall interpreted as non-breaking
+1. The spec may make certain types of "breaking" changes in a `0.x` API version, so long as those changes do not preclude simultaneous lifecycle support for the previous minor or require collaborating buildpacks to implement the same API version.
+1. When the lifecycle implments a new `0.x` API minor version, it will treat the usage of features removed between minor spec versions as a deprecated.
```
twoseat, line 17:
```suggestion
1. When the lifecycle implments a new `0.x` API minor version, it will treat the usage of features removed between minor spec versions as deprecated.
```

jkutner, line 17: are the features actually removed? or are they targeted for removal in the next major version?

jkutner: ah, they are maybe removed from the spec, but not the lifecycle impl itself?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+Because platform authors may want to shield end users from deprecation warning caused by the platform's use of deprecated
+lifecycle features, the lifecycle will provide the following options to configure deprecation warnings:
+
+All lifecycle binaries will have a deprecation mode, configurable in the following way:
+| Flag              | Environment Variable | Default
+|-------------------|----------------------|--------
+| -deprecation-mode | CNB_DEPRECATION_MODE | all
+
+The allowed values of deprecation mode are as follows:
+| Mode      | Behavior
```
jkutner, line 62: should there be a mode where it's errors (and exits non-zero) instead of warns when a deprecated feature is used? I wonder if this is important for automated tests that are future-proofing.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+    - We will likely want to support multiple APIs in the lifecycle eventually to ease the transition to 1.0
+
+# Prior Art
+[prior-art]: #prior-art
+
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Will we allow removal of features in a minor API version in the spec itself post `1.0`?
```
jkutner, line 115: I don't understand this, can you elaborate?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+
+# Summary
+[summary]: #summary
+
+In order enable smoother migrations and upgrades this RFC proposes the following changes to the API version compatibility rules:
+
+1. When the lifecycle increments an API versions in the `0.x` version line, this shall interpreted as non-breaking
+1. The spec may make certain types of "breaking" changes in a `0.x` API version, so long as those changes do not preclude simultaneous lifecycle support for the previous minor or require collaborating buildpacks to implement the same API version.
```
hone, line 16: 2 questions:

1. Also, why is "breaking" in quotesDoes this essentially create 2 ecosystems of buildpacks?
1. Why is "breaking" in quotes?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+This was the correct decision initially b/c we needed a way to indicate breaking changes before the APIs reached 1.0.
+However, this has become annoying over time as it adds considerable friction to the release of non-breaking features. All platforms and buildpacks must explicitly update to support the new APIs before pulling in a new lifecycle, even when a new minor would not break them.
+As the community of platform and buildpack authors grows and matures, this pain is magnified.
+
+Now that the APIs have stabilized we can realistically guarantee that we will not make breaking lifecycle changes before the implementation of the 1.0 APIs are released.
+
+## Deprecations
+We will allow the spec to remove features in a minor API and interperet this as a deprecated feature in the lifecycle.
+Previous we never formalized rules around deprecations. This made sense because all API changes have been breaking.
+However, deprecations can aid platform and buldpack as they prepare to perform a major API upgrade.
```
hone, line 33: This is probably out of scope of this RFC, but there's a question of how we communicate deprecations to platform/buildpacks if it's end users that are seeing those notices through build output.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Platforms
+Platforms that wish to enforce compatibility should now treat all minor API versions as non-breaking with the exception
+of buildpack and platform `0.1` -> `0.2`. These API versions did include breaking changes and will not be changed.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The spec and lifecycle now carry the burden of ensuring there are no breaking changes before `1.0`. `1.0` is an important symbolic
+milestone and there could be unforseen and compelling reasons to make a breaking change before we are ready to release `1.0` APIs.
```
hone, line 98: With all the [potential breaking changes for 1.0](https://docs.google.com/document/d/1ZETueW4JoumOfqa37VICKwcOgEKM2juYtNI9rRv0McM/edit) already outlined, can we achieve this by just doing minor changes and making 1.0 a non event?

text/0000-api-version-compat.md
hone: I believe it's asking, "If we are going to follow this same versioning scheme and semantics for 1.0+?"

text/0000-api-version-compat.md
hone: Maybe `report.toml` :P, but I probably shouldn't treat that as a dumping ground.

text/0000-api-version-compat.md
ekcasey: Can the spec remove a feature from an API between the 1.0 and 1.1 versions. This RFC says feature removal is allowed between 0.4 and 0.5 for example, but makes no comment about spec changes post 1.0.

text/0000-api-version-compat.md
ekcasey: Probably! This RFC doesn't state that `1.0` must be breaking, it keeps the status quo - it can be breaking if it wishes.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,119 @@
...
+[what-it-is]: #what-it-is
+
+## Spec
+The spec will use future `0.x` API release to indicate changes that can be implemented in a non-breaking fashion. These may include:
+- purely additive features
+- simple flag/file/key renames (the previous usage will be supported but deprecated by the lifecycle)
+- removal of features (these will be interpreted as deprecated by the lifecycle)
+
+## Lifecycle
+The lifecycle will continue use the [lifecycle descriptor](https://github.com/buildpacks/rfcs/blob/pack-publish-buildpack/text/0011-lifecycle-descriptor.md)
```
jabrown85, line 48:
```suggestion
The lifecycle will continue to use the [lifecycle descriptor](https://github.com/buildpacks/rfcs/blob/pack-publish-buildpack/text/0011-lifecycle-descriptor.md) 
```

text/0000-api-version-compat.md
ekcasey: Words are hard. To me, breaking implies that things break, and these changes wouldn't result in breaking changes from a user perspective. I wasn't sure what word to use to describe the spec equivalent of a breaking change. I can just remove the quotes

text/0000-api-version-compat.md
ekcasey: Do the deprecation modes address this concern? Platform and buildpack authors can turn on the deprecation output (or enable a failure mode as @jkutner suggested).

text/0000-api-version-compat.md
ekcasey: Even if a `1.0` API version contains breaking changes the `1.0` release of the lifecycle doesn't have to be breaking. I have some idea around facilitating a smooth lifecycle `1.0` upgrade experience but I want to keep them out of scope for this RFC which is focusing on the more immediate problem - future 0.x API versions.

text/0000-api-version-compat.md
ekcasey: added

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,122 @@
...
+and/or filter for deprecations relevant to their use case.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Spec
+The spec will use future `0.x` API release to indicate changes that can be implemented in a non-breaking fashion. These may include:
+- purely additive features
+- simple flag/file/key renames (the previous usage will be supported but deprecated by the lifecycle)
+- removal of features (these will be interpreted as deprecated by the lifecycle)
```
jkutner, line 45: Rather than remove features from the spec, I'd prefer that we mark them as deprecated. We could even move them to special section of the document (rather than keep them inline with non-deprecated features). I think this is important because it's the only way for buildpacks authors and platform operators to assuredly identify deprecated features (i.e. a warning might only appear when a particular code path is executed, and it's difficult to ensure that *every* possible code path is executed in testing). 

As prior art, consider the [Java API specification](https://docs.oracle.com/javase/8/docs/api/overview-summary.html). In particular, the deprecated methods on a class like [java.lang.Thread](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html). There is a tab in the interface that lists all the deprecated methods.

hone:
Ember.js has a great page for handling this here:
https://deprecations.emberjs.com/.

I worry if we remove things from the spec wholesale, it'll be harder for us
to ensure we don't just make breaking changes as it's just gone. Out of
sight, out of mind. There's mental overhead of remembering what is actually
deprecated vs the current state.

text/0000-api-version-compat.md
ekcasey: Adding deprecations to the spec was the path I originally went down but I was struggling to find much prior art (deprecations in specifications vs. implementations). The java API is a good example I didn't find. I am happy to make this change if we think the prior art is solid.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Deprecations
+If a feature is removed from the lifecycle's advertised API version but used by a platform or buildpack the lifecycle will print a deprecation warning.
+Because platform authors may want to shield end users from deprecation warning caused by the platform's use of deprecated
+lifecycle features, the lifecycle will provide the following options to configure deprecation warnings:
+
+All lifecycle binaries will have a deprecation mode, configurable in the following way:
+| Flag              | Environment Variable | Default
+|-------------------|----------------------|--------
+| -deprecation-mode | CNB_DEPRECATION_MODE | warn
```
matthewmcnew, line 59: Is the default `warn` here equivalent to `all`?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,122 @@
...
+| platform            | prints only platform API deprecation warnings
+| none                | silences all deprecation warnings
+| fail                | prints an error message and fail when any deprecated feature is used
+| fail-buildpack      | fail for buildpack deprecations
+| fail-platform       | fail for platform deprecations
+
+# How it Works
+[how-it-works]: #how-it-works
+## Spec
+The core team will ensure that no changes are added to a minor spec release that imply incompatibility with the previous minor.
```
matthewmcnew, line 75: The expectation being that deprecated features can be explicitly removed at `1.0`?

ekcasey:
I changed this RFC to move from a strategy that implies deprecated features exist _within_ an API to a strategy of multiple supported API versions. I think this will be more straightforward, avoids the problem of deprecations in the spec, and provides a better solution for lifecycle->buildpack inputs.

ashwin-venkatesh:
I really like this. It feels like this would make life much easier for kpack.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+1. The spec may make changes it wishes in a new 0.x API so long as those changes do not preclude buildpacks implementing different 0.x APIs from running together (e.g changes to the contractual build plan that cannot be bridged by the lifecycle)
+1. The platform can configure which version of the API the lifecycle implements
+1. When interacting with a buildpack, the lifecycle will use the API declared in each buildpack's `buildpack.toml`
+
+The lifecycle will assume the burden of maintaining compatibility, reducing the burden on platform and buildpack authors.  
+
+# Motivation
+[motivation]: #motivation
+
+Because there are far fewer lifecycle implementation than platforms or buildpacks, the lifecycle should assume the burden of maintaining compatibility. This will free platform authors to consume new lifecycles without worry and free buildpack authors to implement new APIs at their own pace.
```
jkutner, line 25: > buildpack authors to implement new APIs at their own pace

With the caveat that they must implement all or none of the changes for a given API version because there is no concept of deprecation, right?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+## Deprecations
+Once the 1.0 API is supported, we may wish to deprecate older APIs. A future RFC should provide a strategy for deprecations and outline the upgrade path from 0.x to 1.0 APIs. This is purposefully omitted from this RFC so that we can make progress on new 0.x APIs while we work out the details.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Spec
+The spec will use future `0.x` API releases to include any desired changes with one exception: the spec may not change in a way that precludes buildpacks implementing different 0.x APIs from running together.
+
+This RFC does not make changes to the API compatibility rules post 1.0. 1.x APIs version are stil assumed to be purely additive.
```
jkutner, line 44: What about 1.0 itself. Are we ok with precluding 0.x and 1.0 buildpacks from running together?

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+## Platforms
+Platforms that wish to check lifecycle compatibility may now treat all minor API versions as non-breaking with the exception
+of buildpack and platform `0.1` -> `0.2`. Platforms should always set the `CNB_PLATFORM_API` environment variable to indicate
+which version of the API they are using.
+
+## Buildpacks
+Buildpacks should continue to declare the minimum required buildpack API in their `buildpack.toml`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
jkutner, line 84: I think another drawback is that it's difficult for a buildpack author interpret the cumulative set of changes between 0.2 and 0.x API versions.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+The spec may not make certain types of changes to the contractual build plan before 1.0, even if those changes are desirable, because they make concurrent support for multiple APIs impossible.
+
+Buildpacks and platforms are likely to implement new APIs more slowly without a forcing function.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep doing what we are doing
+    - This will continue to cause pain for platform and buildpack authors
+- Do not allow breaking changes in the spec itself. Instead, add deprecations to the spec and print feature level deprecation warnings (the previous version of this RFC)
```
jkutner, line 97: Is there still a commit for this version? Is it possible to link to it?

jkutner:
While I have concerns about this approach, it's important for me to put that in context of my priorities, which are (in order of importance):

1. Release 1.0 is a "non-event" (i.e. the likelihood of someone upgrading to or starting at 1.0 and getting a error is as low as we can possibly make it).
1. Release 1.0 is a "clean room" implementation (i.e. we do not need to carry any deprecated features forward from 0.x)
1. Upgrade paths between *any* versions are clearly documented and discoverable. 

I believe the only way to achieve this things is by supporting modes for each 0.x version and 1.0 (with no statement about what we will do after 1.0). However, in order to make the upgrade path as easy as possible, I think we need a mechanism to document deprecations in the spec, which is the one thing I think is missing from this plan.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Lifecycle
+The lifecycle will use the [lifecycle descriptor](https://github.com/buildpacks/rfcs/blob/pack-publish-buildpack/text/0011-lifecycle-descriptor.md)
+to indicate the implemented API versions as before. However, the compatibility assumptions will change.
+It will be assumed that a lifecycle declaring support for the `0.x` version of the API, supports all 0.2-0.x API versions.
+
+### Platform API
+The lifecycle will use the `CNB_PLATFORM_API` environment variable to decide which API to implement. For now, if `CNB_PLATFORM_API` is not supplied, platform API 0.3 (the currently implemented API) will be assumed, to avoid breaking existing platforms that do not currently set this environment variable.
+
+If the API version declared in `CNB_PLATFORM_API` is higher than the newest supported by a given lifecycle, it will fail.
```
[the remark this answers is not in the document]
ekcasey: updated

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Buildpack API
+When orchestrating buildpacks, the lifecycle will provide inputs to and interpret outputs from each buildpack using the buildpack API indicated in that buildpack's `buildpack.toml` file.
+
+If the buildpack API declared in the buildpack.toml of a non-optional buildpack is higher than the newest supported by a given lifecycle, the lifecycle will fail.
+If the buildpack API declared in the buildpack.toml of an optional buildpack is higher than the newest supported by a given lifecycle, the lifecycle will warn, omit this buildpack from the group, and continue running.
```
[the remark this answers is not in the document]
ekcasey: updated

ekcasey:
@jkutner I added spec deprecations back, with some detail about when warnings would be printed. Let me know if this addresses your concerns.

text/0000-api-version-compat.md
ekcasey: Added this drawback and a possible mitigation (upgrade docs)

text/0000-api-version-compat.md
ekcasey: Added deprecations back.

text/0000-api-version-compat.md
ekcasey: I know this has changed a bit since you left this comment, but, for posterity: Deprecated features will be removed from the 1.0 API. Platforms & buildpacks will not be able to use these features when using the 1.0 API (the lifecycle itself may continue to support them in older API "modes", even after the 1.0 API is supported).

text/0000-api-version-compat.md
ekcasey: I am going to resolve this b/c of major holistic to this section of the RFC

text/0000-api-version-compat.md
ekcasey: done

hone:
I'm happy with this after deprecations are added in. I'm working on a separate RFC to cover the `1.0+` story.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,127 @@
...
+[prior-art]: #prior-art
+
+- k8s [api versioning](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning)
+> We chose to version at the API level rather than at the resource or field level to ensure that the API presents a clear, consistent view of system resources and behavior, and to enable controlling access to end-of-life and/or experimental APIs. ... Note that API versioning and Software versioning are only indirectly related
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How do we navigate the upgrade to 1.0 APIs
+- How do we eventually deprecate older APIs
```
hone, line 117: Out of scope of this RFC I think /cc @ekcasey

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,127 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- k8s [api versioning](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning)
+> We chose to version at the API level rather than at the resource or field level to ensure that the API presents a clear, consistent view of system resources and behavior, and to enable controlling access to end-of-life and/or experimental APIs. ... Note that API versioning and Software versioning are only indirectly related
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How do we navigate the upgrade to 1.0 APIs
```
hone, line 116: I believe deprecations + modes help here.

text/0000-api-version-compat.md
```diff
@@ -0,0 +1,127 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The lifecycle must now deal with the complexity of implementing all 0.2+ APIs.
+
+The spec may not make certain types of changes to the contractual build plan before 1.0, even if those changes are desirable, because they make concurrent support for multiple APIs impossible.
+
+Buildpacks and platforms are likely to implement new APIs more slowly without a forcing function.
+
+If a buildpack or platform wishes to upgrade across several API versions, it may be difficult to easily discern the cumulative set of required changes from the spec alone.
```
hone, line 97: deprecations help mitigate this issue b/c they'll be in the API as you upgrade and marked clearly.

nebhale:
Final Comment Period with merge disposition, closing on 10 June, 2020.

PR #80, opened 2020-05-21
RFC: pack Image Pull Policy

jromero:
* [Readable](https://github.com/buildpacks/rfcs/blob/pack-pull-policy/text/0000-pack-pull-policy.md)

### DIFF
--- a/text/0000-pack-pull-policy.md
+++ b/text/0000-pack-pull-policy.md
@@ -18,7 +18,7 @@ Replace the flag `--no-pull` with `--pull-policy <option>` and default to `if-no
 
 #### Performance
 
-By not having to check for the latest images on every build we improve performance.
+By not having to check for the latest images on every build we improve [performance](https://github.com/buildpacks/pack/issues/632#issuecomment-629458867).
 
 #### Uncommon
 
--- a/text/0000-pack-pull-policy.md
+++ b/text/0000-pack-pull-policy.md
@@ -11,18 +11,14 @@
 # Summary
 [summary]: #summary
 
-Replace the flag `--no-pull` with `--pull-policy <option>` and default to `if-not-present`.
+Replace the flag `--no-pull` with `--pull-policy <option>` and introduce `if-not-present`.
 
 # Motivation
 [motivation]: #motivation
 
-#### Performance
-
-By not having to check for the latest images on every build we improve [performance](https://github.com/buildpacks/pack/issues/632#issuecomment-629458867).
-
-#### Uncommon
-
-The docker CLI, docker compose, and k8s default to similar functionality as described by `if-not-present`.
+1. Enable the use case where the user may locally create an alternative to an image that would otherwise be pulled without having to manually pull all other images if using `--no-pull`.
+1. Provide support for an [expected](#prior-art) pull policy without [polluting the number of flags](#alternatives) available to commands.
+1. Allow for additional, potentially more complex pull policies such as [`periodic=1d`](https://github.com/buildpacks/rfcs/pull/80#issuecomment-644120544) or [`lifecycle=never,buildpack=always`](https://github.com/buildpacks/rfcs/pull/80#discussion_r434711922)
 
 # What it is
 [what-it-is]: #what-it-is
@@ -54,7 +50,9 @@ In the following commands, the images listed will adhere to the mentioned Pull P
 
 ### `pack package-buildpack`
 - Buildpacks - When refered to in `package.toml`
+# Backwards Compatibility
 
+For a period of time, `--no-pull` flag will map to `--pull-policy never` and print a deprecation warning.
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/text/0000-pack-pull-policy.md
+++ b/text/0000-pack-pull-policy.md
@@ -1,0 +1,72 @@
+# Meta
+[meta]: #meta
+- Name: `pack` Image Pull Policy
+- Start Date: 2020-05-21
+- Author(s): @jromero
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+Replace the flag `--no-pull` with `--pull-policy <option>` and default to "IfNotPresent"
+
+# Motivation
+[motivation]: #motivation
+
+#### Performance
+
+By not having to check for the latest images on every build we improve performance.
+
+#### Uncommon
+
+The docker CLI, docker compose, and k8s default to similar functionality as described by "IfNotPresent".
+
+# What it is
+[what-it-is]: #what-it-is
+
+A new flag `--pull-policy` that takes the following possible options:
+
+- `if-no-present` - only attempt to pull the images if it's not present
+- `always` - attempt to always pull images
+- `never`- don't attempt to pull images
+
+# How it Works
+[how-it-works]: #how-it-works
+
+<!-- This is the technical portion of the RFC, where you explain the design in sufficient detail.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work. -->
+
+### `pack build`
+
+// TODO: Detail all the images used during build and ensure that a single flag/policy makes sense for all images or detail exclusions.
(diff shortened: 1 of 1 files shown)


ekcasey:
I know this is still a draft, so maybe I should withhold my comments for now, but I like this idea a lot. Always checking for builder/run-image updates adds a surprising amount of time to each build.

text/0000-pack-pull-policy.md
```diff
@@ -0,0 +1,72 @@
...
+### `pack build`
+
+// TODO: Detail all the images used during build and ensure that a single flag/policy makes sense for all images or detail exclusions.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
```
jkutner, line 52: would it be possible to keep `--no-pull` as a sort of alias for `--pull-policy never`?

text/0000-pack-pull-policy.md
```diff
@@ -0,0 +1,72 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+<!-- This is the technical portion of the RFC, where you explain the design in sufficient detail.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work. -->
+
+### `pack build`
+
+// TODO: Detail all the images used during build and ensure that a single flag/policy makes sense for all images or detail exclusions.
```
natalieparellano, line 45: One to keep in mind is the lifecycle image. Given that this image is relatively behind the scenes, it might not make sense to have the same pull policy as builders, for example (even though this is what we do today).

jromero: I would agree with this but given that it would add complexity and confusion. I would question why they should differ?

What are some cases in which you would specify `--no-pull` (or later `--pull-policy never`) and expect any image to be pulled. Even in the scenario of a lifecycle image I would think that if I say `never`, I mean `never` and would have all the bits necessary locally (eg. offline).

text/0000-pack-pull-policy.md
jromero: Is this purely for backwards compatibility cases in which we should note it as such via a deprecation warning and eventually remove it? Or are you thinking of it as shorthand?

jromero:
This is now ready for review. 😄

text/0000-pack-pull-policy.md
ekcasey: We could always expand this later to accept a more fine-grained pull policy: e.g. `--pull-policy builder=always,lifecycle=not-present`. But I think it's okay to wait until we have validated use cases for this.

hone:
I'm :+1: on the flag, but I'm wary about the default of not pulling. I get people do it for performance, but the default was done for the reason of keeping people up to date so people don't have to think about it.

text/0000-pack-pull-policy.md
nebhale: A vote for deprecated compatibility flag.

sclevine:
I'm fine with the flag UX change, but share @hone and @jkutner's concerns about upgrades. Having a single command that produces a secure, up-to-date app image seems valuable.

Could we keep the default behavior, but allow you to set a pack-wide policy in config.toml?

drewpca:
If checking that we have the latest buildpack didn't take multiple seconds, we might not even be talking about adding pull policies nor about some users falling way behind. From the client side, it's just asking about one resource id and getting a HTTP 304 back most of the time, which should be quite acceptable. Have we ruled out fixing the slowness on the server side?

jromero:
@drewpca to be honest, there hasn't been any real look into diagnosing the root cause for the performance hits. Although from observation, it seems to be a lot more apparent to `gcr.io` registries, whether that be the cred-helpers, docker client (as reported [here](https://github.com/docker/for-mac/issues/3513)) or the actual server, there are just too many variables dependent on the registry being used that I personally feel that from the client-side we need to find a viable solution to alleviate that pain point regardless of registry/server implementation.

At the end of the day, the perception and UX from the end-users' perspective is that pack (or buildpacks) are slow when in reality it could very well be related to external circumstances such as in this case.

With all that said, I would also like to add that I agree that we should look into the root cause and maybe provide feedback or fix to the upstream. The purpose of this proposal is to help aid with the majority concern of the performance hit regardless of registry but doesn't negate the need to improve the performance further.

jromero:
Based on the last WG meeting conversations this RFC will be split into at least 2:

1. Focus on solely adding `--pull-policy <arguments>` along with a new policy `if-not-present`
2. Focus on relieving the performance hit of back-to-back build operations.
    Initial proposal:
    * New policy: `--pull-policy periodic=1d`
    * Setting new policy as the default
    * Making policy globally configurable via `config.toml`

text/0000-pack-pull-policy.md
```diff
@@ -0,0 +1,94 @@
...
+# Motivation
+[motivation]: #motivation
+
+#### Performance
+
+By not having to check for the latest images on every build we improve [performance](https://github.com/buildpacks/pack/issues/632#issuecomment-629458867).
+
+#### Uncommon
+
+The docker CLI, docker compose, and k8s default to similar functionality as described by `if-not-present`.
```
jromero, line 25:
```suggestion
1. Enable the use case where the user may locally create an alternative to an image that would otherwise be pulled without having to manually pull all other images if using `--no-pull`.
1. Provide support for an [expected](#prior-art) pull policy without [polluting the number of flags](#alternatives) available to commands.
1. Allow for additional, potentially more complex pull policies such as [`periodic=1d`](https://github.com/buildpacks/rfcs/pull/80#issuecomment-644120544) or [`lifecycle=never,buildpack=always`](https://github.com/buildpacks/rfcs/pull/80#discussion_r434711922)
```

text/0000-pack-pull-policy.md
```diff
@@ -0,0 +1,94 @@
...
+### `pack create-builder`
+- Builder - Used as the build environment
+- Buildpacks - When refered to in `builder.toml`
+
+### `pack rebase`
+- Run (for stack) - Used as the base for the built app image
+
+### `pack package-buildpack`
+- Buildpacks - When refered to in `package.toml`
+
```
jromero, line 57:
```suggestion
# Backwards Compatibility

For a period of time, `--no-pull` flag will map to `--pull-policy never` and print a deprecation warning.
```

text/0000-pack-pull-policy.md
jromero: Added under "Backwards Compatibility"

jromero:
@sclevine / @jkutner / @hone This PR/RFC has been update to only introduce the new flag and new option of `if-not-present` while leaving the default alone. Proposing the change of the default policy will have a follow-up RFC.

@jabrown85 as we've discussed, we can probably have a parallel RFC for updating the default and another for more granular control after this RFC is accepted.

text/0000-pack-pull-policy.md
natalieparellano: I think `if-not-present` satisfies my concern - e.g., the pack user would like NOT to pull the builder image, but has no idea that the lifecycle image even exists and is subsequently confused when the build fails due to the lifecycle image being missing. With `if-not-present` we'll use the local builder and the lifecycle image will be quietly pulled and the user never has to care.

nebhale:
Final Comment Period with merge disposition, closing on 25 July, 2020.

edmorley:
>     2\. New policy: `--pull-policy periodic=1d`

Hi! Has there been any further discussion of adding `periodic` option to `--pull-policy`?

For both local development and CI workflows, the default of `always` adds multiple seconds to the time to `pack build`. 

In addition, now that Docker Hub has rate limiting, and with a no-op pull still counting against the rate limit (https://docs.docker.com/docker-hub/download-rate-limit/#definition-of-limits), preventing unnecessary pulls is even more important (think running 100 integration tests).

Whilst one can work around this using a manual pull combined with `if-not-present` in these scenarios, it would be great to just be able to use a `periodic=1d` policy everywhere and not have to (a) think about this, (b) have split CI/local workflows etc :-)

edmorley:
> Hi! Has there been any further discussion of adding `periodic` option to `--pull-policy`?

Filed:
https://github.com/buildpacks/pack/issues/1368

PR #81, opened 2020-05-27
Offline Buildpackages RFC

dwillist:
[Readable](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md)

### DIFF
--- a/text/0000-offline-buildpackages.md
+++ b/text/0000-offline-buildpackages.md
@@ -1,4 +1,4 @@
-# Meta 4
+# Meta
 [meta]: #meta
 - Name: Offline Buildpackages
 - Start Date: 2020-04-27
@@ -13,7 +13,7 @@
 
 This change allows buildpacks to access vendored assets during builds to enable offline builds and asset reuse.
 
-To do so we introduce the Asset Image, a distribution artifact that provides vendored assets for buildpacks, and platform mechanisms to provide vendored assets to buildpacks during the `build` and `detect`.
+To do so we introduce the Asset Package, a distribution artifact that provides vendored assets for buildpacks, and platform mechanisms to provide vendored assets to buildpacks during the `build` and `detect`.
 
 # Motivation
 [motivation]: #motivation
@@ -29,94 +29,97 @@ Buildpack user, Platform operator, Buildpack author
 
 ### Explaining the feature largely in terms of examples.
 
-Lets walk through an example `pack build` using Asset Images, and the platform changes proposed below.
+Lets walk through an example `pack build` using Asset Package, and the platform changes proposed below.
 
-We pull an Asset Image, using`docker pull asset-image` (or some equivalent command). The asset image has a `json` label mapping the `sha256` values of all the assets it includes to. a `layer` in the Asset Image.
+This is as simple as specifying any Asset Packages to be used in a build.
 
-We run `pack build --buildpack new-buildpack`
+`pack build --buildpack some-buildpack -asset asset-package`
 
-A `new-buildpack` has label metadata containing the `sha256` of assets it could download during a build. However, `new-buildpack` does not contain any of these assets.
+The `asset-package` will provide a set of asset files all rooted at `/cnb/assets/` to which buildpacks may use.
 
-Before buildpack execution the platform reads the `sha256` values from the label metadata on `new-buildpack`. It then searches through local images to check for:
-  1) Exposed `asset` sha256 values.
-  2) A match between the expose, and the `sha256` value from `new-buildpack`.
-
-If both conditions are met, the platform applies the `layerID` mapped to this `sha256` to the build environment, exposing a vendored asset at `/cnb/assets/<asset-sha256>` to all buildpacks.
-
-
+As with the `--buildpack` flag, multiple `--asset` flags may be passed. These flags may reference an image, local archive filepath, or URL.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-### Asset Image Definition.
+### Asset Package Definition.
 
-An Asset Image is a collection of layers each of which provide one or more assets. These layers are in lexicographic order by `diffID`.
+An Asset Package is a collection of layers each of which provide one or more assets. These layers are in lexicographic order by `diffID`. The distribution of assets into layers is an implementation detail left to the platform.
 
 ```
-<layer1> ━━ /cnb/deps/<java11-asset-sha256>/java11
-<layer2> ━━ /cnb/deps/<java13-asset-sha256>/java13
+<layer1> ┳━ /cnb/deps/<java11-asset-sha256>/java11
+         ┃                                  ┗ (archive contents)
+         ┗━ /cnb/deps/<java13-asset-sha256>/java13.tgz
 
 ...
-<layern> ━━ /cnb/deps/<java15-asset-sha256>/java15
+<layern> ━━ /cnb/deps/<java15-asset-sha256>/java15.tgz
 ```
 
-Each Asset Image will have a `io.buildpacks.buildpack.assets` Label. This label's contents will be a `json` object mapping each `asset` sha256 to a `JSON` object that contains a `layerDiffID`, and other optional keys.
-
+Asset Packages will have a `io.buildpacks.buildpack.assets` Label. This label's contents will be a `json` object mapping each layer to metadata describing the assets it provides.
 #### Example
 
 ``` json
 {
-
-  "<java11-asset-sha256>" : {
-     "layerDiffID": "<layer1-diffID>",
-     "additional_paths": ["other/endpoint1", "other/endpoint2"],
-     "metadata": {}
-  },
-  "<java13-asset-sha256>" : {
-    "layerDiffID": "<layer2-diffID>",
-
-    "additional_paths": [],
-    "metadata": {
-      "name": "java13"
+  "<layer1-diffID>": [
+    {
+      "sha256": "java11-asset-sha256",
+      "uri": "https://path/to/java11.tgz",
+      "additionalPaths": [
+        "other/endpoint1",
+        "other/endpoint2"
+      ],
+      "metadata": {}
+    },
+    {
+      "sha256": "<java13-asset-sha256>",
+      "uri": "/local/path/to/java13.tgz",
+      "additionalPaths": [],
+      "metadata": {
+        "name": "java13"
+      }
     }
-  },
-  "<java15-asset-sha256>" : {
-    "layerDiffID": "<layerk-diffID>",
-    "additional_paths": []
-  }
+  ],
+  "...": [],
+  "<layern-diffID>": [
+    {
+      "sha256": "<java15-asset-sha256>",
+      "uri": "/local/path/to/java15.tgz",
+      "additionalPaths": [],
+      "metadata": {}
+    }
+  ]
 }
-
 ```
 
-#### Schema
-TODO: once above format is validated.
+### Creating Asset Packages.
 
+Asset Image creation will be handled by a new pack command, `pack package-assets --config <packag.toml> <asset-image-name>`.
 
+The `package-assets` command will use a `package.toml` configuration file and produce an `<asset-image-name>` OCI image artifact. `<asset-image-name>` may be a tarball, or a locally available image.
 
-### Creating Asset Images.
+To specify assets included in an Asset Package the `package.toml` file must contain an `assets` array. Each entry in this array may provide:
+  - `uri` (string), (required) local path or URL
+  - `sha256` (string), (required) Must be unique. Used for validation and as an endpoint where an  asset will be provided.
+  - `additional-paths` (array)
+  - `metadata` (arbitrary mapping)
+  - `unzip` (bool), unzip asset when building
 
-Asset Image creating will be performed by a new pack command, `pack packages-assets`. `package-assets` and will use an `assets.toml` configuration file. Each entry in `asset.toml` will provide the following:
-  - `uri` string (required)
-  - `sha256` (required) used for validation
-  - `additional_paths` array
-  - `metadata` arbitrary mapping
+The `additional-paths` array lists endpoints where the platform should provide symlinks to the asset. This is provided for cases where `assets` require absolute paths.
 
-The `additional_paths`array lists additional endpoints where the platform should provide the asset to buildpack during a buildpack run. This is provided for cases where `assets` require absolute paths.
-
-
-Ever assets must be in one of the following archive formats:
+The `unzip` field is used to indicate that an asset should be unarchived then added to the Asset Package. In order to use this flag an asset is  must be in one of the following archive formats:
   - `tar.gz`
-  - ...
+  - (TBD)
 
 
-So the `asset.toml` to produce the asset image [above](#Example) would be:
+The `asset` array that would produce the asset image [above](#Example) would be:
 
 #### Example
 ``` toml
 [[asset]]
 url = "https://path/to/java11.tgz"
 sha256 = "some-sha256"
-additional_paths = ["other/endpoint1", "other/endpoint2"]
+additional-paths = ["other/endpoint1", "other/endpoint2"]
+unzip = true
 
 [[asset]]
 url = "/local/path/to/java13.tgz"
@@ -125,7 +128,7 @@ sha256 = "another-sha256"
     name = "java13"
  
 [[asset]]
-...
+#...
 
 [[asset]]
 url = "/local/path/to/java15.tgz"
@@ -133,71 +136,58 @@ sha256 = "another-nother-sha256"
 ```
 
 The `pack create-asset` command will
-  - download all assets, and validate a sha256 of the artifact.
-  - unarchive the archive, transform it to a [image layer filesystem changeset](https://github.com/opencontainers/image-spec/blob/master/layer.md)
+  - download asset if needed, then validate a sha256 of the artifact
+  - if unzip is true, unzip the asset
+  - transform asset to a [image layer filesystem changeset](https://github.com/opencontainers/image-spec/blob/master/layer.md) where the asset is provided at `/cnb/assets/<artifact-sha256>`
+  - add all `additional-paths` symlinks to the resultant image, in the case of path collisions fail
   - order all assets layers [diffID](`https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md#layer-diffid`)
-  - add all `additional_paths` symlinks to the resultant image, in the case of path collisions fail.
-  - add `io.buildpacks.buildpack.assets` label metadata to image.
+  - add `io.buildpacks.buildpack.assets` label metadata to image
 
-### Buildpackage changes
+## Using Asset Packages
 
-In order for buildpackages to be able to reference the `assets` and pull them into a build, we need to add `assetHash` metadata to the `io.buildpacks.buildpack.layers` `buildpackage` label.
+##### Asset packages as arguments
 
-``` json
-{
-   "buildpack-org/java-buildpack":{
-      "0.0.178":{
-         "api":"0.2",
-         "stacks":[
-            {
-               "id":"io.buildpacks.stacks.bionic"
-            }
-         ],
-         "layerDiffID":"sha256:<some-sha256>",
-         "depDiffIDs":[
-            "sha256:<some-dep-sha256>"
-         ],
-         "assetHash": [
-            "<java12-asset-sha256>",
-            "<java13-asset-sha256>",
-            "<java15-asset-sha256>"
-         ]
-      }
-   }
-}
-```
+Asset Images may be passed as arguments to the platform may be in any of the following formats:
+- OCI image archives
+- locally available images
+- remote images
 
-In order for us to have access to this data our `pack package-buildpack` command should require both a `package.toml` and `assets.toml` configuration file.
+Decisions between local or remote images are left to the platform.
 
-### New `build` behavior
+##### `lifecycle` changes
+The platform should now pass an additional `CNB_ASSETS` environment variable to the `build` phase binary. `CNB_ASSETS` will by default be `/cnb/assets`. This provides a standard variable buildpacks may use when searching for assets.
 
+### `create-builder` behavior
 
-##### Lifecycle changes
+The `pack create-builder` command will accept Asset Packages passed with `asset` flags. In the final builder image all layers from Asset Packages will be the final `k` layers in the image. These `k` layers will be ordered by `diffID`.
 
-The platform should now pass an additional `CNB_ASSETS` environment variable, with the default value of `/cnb/assets`, to the `build` binary.
+The `create-builder` command will perform the following steps on the specified Asset Packages:
+  - Validate any two assets do not provide the same `assets` in both 'zipped' and 'unzipped' form.
+  - Validate any two assets do not have a common element in `additionalPaths` that is linked to different `/cnb/assets/<asset-sha256>` locations.
+  - If any asset layer is a superset of another, only the superset layer is included in the builder.
+  - builders inherit a `io.buildpacks.buildpack.assets` Label containing entries for every asset layer included in the builder.
 
-##### Platform behavior changes
+### `build` behavior
 
-To enable buildpacks to access vendored assets, a platform may:
-  - Read buildpackage `assetHash` metadata.
-  - For every `sha256` in this hash we search all locally available images for the `io.buildpacks.buildpack.assets` label, if the label exists check if it contains the `sha256`.
-      - if so add this layer to the builder, this will add an asset at `/cnb/assets`, then and create symlinks for all `additional_paths`.
-  - Error cases:
-      - two different layers modify the same `/cnb/assets/<sha256>`
-      - two different layers attempt to create the same `additional_path` symlink
+To enable buildpacks to access vendored assets during a `pack build`, a platform should:
+  - accept `assets` flags
+  - perform the same asset validation that occurs in `create-builder`
+  - add asset layers from both the builder and flags to the build environment image
 
-During a buildpacks `build` phase, vendored assets will be accessible at below `/cnb/assets` and buildpacks may use them as needed.
+This allows buildpacks to access vendored assets below `/cnb/assets` during the `build` phase.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 Why should we *not* do this?
 
-- Performance concerns around finding `assets` in local images. This search likely does not scale well (lots of file IO)
+- Increases complexity for users. This adds another artifact to manage.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Should `asset.toml` and `package.toml` be collapsed into a single file? This would break lots of existing
-- Should builders have access to local assets? What type of config should users pass to enable or disable this behavior?
+- Other formats to support when 'unzipping'
+- How should we resolve `zipped` vs `unzipped` assets, providing both a better strategy?
+- is `diffID` a good key to order layers by?
+- should we be able to add asset packages to buildpackages?
 
--- a/text/0000-offline-buildpackages.md
+++ b/text/0000-offline-buildpackages.md
@@ -228,7 +228,7 @@ To keep track of the buildpack to asset mapping defined in `buildpack.toml` we a
 This label let us determine:
 - set of all assets that could be used by a buildpackage
 - set of assets that may be used by an individual buildpack.
-- locations to fetch assets may be from
+- locations to fetch assets from
 
 
 To keep our metadata consistent we also propose adding the `assets` array to the `io.buildpacks.buildpackage.metadata` label.
@@ -410,7 +410,7 @@ When creating a builder we can now specify `asset-cache` packages we wish to inc
 
 Asset caches can be added to a build by two mechanisms
 1) specify an asset cache(s) using the `--asset-cache` flag(s) in a `pack build <...>`.
-2) Assets cache layers can be added to a builder image during its creation. These assets will then be available to all builds that use this builder. These assets will be pulled from
+2) Assets cache layers can be added to a builder image during its creation. These assets will then be available to all builds that use this builder.
 
 
 When asset caches are added to a build/builder we need to verify the following:
--- a/text/0000-offline-buildpackages.md
+++ b/text/0000-offline-buildpackages.md
@@ -57,7 +57,7 @@ Asset caches will also have a `io.buildpacks.asset.layers` Label. This label's c
 {
   "<layer1-diffID>": [
     {
-      "name": "java11"
+      "name": "java11",
       "digest": "sha256:java11-asset-sha256",
       "uri": "https://path/to/java11.tgz",
     },
--- a/text/0000-offline-buildpackages.md
+++ b/text/0000-offline-buildpackages.md
@@ -13,7 +13,7 @@
 
 This change allows buildpacks to access cached assets during builds to enable offline builds and asset reuse.
 
-To do so we introduce the asset cache, a distribution artifact that provides vendored assets for buildpacks, and platform mechanisms to provide vendored assets to buildpacks during `build` and `detect`. We also add a top level `[[assets]]` array in `buildpack.toml`
+To do so we introduce the asset cache, a distribution artifact that provides vendored assets for buildpacks, and platform mechanisms to provide vendored assets to buildpacks during `build`. We also add a top level `[[assets]]` array in `buildpack.toml`
 
 # Motivation
 [motivation]: #motivation
@@ -46,11 +46,11 @@ We define the asset cache artifact. It is a reproducible OCI image. Each layer i
 Asset caches will have two labels. An`io.buildpacks.asset.cache.metadata` label that contains a human readable name.
 ``` json
 {
-  "name": "asset-org/asset-name"
+  "name": "asset-cache-org/asset-cache-name"
 }
 ```
 
-Asset caches will also have a `io.buildpacks.asset.cache.layers` Label. This label maps individual `assets` to the containing layer using `layerDiffID`
+Asset caches will also have a `io.buildpacks.asset.layers` Label. This label maps individual `assets` to the containing layer using `layerDiffID`
 
 ##### General Format:
 ```json
@@ -61,6 +61,9 @@ Asset caches will also have a `io.buildpacks.asset.cache.layers` Label. This lab
     "version": "(required)",
     "layerDiffID": "(required)",
     "uri": "(optional)",
+    "licenses" : ["(optional)"],
+    "description" : "(optional)",
+    "homepage" : "(optional)",
     "stacks": [
       "(optional)",
       "..."
@@ -120,6 +123,9 @@ In order to maintain a link between buildpacks and the assets they may provide w
   id   = "(required)"
   version = "(required)"
   uri = "(optional)"
+  licenses = ["(optional)"]
+  description = "(optional)"
+  homepage = "(optional)"
   stacks = ["(required)"]
   [assets.metadata]
    optional-key = "(optional value)"
@@ -151,6 +157,9 @@ To keep track of the buildpack to asset mapping defined in `buildpack.toml` we a
           "id": "asset-1-id",
           "version": "asset-1-version",
           "uri": "asset-1-uri",
+          "licenses" : ["license1"],
+          "description" : "Some asset",
+          "homepage" : "https://buildpacks.io",
           "name": "Asset 1 Name",
           "stacks": [
             "stack1",
@@ -203,11 +212,14 @@ To keep track of the buildpack to asset mapping defined in `buildpack.toml` we a
           "version": "12.9.0",
           "name": "Node JS",
           "uri": "https://asset-location-url",
+          "licenses" : ["GPLv3"],
+          "description" : "the twelfth version of node",
+          "homepage" : "https://buildpacks.io",
           "stacks": [
             "io.buildpacks.stacks.bionic"
           ],
           "metadata": {
-            "license": "MIT"
+            "packaged-by": "buildpacks.io"
           }
         },
         {
@@ -232,7 +244,7 @@ This label let us determine:
 
 ### Builder Changes
 
-To keep track of what assets are actually in our builder, and what layer they are contained in we add the label, `io.buildpacks.asset.cache.layers`. Note, this label has already been defined above on asset-cache.
+To keep track of what assets are actually in our builder, and what layer they are contained in we add the label, `io.buildpacks.asset.layers`. Note, this label has already been defined above on asset-cache.
 
 Builders will also require the same changes outlined in the Buildpackage section above to the io.buildpacks.buildpack.layers tag.
 
@@ -249,6 +261,11 @@ Builders will also require the same changes outlined in the Buildpackage section
         {
           "sha256": "asset-1-sha256",
           "id": "asset-1-id",
+          "name": "Asset Name",
+          "uri": "https://asset-location-url",
+          "licenses" : ["license1"],
+          "description" : "First Asset Description",
+          "homepage" : "https://buildpacks.io",
           "version": "1.2.3",
           "stacks": [
             "stack1",
@@ -281,7 +298,7 @@ Asset Cache image creation will be handled by the platform. E.g. `pack package-a
 
 There are two cases depending on weather `<buildpack-uri> points to a buildpackage or a directory containing a `buildpack.toml` file.
 1) Directory containing a `buildpack.toml` file. This method uses the `assets` array to locate download and verify assets.
-2) An existing buildpackage. This will download all assets using `uri`s from the `io.buildpacks.buildpack.layers` label, or re-use assets that are included in the specified buildpackage's layers.
+2) An existing buildpackage. This will download all assets using `uri`s from the `io.buildpacks.buildpack.layers` label.
 
 The `--config` flag is always optional but recommended to control naming of the generated asset and control which assets are added to the asset-cache. (May also manually pass naming info)
 
@@ -299,14 +316,14 @@ To provide additional control over asset creation, we may optionally pass  the `
 [asset-cache (optional)]
   name = "(required)"
  
-[[exclude-buildpack-assets (optional)]]
+[[exclude-buildpack (optional)]]
   id = "(required)"
   version = "(required)"
 
-[[exclude-assets (optional)]]
+[[exclude (optional)]]
   sha256 = "(required)"
 
-[[asset-override (optional)]]
+[[override (optional)]]
   uri = "preferred/asset/location"
   sha256 = "asset-sha256"
 ```
@@ -316,15 +333,15 @@ To provide additional control over asset creation, we may optionally pass  the `
 [asset-cache]
   name = "node-asset-cache"
  
-[[exclude-buildpack-assets]]
+[[exclude-buildpack]]
   id = "buildpacks/node-engine"
   version = "1.2.3"
 
-[[exclude-assets]]
+[[exclude]]
   sha256 = "vulnerable-node10-sha256"
   name = "Node js version with security vulnerability. (non-functional)"
 
-[[asset-override]]
+[[override]]
   uri = "../local/path/to/nodejs-asset"
   sha256 = "node13-sha256"
 ```
@@ -334,7 +351,7 @@ Asset cache creation should:
   - filter excluded assets specified in the `config` file.
   - Transform all remaining assets to an image layer filesystem changeset where the asset is provided at `/cnb/assets/<artifact-digest>`.
   - Order all assets layers diffID.
-  - Add the `io.buildpacks.asset.cache.layers` label metadata to the asset image. If no name is specified for `io.buildpacks.asset.cache.metadata`, generate one.
+  - Add the `io.buildpacks.asset.layers` label metadata to the asset image. If no name is specified for `io.buildpacks.asset.cache.metadata`, generate one.
   - set the created time in image config to a constant value.
   - set the modification time of all non-asset files to a constant value.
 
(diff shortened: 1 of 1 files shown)


text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,236 @@
...
+│   └── <id>
+│        ├── v1.2.3.toml
+│        ├── v1.2.3
+│        ├── <version>
+│        └── <version>.toml
+└── lifecycle
+```
+### How will paths to dependencies be exposed to a buildpack?
+
+During the `build` phase of the `paketo-buildpacks/mri` buildpack, the `CNB_DEPS` environment variable will provide a path pointing to a `packaged.toml` file, this will contain information on all all dependencies packaged with a buildpack.
```
hone, line 104:
```suggestion
During the `build` phase of the `paketo-buildpacks/mri` buildpack, the `CNB_DEPS` environment variable will provide a path pointing to a `packaged.toml` file, this will contain information on all dependencies packaged with a buildpack.
```

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,236 @@
...
+    │    └── v1.2.3
+    └── paketo.node
+         ├── v10.20.0.toml
+         └── v10.20.0
+
+```
+
+
+### How are dependencies grouped as layers
+Layers containing dependencies are attached as the final layers to a buildpackage. The exact distribution of dependencies into layers is left up to the packaging utility.
```
hone, line 149: Would we want this spec'd so the images produced are the same/deterministic?

hone:
A few questions:

* Will you also package a non-offline buildpackage at all or is the idea that the buildpack can be offline or online and it's just up to how it's being packaged?
* In the case of the buildpack author, what if my deps are in non-offline mode, do I need to switch and figure out how to download those dependencies myself since I can't assume it's offline necessarily?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,236 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What is the strategy for staying under the 127 Layer limit.
```
hone, line 224: If it's an issue, packaging up the deps could be a single layer, but then you do lose some of the benefits of dedupliciation beteween layers.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,236 @@
...
+- Is `CNB_DEPS` a bad buildpack author experience?
+- should we force some dependency api at each `/cnb/deps/some-identifier/v1.2.3`?
```
hone, line 226: What are you thinking here? What would the dependency API do?

dwillist: Ya I don't think we should do this, old note. Removing.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+Answering the following should outline the behavior of this feature.
+
+### How will dependencies appear on the filesystem during the `build` phase?
+
+Dependencies will be listed underneath `/cnb/deps/` with a schema that has the following properties:
+
+- dependency placement does not depend on the dependencies bits (no hashes, SHA256, ...).
```
hone, line 72: Can you add the reasoning explained during the WG today? /cc @sclevine @dwillist

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+
+One paragraph explanation of the feature.
+
+We need to have a method to attach dependencies to buildpackages to support offline workflows. These dependencies should then be made available via the filesystem during the `build` phase.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+    - Remove need for two separate `ids` for online vs offline buildpacks
```
natalieparellano, line 21: Would it be worth describing further how we'll preserve online vs. offline behavior? I assume something like, if a buildpack doesn't see its dependencies in the file system, go download them... or whatever is the intention.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+    - Performance optimizations
```
natalieparellano, line 22: How?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+│           │   └── detect
+│           ├── dependencies
+│           │   ├── ruby-v2.7.1
+│           │   └── ruby-v2.6.8
+│           └── buildpack.toml
+└── lifecycle
+```
+There are problems with the above:
+
+- Buildpacks packaged with dependencies & packaged without dependencies have different `sha256` values and therefore must have different buildpack `id` fields to avoid registry collision issues.
```
natalieparellano, line 54: This is just one problem - are there others? Just curious :)

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+│   └── <id>
+│        ├── v1.2.3.toml
+│        ├── v1.2.3
+│        ├── <version>
+│        └── <version>.toml
+└── lifecycle
+```
+### How will paths to dependencies be exposed to a buildpack?
+
+During the `build` phase of the `paketo-buildpacks/mri` buildpack, the `CNB_DEPS` environment variable will provide a path pointing to a `packaged.toml` file, this will contain information on all dependencies packaged with a buildpack.
```
natalieparellano, line 104: Who writes packaged.toml?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+
+[[dependencies]]
+uri = "path/to/dependency"
+id = "org.some-identifier"
+version = "v1.2.3"
+  [dependencies.metadata]
+  # dependency-specific data
+
+[[dependencies]]
+uri = "https://buildpacks.cloudfoundry.org/dependencies/node/node_10.20.0_linux_x64_cflinuxfs3_76f4a198.tgz"
```
natalieparellano, line 124: Is this for an "online" buildpack? When does this dependency get downloaded?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+            {
+               "id":"org.cloudfoundry.stacks.cflinuxfs3"
+            },
+            {
+               "id":"io.buildpacks.stacks.bionic"
+            }
+         ],
+         "layerDiffID":"sha256:9ae53c280ea60b30b8c03978fbd2f761769fe0c15e421a9ece68e2f52926ce77",
+         "depLayers":{
+              "sha256:sha256-to-dep-layer": {
```
natalieparellano, line 169: Where do we specify where the dependency layer lives (i.e. in the registry)?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+ID = "paketo.node"
+  [dependencies.metadata]
+  cool = "beans"
+```
+
+This file gives us a single place to read all offline dependency information during a build.
+
+
+### Other benefits?
+Build times could be further optimized, with dependencies attached as layers, there is the option to symlink a file at `cnb/dependencies/org.dep-id/stack.v1.2.3` directly to a buildpack provided layer.
```
natalieparellano, line 214: Could you describe in a little more detail how this would work? I believe there was also some discussion around when this would NOT be possible and what sort of actions (copying, etc.) would be necessary in that case. Maybe not essential but would be good not to lose that context.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+                 "metadata": {"cool": "beans"}
+              }
+         }
+    
+      }
+```
+Each entry in the `depLayers` object maps a layer SHA to it's path. As well as preserves the `id`, `version` and `metadata` fields from `package.toml`
+
+### How do we determine if a dependency layer should be added to a build?
+The file tree rooted at the `/cnb/deps/` is only available at `build` time. So we add all layers to a build that match the following criteria:
```
natalieparellano, line 187: Does this have any implications for caching? Currently for `pack` we get a fair number of questions around e.g., why does the JDK get dowloaded separately for all of my apps... presumably if the dependency SHA is unchanged then that layer could get re-used across builds?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+                 "path": "/cnb/deps/path/to/dep1",
+                 "id": "org.some-identifier",
+                 "version": "v1.2.3", "metadata": {}
+              },
+              "sha256:sha256-to-other-dep-layer": {
+                 "path": "/cnb/deps/path/to/dep1",
+                 "id": "paketo.node",
+                 "version": "v10.20.0",
+                 "metadata": {"cool": "beans"}
+              }
```
ekcasey, line 179: This format seems to preclude adding multiple deps in a single layer in the future. Should we make each `depLayer` contain an array of dependencies so that we have some breathing room if/when we need to design a strategy for dealing with the layer limit?

ekcasey: Also, maybe we should pull the `depLayers` out the buildpack map,  in case multiple buildpacks want to share an identical dep layer?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+         }
+    
+      }
+```
+Each entry in the `depLayers` object maps a layer SHA to it's path. As well as preserves the `id`, `version` and `metadata` fields from `package.toml`
+
+### How do we determine if a dependency layer should be added to a build?
```
ekcasey, line 186: It seems like we are implying that an ephemeral builder images will be generated per build. I think this is impossible in many platforms and might be slow in `pack` (generating new daemon images is slow without a lot of complex caching by the platform).

text/0000-offline-buildpackages.md
ekcasey, line 104: Instead of adding an additional `packaged.toml` file, I would prefer that we simply given the buildpack the path to the dependency directory in an environment variable and lay out dependencies within that directory in a predictable/specified way.

text/0000-offline-buildpackages.md
dwillist: Ya definitely, I'll add a total ordering scheme to these additional layers.

text/0000-offline-buildpackages.md
dwillist: Yep, that was the initial idea, online vs offline behavior would require logic in each buildpack ( buildpack authors would be responsible). There was some discussion about trying to reduce this and provide a better experience for buildpack authors.

text/0000-offline-buildpackages.md
dwillist: Ya this should be 'potential performance optimizations' because it requires some additional lifecycle logic. 
How this 'could' work:
- we have a dependency layer in our buildpackage located at `/cnb/deps/my-org/my-dep`, where the contents of `my-dep` look something like this:
```
├─ /cnb/deps/some-org/my-dep
    ├── bin
    │    └── some-executable
    └── lib
         └── header-file.so
         └── header-file.so
```

This dependency has a `bin` and `lib` folder and looks a lot like a [`<layer>`](https://github.com/buildpacks/spec/blob/master/buildpack.md#build). We could have `<layers>/<layer>` symlink to `/cnb/deps/some-org/my-dep`. The lifecycle would have to add a feature where we add symlinked layers to the final image in order to make use of this symlink. This would let buildpack authors just write a symlink file instead of copying all the `my-dep` bits into a layer. (not too sure about how the lifecycle but guessing this might also save us from creating a new OCI layer for `<layers>/layer>` to be added to final image?)

text/0000-offline-buildpackages.md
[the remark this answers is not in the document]
sclevine: The goal is 1:1 correspondence between the FS blob digest for the buildpack bits and the ID/version combo. I.e., when you build a builder, sourcing both the online and offline buildpackages would result in the offline buildpackage bits being included by virtual of it being a superset of the online buildpackage layers.

text/0000-offline-buildpackages.md
matthewmcnew: I think it would be valuable if this RFC provided a discussion or proposal of how the lifecycle would be able to pull off this performance optimization. 

It seems that the lifecycle would need access to metadata about where deps could be cross repository blob mounted from. Perhaps, when the builder is created a "deps.toml" could be constructed for the lifecycle that lists where each dep is available on a registry so the lifecycle knows if it should attempt to perform a cross repository blob mount on export. 

Additionally, will the lifecycle create one layer with all deps symlinks or will each symlink be its own layer?

cc @ekcasey

text/0000-offline-buildpackages.md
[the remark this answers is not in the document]
matthewmcnew: While it certainly feels cleaner that resulting builder has an identical layer digest for all buildpack id/versions, it seems disingenuous to imply that this actually avoids avoid registry collision issues.Buildpack tooling will still need to differentiate/choose between multiple versions of a buildpackage in a registry or buildpack repository.

matthewmcnew:
Depending on the fate of `packaged.toml`, I think we need to clarify what info needs to be made available to the lifecycle and how it is done (perhaps a specific deps-info.toml for the lifecycle only to consume?)

text/0000-offline-buildpackages.md
dwillist: The lifecycle.

text/0000-offline-buildpackages.md
dwillist: Think this is talked about [above](https://github.com/buildpacks/rfcs/pull/81#discussion_r432077984)

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+[[dependencies]]
+uri = "path/to/dependency"
+id = "org.some-identifier"
+version = "v1.2.3"
+  [dependencies.metadata]
+  # dependency-specific data
+
+[[dependencies]]
+uri = "https://buildpacks.cloudfoundry.org/dependencies/node/node_10.20.0_linux_x64_cflinuxfs3_76f4a198.tgz"
+ID = "paketo.node"
```
dfreilich, line 125: nit: 
```suggestion
id = "paketo.node"
```

text/0000-offline-buildpackages.md
yaelharel, line 21: I think it's worth adding when users are using offline vs online buildpacks

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+- Start Date: 2020-04-27
+- Author(s): Dan Thornton
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
```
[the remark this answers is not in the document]
yaelharel: The same comment in every section (for example - line 27)

text/0000-offline-buildpackages.md
dwillist: Believe this is clarified in the updated version [here](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md#id--version-vs-hash-as-a-key)

text/0000-offline-buildpackages.md
dwillist: Ya, going to table these optimizations to a subsequent rfc

text/0000-offline-buildpackages.md
dwillist: There is now an alternative proposal that has behavior like this [here](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md#always-require-offline_dependencies-would-need-a-name-change)

text/0000-offline-buildpackages.md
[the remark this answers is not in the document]
dwillist: Think this change is that for buildpack authors. No longer do they need to have 2 separate `buildpack.toml` id's for these two artifacts.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,234 @@
...
+[how-it-works]: #how-it-works
+
+Answering the following should outline the behavior of this feature.
+
+### How will dependencies appear on the filesystem during the `build` phase?
+
+Dependencies will be listed underneath `/cnb/deps/` with a schema that has the following properties:
+
+- dependency placement does not depend on the dependencies bits (no hashes, SHA256, ...).
+- dependency placement is calculable by the lifecycle (needed to provide through `CNB_DEPS`).
```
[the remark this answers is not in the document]
dwillist: Ya this is was confusingly worded. There is now an amended explanation of this [here](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md#id--version-vs-hash-as-a-key)

text/0000-offline-buildpackages.md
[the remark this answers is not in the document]
dwillist: Thanks, think these are resolved in the updated version 👍

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,399 @@
...
+[how-it-works]: #how-it-works
+
+Lets walk through the following steps:
+1) How to create an offline buildpackage.
+2) Changes to the buildpackage format.
+3) How buildpacks access offline dependencies during execution.
+
+
+## How to Create an offline buildpackage
+This is currently done via the `pack package-buildpack` command, which consumes a [`package.toml`](https://buildpacks.io/docs/reference/package-config/) file. We propose adding an `offline_dependencies` array. We require that the `id` + `version` combination will be unique among all `offline_dependencies` entries.
```
danielleadams, line 66: Does the version always have to be exact? Or can this support a range? Also, would version be optional?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,399 @@
...
+│       ├── <paketo.bundler_1.17.1_diff_sha256>
+│       ├── <image_config_json_sha256>
+│       └── out
+├── index.json
+└── oci-layout
+```
+
+We will refer the the `<paketo.ruby_*..>` entries in the above OCI image as dependency layers.
+
+When this all of the layers of the above image are applied we should have the following directory structure:
```
danielleadams, line 110:
```suggestion
When all of the layers of the above image are applied we should have the following directory structure:
```

text/0000-offline-buildpackages.md
dwillist: It would totally be up to the buildpack author what to put in the `version` field. It could really be any string. This is just a piece metadata telling your buildpack where it should look for this dependency in case of an offline build `/cnb/deps/<id>/<version>`. The only request is that it be unique.

text/0000-offline-buildpackages.md
dwillist: Thanks 👍

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,398 @@
...
+
+## How to Create an offline buildpackage
+This is currently done via the `pack package-buildpack` command, which consumes a [`package.toml`](https://buildpacks.io/docs/reference/package-config/) file. We propose adding an `offline_dependencies` array. We require that the `id` + `version` combination will be unique among all `offline_dependencies` entries.
+
+#### package.toml
+``` toml
+[buildpack]
+uri = "./path/to/buildpack.tgz"
+
+[[offline_dependencies]]
```
jkutner, line 73: I think this is useful beyond the offline buildpacks use case. Could we name this table `[[assets]]`?

Another reason to change the name is the a "dependency" in the context of this file is a buildpack, but these are not buildpacks.

dwillist:
Summary of discussions around this RFC from previous working group meetings etc:
- Would be useful to have `assets` in a separate image and completely separate from the buildpackage.
  - would make it easier to create offline builders with many metabuildpacks (less artifact management)
  - would let buildpacks be more intelligent about using dependencies if they are present on the registry.
- listing dependencies at `/cnb/deps/<id>/<version>` is more awkwards that using `/cnb/deps/<sha256>`. 
  - Just providing a work-around for dependencies that need absolute paths may be preferable.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,203 @@
...
+[[asset]]
+...
+
+[[asset]]
+url = "/local/path/to/java15.tgz"
+sha256 = "another-nother-sha256"
+```
+
+The `pack create-asset` command will
+  - download all assets, and validate a sha256 of the artifact.
```
nebhale, line 136: Is download reasonable?  What if you want to build an asset image with an asset you currently have, but cannot put on then HTTP endpoint?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,203 @@
...
+  - unarchive the archive, transform it to a [image layer filesystem changeset](https://github.com/opencontainers/image-spec/blob/master/layer.md)
```
nebhale, line 137: What if you don't want to unarchive the file?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,203 @@
...
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+- Performance concerns around finding `assets` in local images. This search likely does not scale well (lots of file IO)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should `asset.toml` and `package.toml` be collapsed into a single file? This would break lots of existing
```
nebhale, line 201: Would it?  Seems like it's a purely additive change:

```yaml
dependencies = [
  { image = "gcr.io/paketo-buildpacks/apache-tomcat:1.5.0" },
  { image = "gcr.io/paketo-buildpacks/azure-application-insights:1.10.0" },
  ...
]

assets = [
  { image = "gcr.io/paketo-assets/apache-tomcat:8.0.0" },
  { image = "gcr.io/paketo-assets/apache-tomcat:8.0.1" },
]
```

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,203 @@
...
+- Should builders have access to local assets? What type of config should users pass to enable or disable this behavior?
```
nebhale, line 202: Yep, they should.

dwillist:
Summarizing conversation from last working group:
- The current specification for the `[[asset]]` array has some undesirable behavior. Namely that two assets can be placed at the same `/cnb/assets/<sha256>` location. This occurs when one asset is packaged with `unzip = true` and the other is not. This is particularly undesirable for un-archived assets (where `unzip = true`) as the contents are unzipped and we lose information about how they were packaged.
Possible solutions discussed:
- GUID for each asset
- lookup table in metadata

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,193 @@
...
+To specify assets included in an Asset Package the `package.toml` file must contain an `assets` array. Each entry in this array may provide:
+  - `uri` (string), (required) local path or URL
+  - `sha256` (string), (required) Must be unique. Used for validation and as an endpoint where an  asset will be provided.
+  - `additional-paths` (array)
+  - `metadata` (arbitrary mapping)
+  - `unzip` (bool), unzip asset when building
+
+The `additional-paths` array lists endpoints where the platform should provide symlinks to the asset. This is provided for cases where `assets` require absolute paths.
+
+The `unzip` field is used to indicate that an asset should be unarchived then added to the Asset Package. In order to use this flag an asset is  must be in one of the following archive formats:
```
hone, line 109: When I think of `unzip` it makes me think of the `.zip` format explicitly. What do you think of `extract` which is what is used in the [tar manpage](https://linux.die.net/man/1/tar) for GNU flags?

dwillist:
Here are three proposals, let me know what your thoughts are on them. I would love to reach some consensus and finish off this RFC.

### ecasey's idea:
Two types of assets; layer assets and file assets.
1) Layer assets:
    * must be tgz, tar files, or local directories
    * get extracted when the asset layer is created
    * allow for symlinking-reuse magic
2. File assets:
    * can be any single file
    * never get extracted
    * must be copied into a layer to be used
    * Then we just map them under two separate hierarchies

### Possible Separation Schemes: 

#### Totally different paths: 

```
/cnb/file-assets/<sha256-file-asset> (archived files)
  - this asset may be double compressed

/cnb/layer-assets/<sha256-layer-asset> (unarchived files)
  - asset has been uncompressed, then saved as a layer.
```

Tradeoff is we lose the chain of custody, `<sha256-asset>` will not be the `sha256` of the actual contents under `/cnb/layer-assets/*`.

#### All under assets:
A blend of the above and a proposal Stephen had.

Again 2 types of assets, Layer Assets & File Assets. Both are made available under the same `/cnb/file-assets`

```
/cnb/assets/<file-asset-sha256> (archived files)
  - asset has been uncompressed, then saved as a layer.

/cnb/assets/<layer-asset-sha256> (unarchived files)
   - this asset may be double compressed
```

Tradeoff is `<layer-asset-sha256>` is calculated by pack & makes buildpack lookup more complicated.


#### Bit more nesting.
Or something with a bit more nesting but may make the buildpack implementation a bit easier.

Again same file-asset, layer-asset split.

```
/cnb/assets/<file-sha256>/asset/file-asset (archived files)
  - this asset may be double compressed
  
/cnb/assets/<file-sha256>/<layer-sha256>/layer-asset (unarchived files)
  - asset has been uncompressed, then saved as a layer.
```

Tradoff is file structure gets more complicated. Kind of gross extra `asset` folder for `file-assets`.

jkutner (approved):

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+Asset packages will have a `io.buildpacks.asset.layers` Label. This label's contents will be a `json` object mapping each layer to metadata describing the assets it provides.
+
+``` json
+{
+  "<layer1-diffID>": [
+    {
+      "sha256": "java11-asset-sha256",
+      "uri": "https://path/to/java11.tgz",
```
natalieparellano, line 50: Can you clarify what `uri` is for? The package already contains the data, is it there for audit purposes?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,221 @@
...
+
+```
+cnb
+ └── assets
+     ├── <java11-asset-sha256>
+     ├── <java13-asset-sha256>
+     └── <java15-asset-sha256>
+```
+
+Buildpacks should then be able to quickly check for available assets using a `sha256` of the asset they wish to use in tandem with the `CNB_ASSETS` environment variable.
```
natalieparellano, line 202: Can you describe further what `CNB_ASSETS` will look like?

natalieparellano (commented):
Excited the see this incarnation :)

text/0000-offline-buildpackages.md
dwillist: Yep, it just there so that we have some human readable information about what is in the layer.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,426 @@
...
+    }
+  ]
+}
+```
+
+##### Builder Changes
+
+Builders that contain asset caches will have an `io.buildpacks.asset.layers` with the same format as the asset cache label. Again this label will contain a `<layer1-diffID>` entry for each asset layer contained in the builder.
+
+Builders will additionally need to retain the asset references for each buildpackage in the builder. For this we will extend the existing `io.buildpacks.buildpack.layers` label to contain mappings to contain an `cachedAssets` field identical to the one above.
```
matthewmcnew, line 140: Why will builders need this information as well. Would it be sufficient for the builder to have a label with information equivalent to: `io.buildpacks.buildpackage.assets`?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,426 @@
...
+    {
+      "digest": "sha256:<java15-asset-sha256>",
+    }
+  ]
+}
+```
+
+##### Buildpackage Changes
+
+To enable buildpackages to reference asset caches, we define a new buildpackage label, `io.buildpacks.buildpackage.assets`.
```
matthewmcnew, line 82: Could this be named `io.buildpacks.asset.layers` to maintain the same name across buildpackages/builders/asset packages?

text/0000-offline-buildpackages.md
dwillist: Ya, I was thinking this would be good information to let platforms have access to allow them to make decisions about pulling asset images before a build. This asset image would then be available locally for all subsequent builds.

dwillist:
From the last discussion on the RFC:
- There is no way to build a mapping between each `buildpacks` and its provided `assets` in a metabuildpackage. This is a problem for `kpack` which has the ability to break apart metabuildpacks, into it's component pieces. It is then impossible to determine what `assets` should going along with these component pieces.

The latest revision of this proposal remedies this by 
- Adding `assets` as a defined key in `buildpack.toml` To denote dependencies the buildpack may install.  This defines a mapping between a buildpack and it's `assets`
- Keeping the above mapping in the `io.buildpacks.buildpack.layers` metadata.
- Require all metabuildpacks to keep a list of all assets that any of it's constituent buildpacks could require.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+    }
+  }
+}
+```
+
+
+This label let us determine:
+- set of all assets that could be used by a buildpackage
+- set of assets that may be used by an individual buildpack.
+- locations to fetch assets may be from
```
ryanmoran, line 231:
```suggestion
- locations to fetch assets from
```

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Builder Creation
+
+When creating a builder we can now specify `asset-cache` packages we wish to include in the builder image. We may also pass an `--asset-config asset.toml` file to excluded particular assets that would normally be included. Additionally there should be the option to include all assets from specified buildpackages using the `--offline` flag (the resulting builders will be very large).
+
+## Using Asset Caches
+
+Asset caches can be added to a build by two mechanisms
+1) specify an asset cache(s) using the `--asset-cache` flag(s) in a `pack build <...>`.
+2) Assets cache layers can be added to a builder image during its creation. These assets will then be available to all builds that use this builder. These assets will be pulled from
```
ryanmoran, line 413: I feel like the end of this line, "These assets will be pulled from", was going somewhere, but then it just stopped. Was there a thought you wanted to finish here?

dwillist: Hey thanks! With transition to using `uri`'s as the way to reference assets, instead of using a mutable `asset-cache` reference `These assets will be pulled from ...` is no longer needed. Thanks for catching this.

dwillist:
At this point I think we have enough consensus around what this feature should look like & would like this RFC to be voted on.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+  "name": "asset-org/asset-name"
+}
+```
+
+Asset caches will also have a `io.buildpacks.asset.layers` Label. This label maps individual `assets` to the containing layer using `layerDiffID`
+
+##### General Format:
+```json
+{
+  "asset-sha256": {
```
hone, line 58:
```suggestion
  "<asset-sha256>": {
```

hone (approved):

schneems:
I'm interested in being able to inspect what offline assets are available. I.e. if I want to warn people that they're using Ruby 2.5.5 provided by the platform for <stack-name> but it's not available in <stack-name + 1>

dwillist:
@schneems interesting, not entirely sure I follow the example. But there will be label metadata on both builders and asset packages that lists all included offline assets [defined here](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md#asset-cache-image-labels).

sclevine (approved):

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Asset cache creation
+
+Asset Image creation will be handled by the platform. E.g. `pack package-asset <asset-cache-name> --asset-config <optional-path-to-asset.toml>`.
+
+There are two ways to package assets
+1) from a `buildpack.toml` file. This method uses the `assets` array to locate download and verify assets.
+    - Ex: `pack package-asset <asset-cache-name> <path-to-buildpack.toml>`
+2) from an already existing buildpackage. This will download all assets using `uri`s from the `io.buildpacks.buildpack.layers` label, or re-use assets that are included in the specified buildpackage's layers.
+    - Ex: `pack package-asset <asset-image-name> <path-to-package.toml>`
```
ekcasey, line 326: If we want to package from an already existing buildpackage, it might make more sense to:
`pack package-asset <asset-image-name> <buildpack-uri>`
As the buildpackage itself is the artifact that is passed around (and stored in the registry) rather than `package.toml`.

What is nice about this is that since a directory containing a `buildpack.toml` is also a buildpack URI (as pack understands it) so this usage could also satisfy the above case.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+
+The `--asset-config` flag is optional but recommended to control naming of the generated asset. (May also manually pass naming info)
```
ekcasey, line 328: I think new `pack` conventions dictate that this should be just `--config` (short `-c`).

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+## Asset cache creation
+
+Asset Image creation will be handled by the platform. E.g. `pack package-asset <asset-cache-name> --asset-config <optional-path-to-asset.toml>`.
+
+There are two ways to package assets
+1) from a `buildpack.toml` file. This method uses the `assets` array to locate download and verify assets.
+    - Ex: `pack package-asset <asset-cache-name> <path-to-buildpack.toml>`
```
ekcasey, line 324: Is there a colorable use case for packaging assets from multiple unrelated buildpacks into a single asset cache? I am assuming not.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+1) specify an asset cache(s) using the `--asset-cache` flag(s) in a `pack build <...>`.
+2) Assets cache layers can be added to a builder image during its creation. These assets will then be available to all builds that use this builder.
+
+
+When asset caches are added to a build/builder we need to verify the following:
+  - The asset is not excluded through a `asset.toml` file passed to `pack build`.
+  - If two assets provide the same `/cnb/assets/<asset-digest>` these two files must have identical contents. Decisions about rewriting these layers to optimize space are left to the platform.
+
+When creating a builder with asset caches:
+  - Asset cache layers should be the final k layers in the builder. These should be ordered by `diff-ID`.
```
ekcasey, line 421: It probably makes sense to implement this such that the asset layers go last, but maybe we should leave it out of the RFC because it's an implementation detail that doesn't matter to users and might cause performance problems if someone was extending an existing builder at build time or with some future pack feature.
```suggestion
```

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,455 @@
...
+When creating a metabuildpack, we want to retain all of the asset information from the child buildpacks.
+
+As such `assets` array in `io.buildpacks.buildpack.layers` `io.buildpacks.buildpack.metadata` of the resulting metabuildpack will be a sum of all of the `asset` arrays in its children.
+
+If there are conflicts when merging these entries, the platform may decide to continue or fail.
+
+
+## Builder Creation
+
+When creating a builder we can now specify `asset-cache` packages we wish to include in the builder image. We may also pass an `--asset-config asset.toml` file to excluded particular assets that would normally be included. Additionally there should be the option to include all assets from specified buildpackages using the `--offline` flag (the resulting builders will be very large).
```
ekcasey, line 407: Would it make sense to basically inline the `asset.toml` schema under an `[assets]` table in `builder.toml`? And maybe also include the asset-cache URI as a builder.toml option. It might be nice for users if all the config for a given builder could be specified in one file, even if some things like `asset-cache` can be modified on the fly with a flag.

ekcasey (commented):
I _love_ how this proposal has developed. I think this plan will greatly benefit `pack` users and buildpack authors.

At this point all of my comments are essentially about names and other details (I may suggest some minor `asset.toml` schema changes shortly). This is because flags and label/file schemas are expensive to change once they ship.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,426 @@
...
+...
+<layern> ━━ /cnb/assets/<java15-asset-digest>
+```
+
+##### Asset Image Labels
+Asset caches will have two labels. A simple `io.buildpacks.asset.metadata` label that contains the asset's name.
+``` json
+{
+  "name": "asset-org/asset-name",
+}
```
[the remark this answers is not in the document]
dwillist: I don't think this implies 1 asset per cache. This is just a bit of extra metadata so the we can name an asset cache in a way that is more permanent than its tag in a registry.

text/0000-offline-buildpackages.md
dwillist: I can't think of any, seems like it would be simple enough to amend this RFC & add the functionality if the need arose.

text/0000-offline-buildpackages.md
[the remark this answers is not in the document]
dwillist: I could change this to `asset-cache-org/asset-cache-name` to clarify this.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,463 @@
...
+...
+<layern> ━━ /cnb/assets/<java15-asset-sha256>
+```
+
+### Asset Cache Image Labels
+Asset caches will have two labels. An`io.buildpacks.asset.cache.metadata` label that contains a human readable name.
+``` json
+{
+  "name": "asset-org/asset-name"
+}
```
ekcasey, line 50:
```suggestion
{
  "name": "asset-org/asset-name",
  "buildpack": {
    "id": "(required)",
    "version": "(required)"
  }
}
```
Given in the proposed workflows the asset cache is for a specific buildpack, does it make sense to the buildpack coordinates (id and version) to the metadata here?

ekcasey: Also, if we tied the cache to a buildpack id/version do we still want or need a `name`? When would the name be useful?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,463 @@
...
+[buildpack]
+  id = "paketo-buildpacks/node-engine"
+  version = "0.0.1"
+ 
+[[assets]]
+  sha256 = "(required)"
+  name = "(optional)"
+  id   = "(required)"
+  version = "(required)"
+  uri = "(optional)"
```
ekcasey, line 122: What are the cases where a `uri` would not be provided. An asset that is not available for public download? In that case is an asset override required to create the asset cache? What happens if I try to create a builder from this buildpack without an asset cache? Does it fail to create the builder? Or will we wait for the buildpack to fail in cases where the asset is needed?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,463 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+## Asset cache creation
+
+Asset Cache image creation will be handled by the platform. E.g. `pack package-asset-cache <image-name> <buildpack-uri> --config <optional-path-to-asset-cache.toml>`.
+
+There are two cases depending on weather `<buildpack-uri> points to a buildpackage or a directory containing a `buildpack.toml` file.
+1) Directory containing a `buildpack.toml` file. This method uses the `assets` array to locate download and verify assets.
+2) An existing buildpackage. This will download all assets using `uri`s from the `io.buildpacks.buildpack.layers` label, or re-use assets that are included in the specified buildpackage's layers.
```
ekcasey, line 284:
```suggestion
2) An existing buildpackage. This will download all assets using `uri`s from the `io.buildpacks.buildpack.layers` label.
```
I am assuming this is out of date becaues there is no mechanism for including asset layers in a buildpackage. Is that correct?

ekcasey (approved):
I am approving. I left a couple clarifying questions and suggested some extra metadata. If we want to change the RFC to include them that would be great but I imagine some of the details will inevitably get tweaked as we make discoveries during specification and implementation anyways.

text/0000-offline-buildpackages.md
dwillist: thanks, missed this edit.

text/0000-offline-buildpackages.md
dwillist: I believe VMware has a internal dynatrace buildpack that constructs the `uri` dynamically. Thought it might be useful to still have an entry in `buildpack.toml` to indicate that an asset is downloaded as part of buildpack execution, or use it for some `sha256` validation if needed.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,463 @@
...
+```toml
+[asset-cache (optional)]
+  name = "(required)"
+ 
+[[exclude-buildpack-assets (optional)]]
+  id = "(required)"
+  version = "(required)"
+
+[[exclude-assets (optional)]]
+  sha256 = "(required)"
```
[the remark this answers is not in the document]
dwillist: Ya, when adding this I was thinking more about dev/testing workflows, I don't think we will distribute incomplete caches. But having a way to trim down the cache I make seems useful if I'm running tests against my buildpack to check if offline functionality works and I don't want to download generate the entire cache or generate a one off test builder.

text/0000-offline-buildpackages.md
dwillist: I had envisioned cache-images as having little info about how they are made or who is going to consume them, but I guess the only way we generate them now is with a `buildpack uri`.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,479 @@
...
+  - set the created time in image config to a constant value.
+  - set the modification time of all non-asset files to a constant value.
+
+## Metabuildpackage Creation
+
+### Metadata
+
+When creating a metabuildpack, we want to retain all of the asset information from the child buildpacks.
+
+As such `assets` array in `io.buildpacks.buildpack.layers` of the resulting metabuildpack will be a sum of all of the `asset` arrays in its children.
```
matthewmcnew, line 364: Why is this neccessary? The `io.buildpacks.buildpack.layers` will already include the children buildpacks as their own entry. It seems that these buildpack entries would be the natural place for the assets. A platform can find the assets for a metabuildpack by looking at each child buildpack's assets.

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,479 @@
...
+  - The asset is not excluded through a `asset-cache.toml` file passed to `pack build`.
+  - If two assets provide the same `/cnb/assets/<asset-digest>` these two files must have identical contents. Decisions about rewriting these layers to optimize space are left to the platform.
+
+When creating a builder with asset caches:
+  - If any asset layer is a superset of another, only the superset layer is included in the builder.
+  - builders have a `io.buildpacks.buildpack.assets` Label containing entries for every asset included in the builder.
+
+
+### `lifecycle` changes
+Builders should now provide a `CNB_ASSETS` environment variable to the `build` phase`. This provides a standard variable buildpacks may use when looking up assets. The value of `CNB_ASSETS` will default to `/cnb/assets`.
```
matthewmcnew, line 451: Won't any alternative to `/cnb/assets` require rewriting the structure of an asset image? What is the usecase for this?

text/0000-offline-buildpackages.md
```diff
@@ -0,0 +1,479 @@
...
+```
+
+Buildpacks should then be able to quickly check for available assets using a `sha256` of the asset they wish to use in tandem with the `CNB_ASSETS` environment variable. The `sha256` value used to perform this lookup should be pulled from the relevant `[[asset]]` entry in `buildpack.toml`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+- Increases complexity for users/buildpack authors. This adds another artifact to manage and new code paths for checking for vendored assets.
```
matthewmcnew, line 472: It is worth noting that this will add considerable complexity for kpack users who now need to provide a buildpackage image and a asset image when looking to create an "offline" builder. Would it make sense in a future rfc to explore a buildpackage format that can include assets?

text/0000-offline-buildpackages.md
dwillist: This is more for convenience so buildpack authors have a environment variable they can refer to versus. It also gives us a level flexibility if we come up with a more efficient way to provide unarchived assets (which was a part of this RFC but was pulled out)

text/0000-offline-buildpackages.md
dwillist: Good point, This is duplicated data.

dwillist: Good point this will be duplicated data, we can remove.

text/0000-offline-buildpackages.md
dwillist: I'm hopeful the ability to construct an asset-cache from a buildpackage, and the ability to build offline builders from online buildpackages will ease this pain point.  But if not It might be worthwhile to think about merging buildpackages & assets.

dwillist:
Issues linked here:
buildpacks/spec#184
buildpacks/pack#1055
buildpacks/lifecycle#521

jkutner:
Merged as [RFC-0073](https://github.com/buildpacks/rfcs/blob/main/text/0073-offline-buildpackages.md)

abdennour:
offline builds are not only about private image registries.. but also app dependencies.
For example, maven should download/deploy from/to a private maven repo.
Think about like you are in offline env however you have access to Nexus Sonatype.

ekcasey:
@abdennour I think pulling app dependencies from a private repo should be the responsibility of a given buildpack and the CNB project already provides the tools buildpacks need to support this. Example: https://paketo.io/docs/buildpacks/language-family-buildpacks/java/#connecting-to-a-private-maven-repository

PR #83, opened 2020-05-30
RFC: Process specific env and profile.d

jabrown85:


hone:
thanks for picking this up.

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,67 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Each env or profile layer dir can have subdirs like `<layer>/env.launch/web/` and `<layer>/profile.d/web/`. These behave the same way as `<layer>/env` but are specific to the launch of the `web` process.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Lifecycle would use the identifier of the intended launch process to bring in additional `env` and `profile.d` files. The launcher would process these process specific values _after_ less specific values to allow [Environment Variable Modification Rules](https://github.com/buildpacks/spec/blob/master/buildpack.md#environment-variable-modification-rules) to apply in the context of a single buildpack.
```
jromero, line 42: 1. Does this proposal inherit the dot delimited suffix annotations? ex. `.default`, `.override`, `.etc`?
2. If it does, can we get more concrete examples on how the logic would be applied in cases where something like `.override` is present in the "non-process" specific but not specified in the process specific one?

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,67 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- Copied from [here](https://github.com/buildpacks/rfcs/pull/72) and updated to include additional details and examples.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should this idea be extended to `bin`, `lib`, and any other buildpack output intended for launch?
+- Should lifecycle ignore invalid entries like `<layers>/<layer>/env.build/web/NODE_OPTIONS`?
```
jromero, line 64: I'm failing to see what is invalid in this entry...

jabrown85: For this entry it specifies build and a process. Since env vars would be process specific this wouldn’t make since as you aren’t building a specific process.

text/0000-process-specific-env.md
jabrown85: I assume it would inherit. Let me think about it and add some examples.

text/0000-process-specific-env.md
jromero: Ahh, I see now.

text/0000-process-specific-env.md
jabrown85: I updated the section just now. Hopefully it makes more sense. I'm sure the wording could be better.

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,81 @@
...
```
hone, line 78: I wonder if it should just error?

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+- Copied from [here](https://github.com/buildpacks/rfcs/pull/72) and updated to include additional details and examples.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should this idea be extended to `bin`, `lib`, and any other buildpack output intended for launch?
```
hone, line 77: I'm not sure what the use case is? @jabrown85

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,81 @@
...
+
+```
+<layers>/<layer>/env/NODE_OPTIONS.append
+<layers>/<layer>/env.launch/NODE_OPTIONS.append
+<layers>/<layer>/env.launch/web/NODE_OPTIONS.append
+```
+
+Example spec update:
+
+- Environment variable file contents originating in the same layer MUST be sorted such that file contents in <layers>/<layer>/env.launch/<process>/ precede file contents in <layers>/<layer>/env.launch/, which must precede <layers>/<layer>/env/.
```
hone, line 54:
```suggestion
- Environment variable file contents originating in the same layer MUST be sorted such that file contents in `<layers>/<layer>/env.launch/<process>/` precede file contents in `<layers>/<layer>/env.launch/`, which must precede `<layers>/<layer>/env/`.

```

text/0000-process-specific-env.md
```diff
@@ -0,0 +1,81 @@
...
+- Environment variable file contents originating in the same layer MUST be sorted such that file contents in <layers>/<layer>/env/ precede file contents in <layers>/<layer>/env.build/ or <layers>/<layer>/env.launch/ which must preced file contents in <layers>/<layer>/env.launch/<process>/.
```
hone, line 55:
```suggestion
- Environment variable file contents originating in the same layer MUST be sorted such that file contents in <layers>/<layer>/env/ precede file contents in <layers>/<layer>/env.build/ or <layers>/<layer>/env.launch/ which must preced file contents in <layers>/<layer>/env.launch/<process>/.
```suggestion
- Environment variable file contents originating in the same layer MUST be sorted such that file contents in `<layers>/<layer>/env/` precede file contents in `<layers>/<layer>/env.build/` or `<layers>/<layer>/env.launch/` which must precede file contents in `<layers>/<layer>/env.launch/<process>/`.

```
```

File: text/0000-process-specific-env.md [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+- Copied from [here](https://github.com/buildpacks/rfcs/pull/72) and updated to include additional details and examples.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should this idea be extended to `bin`, `lib`, and any other buildpack output intended for launch?
+- Should lifecycle ignore invalid entries like `<layers>/<layer>/env.build/web/NODE_OPTIONS`?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
```
hone, line 81: Can you detail out some of the spec changes here?

text/0000-process-specific-env.md
jabrown85: A user might want to add a credential helper binary for `web` but not have that accessible for `debug`. They could put the binary at `bin/web/credential-helper` to achieve this if we decided to allow this.

hone: Is that a different image at that point?

jabrown85: I'm not sure. I think it's reasonable to think that perhaps a `web` process is restricted to having access to something in `bin/lib` that perhaps you wouldn't want to be immediately accessible for non-defined launcher executions `/cnb/lifecycle/launcher -- bash` for instance.

text/0000-process-specific-env.md
jabrown85: Done. I bolded the parts I think need to change.

text/0000-process-specific-env.md
hone: Yeah, I guess without changing the image it's still technically there. I'm happy to mark this as out of scope of the RFC unless you feel strongly about it.

text/0000-process-specific-env.md
hone: @ekcasey @sclevine can you review this?

PR #84, opened 2020-06-02
Launcher Arguments

nebhale:
This RFC proposes a change to the lifecycle execution behavior to allow arguments to be appended to buildpack-contributed process types.

[Rendered](https://github.com/buildpacks/rfcs/blob/arguments/text/0000-process-type-arguments.md)


ekcasey:
I really like this idea overall, but I think the big outstanding question here is the UX for `direct=false` processes. When a `direct=false` process is launched with a shell, the process arguments are actually arguments to `bash -c <cmd>` To make this work, each `direct=false` command  would need to explicitly account for the possibility of arguments to take advantage of this feature.

As a simple/contrived example, imagine "hello world" type process.

A directly executed version of this command would look like the following:
```
[[processes]]
  type = "hello"
  command = "echo"
  args = ["hello"]
  direct = true
```
If we implemented this RFC and I passed "world" to the launcher the command essentially becomes
```
[[processes]]
  type = "hello"
  command = "echo"
  args = ["hello", "world"]
  direct = true
```
what gets executed by the launcher is essentially just `echo hello world`. Everything works great.

BUT a `direct=false` hello process could easily look like:
```
[[processes]]
  type = "hello"
  command = "echo hello"
  direct = false
```
Passing the same "world" argument to the launcher would result in
```
[[processes]]
  type = "hello"
  command = "echo hello"
  args = ["world"]
  direct = false
```
and the launcher would essentially execute `bash -c "echo hello" "world"`. This doesn't work.
The command itself would have to be carefully constructed to accept bash arguments intuitively:
```
[[processes]]
  type = "hello"
  command = "echo $@"
  args = ["echo", "hello"]
  direct = false
```
AND even the above doesn't work in practice when read from a `config.toml` file. After experimenting with every permutation of quoting the only thing i could get to work is toml unicode escape sequences for the `$@` so it is evaluated at the correct time...
```
[[processes]]
type = "hello"
command = "echo \u0024\u0040"
args = ["echo", "hello"]
direct = false
EOF
```
 (this last part is probably to some quoting, toml parsing, nested bash complexity in the launcher implementation, that I don't fully understand, and might independently resolvable) 

If we fix quoting edge cases this UX might be okay? Folks that want to construct `direct=false` processes that accept args have a mechanism, but they will need to be careful and I imagine this could trick a lot of people.

nebhale:
I'd go with something significantly simpler, appending the arguments, space delimited to the command itself.

text/0000-arguments.md
```diff
@@ -0,0 +1,76 @@
...
+
+1.  0 args: execute the default process type
+1.  1 arg:  if `$1` matches a process type, execute that process type
+1.  N args: execute `$1` as the command with `${@:2}` as the arguments appended to that command
+
+This proposal is to augment that command resolution flow with a fourth step:
+
+1.  0 args: execute the default process type
+1.  N args, if `$1` matches a process type: execute that process type with `${@:2}` as the arguments appended to that process type
+1.  N args, if `$1` does not match a process type and does not match something on the resolved path: execute the default process type with `${@}` as the arguments appended to that process type
```
hone, line 43: There is technically a breaking change that previously if `$1` matched but you provided additional args it would not match the process type. This isn't a huge concern, but just wanted to note for completeness.

hone:
I'm :+1: on this idea as well, but pending resolving the `direct=false` case.

sclevine:
I'm strongly opposed to modifying the command automatically. It seems very confusing to provide discreet arguments that don't map to proper argument inputs. Could we just disallow `direct = false` and non-empty `args`?

My other concern is that overloading the first argument could lead to unexpected behavior. E.g., if an argument happens to match something on the path, the command would unexpected execute that binary instead.

Could we use environment variables to provide process-type selection and/or shell command overrides instead of trying to fit all of this into a single list?

natalieparellano:
@ekcasey in the example you provide:

```
[[processes]]
  type = "hello"
  command = "echo $@"
  args = ["echo", "hello"]
  direct = false
```

Do you mean to say `args = ["hello", "world"]`? Just trying to follow along here, want to make sure I'm not missing something.

sclevine:
Request for prior art: what other tools solve this problem?

jabrown85:
https://docs.npmjs.com/cli/run-script#description allows arguments via `--`. Looks like the parsing is described as platform dependent.

hone:
> Request for prior art: what other tools solve this problem?

[`npm run-script`](https://docs.npmjs.com/cli/run-script) does let you pass in extra args:
> As of npm@2.0.0, you can use custom arguments when executing scripts. The special option -- is used by getopt to delimit the end of the options. npm will pass all the arguments after the -- directly to your script:
>
```
npm run test -- --grep="pattern"
```
> The arguments will only be passed to the script specified after npm run and not to any pre or post script.

jabrown85:
+1 to this idea. I've been wanting to be able to define `npm install` task that takes arbitrary arguments for a development mode.

text/0000-arguments.md
nebhale: I concur with the spirit of this but I’m choosing (conveniently for me) to view this as changing unspecified behavior.  We’ve never defined what _should_ happen in that case so changing it, and more importantly specifying it, wouldn’t be a breaking change to the spec but rather the implementation.  Semantics, I know.

text/0000-arguments.md
```diff
@@ -0,0 +1,195 @@
...
+If `CNB_PROCESS_TYPE` is unset, it will continue to default to `web`.
+
+This functionality has existed for a long time, but we will remove the second way to set toggle the same behavior, by passing a single argument containing a process-type to the launcher (i.e. `docker run <image> my-process-type` won't work anymore).
+
+### Specifying a custom process at runtime
+If users wish to provide a custom command instead of using one of the buildpack generated process types, they must set `CNB_PROCESS_TYPE=override` to explicitly indicate that they wish to override the entire command, rather than to append args.
+
+Running `docker run --env CNB_PROCESS_TYPE=override <image> echo hello world`, will be equivalent to running a process type where `direct = false`, `command = "echo"` and `args = ["hello", "world"]`
+
+Running `docker run --env CNB_PROCESS_TYPE=override <image> -- echo hello world`, will be equivalent to running a process type where `direct = true`, `command = "echo"` and `args = ["hello", "world"]`
```
nebhale, line 86: This `--` indicates `direct = true` sort of pops up out of nowhere.  What's the background/prior art/justification for this syntax?

ekcasey: It is the current syntax for specifying a `direct=true` command at runtime

nebhale: Is that documented anywhere, or just implemented?

ekcasey: It's in the Buildpack API spec -https://github.com/buildpacks/spec/blob/master/buildpack.md#launch

imho the override behavior here belongs in the Platform API

text/0000-arguments.md
```diff
@@ -69,49 +69,54 @@ produce results equivalent to the execution of `"<CMD>" "<ARG1>" "<ARG2>"` in a
...
 
 ## Behavior
 ### Selecting a process type at runtime
-Users will no longer be able to pass a process-type to the launcher, instead they must set `CNB_PROCESS_TYPE` in the running container to select a process type.
+Users will no longer be able to pass a process-type to the launcher as a positional argument. Instead, they must set `CNB_PROCESS_TYPE` in the running container to select a process type.
 If `CNB_PROCESS_TYPE` is unset, it will continue to default to `web`.
 
-This functionality has existed for a long time, but we will remove the second way to set toggle the same behavior, by passing a single argument containing a process-type to the launcher (i.e. `docker run <image> my-process-type` won't work anymore).
+The `CNB_PROCESS_TYPE` functionality has existed for a long time and is the only specified way to select a process-type. We will simply remove the secondary, unspecified way to select a process-type by passing a single argument containing a process-type to the launcher (i.e. `docker run <image> my-process-type` won't work anymore).
```
nebhale:
```suggestion
The `CNB_PROCESS_TYPE` functionality has existed for a long time and is the only specified way to select a process-type. We will remove the secondary, unspecified way to select a process-type by passing a single argument containing a process-type to the launcher (i.e. `docker run <image> my-process-type` won't work anymore).
```

text/0000-arguments.md
```diff
@@ -120,16 +125,18 @@ The `launcher` currently has a three step command resolution flow:
 
 1.  0 args: execute the default process type, where the default process type is `${CNB_PROCESS_TYPE:-web}`
 1.  1 arg:  if `$1` matches a process type, execute that process type
-1.  N args: execute `$1` as the command with `${@:2}` as the arguments appended to that command
+1.  N args if `$1` is `--`: directly execute `$2` as the command with `${@:3}` as the arguments appended to that command
+1.  N args: execute `$1` as the command with `${@:2}` as the arguments appended to that command in a `bash` shell
 
-This proposal is to augment that command resolution flow to:
+This proposal change the command resolution flow to:
```
nebhale:
```suggestion
This proposal changes the command resolution flow to:
```

text/0000-arguments.md
```diff
@@ -120,16 +125,18 @@ The `launcher` currently has a three step command resolution flow:
...
 
 1.  If `CNB_PROCESS_TYPE=override`
     1. 0 args: fail
-    1. N args: execute `$1` as the command with `${@:2}` as the arguments
+    1.  N args if `$1` is `--`: directly execute `$2` as the command with `${@:3}` as the arguments appended to that command
+    1.  N args: execute `$1` as the command with `${@:2}` as the arguments appended to that command in a `bash` shell
 1.  Else
     1. 0 args: execute the process type `${CNB_PROCESS_TYPE:-web}`
-    1. N args: Execute process type `${CNB_PROCESS_TYPE:-web}`, after appending `"$@"` to the arguments
+    1. N args: Execute process type `${CNB_PROCESS_TYPE:-web}`, after appending `$@` (all positional arguments) to the process arguments
```
nebhale: Should the `"` be removed from this?  Ignoring, for a moment, the actual implementation, is your point that with this system, property tokenization is already handled and it's no longer a concern that each argument is properly quoted for tokenization?

text/0000-arguments.md
```diff
@@ -138,31 +145,32 @@ Ensuring all arguments are tokenized and parsed correctly requires some `bash` c
 'source <app-dir>/.profile
 source <layers>/<bp.id>/profile.d/<profile-script>
 ...
-exec bash -c '$(eval echo \"$0\") ...$(eval echo \"$n\")' "$@"' /cnb/lifecycle/launcher <cmd> <args>...
+exec bash -c '$(eval echo \"$0\") ...$(eval echo \"$n\")' "$@"' /CNB/lifecycle/launcher <cmd> <args>...
```
nebhale:
```suggestion
exec bash -c '$(eval echo \"$0\") ...$(eval echo \"$n\")' "$@"' /cnb/lifecycle/launcher <cmd> <args>...
```

text/0000-arguments.md
yaelharel: Why we don't allow the `--` option as we do when `CNB_PROCESS_TYPE=override`?

text/0000-arguments.md
```diff
@@ -69,49 +69,54 @@ produce results equivalent to the execution of `"<CMD>" "<ARG1>" "<ARG2>"` in a
 ## Philosophy
 The goal is to make the runtime usage of a CNB container as similar as possible to the runtime usage on a non-CNB container, without removing functionality or violating assumptions that have been firmly established in the buildpack community (e.g. defaulting to `web`).
 
-Because the `launcher` must occupy the container entrypoint we will instead use `CNB_PROCESS_TYPE` as a CNB-specific analog to the entrypoint. Users can already set `CNB_PROCESS_TYPE` to toggle between processes. Setting `CNB_PROCESS_TYPE=override` will be analogous to clearing the entrypoint.
+Because the `launcher` must occupy the container `entrypoint` we will instead use `CNB_PROCESS_TYPE` as a CNB-specific analog to the `entrypoint`. Users can already set `CNB_PROCESS_TYPE` to toggle between processes. Setting `CNB_PROCESS_TYPE=override` will be analogous to clearing the `entrypoint`.
```
yaelharel: I think, there should be "," after the first `entrypoint`.

text/0000-arguments.md
```diff
@@ -138,31 +145,32 @@ Ensuring all arguments are tokenized and parsed correctly requires some `bash` c
...
 
-For the changes proposed in this RFC, changing the default interpretation of `launcher` arguments from a custom command to additional arguments, will likely have the biggest impact on platforms. Platforms will need to make changes to runtime logic and/or documentation before enabling the platform API containing this change.
+Of the changes proposed in this RFC, changing the default interpretation of `launcher` arguments (from a custom command to additional arguments), will likely have the biggest impact on platforms. Platforms will need to make changes to runtime logic and/or documentation before enabling the platform API containing this change.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 * Breaking changes!
-* Some functionality is only accessible by setting environment variables, when previously all functionality was available with launcher arguments.
+* Some functionality is only accessible by setting environment variables, when, previously, all functionality was available via the command line.
```
yaelharel: I would break it to 2 sentences:
Some functionality is only accessible by setting environment variables. Previously, all functionality was available via the command line.

text/0000-arguments.md
ekcasey: guess my search/replace was a little too aggressive!

text/0000-arguments.md
ekcasey: I just wanted a shorthand for all args. Adding the quotes seems to imply something about the implementation. Maybe we should just skip the bash shorthand in this section all together? It potentially get's confusing in context.

ekcasey: When you are appending arguments to an existing process-type `direct=true` vs. `direct=false` is already defined for the process. It wouldn't be safe to change the value of `direct` for a buildpack-provided process.

ekcasey: I am trying to describe appending to an array of strings here and with the quotes it seems to imply something else? But I'll defer to my readers

nebhale: Having the quotes is a specific bash-ism that implies that the shell will properly quote everything so a later tokenization gets the same collection.  Without it, arguments with spaces would be re-tokenized as multiple arguments.  So the reason for me originally adding that syntax is to indicate that we would do the appropriate things necessary to ensure that the original tokenization would be preserved through out.  It may not be necessary any longer, if you believe the proposal already accommodates that behavior before describing it here.

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,207 @@
...
+[drawbacks]: #drawbacks
+
+* Breaking changes!
+* Some functionality is only accessible by setting environment variables. Previously, all functionality was available via the command line.
+* Increased complexity of launcher shell logic.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Previously in Cloud Foundry which did not support a system like this, the solution that was chosen was to use an environment variable to represent arguments:
```
hone, line 179: I think you meant did here.

```suggestion
Previously in Cloud Foundry which did support a system like this, the solution that was chosen was to use an environment variable to represent arguments:
```

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,207 @@
...
+### Specifying a custom process at runtime
+If users wish to provide a custom command instead of using one of the buildpack provided process types, they must set `CNB_PROCESS_TYPE=override`. `override` is a special keyword indicating that the entire process shall be defined by the user.
+
+If `CNB_PROCESS_TYPE` is set to `override`, the `launcher` will construct a process from the positional arguments it receives. If the first argument is `--` this will continue to signify a `direct` process. The first (when it is anything other than `--`) or second (if the first is `--`) argument provided to the `launcher`  will become the `command` of the resultant process. Any subsequent arguments to the launcher will become `args` of the resultant process.
+
+Example:
+
+Running `docker run --env CNB_PROCESS_TYPE=override <image> echo hello world`, will be equivalent to running a process type where `direct = false`, `command = "echo"` and `args = ["hello", "world"]`
+
+Running `docker run --env CNB_PROCESS_TYPE=override <image> -- echo hello world`, will be equivalent to running a process type where `direct = true`, `command = "echo"` and `args = ["hello", "world"]`
```
jkutner, line 90: What the behavior when I run `docker run -it <image> bash`? An error that no process type is found for "bash"?

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,207 @@
...
+* shell process - a process with `direct=false`
+
+## Philosophy
+The goal is to make the runtime usage of a CNB container as similar as possible to the runtime usage on a non-CNB container, without removing functionality or violating assumptions that have been firmly established in the buildpack community (e.g. defaulting to `web`).
+
+Because the `launcher` must occupy the container `entrypoint`, we will instead use `CNB_PROCESS_TYPE` as a CNB-specific analog to the `entrypoint`. Users can already set `CNB_PROCESS_TYPE` to toggle between processes. Setting `CNB_PROCESS_TYPE=override` will be analogous to clearing the `entrypoint`.
+
+## Behavior
+### Selecting a process type at runtime
+Users will no longer be able to pass a process-type to the launcher as a positional argument. Instead, they must set `CNB_PROCESS_TYPE` in the running container to select a process type.
```
jkutner, line 76: I would rather see us *require* the process name in the CLI. For an image with a `web` process only, this would yield the following cases:

* `docker run <image>` - works as it does today
* `docker run <image> web` - works as it does today
* `docker run <image> web asdf` - passes `["asdf"]` as args
* `docker run <image> bash` - fails to locate a process named "bash" and sets `command = "bash"` instead
* `docker run <image> -- bash` - fails to locate a process named "bash" and sets `command = "bash"` and `direct = true`

ekcasey: If we went this route I don't think process type lookup should be attempted if the first arg is `--`. `--` means `direct` and we should not change the value of `direct` for a buildpack provided process. If `--` is provided it is always a custom command

sclevine: I feel pretty strongly that this makes our images less compliant / normal compared to images built with Dockerfiles. Most app/service images I can find allow you to provide args. K8s calls the field `args`, and the assumption that entrypoint == `command` is built-in to k8s: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

I feel like nearly every K8s user is going to be confused by this.

jkutner: But the args we're talking about aren't args to the launcher. They are args to the process it's run. I feel like it makes sense to always include the process name in what we pass to launcher.

This is the way it works today. Do we have feedback today that it confusing k8s users?

hone:
As part of the WG discussion today @ekcasey is going to do another rewrite of this idea that creates multicall binaries for each process type that can be set as `entrypoint` which will allow them to have args appended.

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Philosophy
+The goal is to make the runtime usage of a CNB container as similar as possible to the runtime usage on a non-CNB container, without removing functionality or violating assumptions that have been firmly established in the buildpack community (e.g. defaulting to `web`).
+
+Because the `launcher` must occupy the container `entrypoint`, we will make the `launcher` a multicall binary that derives the intended process-type, or lack there of, from `argv0`. This allows for intuitive use of both `entrypoint` and `cmd`: `entrypoint` will be used to select a process and `cmd` will be used to supply additional argument to that process.
+
+## Behavior
+
+### Exporter
+When `exporter` creates an app image it will create one symlink per buildpack-contributed process type. The symlink files will be named `/cnb/process/<process-type>` and will point at the launcher.
```
jabrown85, line 77: I couldn't find anything in the spec that says `[[process]] type = <process-type>` has to be a certain format. Is that an implementation detail somewhere perhaps?

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,221 @@
...
+
+The `exporter` will prepend `/cnb/process` to the image `PATH`.
```
jabrown85, line 79: I think we discussed that this `PATH` change would be restricted to starting the image and stripped from the `env` of the actual process.

text/0000-process-type-arguments.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Selecting a process type at runtime
+Users will no longer be able to pass a process-type to the launcher as a positional argument or with the `CNB_PROCESS_TYPE` environment variable. Instead, they must set the `entrypoint` in the running container to select a process type other than the default.
+
+For Example:
+`docker run --entrypoint web <image>`
+`docker run --entrypoint /cnb/process/web <image>`
+Will both launch the process with `type` `web`
+
+Buildpack-contributed processes will be forbidden from having `type` equal to `"launcher"`.
```
jabrown85, line 92: Are we saying that buildpack authors will have no way of getting the `--entrypoint` to be `launcher` if any other buildpack defined a process?

ekcasey: We don't want to allow ambiguity in `$0` between a process with type `launcher` and the normal `launcher`

text/0000-process-type-arguments.md
ekcasey: yep, I'll update that

nebhale:
Final Comment Period with merge disposition, closing on 26 June, 2020.

dfreilich:
Simple question, but to make sure I understand:
With this RFC, if I would build an image of `pack` using `pack` (e.g. `pack build pack -B gcr.io/paketo-buildpacks/builder:tiny -p .`), I would be able to use it just by calling 
```
docker run --rm -it pack version
````
rather than what we currently have to do (`docker run --rm -it pack -- pack version`)?

jabrown85:
> Simple question, but to make sure I understand:
> With this RFC, if I would build an image of `pack` using `pack` (e.g. `pack build pack -B gcr.io/paketo-buildpacks/builder:tiny -p .`), I would be able to use it just by calling
> 
> ```
> docker run --rm -it pack version
> ```
> 
> rather than what we currently have to do (`docker run --rm -it pack -- pack version`)?

Assuming the built image has a `pack` process type, yes. The `pack` launcher process would become the entrypoint and the `version` would be the argument.

jromero:
@ekcasey / @nebhale can we rename the PR title and the RFC name to reflect that this is talking "image launch flags"

text/0000-process-type-arguments.md
ekcasey: I am going to resolve this b/c the RFC has changed considerably since this question was asked.

PR #85, opened 2020-06-03
Read/Write Platform Volume Mount

nebhale:
This change proposes a well-known location in /platform for mounting read/write filesystems.

### DIFF
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -16,7 +16,7 @@ The `pack` CLI should expand its volume mounting functionality to allow arbitrar
 # Motivation
 [motivation]: #motivation
 
-Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisifies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downlaoded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
+Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisfies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downloaded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
 
 This benefit is generally useful and extends beyond build caches to nearly every aspect of buildpack usage including development of both buildpacks and the lifecycle.
 
@@ -35,7 +35,7 @@ The technical details are opaque and implementation dependent as this is a chang
 
 * There is an amazing amount of danger in using a shared filesystem in distributed systems.
 * Performance problems accessing shared filesystems in distributed systems.
-* Performance problems accessing a local filessystem from within the MacOS Docker Daemon.
+* Performance problems accessing a local filesystem from within the MacOS Docker Daemon.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -52,7 +52,7 @@ The technical details are opaque and implementation dependent as this is a chang
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-* Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their lcoations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that signficantly.
+* Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their locations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that significantly.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -36,6 +36,7 @@ The technical details are opaque and implementation dependent as this is a chang
 * There is an amazing amount of danger in using a shared filesystem in distributed systems.
 * Performance problems accessing shared filesystems in distributed systems.
 * Performance problems accessing a local filesystem from within the MacOS Docker Daemon.
+* There are signficant risks around file ownership on Linux hosts
 
 # Alternatives
 [alternatives]: #alternatives
@@ -53,6 +54,7 @@ The technical details are opaque and implementation dependent as this is a chang
 [unresolved-questions]: #unresolved-questions
 
 * Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their locations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that significantly.
+* Should this be included behind experimental support of the `pack` CLI?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -25,6 +25,13 @@ This benefit is generally useful and extends beyond build caches to nearly every
 
 Currently, the `pack` CLI has a `--volume` flag that allows users to expose a local filesystem location as a read-only volume mount into the `/platform` directory.  This change should generalize that flag and allow it to mount a local filesystem location as a read-write volume mount into any location within the build container.  If a user chooses to mount at any specification-reserved files system location (e.g. `/cnb` or `/layers`), a warning should be displayed.
 
+`--volume` will allow `mode` as an optional third colon separated parameter.
+
+Format: `--volume=<src>:<target>[:mode]`
+
+Mode values:
+* `ro` - Read-only (default)
+* `rw` - Read/Write
 # How it Works
 [how-it-works]: #how-it-works
 
@@ -36,7 +43,8 @@ The technical details are opaque and implementation dependent as this is a chang
 * There is an amazing amount of danger in using a shared filesystem in distributed systems.
 * Performance problems accessing shared filesystems in distributed systems.
 * Performance problems accessing a local filesystem from within the MacOS Docker Daemon.
-* There are signficant risks around file ownership on Linux hosts
+* There are significant risks around file ownership on Linux hosts
+* There would be a small transition period of incompatibility for users that already use `--volume` for mounting platform directories.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -50,7 +50,7 @@ A folder named `/platform/unsafe` should be reserved by specification.  Within t
 # Prior Art
 [prior-art]: #prior-art
 
-No relavant priort art.
+Currently, arbitrary read-only volumes can be mounted under `/platform`.  The outcome of this RFC would be similar with looser permissions and more risk.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -1,6 +1,6 @@
 # Meta
 [meta]: #meta
-- Name: Read/Write Platform Volume Mount
+- Name: Read-Write Volume Mount in Pack
 - Start Date: 2020-06-02
 - Author(s): nebhale
 - RFC Pull Request: (leave blank)
@@ -11,27 +11,29 @@
 # Summary
 [summary]: #summary
 
-Buildpacks should have access to a read/write file system location provided by the platform.
+The `pack` CLI should expand its volume mounting functionality to allow arbitrary read-write mounting.
 
 # Motivation
 [motivation]: #motivation
 
-Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisifies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downlaoded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into a safe location within the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
+Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisifies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downlaoded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
+
+This benefit is generally useful and extends beyond build caches to nearly every aspect of buildpack usage including development of both buildpacks and the lifecycle.
 
 # What it is
 [what-it-is]: #what-it-is
 
-A folder named `/platform/unsafe` should be reserved by specification.  Within that folder, an arbitrary collection of directories could be mounted (e.g. `/platform/unsafe/{gradle,maven}`).  Each of the mounts at those specified children would be ready/write instead of read only as all other mounts are configured.  Buildpacks have complete access during both `detect` and `build` and no other enforcment is required.
+Currently, the `pack` CLI has a `--volume` flag that allows users to expose a local filesystem location as a read-only volume mount into the `/platform` directory.  This change should generalize that flag and allow it to mount a local filesystem location as a read-write volume mount into any location within the build container.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-The technical details are platform implementation dependent as this is a change to the specification.
+The technical details are opaque and implementation dependent as this is a change to the behavior of a `pack`-specific flag.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-* There is an amazing amount of danger in using a shared filessystem in distributed systems.
+* There is an amazing amount of danger in using a shared filesystem in distributed systems.
 * Performance problems accessing shared filesystems in distributed systems.
 * Performance problems accessing a local filessystem from within the MacOS Docker Daemon.
 
@@ -44,15 +46,15 @@ The technical details are platform implementation dependent as this is a change 
 # Prior Art
 [prior-art]: #prior-art
 
-Currently, arbitrary read-only volumes can be mounted under `/platform`.  The outcome of this RFC would be similar with looser permissions and more risk.
+* Currently, arbitrary read-only volumes can be mounted under `/platform`.  The outcome of this RFC would be similar with looser permissions and more risk.
+* The `docker run` command presents a good model, both in behavior and syntax for the analogous `pack` CLI `--volume` flag.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-* Should the folder actually be called `/platform/danger-zone`?
+* Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their lcoations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that signficantly.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
-* The buildpack specification will need to describe a reserved folder in `/platform` where read/write filesystems can be mounted.
-* The platform specification will need to describe how to communicate to the `lifecycle` what folders to mount.
+This is a `pack`-specific feature, so no specification will need to be changed.
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -21,18 +21,12 @@ Currently, every part of the filesystem available during `detect` and every part
 # What it is
 [what-it-is]: #what-it-is
 
-This provides a high level overview of the feature.
-
-- Define any new terminology.
-- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
-- Explaining the feature largely in terms of examples.
-- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
-- If applicable, describe the differences between teaching this to existing users and new users.
+A folder named `/platform/unsafe` should be reserved by specification.  Within that folder, an arbitrary collection of directories could be mounted (e.g. `/platform/unsafe/{gradle,maven}`).  Each of the mounts at those specified children would be ready/write instead of read only as all other mounts are configured.  Buildpacks have complete access during both `detect` and `build` and no other enforcment is required.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-A folder named `/platform/unsafe` should be reserved by specification.  Within that folder, an arbitrary collection of directories could be mounted (e.g. `/platform/unsafe/{gradle,maven}`).  Each of the mounts at those specified children would be ready/write instead of read only as all other mounts are configured.  Buildpacks have complete access during both `detect` and `build` and no other enforcment is required.
+The technical details are platform implementation dependent as this is a change to the specification.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-danger-zone.md
+++ b/text/0000-danger-zone.md
@@ -1,0 +1,64 @@
+# Meta
+[meta]: #meta
+- Name: Read/Write Platform Volume Mount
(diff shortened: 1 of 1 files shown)


text/0000-danger-zone.md
```diff
@@ -0,0 +1,64 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+* There don't appear to be any viable strategies for exposing a local filesystem to the build container.
+* Not doing this leaves things with the status quo which isn't horrible.
+
+# Prior Art
+[prior-art]: #prior-art
+
+No relavant priort art.
```
natalieparellano, line 53: This is probably tangential to the discussion, but it might be worth noting that today `pack` allows mounting arbitrary **read only** volumes under `/platform`.

nebhale: Happy to add it.

text/0000-danger-zone.md
```diff
@@ -0,0 +1,64 @@
...
+[how-it-works]: #how-it-works
+
+A folder named `/platform/unsafe` should be reserved by specification.  Within that folder, an arbitrary collection of directories could be mounted (e.g. `/platform/unsafe/{gradle,maven}`).  Each of the mounts at those specified children would be ready/write instead of read only as all other mounts are configured.  Buildpacks have complete access during both `detect` and `build` and no other enforcment is required.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* There is an amazing amount of danger in using a shared filessystem in distributed systems.
+* Performance problems accessing shared filesystems in distributed systems.
+* Performance problems accessing a local filessystem from within the MacOS Docker Daemon.
```
ekcasey, line 42: I confirmed with a hacked together POC that despite the performance drawbacks of bind mounts vs. regular volumes, even a partially prepopulated .m2 cache on a bind mount from my MacOs host results in significantly faster build times vs. downloading the dependencies. Your milage may very based on individual disk performance and network speeds but the difference was dramatic enough that I imagine it would be considerably faster in the vast majority of cases.

text/0000-danger-zone.md
```diff
@@ -0,0 +1,64 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
```
ekcasey, line 30: did you miss this section?

text/0000-danger-zone.md
ekcasey: For context, this one-off test was done building petclinic from source. A couple newer libraries were downloaded, but most of largest dependencies existed in the bind mounted cache already.

text/0000-danger-zone.md
```diff
@@ -0,0 +1,64 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+A folder named `/platform/unsafe` should be reserved by specification.  Within that folder, an arbitrary collection of directories could be mounted (e.g. `/platform/unsafe/{gradle,maven}`).  Each of the mounts at those specified children would be ready/write instead of read only as all other mounts are configured.  Buildpacks have complete access during both `detect` and `build` and no other enforcment is required.
+
```
jromero, line 36: The motivation makes sense. I'm trying to wrap my mind around how this directory/mount would be shared across multiple builds. I've written down a few thought experiments. Could you validate whether I'm thinking about this right?

### pack (local mount share)

In the case of something like `pack` it could make sense to **provide the same `unsafe` mount for every build**. This would enable the use case of: As a java developer with app A and app B, I don't want to have to download the same JDK or dependencies across builds.

Are there any concerns about sharing the same `unsafe` mount across multiple _very_ unrelated builds like a power dev that builds X amount of languages, uses various buildpacks/builders vendors? Or even more probable, using `pack` in a centralized CI system host like Jenkins.

### multi-tenant

For other multi-tenant platforms, they could theoretically share the same `unsafe` mount via additional configuration (such as specifying what additional directories inside `unsafe` to share) or a higher level mechanism (such as per "app", per "team", etc).

text/0000-danger-zone.md
nebhale: Apparently writing this while on a call wasn't great.

text/0000-danger-zone.md
nebhale: @jromero

### `pack`

This wouldn't apply to dependencies that are included in buildpacks like the JDK addressed via #81.  Those have very different sharing and caching strategies (among other things, they're much safer to cache and share, and perform better when sharing is available) from what is being addressed by this PR, so you should focus on the downloading of application-specific dependencies.  The key use-case is reusing a common `.m2/repository` with tons of dependencies already downloaded in it.  This would be across unrelated single-language builds but typically not across languages.  It would also enable the use of locally-built snapshots of dependencies without needing to publish them to another location available to the build container.

Other than the same performance concerns of local mounting on MacOS generally, there are no concerns about mounting something broader than just a download cache.  You _could_ mount `$HOME` in there and while it wouldn't be intuitively useful (for example a `paketo-buildpacks/maven` would only be interested in seeing `/platform/unsafe/maven`), a custom buildpack built by a user could work to its heart's content.

### Multi-tenant

In the olden days (👴🏻) Jenkins used to run multiple builds, often simultaneously, on the same VM.  In these cases there *was* a shared `~/.m2/repository` and it **SUCKED**.  Concurrent builds against file systems where the build system assumed exclusive access is a recipe for failure.  To this day, it's unlikely that newer languages (like say Go) would hand this, and the necessary lag of going to an NFS, very well.  This was one of the foundational drivers behind the switch to building in containers across the entire industry.  So I don't see this is a feature for co-mingling multi-tenant builds as we'd re-introduce many of the same issues that containerization eliminated.

I'd argue that the `cache = true` layer for downloaded application dependencies in a CI system **is** sufficient in a way that it's not for local-development with something like `pack`.  In that case, the application is built slowly once, but then happens in an asynchronous way (triggered by a commit while you're off getting coffee), and you don't have the slowness of the first and the quickness of subsequent ones staring you in the face.  In addition, there is much more continuity between CI'd builds (i.e. you always build the same images over and over) leading to cache hits, rather than local development where you give images new, unique names all the time and the `cache = true` misses most of the time because it was for another "app" (image).

sclevine:
Alternative for consideration: what if we added arbitrary read/write volume mounts to pack via an explicit `-v` command line argument. E.g., you could mount your local `~/.m2/` directory to `~/.m2/` in the container. We could output a warning when users use this flag.

If we're going to introduce a security/interoperability risk, why not make the user introduce this risk explicitly?

This feature could be used for many out-of-spec things, but I'm starting to feel like we'll end up fighting users who know what they're doing in many cases. (e.g., mounting the ssh-agent socket)

nebhale:
@sclevine it’s a reasonable alternative, but is single-platform specific.  It is the platform where it most matters, admittedly, but if we think that it is a behavior that is likely to be adopted by other platforms, should it be more structured?

sclevine:
Other platforms could implement the same feature (e.g., it would be easy to implement with k8s PVCs).

I'm not trying to restrict the feature to pack -- just trying to avoid contractualizing and hiding something that is unsafe. Supporting an unsafe feature generally means trusting that users of the feature know what they're doing. Making the owner of the individual image (or platform) specify the volume mount explicitly seems safer than hiding this feature behind the buildpack.

ekcasey:
> Alternative for consideration: what if we added arbitrary read/write volume mounts to pack via an explicit -v command line argument. E.g., you could mount your local ~/.m2/ directory to ~/.m2/ in the container. We could output a warning when users use this flag.

I worry that we can't always do the most efficient thing if the unsafe mounts aren't in a common predictable place where buildpacks can find them and change their behavior accordingly (without making a lot of brittle assumptions). For example, the paketo maven buildpack creates a cache layer to store the `.m2` cache and then creates a symlink from `$HOME/.m2` to that layer. If the buildpack could find a `maven` cache in `platform/unsafe` it can change it's behavior.

If we went with your suggestion we could make the prior existence of `$HOME/./m2` a conditional that turns off the creation/linking of the cache layer, but I wonder if that type of logic is predictable enough or possible in all relevant cases.

I also worry about folks mounting into `/cnb` `/layers` or other unsafe locations.

chrono2002:
> Alternative for consideration: what if we added arbitrary read/write volume mounts to pack via an explicit `-v` command line argument. E.g., you could mount your local `~/.m2/` directory to `~/.m2/` in the container. We could output a warning when users use this flag.

That tarring/untarring circle Javier talked about is a disaster. Please provide a way to mount cache dirs like buildkit do.

nebhale:
@jromero Can you please re-review this?

text/0000-danger-zone.md
```diff
@@ -11,27 +11,29 @@
...
+Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisifies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downlaoded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
```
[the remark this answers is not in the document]
yaelharel: Thank you!

ekcasey:
> Writing files back to the volume could lead to something not accessible by the host user. 

This is a good point, when I was playing with this I did not check the ownership of the files written back to the host. I suspect it would create problems on Mac as well as linux. In general I think we should expose the full `--volume` syntax provided by docker, but that won't let users address use every use case. It will have problems for the same reasons volume permission generally cause problems in docker.

`.m2` could be handled with a normal r/w volume (not a bind mount) that is shared across builds and perhaps pre-populated by the user.

The ssh agent case is actually dicier imho. The CNB user likely won't be able to access the socket b/c of ownership (for example, when connecting to a docker daemon socket inside the container the lifecycle actually has to run and connect as root before dropping privileges). Mounting a socket is a common workflow in containers that run as `root` but running as an unprivileged user breaks.

I am not sure it is in own power to fix volume permission problems that docker itself can't fix (particularly in containers with a non-root user).

Theoretically `userns-remap` could potentially be used to address some of these issues...but I haven't never tried it and don't know what all the limitations are.

text/0000-danger-zone.md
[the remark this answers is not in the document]
dfreilich: This was helpful for me as well, thanks Ben!

text/0000-danger-zone.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisifies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downlaoded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
```
dfreilich, line 19:
```suggestion
Currently, every part of the filesystem available during `detect` and every part of the filesystem other than layers during `build` are read-only.  This satisfies most use-cases, but specifically does not support the possibility of multiple builds sharing a communal filesystem location.  While this would likely to be a disaster in many situations, especially in production environments, being able to share a filesystem between a development laptop and a build is _very_ useful.  For example, the single longest bit of building a Java application from source is the initial population of a Maven or Gradle cache.  Hundreds and perhaps thousands of artifacts are downloaded to do the very first build and subsequently never need to be downloaded again.  Caching these artifacts in a `cache = true` layer helps, but doesn't solve the first build speed problem and results in a pretty poor demo experience.  If a user could mount their `~/.m2` or `~/.gradle` folders into the build container, accepting the overall risk of such a choice, the experience would be vastly superior.
```
Fix spelling mistakes

text/0000-danger-zone.md
```diff
@@ -0,0 +1,60 @@
...
+[how-it-works]: #how-it-works
+
+The technical details are opaque and implementation dependent as this is a change to the behavior of a `pack`-specific flag.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* There is an amazing amount of danger in using a shared filesystem in distributed systems.
+* Performance problems accessing shared filesystems in distributed systems.
+* Performance problems accessing a local filessystem from within the MacOS Docker Daemon.
```
dfreilich, line 38:
```suggestion
* Performance problems accessing a local filesystem from within the MacOS Docker Daemon.
```

text/0000-danger-zone.md
```diff
@@ -0,0 +1,60 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+* Currently, arbitrary read-only volumes can be mounted under `/platform`.  The outcome of this RFC would be similar with looser permissions and more risk.
+* The `docker run` command presents a good model, both in behavior and syntax for the analogous `pack` CLI `--volume` flag.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+* Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their lcoations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that signficantly.
```
dfreilich, line 55:
```suggestion
* Should certain CNB-reserved folders be mountable?  I believe that preventing mounts at their locations restricts the upside to this feature, without adding a significant amount of safety.  The overall feature is quite dangerous and special-casing two folders isn't going to change that significantly.
```

nebhale:
@jabrown85 @ekcasey investigated how much of a difference it makes, even on slow systems like macOS and [it is significant](https://github.com/buildpacks/rfcs/pull/85#discussion_r434704102).

In addition, the problem with a totally controlled volume is that it precludes the contribution of locally built artifacts that are not hosted at some network location.  For example, I'm working on a change to Spring Boot. Outside of the container, I'd have a Spring Boot application that pointed at `2.3.2.BUILD-SNAPSHOT`.  I could `./gradlew publishMavenPublicationToMavenLocal` from `spring-projects/spring-boot` and then go over to my test application to `./gradlew clean build` and get the new dependency for testing.

Without a mount to my local file system, I cannot do this same behavior.  I'd have to publish this to some network-available location so that the build inside my container could access it.  And network-available location probably means a commit, and on and on.

Note that this same behavior exists in Go with the `mod` `replace` directive and I'd bet NodeJS has something similar.

jabrown85:
> In addition, the problem with a totally controlled volume is that it precludes the contribution of locally built artifacts that are not hosted at some network location. For example, I'm working on a change to Spring Boot. Outside of the container, I'd have a Spring Boot application that pointed at `2.3.2.BUILD-SNAPSHOT`. I could `./gradlew publishMavenPublicationToMavenLocal` from `spring-projects/spring-boot` and then go over to my test application to `./gradlew clean build` and get the new dependency for testing.

We could do this without bind mount, but it would involve copying files into a totally controlled volume at container build time. It would also probably involve relying on something rsync-like to make it speedier between subsequent builds that use that same mount.

I still worry about all the permission problems we may be expected to chase if we use bind mounts. Our experience on linux recently has made bind mounts a scary proposition to help maintain. The corp machines we have don't use user 1000 and this means that basically nothing can be accessed by a pack build container if the command isn't ran as root and `chmod` to the current host linux user. This also, of course, modifies the files on the host system and that can also be confusing and potentially break things. You can't also safely do this just once, since the host can constantly be modifying and writing files. I think @hone poked at `userns-remap` but found it not to be useful for his particular situation involving a bind mount.

jabrown85:
It's probably worth pointing out that [buildkit has a few mount options](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#experimental-syntaxes) that are similar to what is being discussed.

`RUN --mount=type=cache,target=/root/.cache/go-build go build ...`

This sort of syntax doesn't appear to have a way to load this new buildkit controlled volume from the host machine, however. They do also have an explicit `RUN --mount=type=ssh`, which I thought was interesting.

nebhale:
> We could do this without bind mount, but it would involve copying files into a totally controlled volume at container build time. 

Using my current `~/.m2` as an anecdote, I've got 6.5G of data which might not be a huge slow down, but spread out across 82,853 files.  `rsync`ing that into a controlled container would be significantly slower than just downloading all of the necessary build artifacts the first time you build an image; e.g. it'd take longer than the ~4 minutes my "first builds" take.  It not only doesn't improve the situation from where we are now, it likely is worse than the status quo.

I want to point out that the latest incarnation of this RFC makes this change a `pack`-specific feature.  Heroku and `kpack` **should** not support, and would not be required to support, this behavior.

text/0000-danger-zone.md
natalieparellano: This helps a lot to clarify, thank you!

jabrown85:
This being `pack` specific is 👍  

Writing files as one user from host to `~/.m2`, referencing those files during build that aren't owned by the `CNB_USER` as the `CNB_USER`, and also writing/updating files as `CNB_USER` will almost certainly make these directories _weird_ if not entirely broken.

I think `pack` will have to do something to normalize the permissions for these directories. Like `chown -R` for the builder user, as root. Otherwise we have to be ok with linux users that want this mount behavior having to potentially set permissions on the bind mounted folders to the `CNB_USER` in the builder(s) they expect to use the volume with. That or perhaps guiding them to `userns-remap` if it works out?

ekcasey:
**Speed**:
> We could do this without bind mount, but it would involve copying files into a totally controlled volume at container build time. It would also probably involve relying on something rsync-like to make it speedier between subsequent builds that use that same mount.

I don't think `pack` should support copying into a volume or any complicated `rsyncing`, users can do this in advance of course.

If we support the full docker `--volume` syntax users can set `consistency` to tweak the performance of their bind mounts on docker for mac (with `consistency=delegated` you can essentially make the host fs eventually consistent with the container instead of paying the costs of consistency on each write)

Even w/o configuring `consistency`, using the bind mount sped up my normal use cases considerably vs. a full download and was less of a headache than copying my `.m2` cache into a volume. If we give folks they full spectrum of options they can make those tradeoffs for themselves

**Ownership Issues**:
After playing with bind mounts using docker-desktop on Mac I observed no permission issues when bind mounting. It seems docker desktop does some nice magic for us here. Files were owned by the container user within the container and owned by the host user on the host. In the past I observed that bind mounted files had permission issues when files in bind mounted dirs did not. But I actually can't reproduce that with the latest version of docker desktop. So we might be good here.

Given that this works very nicely on Mac and `pack` has many mac users I think this feature is worth supporting.

Linux does complicate the picture a bit. Bind mounts on linux have a potential to create file ownership problems. If we are worried we could disable bind mount on linux in particular? Personally I am not inclined to do this. The limitations here are the same for `docker run --volume` and folks can use or not use it as they see fit.

Maybe some of the linux permissions problems here are a point in favor of @sclevine 's opinion that CI systems should just run the lifecycle directly, instead of orchestrating with `pack`.

**TLDR**: I think this is worth doing despite the limitations

jabrown85:
> **TLDR**: I think this is worth doing despite the limitations

I do agree with this. I think this can greatly speed up a lot of very typical build use cases. I just wanted to point out the linux issues.

nebhale:
Seems like we've reached consensus.  @jromero Can you please have the final say before I move this to FCP?

hone:
> The limitations here are the same for docker run --volume and folks can use or not use it as they see fit.
Using Docker as the gold standard shouldn't be what we hold ourselves to here. `pack` is a PLATFORM. We need to do our best to ensure a great experience across every OS. I believe prioritization around Mac specific features is a dangerous line of thinking, though tbf I believe Win should have the same experience.

> SSH/other socket mount. I think supporting a raw bind mount makes sense. I could also see making this a first party option for pack instead of mounting it via the --volume flag.
From talking with @ekcasey  (correct me if I'm wrong), that this feature does not actually solve this b/c of the need for `--privileged` flag.

nebhale:
@hone [Warning Message](a40ade6)

jabrown85:
How would we feel about marking this `pack` feature experimental?

jromero:
Thank you @nebhale for forcing me to look at this.

To summarize my understandings:

### Motivation

#### My understanding

The real motivation for this goes beyond subsequent build operations and incorporates some desired **development** experience.

#### My feedback

I'm not sure that "development" experience should be a guiding principle given that we've had other features in our sights (eg. `/bin/develop`) that would likely be better suited. If we focus on the primary motivation, I'm not sure that the UX is ideal. 

My take on motivations (ranked):

1. Subsequent builds (of related language families) should be able to share cache
2. First build should be able to have prepopulated cache.
3. ~~sockets~~
    I'm not certain that we have enough information to provide a solution here. Yes, the current proposal would work but given that it's limited to `pack`, is it the right solution? Furthermore, the original contributor solved this via `/platform` mount.


### UX

We would be proposing a recommendation to the end user to execute commands knowing what directories to "share":
```bash
pack --volume=/.../.m2:~/.m2 build
```

#### My feedback

Personally, I'm not keen on requiring a parameter. I would rather it be something that "just works" by default and then could potentially allow you to provide _your_ sources.

### Limitation

#### My understanding

Volume mounting would "just work" on macOS due to permissions being persisted in attributes on the FS and not overriding the actual host permissions. This doesn't work for linux users. This is not a current issue for `pack ... --volume` because it's read-only.

#### My feedback

This makes me personally feel uneasy about mounting anything to buildpacks that can wreak in the mount I've then provided. Specially, if we open it to ANY directory in the build process.

## Conclusion

Overall I want to look at this a little differently. I'm bouncing around some ideas that may be more specific to the original/primary motivation. My initial instinct is to think about only enabling this for the home directory as it's already writable via the spec and based on most, if not all, of the examples here that's where the main use case comes into play.

ekcasey:
@hone 
> From talking with @ekcasey (correct me if I'm wrong), that this feature does not actually solve this b/c of the need for --privileged flag.

From my latest tests bind mounting permission appear to work with single files (including sockets). One thing to note from a `pack` implementation perspective is that (to take advantage of the permission fixes) we can't run containers as `root` and then drop privileges. Practically, this mean that you won't want to use `creator` with a bind mounted volume (unless we fundamentally change something in the daemon case).

> Using Docker as the gold standard shouldn't be what we hold ourselves to here. pack is a PLATFORM. We need to do our best to ensure a great experience across every OS. I believe prioritization around Mac specific features is a dangerous line of thinking, though tbf I believe Win should have the same experience. I believe if this goes in we should be willing to put in the gross linux bind volume mount permission hacks into pack.

I am not opposed to doing something clever to fix this on linux eventually, but I am okay taking the easy win and then building on top of that w/ linux improvements. Are you opposed to supporting this on Mac before we support it on linux? Or do you believe we must support all OSes equally when the feature lands?

hone:
> From my latest tests, bind mounting permission appear to work with single files (including sockets). One thing to note from a `pack` implementation perspective is that (to take advantage of the permission fixes) we can't run containers as `root` and then drop privileges. Practically, this mean that you won't want to use `creator` with a bind mounted volume (unless we fundamentally change something in the daemon case).

So we can just bind in sockets then?

> I am not opposed to doing something clever to fix this on linux eventually, but I am okay taking the easy win and then building on top of that w/ linux improvements. Are you opposed to supporting this on Mac before we support it on linux? Or do you believe we must support all OSes equally when the feature lands?

I think that's the ideal scenario, but also understand in practice that may not work. RFCs are an agreed upon idea and I'd like to see us have a plan to support all OSes as part of that even if they're done separately if that make sense. This to some degree comes back to the idea of "editions" we've talked about. We could ship support for something over many releases, but when it's fully supported we'd be able to roll it all together to market it.

text/0000-danger-zone.md
```diff
@@ -0,0 +1,60 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+* There don't appear to be any viable strategies for exposing a local filesystem to the build container.
+* Not doing this leaves things with the status quo which isn't horrible.
+
+# Prior Art
+[prior-art]: #prior-art
+
+* Currently, arbitrary read-only volumes can be mounted under `/platform`.  The outcome of this RFC would be similar with looser permissions and more risk.
```
jromero, line 49: I'm curious how this would behave if we were to put it behind experimental as [proposed](https://github.com/buildpacks/rfcs/pull/85#issuecomment-649007276) by @jabrown85 . 

Side note: Honestly, based on the concerns listed in this [comment](https://github.com/buildpacks/rfcs/pull/85#issuecomment-649613027), I would be okay with this if it was behind experimental (and with decent amount of warnings based on the usage - eg. linux) otherwise it just seems too confusing and dangerous to use. For example, if I recall correctly based on a discussion of the WG meeting, some buildpacks `rm -rf` directories from home such as `.m2`.

Currently, we map the `--volume=<src>:<dest>` where `<dest>` is then prefixed with `/platform` automatically.

> --volume stringArray       Mount host volume into the build container, in the form '<host path>:<target path>'. Target path will be prefixed with '/platform/'

If we were to put this behind experimental how does this feature behave:

A. ~~The `--volume` flag changes behaviour based on `experimental`. When not experimental it prefixes to `/platform`.~~ (just thinking about this made me discard the idea)
B. Drop the UX for mounting to platform and then require users to provide `--volume=<src>:/platform/<dest>`
C. Separate the features as `--volume` being the experimental (I can mount anywhere) and move our current impl to a new `--platform-volume`

hone:
I know we talked in the WG about this being experimental feature of `pack` is that something we're still considering for this?

nebhale:
@hone I'd prefer for it to _not_ be experimental, but if that's what it takes to move this along, I'll update the RFC to it.

nebhale:
Final Comment Period with merge disposition, closing on 30 July, 2020.

text/0000-danger-zone.md
```diff
@@ -0,0 +1,62 @@
...
+
+This benefit is generally useful and extends beyond build caches to nearly every aspect of buildpack usage including development of both buildpacks and the lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently, the `pack` CLI has a `--volume` flag that allows users to expose a local filesystem location as a read-only volume mount into the `/platform` directory.  This change should generalize that flag and allow it to mount a local filesystem location as a read-write volume mount into any location within the build container.  If a user chooses to mount at any specification-reserved files system location (e.g. `/cnb` or `/layers`), a warning should be displayed.
+
```
jromero, line 27:
```suggestion

`--volume` will allow `mode` as an optional third colon separated parameter.

Format: `--volume=<src>:<target>[:mode]`

Mode values:
* `ro` - Read-only (default)
* `rw` - Read/Write
```

text/0000-danger-zone.md
```diff
@@ -0,0 +1,62 @@
...
+The technical details are opaque and implementation dependent as this is a change to the behavior of a `pack`-specific flag.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* There is an amazing amount of danger in using a shared filesystem in distributed systems.
+* Performance problems accessing shared filesystems in distributed systems.
+* Performance problems accessing a local filesystem from within the MacOS Docker Daemon.
+* There are signficant risks around file ownership on Linux hosts
```
jromero, line 39:
```suggestion
* There are significant risks around file ownership on Linux hosts
* There would be a small transition period of incompatibility for users that already use `--volume` for mounting platform directories.
```

PR #86, opened 2020-06-11
Inline Buildpacks

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/inline-buildpack/text/0000-inline-buildpack.md)

### DIFF
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -119,7 +119,7 @@ inline = "<script contents (optional)>"
 
 This defines the buildpacks that a platform should use on the repo.
 
-Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided the `version`, `inline`, and `shell` MUST NOT be allowed.
+Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided then `api`, `version`, `inline`, and `shell` MUST NOT be allowed.
 
 The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provide the `version` and `uri` MUST NOT be allowed.
 
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -29,7 +29,7 @@ The target personas for this RFC are the buildpack author and buildpack user who
 
 We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
 
-- `inline` - a script that will be run as the `bin/build` of the inline buildpack
+- `inline` - the contents of a script that will be run as the `bin/build` of the inline buildpack
 - `shell` - defines the shebang line of the script in `inline`
 
 (Note: there is no need to implement `bin/detect`. The detect phase will always pass because there's no foreseeable reason that a user would create an inline buildpack that they don't want to use)
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -112,7 +112,7 @@ id = "<buildpack ID (optional)>"
 version = "<buildpack version (optional default=latest)>"
 uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
 shell = "<string (optional default=/bin/sh)>"
-inline = "<script (optional)>"
+inline = "<script contents (optional)>"
 ```
 
 This defines the buildpacks that a platform should use on the repo.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -37,7 +37,7 @@ We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
 # How it Works
 [how-it-works]: #how-it-works
 
-When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the lifecycle WILL NOT attempt to download the buildpack `id@version`. Instead, it will create an ephemeral buildpack, with the following:
+When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the platform WILL NOT attempt to download the buildpack `id@version`. Instead, it will create an ephemeral buildpack, with the following:
 
 - a `bin/detect` that always passes (i.e. `exit 0`)
 - a `bin/build` that contains the value of `inline` script without any changes (i.e. no magic)
@@ -94,6 +94,8 @@ inline = """
 - Can inline buildpacks be root buildpacks?
 - Can a `project.toml` contain more than one inline buildpack?
 - Should `version` be required or should it default to `0.0.0`?
+- How will platforms implement this behavior?
+    - A future `/lifecycle/prepare` could generate the temporary buildpack from the inline script.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -93,6 +93,7 @@ inline = """
 - Can an inline buildpack be published to a registry?
 - Can inline buildpacks be root buildpacks?
 - Can a `project.toml` contain more than one inline buildpack?
+- Should `version` be required or should it default to `0.0.0`?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -65,12 +65,14 @@ id = "example/ruby"
 version = "1.0"
 
 [[build.buildpacks]]
-api = "0.3"
 id = "me/rake-tasks"
-inline = """
-  rake war
-  rake db:migrate
-"""
+
+  [build.buildpacks.script]
+  api = "0.3"
+  inline = """
+    rake war
+    rake db:migrate
+  """
 ```
 
 # Drawbacks
@@ -91,6 +93,7 @@ inline = """
 
 - `Dockerfile`
 - [Packfile](https://github.com/sclevine/packfile/)
+- [HTML `<script>` element](https://www.w3.org/html/wiki/Elements/script)
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
@@ -112,6 +115,8 @@ The build table MAY contain an array of buildpacks. The schema for this table is
 
 ```toml
 [[build.buildpacks]]
+
+[build.buildpacks.script]
 api = "<buildpack api>"
 shell = "<string (optional default=/bin/sh)>"
 inline = "<script contents (optional)>"
@@ -119,6 +124,6 @@ inline = "<script contents (optional)>"
 
 This defines the buildpacks that a platform should use on the repo.
 
-Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided then `api`, `version`, `inline`, and `shell` MUST NOT be allowed.
+Either an `version`, `uri`, or `script` table MUST be included, but MUST NOT include any combination of these elements.
 
-The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provided the `version` and `uri` MUST NOT be allowed.
+The `api` and `inline` key MUST be defined in the `script` table.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -122,5 +122,3 @@ This defines the buildpacks that a platform should use on the repo.
 Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided then `api`, `version`, `inline`, and `shell` MUST NOT be allowed.
 
 The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provided the `version` and `uri` MUST NOT be allowed.
-
-When an `inline` script is provided, the `CNB_BUILDPACK_DIR` will be set to `<app-dir>`.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -118,3 +118,5 @@ inline = "<script contents (optional)>"
 This defines the buildpacks that a platform should use on the repo.
 
 Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided the `version`, `inline`, and `shell` MUST NOT be allowed.
+
+When an `inline` script is provided, the script will be executed using the app directory as the buildpack root directory.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -109,7 +109,7 @@ The build table MAY contain an array of buildpacks. The schema for this table is
 id = "<buildpack ID (optional)>"
 version = "<buildpack version (optional default=latest)>"
 uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
-shell = "<string (optional default=/bin/sh)"
+shell = "<string (optional default=/bin/sh)>"
 inline = "<script (optional)"
 ```
 
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -123,4 +123,4 @@ Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` 
 
 The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provided the `version` and `uri` MUST NOT be allowed.
 
-When an `inline` script is provided, the script will be executed using the app directory as the buildpack root directory.
+When an `inline` script is provided, the `CNB_BUILDPACK_DIR` will be set to `<app-dir>`.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -121,6 +121,6 @@ This defines the buildpacks that a platform should use on the repo.
 
 Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided then `api`, `version`, `inline`, and `shell` MUST NOT be allowed.
 
-The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provide the `version` and `uri` MUST NOT be allowed.
+The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provided the `version` and `uri` MUST NOT be allowed.
 
 When an `inline` script is provided, the script will be executed using the app directory as the buildpack root directory.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -97,4 +97,21 @@ inline = """
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
+The [Project Descriptor extension](https://github.com/buildpacks/spec/blob/master/extensions/project-descriptor.md) will be ammended to include:
 
+## [[build.buildpacks]]
+
+The build table MAY contain an array of buildpacks. The schema for this table is:
+
+```toml
+[[build.buildpacks]]
+id = "<buildpack ID (optional)>"
+version = "<buildpack version (optional default=latest)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+shell = "<string (optional default=/bin/sh)"
+inline = "<script (optional)"
+```
+
+This defines the buildpacks that a platform should use on the repo.
+
+Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided the `version`, `inline`, and `shell` MUST NOT be allowed.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -29,8 +29,11 @@ The target personas for this RFC are the buildpack author and buildpack user who
 
 We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
 
+- `api` - the API compatibility version of the buildpack. Matches the schema of `buildpack.toml`
 - `inline` - the contents of a script that will be run as the `bin/build` of the inline buildpack
 - `shell` - defines the shebang line of the script in `inline`
+
+When any of these keys are provided the following keys are disallowed: `version`, `uri`.
 
 (Note: there is no need to implement `bin/detect`. The detect phase will always pass because there's no foreseeable reason that a user would create an inline buildpack that they don't want to use)
 
@@ -62,6 +65,7 @@ id = "example/ruby"
 version = "1.0"
 
 [[build.buildpacks]]
+api = "0.3"
 id = "me/rake-tasks"
 inline = """
   rake war
@@ -74,6 +78,7 @@ inline = """
 
 - It discourages buildpack reuse (i.e. if it's really easy to create a one-off buildpack, people are less like to make reuseable buildpacks and share them)
 - It encourages the copy-paste problems created by `Dockerfile`
+- There is no built-in mechanism to provide dependencies in the build plan. However, this can be worked-around with a generic build-plan-buildpack that either statically or dynamically creates entries in the build plan.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -93,7 +98,6 @@ inline = """
 - Can an inline buildpack be published to a registry?
 - Can inline buildpacks be root buildpacks?
 - Can a `project.toml` contain more than one inline buildpack?
-- Should `version` be required or should it default to `0.0.0`?
 - How will platforms implement this behavior?
     - A future `/lifecycle/prepare` could generate the temporary buildpack from the inline script.
 
@@ -108,9 +112,7 @@ The build table MAY contain an array of buildpacks. The schema for this table is
 
 ```toml
 [[build.buildpacks]]
-id = "<buildpack ID (optional)>"
-version = "<buildpack version (optional default=latest)>"
-uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+api = "<buildpack api>"
 shell = "<string (optional default=/bin/sh)>"
 inline = "<script contents (optional)>"
 ```
@@ -118,5 +120,7 @@ inline = "<script contents (optional)>"
 This defines the buildpacks that a platform should use on the repo.
 
 Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided the `version`, `inline`, and `shell` MUST NOT be allowed.
+
+The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provide the `version` and `uri` MUST NOT be allowed.
 
 When an `inline` script is provided, the script will be executed using the app directory as the buildpack root directory.
--- a/text/0000-inline-buildpack.md
+++ b/text/0000-inline-buildpack.md
@@ -110,7 +110,7 @@ id = "<buildpack ID (optional)>"
 version = "<buildpack version (optional default=latest)>"
 uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
 shell = "<string (optional default=/bin/sh)>"
-inline = "<script (optional)"
+inline = "<script (optional)>"
 ```
 
 This defines the buildpacks that a platform should use on the repo.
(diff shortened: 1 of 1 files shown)


text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,117 @@
...
+
+- `inline` - a script that will be run as the `bin/build` of the inline buildpack
+- `shell` - defines the shebang line of the script in `inline`
+
+(Note: there is no need to implement `bin/detect`. The detect phase will always pass because there's no foreseeable reason that a user would create an inline buildpack that they don't want to use)
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the lifecycle WILL NOT attempt to download the buildpack `id@version`. Instead, it will create an ephemeral buildpack, with the following:
```
ekcasey, line 40: This makes it sounds like the lifecycle currently downloads missing buildpacks. It does not do this. Buildpacks must be present on the filesystem when the lifecycle is executed.

ekcasey: Also I believe we have deliberately avoided have the lifecycle directly read `project.toml`, do we want the platform to create an ephemeral buildpack? I lean towards providing this information to the lifecycle in another format.

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,117 @@
...
+[what-it-is]: #what-it-is
+
+- **inline** - existing in the same repository
+
+The target personas for this RFC are the buildpack author and buildpack user who need to create a custom, one-off build or are just getting started with buildpacks and do not want the overhead of a new repo.
+
+We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
+
+- `inline` - a script that will be run as the `bin/build` of the inline buildpack
+- `shell` - defines the shebang line of the script in `inline`
```
ekcasey, line 33: Why not let the shebang live inside the contents of `inline`?

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,117 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the lifecycle WILL NOT attempt to download the buildpack `id@version`. Instead, it will create an ephemeral buildpack, with the following:
+
+- a `bin/detect` that always passes (i.e. `exit 0`)
+- a `bin/build` that contains the value of `inline` script without any changes (i.e. no magic)
+- a `buildpack.toml` using the `id` and `version` from the buildpack entry in `project.toml`
+
+When the emphemeral buildpack is executed, its working directory will contain the artifacts in the app repo. In this way, an `inline` script MAY reference other scripts in the repo. For example, a `lib/utils.sh` may be sourced by an inline Bash script.
```
ekcasey, line 46: This isn't specific to inline buildpacks right? It's just generally true.

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,117 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- `Dockerfile`
+- [Packfile](https://github.com/sclevine/packfile/)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Can an inline buildpack be published to a registry?
```
ekcasey, line 93: Probably no

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,118 @@
...
+## [[build.buildpacks]]
+
+The build table MAY contain an array of buildpacks. The schema for this table is:
+
+```toml
+[[build.buildpacks]]
+id = "<buildpack ID (optional)>"
+version = "<buildpack version (optional default=latest)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+shell = "<string (optional default=/bin/sh)"
```
jabrown85, line 112:
```suggestion
shell = "<string (optional default=/bin/sh)>"
```

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,118 @@
...
+inline = "<script (optional)"
```
jabrown85, line 113:
```suggestion
inline = "<script (optional)>"
```

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,118 @@
...
+[prior-art]: #prior-art
+
+- `Dockerfile`
+- [Packfile](https://github.com/sclevine/packfile/)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Can an inline buildpack be published to a registry?
+- Can inline buildpacks be root buildpacks?
```
jabrown85, line 94: I vote yes on this. `apt-get install <>`

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,118 @@
...
+- Can a `project.toml` contain more than one inline buildpack?
```
jabrown85, line 95: Why wouldn't it? Maybe I'm missing something.

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,118 @@
...
+- Should `version` be required or should it default to `0.0.0`?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+The [Project Descriptor extension](https://github.com/buildpacks/spec/blob/master/extensions/project-descriptor.md) will be ammended to include:
```
jabrown85, line 101: What's the driver behind limiting this to `project.toml`? Should this also be available when in `buildpack.toml`?

text/0000-inline-buildpack.md
jkutner: oh yea, i just meant to put platform/pack here

text/0000-inline-buildpack.md
jkutner: because @sclevine

text/0000-inline-buildpack.md
jkutner: We batted this around when we first spec'ed out `project.toml` and `buildpack.toml`. I think it probably makes sense to have a similar concept in `buildpack.toml`, but the schema is different enough (and so are the use cases) that I think we should make that it's own RFC.

text/0000-inline-buildpack.md
jkutner: Fixed in c74c093

text/0000-inline-buildpack.md
jkutner: Yea, I guess what i'm trying to call out here is that the app source will be available to the buildpack. I don't think can be assumed (and it's unclear if we'll actually copy it, or symlink, or just change the working dir when the bp executes)

edmorley: > it's unclear if we'll actually copy it, or symlink, or just change the working dir when the bp executes

It feels like this choice needs to be documented in the spec (presuming we expect there to be multiple implementations of the spec?), since otherwise inline buildpack authors might inadvertently rely on the fact that changes made to CWD do/do not affect BUILD_DIR in some implementations.

text/0000-inline-buildpack.md
jkutner: We talked about this in the working group, and Stephen reminded us that without the `shell`, it becomes very difficult to know how to run the script. We could require a shebang, but then we may have to parse it. We can give `shell` a default, like `bash`, and then most people won't need to use it at all.

text/0000-inline-buildpack.md
jkutner: We've been discussing some options around using `mixins` to define custom packages, so I hope an explicit `apt install` won't be needed. But I could see this being used to install a cert?

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,120 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+- **inline** - existing in the same repository
+
+The target personas for this RFC are the buildpack author and buildpack user who need to create a custom, one-off build or are just getting started with buildpacks and do not want the overhead of a new repo.
+
+We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
+
+- `inline` - a script that will be run as the `bin/build` of the inline buildpack
```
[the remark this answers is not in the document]
jkutner: addressed in 681e17b

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,120 @@
...
+- `Dockerfile`
+- [Packfile](https://github.com/sclevine/packfile/)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Can an inline buildpack be published to a registry?
+- Can inline buildpacks be root buildpacks?
+- Can a `project.toml` contain more than one inline buildpack?
+- Should `version` be required or should it default to `0.0.0`?
```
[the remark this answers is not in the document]
jkutner: the version of the resultant ephemeral buildpack.

The reasons that come to mind are:
1. so that it's not 0.0.0
2. so you can tell what version by looking at the image metadata w/o trying to cross reference the code version to the project.toml

Those aren't great reasons.

text/0000-inline-buildpack.md
jkutner: addressed in db546b5

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,122 @@
...
+- **inline** - existing in the same repository
+
+The target personas for this RFC are the buildpack author and buildpack user who need to create a custom, one-off build or are just getting started with buildpacks and do not want the overhead of a new repo.
+
+We propose two new keys in the `[[build.buildpacks]]` table in `project.toml`:
+
+- `inline` - the contents of a script that will be run as the `bin/build` of the inline buildpack
+- `shell` - defines the shebang line of the script in `inline`
+
+(Note: there is no need to implement `bin/detect`. The detect phase will always pass because there's no foreseeable reason that a user would create an inline buildpack that they don't want to use)
```
hone, line 35: I'd also add that `bin/detect` doesn't work like `Dockerfile` and probably not what people expect: run in parallel, doesn't have access to build artifacts/layers since run first.

text/0000-inline-buildpack.md
hone: I disagree. I actually think you can solve re-use by writing an inline buildpack buildpack. You could easily bundle a thing in to make it easily exportable.

text/0000-inline-buildpack.md
hone: I would vote yes, but also wonder if this is out of scope of this RFC until root buildpack makes it in.

text/0000-inline-buildpack.md
hone: Basically what @edmorley said. Being able to inject in your example, inline code b/t the ruby/java buildpacks or even / before after though may not be common is powerful. I don't see a reason to restrict it.

text/0000-inline-buildpack.md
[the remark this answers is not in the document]
hone: I think if you were to do it, you'd have to break the sem ver version constraints here b/c the only way to automatically do it is by doing a sha/checksum of the contents of the string.

text/0000-inline-buildpack.md
hone: @jabrown85 are you saying as a buildpack author i can just write an inline buildpack instead of a full one?

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,122 @@
...
+[spec-changes]: #spec-changes
+
+The [Project Descriptor extension](https://github.com/buildpacks/spec/blob/master/extensions/project-descriptor.md) will be ammended to include:
+
+## [[build.buildpacks]]
+
+The build table MAY contain an array of buildpacks. The schema for this table is:
+
+```toml
+[[build.buildpacks]]
```
hone, line 110: does it make sense to use an inline table since it's a separate concept and make it distinct.

```toml
[[build.buildpacks]]
shell = "<string (optional default=/bin/sh)>"
inline = "<script contents (optional)>"
```

text/0000-inline-buildpack.md
jabrown85: Yeah, I like the idea of a `buildpack.toml` being the entire buildpack, pushed to a registry if it is simple enough.

text/0000-inline-buildpack.md
jkutner: I guess the one drawback of not supporting `bin/detect` is that you can use `[[provides]]` in the build plan to satisfy some dependency of another buildpack. But as @nebhale mentioned in the WG discussion, this could be worked-around with a generic build-plan-buildpack that either statically or dynamically creates entries in the build plan.

text/0000-inline-buildpack.md
ekcasey: @jabrown85 I agree that an inline `buildpack.toml` is a good idea but I also agree with @jkutner that it probably belongs in its own RFC

text/0000-inline-buildpack.md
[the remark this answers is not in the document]
ekcasey: Versioning the inline buildpack feels weird to me (akin to versioning your `Makefile` separately from your source). 

> so you can tell what version by looking at the image metadata w/o trying to cross reference the code version to the project.toml

I think users are in danger of updating the script w/o updating the buildpack version and then encountering this same problem in a more confusing way.

I wonder if we should do something special here like have an `inline` version, to indicate explicitly to users that they must cross reference the source metadata to find the exact buildpack that was used.

text/0000-inline-buildpack.md
ekcasey: I think optional versioning for inline buildpacks makes publishing potentially flaky/dangerous. Also, info like `stacks` that you would need in a published buildpack is intentionally missing here (b/c it this is designed to address a specific type of use case). Maybe an `inline` option in `buildpack.toml` (like @jabrown85 was suggesting in a different thread) would better address this need (easy but redistributable buildpacks).

hone: yeah, that's a better idea and good points.. I just don't think an inline buildpack shouldn't have a path to publishing b/c there could be very useful small/simple buildpacks that warrant being shared.

text/0000-inline-buildpack.md
ekcasey: I'm sorry but I still don't understand. Don't we always execute `bin/build` and `bin/detect` with `bash` (per the spec). If there was a shebang in the script, wouldn't `bash` do the right thing for us?

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,122 @@
...
+id = "<buildpack ID (optional)>"
+version = "<buildpack version (optional default=latest)>"
+uri = "<url or path to the buildpack (optional default=urn:buildpack:<id>)"
+shell = "<string (optional default=/bin/sh)>"
+inline = "<script contents (optional)>"
```
ekcasey, line 115: Do we need an `api` key to indicate Buildpack API for inline buildpacks?

text/0000-inline-buildpack.md
jkutner: Actually it would be `[build.buildpacks.inline]` (not an array), which is one of the weirder things in TOML. I'd rather avoid it for that reason, but I admit this kinda makes sense. In addition to `shell` and `script`, we could also add `api` here and then not worry about those keys polluting the top-level `[[build.buildpacks]]`.

That said, since most of these are will now be mutually exclusive (i.e. only one of `version`, `uri`, `inline` are allowed), I wonder if we even care about polluting the top-level `[[build.buildpacks]]`.

jkutner: For most inline buildpacks, it would probably look like:

```toml
[[build.buildpacks]]
[build.buildpacks.inline]
api = "0.3"
script = "rake task"
```

I find that pretty dang cumbersome

text/0000-inline-buildpack.md
jkutner: Added this as a drawback in 267b45b

text/0000-inline-buildpack.md
jkutner: Fixed in 267b45b

text/0000-inline-buildpack.md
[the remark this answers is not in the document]
jkutner: Addressed in 267b45b. `version` and `inline` are mutually exclusive. However, I have not specified what value of `version` the platform should use for inline buildpacks. I think it might be okay to leave this undefined.

jkutner:
I believe there are two outstanding discussion items on this proposal:

1. [Using (or not using) `shell`](https://github.com/buildpacks/rfcs/pull/86#discussion_r445705180)
1. [Using (or not using) a table for `inline`](https://github.com/buildpacks/rfcs/pull/86#discussion_r445070353)

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,126 @@
...
+```toml
+[[build.buildpacks]]
+api = "<buildpack api>"
+shell = "<string (optional default=/bin/sh)>"
+inline = "<script contents (optional)>"
+```
+
+This defines the buildpacks that a platform should use on the repo.
+
+Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided the `version`, `inline`, and `shell` MUST NOT be allowed.
```
ekcasey, line 122:
```suggestion
Either an `id` or a `uri` MUST be included, but MUST NOT include both. If `uri` is provided then `api`, `version`, `inline`, and `shell` MUST NOT be allowed.
```

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,126 @@
...
+
+The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provide the `version` and `uri` MUST NOT be allowed.
```
ekcasey, line 124:
```suggestion
The `api` and `inline` key MUST be used together (i.e. one is not valid without the other). When any of the keys `api`, `inline`, or `shell` are provided the `version` and `uri` MUST NOT be allowed.
```

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,126 @@
...
+
+When an `inline` script is provided, the script will be executed using the app directory as the buildpack root directory.
```
ekcasey, line 126: As is `CNB_BUILDPACK_DIR` will be set to `<app-dir>` rather than to `<buildpacks-dir>/<id>/<version>`? I think I agree with this but just want to clarify.

ekcasey: I think this may require an additional key in `buildpack.toml` so that the detect and build phases know to do this.

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,126 @@
...
+- `Dockerfile`
+- [Packfile](https://github.com/sclevine/packfile/)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Can an inline buildpack be published to a registry?
+- Can inline buildpacks be root buildpacks?
+- Can a `project.toml` contain more than one inline buildpack?
+- How will platforms implement this behavior?
```
matthewmcnew, line 101: Does this rfc assume that a platform will provide the inline buildpack as an ephemeral/temporary buildpack to the lifecycle?

matthewmcnew: Or will the lifecycle need to become project.toml aware?

text/0000-inline-buildpack.md
dfreilich: Previous art for `shell`: [Github Actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell)

text/0000-inline-buildpack.md
jkutner: Ultimately, `project.toml` is a platform concern. But I hope that in the future, lifecycle will provide some tools that make it easier for a platform to support `project.toml`. That means for now, a platform must handle this (and other project.toml features) on it's own and would need to provide the inline buildpack as an ephemeral/temporary buildpack to the lifecycle.

text/0000-inline-buildpack.md
jkutner: @ekcasey i agree that there needs to be a way for the buildpack to tell the lifecycle where the dir is (or that it's an inline buildpack), but i worry about a key like that `buildpack.toml` because i can't think of reason it would configurable by a buildpack author. 

Is there some other metadata file we could use to indicate this?

jkutner: I'm wondering if the platform can define the dir in `order.toml` (`--order`) for detect, and the lifecycle would pass it on to `group.toml` for build.

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,126 @@
...
+- `shell` - defines the shebang line of the script in `inline`
+
+When any of these keys are provided the following keys are disallowed: `version`, `uri`.
+
+(Note: there is no need to implement `bin/detect`. The detect phase will always pass because there's no foreseeable reason that a user would create an inline buildpack that they don't want to use)
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the platform WILL NOT attempt to download the buildpack `id@version`. Instead, it will create an ephemeral buildpack, with the following:
```
sclevine, line 43: Could we leave this unspecified? It may be significantly more efficient to have the lifecycle execute `inline` directly with `shell`. This will also make it easier to `CNB_BUILDPACK_DIR` to the app dir, which wouldn't otherwise be possible.

jkutner: Yea, that makes sense and it would resolve the discussion in https://github.com/buildpacks/rfcs/pull/86#discussion_r450251899. But then we're talking about lifecycle understanding this inline construct.

text/0000-inline-buildpack.md
ekcasey: I originally thought supporting `inline` in `buildpack.toml` was a good idea **but** it should be a separate RFC. However, if we include that in this RFC we could solve this problem and a separate problem preserving the value of `shell`.

For example: if, instead of writing a `bind/build` and `bin/detect` script for each inline buildpack, we wrote a `buildpack.toml` with an `inline` and a `shell` field, the lifecycle could run those buildpacks appropriately.

Adding a change to the `order.toml` schema to work around this use case when we may want to remove it in favor of `buildpack.toml` changes shortly thereafter doesn't seem ideal.

text/0000-inline-buildpack.md
ekcasey: If we specify an `inline` version of `buildpack.toml` we could solve both of these problems? The inline buildpack in `project.toml` could be tranforms into an "inline" `buildpack.toml`. https://github.com/buildpacks/rfcs/pull/86#discussion_r451643134

text/0000-inline-buildpack.md
ekcasey: If we specify an `inline` version of `buildpack.toml` we could solve both of these problems? The inline buildpack in `project.toml` could be tranforms into an "inline" `buildpack.toml`. https://github.com/buildpacks/rfcs/pull/86#discussion_r451643134

text/0000-inline-buildpack.md
ekcasey: Honestly, after thinking about this more, how important is it that `CNB_BUILDPACK_DIR` is the app dir? It does make sense on some level... but given that the working dir is the app dir and these are intended to be simple scripts, would a user realistically be using `CNB_BUILDPACK_DIR` instead of `PWD`?

text/0000-inline-buildpack.md
hone: I don't actually find that that cumbersome especially b/c it makes it clearer what is happening, but not a hill I'm going to die on.

hone: Didn't we talk about using `api` and `id`?

```toml
[[build.buildpacks]]
id = "inliner"
api = "0.3"
[build.buildpacks.inline]
shell = "/bin/bash"
script = "rake task"
```

text/0000-inline-buildpack.md
jkutner: @ekcasey yea, i think you're right. The most important thing is that a line like `source lib/helpers.sh` in the `inline` script works (not that the buildpack root is the app dir). And that example _just works_ because the working dir is the app dir.

text/0000-inline-buildpack.md
jkutner: @ekcasey i want `inline` in `buildpack.toml` too, but was hoping to punt on that for now. I think i can leave this unspecified, and we can deal with it later.

jkutner: addressed in de7783a

text/0000-inline-buildpack.md
jkutner: addressed in f1b6564

jkutner:
After more updates, the remaining TBD is [using (or not using) a table for `inline`](https://github.com/buildpacks/rfcs/pull/86#discussion_r445070353).

text/0000-inline-buildpack.md
jkutner: `api` shouldn't go in `[build.buildpacks]` if we have the `inline` table. And requiring `id` just feels like making the user configure something they're never going to need. The `inline` table also doesn't solve the conflict between `uri` and `version`, so it's a half measure. 

Ultimately, the "table within an array of tables" syntax is the ugliest part of TOML, and I think it's best to avoid it unless we really don't have a choice.

jkutner: we'd still have to figure out what to do with this (which is invalid):

```toml
[[build.buildpacks]]
id = "inliner"
api = "0.3"
version = "1.0"
uri = "docker://gcr.io/projectriff/node-function:0.6.1"

[build.buildpacks.inline]
shell = "/bin/bash"
script = "rake task"
```

I feel like the `inline` table doesn't really solve the problem, unless we force the other attributes into their own table.

hone: > The inline table also doesn't solve the conflict between `uri` and `version`, so it's a half measure.
well, that's something we could solve, but I think maybe out of scope of this RFC. Ideally, I'd like this to not compound the problem. `uri` and `version` are the common case, so I expect that's fine to have that at the top level. `inline` is a special case to some degree. If we make `api` a valid key, is that confusing to users? What would a user expect if I define `api` for a non inline buildpack since it applies to a thing in `buildpack.toml`?

> And requiring id just feels like making the user configure something they're never going to need.
I pulled `id` from your example above.

> I think it's best to avoid it unless we really don't have a choice.
I basically don't agree with this.

```toml
[[build.buildpacks]]
id = "inliner-buildpack"
[build.buildpacks.inline]
api = "0.3"
shell = "/bin/bash"
script = "rake task"
```

If you have a one liner script you could even do:

```toml
[[build.buildpacks]]
id = "inliner-buildpack"
inline = {api = "0.3", shell = "/bin/bash", script = "rake task"}
```

I think technically you can even multiline the script for inline, but that's discouraged in toml land.

sclevine: My strongest opinion is that `inline` should be the field name for the literal inline string that specifies the code. That way `script` or `file` could be introduced later to point a file in the app directory that contains the code.

jkutner:
Proposal has been updated with the following `project.toml` schema

```
[[build.buildpacks]]

[build.buildpacks.script]
api = "<buildpack api>"
shell = "<string (optional default=/bin/sh)>"
inline = "<script contents (optional)>"
```

jkutner:
Open question:: is `id` a required element of `[[build.buildpacks]]`?

text/0000-inline-buildpack.md
```diff
@@ -0,0 +1,129 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+When an entry in the `[[build.buildpacks]]` table contains an `inline` value, the following is implied:
+
+- a `bin/detect` that always passes (i.e. `exit 0`)
+- a `bin/build` that contains the value of `inline` script without any changes (i.e. no magic)
+- a `buildpack.toml` using the `id` and `version` from the buildpack entry in `project.toml`
+
+When the inline buildpack is executed, the working directory will the app directory (as normal). In this way, an `inline` script MAY reference other scripts in the repo. For example, a `lib/utils.sh` may be sourced by an inline Bash script.
```
danielleadams, line 49:
```suggestion
When the inline buildpack is executed, the working directory will be the app directory (as normal). In this way, an `inline` script MAY reference other scripts in the repo. For example, a `lib/utils.sh` may be sourced by an inline Bash script.
```

hone:
> Open question:: is `id` a required element of `[[build.buildpacks]]`?

I want to say no. I believe that the common case will be 1 inline buildpack, so there isn't a need to differentiate. `lifecycle` could also auto generate a buildpack id based on the project id if we need it.

danielleadams:
Not sure if I saw this anywhere - but is there any indication in the build output that an inline script has been run? If the script's output is silent and the "buildpack" isn't listed in the detect phase, could this be confusing?

PR #87, opened 2020-06-22
Application Mixins

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/app-mixins/text/0000-app-mixins-stack-buildpacks.md)

### DIFF
--- a/text/0000-app-mixins-stack-buildpacks.md
+++ b/text/0000-app-mixins-stack-buildpacks.md
@@ -1,6 +1,6 @@
 # Meta
 [meta]: #meta
-- Name: Application Mixins
+- Name: Application Mixins and Stack Buildpacks
 - Start Date: 2020-06-22
 - Author(s): [Joe Kutner](https://github.com/jkutner/)
 - RFC Pull Request: (leave blank)
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -51,9 +51,9 @@ When a list of mixins are required by buildpacks via the build plan and the buil
 1. If at least one stackpack passes detect and provides the required mixin(s), the lifecycle will execute the stackpack build phase for all passing stackpack(s). If no stackpacks pass detect, or no stackpacks provide the required mixin, the build will fail with a message that describes the mixins that could not be provided.
 1. During the lifecycle's build phase, the stackpacks that passed detection will run against the build and run images accordingly (see details below). All stackpacks will be run before the user's buildpacks.
 
-## Stackpacks
+## Stack Buildpacks
 
- A stackpacks is a special case of buildpack that has the following properties:
+ A stack buildpacks (a.k.a. stackpacks) is a special case of buildpack that has the following properties:
 
 * Is run as the `root` user
 * Configured with `privileged = true` in the `buildpack.toml`
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -149,6 +149,7 @@ EOF
 [prior-art]: #prior-art
 
 - [RFC-0006: Stage specific mixins](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)
+- The term "non-idempotent" is used in section 9.1.2 of [Hypertext Transfer Protocol -- HTTP/1.1 (RFC 2616)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
@@ -156,7 +157,6 @@ EOF
 - How should a buildpack be identified as a stackpack?
     - a special `key` in the `buildpack.toml`?
     - a `stackpack.toml`?
-- What is a word that means "not-idempotent"?
 - How do we exclude/include from the cache?
 - Should the stackpack's detect have read-only access to the app?
 
@@ -206,8 +206,8 @@ Where:
  ```
  [buildpack]
  privileged = <boolean (default=false)>
- not-idempotent = <boolean (default=false)>
+ non-idempotent = <boolean (default=false)>
  ```
 
 * `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user.
-* `not-idempotent` - when set to `true`, indicates that the buildpack is not idempotent. The lifecycle will provide a clean filesystem from the stack image(s) before each run (i.e. no cache).
+* `non-idempotent` - when set to `true`, indicates that the buildpack is not idempotent. The lifecycle will provide a clean filesystem from the stack image(s) before each run (i.e. no cache).
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -68,7 +68,7 @@ The stackpack interface is identical to the buildpack interface (i.e. the same `
 
 For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
 
-A stack can provide stackpacks by including them in the `/cnb/stackpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
+A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
 
 ```
 [[stack.order]]
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -90,7 +90,7 @@ Where:
 
 ## Rebasing an App
 
-App that uses stackpacks can be rebased as normal. Stackpacks are expected to retain ABI compatibility.
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
 
 ## Example: Apt Buildpack
 
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -151,12 +151,8 @@ EOF
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- How should a buildpack be identified as a stackpack?
-    - Not everyone likes `privileged`
-    - a special `key` in the `buildpack.toml`?
-    - a `stackpack.toml`?
 - Should the stackpack's detect have read-only access to the app?
-    - In this proposal it does, and in the future a more generic "root buildpack" construct would need it too.
+    - This would likely be driven by a stackpack that does not provide mixins, but instead dynamically contributes to the build plan based on the contents of the app source code. For example, a stackpack that adds custom cacerts.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -155,10 +155,11 @@ EOF
 [unresolved-questions]: #unresolved-questions
 
 - How should a buildpack be identified as a stackpack?
+    - Not everyone likes `privileged`
     - a special `key` in the `buildpack.toml`?
     - a `stackpack.toml`?
-- How do we exclude/include from the cache?
 - Should the stackpack's detect have read-only access to the app?
+    - In this proposal it does, and in the future a more generic "root buildpack" construct would need it too.
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -103,7 +103,11 @@ A buildpack that can install an arbitrary list of mixins would have a `buildpack
 [buildpack]
 id = "example/apt"
 privileged = true
-provides-mixins = true
+
+[buildpack.mixins]
+all = "<boolean (default=false)>"
+names = [ "<mixin name>" ]
+
 
 [[stacks]]
 id = "io.buildpacks.stacks.bionic"
@@ -113,14 +117,6 @@ Its `bin/detect` would have the following contents:
 
  ```bash
 #!/usr/bin/env bash
-
-# TODO read the mixins.toml (which has same schema as build plan)
-for package in $(cat mixins.toml | yj -t | jq -r ".requires | .[] | .name"); do
-  cat << EOF >> $2
-[[provides]]
-name = "${package}"
-EOF
-done
 
 exit 0
 ```
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -25,7 +25,7 @@ For that reason, this proposal defines a mechanism to implement dynamic installa
 
 - *application developer* - a person who is using buildpacks to transform their application code into an OCI image
 - *root buildpack* - a new type of buildpack that runs as the root user
-- *stackpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
+- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
 
 An application developer may specify a list of mixins in their application's `project.toml` file like this:
 
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -91,27 +91,7 @@ A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` 
 
 A stackpack will only execute if it passes detection. When the stackpack is executed, it's detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. FOr example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
 
-The stackpack's snapshot layer may be modified by writing a `stack.toml` file with the following schema:
-
-```
-[[build.restores]]
-paths = ["<sub-path glob>"]
-
-[[run.excludes]]
-paths = ["<sub-path glob>"]
-cache = false
-```
-
-Where:
-
-`[[build.restores]]` defines the directories and files that will be restored at build time even when the base image is modified:
-
-* `paths` = a list of paths to always restore
-
-`[[run.excludes]]` defines the directories and files that will be excluded from the launch image:
-
-* `paths` = a list of paths to exclude from the layer
-* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in the cache layer.
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
 
 ## Rebasing an App
 
@@ -327,3 +307,27 @@ Under the `[buildpack.mixins]` table:
 
 * `any` - a boolean that, when true, indicates that the buildpack can provide all mixins
 * `names` - a list of names that match mixins provided by this buildpack
+
+## stack.toml (TOML)
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file with the following schema:
+
+```
+[[build.restores]]
+paths = ["<sub-path glob>"]
+
+[[run.excludes]]
+paths = ["<sub-path glob>"]
+cache = false
+```
+
+Where:
+
+`[[build.restores]]` defines the directories and files that will be restored at build time even when the base image is modified:
+
+* `paths` = a list of paths to always restore
+
+`[[run.excludes]]` defines the directories and files that will be excluded from the launch image:
+
+* `paths` = a list of paths to exclude from the layer
+* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in the cache layer.
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -188,14 +188,9 @@ Where:
 [buildpack]
 privileged = <boolean (default=false)>
 non-idempotent = <boolean (default=false)>
-
-[[mixins]]
-name = "<mixin name pattern>"
+mixins = [ "<mixin name pattern>" ]
  ```
 
 * `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user.
 * `non-idempotent` - when set to `true`, indicates that the buildpack is not idempotent. The lifecycle will provide a clean filesystem from the stack image(s) before each run (i.e. no cache).
-
-The `[[mixins]]` array of tables defines the list of mixins provided by this buildpack.
-
-* `name` - a pattern describing the mixins provide by this buildpack
+* `mixins` - a list of patterns that match mixins provided by this buildpack
--- a/text/0000-app-mixins.md
+++ b/text/0000-app-mixins.md
@@ -137,13 +137,17 @@ EOF
 
 ## Example: CA Certificate Buildpack
 
+Support for custom CA Certificates can be accomplished with two buildpacks: a stackpack that can install the cert, and a normal buildpack that can provide a cert in the build plan.
+
+### CA Cert Installer Stackpack
+
 A buildpack that works on the `ubuntu-20` stack, which defines a `type=` prefix for mixins, and installs custom CA Certificates would have a `buildpack.toml` that looks like this:
 
 ```toml
 [buildpack]
 id = "example/cacerts"
 privileged = true
-mixins = ["type=cacert"]
+mixins = ["cacert"]
 
 [[stacks]]
 id = "ubuntu-20"
@@ -154,28 +158,76 @@ Its `bin/detect` would have the following contents:
  ```bash
 #!/usr/bin/env bash
 
+cat << EOF > $2
+[[provides]]
+name = "cacert"
+EOF
+
 exit 0
 ```
 
-Its `bin/build` would have the following contents:
+Its `bin/build` would would install each `cacert` in the build plan. It would have the following contents:
 
  ```bash
 #!/usr/bin/env bash
 
-for cert_path in $(cat $3 | yj -t | jq -r ".entries | .[] | .name"); do
-  cert_file=$(basename $cert_path)
-  cp $cert_path /usr/share/ca-certificates/$cert_file
-  chmod 644 /usr/share/ca-certificates/$cert_file
+# filter to the cert
+for file in $(cat $3 | yj -t | jq -r ".entries | .[] | select(.name==\"cacert\") | .metadata | .file"); do
+  cert="$(cat $3 | yj -t | jq -r ".entries | .[] | select(.name==\"cacert\") | .metadata | select(.file==\"$file\") | .content")"
+  echo $cert > /usr/share/ca-certificates/$file
+  chmod 644 /usr/share/ca-certificates/$file
 done
 
 update-ca-certificates
 ```
 
-An application that is using this stackpack would the following `project.toml`:
+### CA Cert Provider Buildpack
+
+The stackpack must be used with a buildpack that provides a certificate(s) for it to install. That buildpack would have the following `buildpack.toml`:
 
 ```toml
-[build]
-mixins = [ "type=cacert:mycerts/database.crt" ]
+[buildpack]
+id = "my/db-cert"
+
+[[stacks]]
+id = "ubuntu-20"
+mixins = ["cacert"]
+```
+
+Its `bin/detect` would require a certificate with the following contents:
+
+ ```bash
+#!/usr/bin/env bash
+
+cat << EOF > $2
+[[requires]]
+name = "cacert"
+
+[requires.metadata]
+file = "database.crt"
+content = """
+$(cat $CNB_BUILDPACK_DIR/database.crt)
+"""
+
+[[requires]]
+name = "cacert"
+
+[requires.metadata]
+file = "server.crt"
+content = """
+$(cat $CNB_BUILDPACK_DIR/server.crt)
+"""
+EOF
+
+exit 0
+```
+
+Its `bin/build` would do nothing, and `have the following contents:
+
+ ```bash
+#!/usr/bin/env bash
+
+exit 0
 ```
 
 # Drawbacks
@@ -205,14 +257,7 @@ mixins = [ "type=cacert:mycerts/database.crt" ]
 - Should stackpacks be able to define per-stack mixins?
     - We could support a top-level `mixins` array, and allow refinements under `[[stacks]] mixins`. If we do this, we need to distinguish between provided and required mixins (at least under `[[stacks]]`).
     - If buildpacks can require mixins from `bin/detect`, the stackpack could use this mechanism to require per-stack mixins.
-- Should we use regex to match mixins, or should we double-down on `type=` (or similar)?
-    - regexs make it likely that a stackpack will be unable to distinguish packages from other types of mixins
-    - It will be difficult for the platform to distinguish between regex patterns and explict value to match.
-    - Alternatively, we could define a URN for mixins, like:
-        - `urn:package:libpq`
-        - `urn:package:build:libpq-dev`
-        - `urn:cacert:mycerts/database.crt`
-        - Naked strings, like `libpq`, could be assumed to represent packages.
+- Should we use regex to match mixins?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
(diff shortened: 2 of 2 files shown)


text/0000-app-mixins.md
```diff
@@ -0,0 +1,214 @@
...
+
+Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time, but it's common for application code to depend on OS packages too. But allowing buildpacks to arbitrarily install OS packages during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements. For that reason, this proposal defines a mechanism to implement mixin support for application developers without significant impact on build performance.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *application developer* - a person who is using buildpacks to transform their application code into an OCI image
+- *stackpack* - a new type of buildpack that runs against the stack image(s) instead of an app
+
+A application developer may specify a list of mixins in their application's `project.toml` file like this:
```
yaelharel, line 27: **An** application developer

text/0000-app-mixins.md
```diff
@@ -0,0 +1,214 @@
...
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
+
+A stackpack is included in a builder by defining it in the `builder.toml` in the `[[stack.buildpacks]]` array of tables:
+
+```
+[[stack.buildpacks]]
+id = "<stackpack id>"
+uri = "<uri to stackpack>"
+```
```
ekcasey, line 74: I feel like this has been a persistent point of confusion for me. In working group I thought the consensus was that `stackpacks` would ship on the stack images (build and run images) rather than on the builder image.

This would allow us to avoid having to output two images from `create-builder`. We could exclude the stackpacks from the final app image either by adding metadata to the run-image that will allow us to selectively exclude these layers or by adding a layer that [whitesout](https://github.com/opencontainers/image-spec/blob/master/layer.md#whiteouts) the buildpack dir

ekcasey, line 74: I feel like this has been a persistent point of confusion for me. In working group I thought the consensus was that `stackpacks` would ship on the stack images (build and run images) rather than on the builder image.

This would allow us to avoid having to output two images from `create-builder`. We could exclude the stackpacks from the final app image either by adding metadata to the run-image that will allow us to selectively exclude these layers or by adding a layer that [whitesout](https://github.com/opencontainers/image-spec/blob/master/layer.md#whiteouts) the buildpack dir

text/0000-app-mixins.md
```diff
@@ -0,0 +1,214 @@
...
+[[excludes]]
+paths = ["<sub-path glob>"]
+cache = false
+```
+
+Where:
+
+* `paths` = a list of paths to exclude from the layer
+* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in the cache layer.
```
ekcasey, line 188: Doesn't this conflict with line 175

> A `launch.toml` WILL NOT be honored.

Sorry, I know this is still a draft, just trying to get a sense of it.

jkutner: Addressed in 9f39572

text/0000-app-mixins.md
jkutner: Fixed in 9f39572

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+* If the mixin name is prefixed with `run:` (as per [RFC-0006](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)) the mixin will be dynamically added to the run image.
+* If the mixin name does not have a prefix it will be dynamically added to both the build and run stack images.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When a list of mixins are defined in a `project.toml` and `pack build` is run against that project, the following will happen:
+
+1. The lifecycle will compare the list of mixins to those provided by the stack. If all mixin names are provided by the stack, no further action is required.
+1. In any requested mixin is not provided by the stack, the lifecycle will run the detect phase for all stackpacks defined in the builder.
```
edmorley, line 47:
```suggestion
1. If any requested mixin is not provided by the stack, the lifecycle will run the detect phase for all stackpacks defined in the builder.
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+1. The lifecycle will compare the mixins added to the build plan to see if they match they required mixins.
+1. If at least one stackpack passes detect and provides the required mixin(s), the lifecycle will execute the stackpack build phase for passing stackpack(s). If no stackpacks pass detect, or no stackpacks provide the required mixin, the build will fail with a message that describes the mixins that could not be provided.
```
edmorley, line 49:
```suggestion
1. If at least one stackpack passes detect and provides the required mixin(s), the lifecycle will execute the stackpack build phase for all passing stackpack(s). If no stackpacks pass detect, or no stackpacks provide the required mixin, the build will fail with a message that describes the mixins that could not be provided.
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+A stackpack is included in a builder by defining it in the `builder.toml` in the `[[stack.buildpacks]]` array of tables:
+
+```
+[[stack.buildpacks]]
+id = "<stackpack id>"
+uri = "<uri to stackpack>"
+```
+
+Each stackpack included in the builder will execute until all of a project's mixins have been provided.
```
edmorley, line 77: This wording to me reads as ~"once a project's mixins have been provided by an earlier stackpack in the execution order, remaining stackpacks won't be executed".

However in the "how it works" section, the wording is:

> If at least one stackpack passes detect and provides the required mixin(s), the lifecycle will execute the stackpack build phase for passing stackpack(s).

...which suggests that instead there is no "early return" and all passing stackpacks will be run, even if an earlier one in the sequence provides the required mixin?

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+* `paths` = a list of paths to exclude from the layer
+* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in the cache layer.
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
+
+## Example: Apt Buildpack
+
+A buildpack that installs a custom certificate would have a `buildpack.toml` like this:
```
edmorley, line 98: Is the description here wrong? It seems to be describing the second example (custom certificate) rather than the APT buildpack example?

I think it might also be helpful to include an example `project.toml` of how one might use this APT buildpack, similar to the `project.toml` example for the CA certificate case?

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+ ```toml
+[buildpack]
+id = "example/apt"
+privileged = true
+mixins = ["[^=]+"]
```
edmorley, line 104: This `mixins` pattern seems quite broad - won't it mean this buildpack matches against too many mixin names? (It's quite possible I'm misunderstanding how the matching works here)

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+```toml
+[build]
+mixins = [ "type=cacert:mycerts/database.crt" ]
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Mixins are less flexible that a generic [root buildpack](https://github.com/buildpacks/rfcs/blob/app-image-ext-buildpacks/text/0000-app-image-extensions.md).
```
edmorley, line 182:
```suggestion
- Mixins are less flexible than a generic [root buildpack](https://github.com/buildpacks/rfcs/blob/app-image-ext-buildpacks/text/0000-app-image-extensions.md).
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- [Add root buildpack as interface for app image extension RFC](https://github.com/buildpacks/rfcs/pull/77)
+- [App Image Extensions (OS packages)](https://github.com/buildpacks/rfcs/pull/23)
+- [Root Buildpacks](https://github.com/buildpacks/rfcs/blob/root-buildpacks/text/0000-root-buildpacks.md): Allow application developers to use root buildpacks in `project.toml`. This would have significant performance implications, and creates a "foot gun" in qhich end users could build images they are not able to maintain. For this reason, we are holding off on a generic root buildpack feature.
```
edmorley, line 189:
```suggestion
- [Root Buildpacks](https://github.com/buildpacks/rfcs/blob/root-buildpacks/text/0000-root-buildpacks.md): Allow application developers to use root buildpacks in `project.toml`. This would have significant performance implications, and creates a "foot gun" in which end users could build images they are not able to maintain. For this reason, we are holding off on a generic root buildpack feature.
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+    - This would likely be driven by a stackpack that does not provide mixins, but instead dynamically contributes to the build plan based on the contents of the app source code. I don't know if we have a use case for this, but I can imaging a buildpack that reads environment variables as input to some function.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+## Stackpacks
+
+Stackpacks are identical to other buildpacks, with the following exceptions:
+
+1. The `<layers>` directory is NOT writable
```
edmorley, line 211:
```suggestion
1. The `<layers>` directory is NOT writable.
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+1. The working directory WILL NOT contain application source code during the build phase.
+1. All changes made to the filesystem (with the exception of `/tmp`) during the execution of the stackpack's `bin/build` will be snaphotted and stored as a single layer.
```
edmorley, line 213:
```suggestion
1. All changes made to the filesystem (with the exception of `/tmp`) during the execution of the stackpack's `bin/build` will be snapshotted and stored as a single layer.
```

text/0000-app-mixins.md
jkutner: The mixins concept is intentionally flexible as a whole. But it can be more narrowly scoped for each stack. For example, [a mixin is defined as a "package" for the bionic stack](https://github.com/buildpacks/rfcs/blob/main/text/0031-bionic-mixins.md). Because the stackpack in this example only works with the bionic stack, it can safely assume the mixin is a package.

text/0000-app-mixins.md
jkutner: whoops. fixed in efb1fa6

text/0000-app-mixins.md
jkutner: Yea, that's right. I'm wrong in the text of the RFC. Will update

jkutner: @ekcasey do you think the stack should contain the buildpack before the builder image is created, or can the `create-builder` command inject the stackpack(s) (outputting new build and run images w/ the stackpack(s)).

ekcasey: I think the build and run images should ship with the stackpacks pre-installed.

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+# Summary
+[summary]: #summary
+
+This is a proposal for allowing application developers (buildpack users) to specify mixins that will be dynamically included with their build and run images.
+
+# Motivation
+[motivation]: #motivation
+
+Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time, but it's common for application code to depend on OS packages too. But allowing buildpacks to arbitrarily install OS packages during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements. For that reason, this proposal defines a mechanism to implement mixin support for application developers without significant impact on build performance.
```
sclevine, line 19:
```suggestion
Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time, but it's common for application code to depend on OS packages too. But allowing buildpacks to arbitrarily install OS packages during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements.

For that reason, this proposal defines a mechanism to implement dynamic installation of mixins at build time while minimizing the performance impact. This is accomplished by allowing stack images to satisfy mixin requires statically and/or reject dynamic mixin installation early in the build process.
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+
+ A stackpacks is a special case of buildpack that has the following properties:
+
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
+* Does not have access to the application source code
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the builder image
```
sclevine, line 62:
```suggestion
* Is distributed with the stack run and/or build image
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,241 @@
...
+When a command like `pack build` is run, the list of mixins will be processed before buildpacks are run. For each mixin name, the following will happen:
+
+* If the mixin name is prefixed with `build:` (as per [RFC-0006](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)) the mixin will be dynamically added to the build image.
+* If the mixin name is prefixed with `run:` (as per [RFC-0006](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)) the mixin will be dynamically added to the run image.
+* If the mixin name does not have a prefix it will be dynamically added to both the build and run stack images.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When a list of mixins are defined in a `project.toml` and `pack build` is run against that project, the following will happen:
```
sclevine, line 44: I thought they're supposed to come from the build plan (contributed by other buildpacks)? 
```suggestion
When a list of mixins are required by buildpacks via the build plan the build phase starts:
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,243 @@
...
+done
+
+update-ca-certificates
+```
+
+An application that is using this stackpack would the following `project.toml`:
+
+```toml
+[build]
+mixins = [ "type=cacert:mycerts/database.crt" ]
```
jabrown85, line 178: What do you think about using `,` instead of `:`
```suggestion
mixins = [ "type=cacert,mycerts/database.crt" ]
```

Sort of like docker
`docker run --read-only --mount type=volume,target=/icanwrite busybox`

text/0000-app-mixins.md
```diff
@@ -0,0 +1,243 @@
...
+An application developer may specify a list of mixins in their application's `project.toml` file like this:
+
+```toml
+[build]
+mixins = [ "<mixin name>" ]
+```
+
+When a command like `pack build` is run, the list of mixins will be processed before buildpacks are run. For each mixin name, the following will happen:
+
+* If the mixin name is prefixed with `build:` (as per [RFC-0006](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)) the mixin will be dynamically added to the build image.
```
hazemm96, line 39: Newbie question: I'm assuming this happens prior to any phase of the buildpack lifecycle happening. As in, it happens when we `pack create-builder ...`. Which means all the buildpack lifecycle steps will never read to read/write from/to `/`. Is this accurate?

hazemm96: (By lifecycle steps I mean: detector, analyzer, restorer, builder, exporter)

jkutner: No, the stackpacks would not execute during `create-builder`. They run during the normal lifecycle phases, but in as the `root` user, which allows them to write to `/`). Does that help?

hazemm96: I see. So in the line above you mean the mixin will be dynamically added to the builder* image when prefixed with `build:`. 

I'm curious, if implemented, which lifecycle phase will need to read/write to/from `/`. All of them or just the `builder` phase?

jkutner: Only the build phase will need (or more correctly "be able") to write to `/`

text/0000-app-mixins.md
jkutner: This has changed in 6a3cfba

text/0000-app-mixins.md
jkutner: @ekcasey should i change `uri` to `path`?

ekcasey: @jkutner Do we even need to list `stack.buildpacks` if they are already in the image? The `[[buildpacks]]` section typically describe where to fetch buildpacks that should be added to the image.

Is the point to enable/disable them (I wonder if default to enabled is the right call)? In any case I think we only need id/version.

ekcasey: I am imagining that stackpacks live either in `/cnb/buildpacks/<id>/<version>` or in an analagous `/cnb/stackpacks`, and can therefore be looked up by `id` and `version`.

text/0000-app-mixins.md
```diff
@@ -0,0 +1,299 @@
...
+* If the mixin name is prefixed with `run:` (as per [RFC-0006](https://github.com/buildpacks/rfcs/blob/main/text/0006-stage-specific-mixins.md)) the mixin will be dynamically added to the run image.
+* If the mixin name does not have a prefix it will be dynamically added to both the build and run stack images.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When a list of mixins are required by buildpacks via the build plan and the build phase starts:
+
+1. The lifecycle will compare the list of mixins to those provided by the stack. If all mixin names are provided by the stack, no further action is required.
+1. If any requested mixin is not provided by the stack, the lifecycle will run the detect phase for all stackpacks defined in the builder.
```
sclevine, line 49: Can we run `/bin/detect` in parallel with other buidpacks' `/bin/detect`? Then the stackpack could opt-out entirely (be incapable of providing mixins) if its detect fails.

(Should stackpacks have an optional flag and fail all builds if they fail detection?)

text/0000-app-mixins.md
```diff
@@ -0,0 +1,299 @@
...
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not write to the `/layers` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
+
+A builder can use a stackpack by defining it in the `builder.toml` in the `[[stack.buildpacks]]` array of tables:
```
sclevine, line 71: Given that the run image must also contain stackpacks, I suggest we require the stackpacks be present on the stack images themselves and not added as part of `pack create-builder`.

text/0000-app-mixins.md
```diff
@@ -0,0 +1,299 @@
...
+id = "ubuntu-20"
+```
+
+Its `bin/detect` would have the following contents:
+
+ ```bash
+#!/usr/bin/env bash
+
+cat << EOF > $2
+[[provides]]
```
sclevine, line 162: I support `/bin/detect` for stackpacks, so that they can validate that the stack image itself is in a valid state. However, I'm reluctant to allow mixins to be specified dynamically during detect, given that the app directory is not available.

text/0000-app-mixins.md
```diff
@@ -0,0 +1,299 @@
...
+* Can only create one layer
+* Does not have access to the application source code
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not write to the `/layers` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
micahyoung, line 69: This feature would be super valuable for Windows images for all the same reasons, perhaps more-so since Windows artifacts are often more strict about their location. 

Kaniko appears to have no plans to support Windows containers (https://github.com/GoogleContainerTools/kaniko/issues/111), but does this spec language leave open the option for lifecycle to use alternate implementations? Is that the intent of: 
> Alternatively, a platform may store a new stack image to cache the changes?.

text/0000-app-mixins.md
jkutner: @sclevine what if it's not based on app source, but some other input from the platform (like a list of mixins that are needed)

text/0000-app-mixins.md
```diff
@@ -0,0 +1,305 @@
...
+EOF
+```
+
+## Example: CA Certificate Buildpack
+
+Support for custom CA Certificates can be accomplished with two buildpacks: a stackpack that can install the cert, and a normal buildpack that can provide a cert in the build plan.
+
+### CA Cert Installer Stackpack
+
+A buildpack that works on the `ubuntu-20` stack, which defines a `type=` prefix for mixins, and installs custom CA Certificates would have a `buildpack.toml` that looks like this:
```
jabrown85, line 152: This is no longer true in the example right?

text/0000-app-mixins.md
```diff
@@ -0,0 +1,305 @@
...
+id = "io.buildpacks.stacks.bionic"
+```
+
+Its `bin/detect` would have the following contents:
+
+ ```bash
+#!/usr/bin/env bash
+
+cat << EOF > $2
+[[provides]]
```
jabrown85, line 169: Is there any way for this buildpack to _only_ run on run images and not on the build image? The mixins has `build:` and `run:`, can this example only run on run images?

text/0000-app-mixins.md
```diff
@@ -0,0 +1,305 @@
...
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not write to the `/layers` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase (excluding `/tmp`, `/cnb`, and `/layers`). Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
+
+A stack can provide stackpacks by including them in the `/cnb/stackpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
```
jromero, line 71:
```suggestion
A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
```

text/0000-app-mixins.md
```diff
@@ -0,0 +1,305 @@
...
+Mixins already allow buildpack authors to create buildpacks that depend on an extended set of OS packages without affecting build time, but it's common for application code to depend on OS packages too. But allowing buildpacks to arbitrarily install OS packages during build would drastically increase build time, especially when CNB tooling is used to build or rebase many apps with similar package requirements.
+
+For that reason, this proposal defines a mechanism to implement dynamic installation of mixins at build time while minimizing the performance impact. This is accomplished by allowing stack images to satisfy mixin requires statically and/or reject dynamic mixin installation early in the build process.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *application developer* - a person who is using buildpacks to transform their application code into an OCI image
+- *root buildpack* - a new type of buildpack that runs as the root user
+- *stackpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
```
jromero, line 28:
```suggestion
- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
```

I propose keeping the naming more inline with "buildpacks" and simply prefix the type as the name as a whole. Specially, if we can envision being able to provide them at build time from a platform's perspective.

text/0000-app-mixins.md
jkutner: Fixed in 8cd657fc613769f29f84ca88673415504d57ea4c

text/0000-app-mixins.md
jkutner: Addressed in 8ee98d430e2ce55652cd53ae5ed9f344bd888889

text/0000-app-mixins.md
jkutner: Addressed in 8cd657fc613769f29f84ca88673415504d57ea4c

text/0000-app-mixins.md
jkutner: Addressed in 1881dd3dead0a35b3335c99cfad893d1e2ba24b1

jkutner:
I've changed the schema to:

```
[buildpack.mixins]
any = "<boolean (default=false)>"
names = [ "<mixin name>" ]
```

While this is functionally what I want, I'm unsure about the table and key names. I also considered:
- Naming the table `[buildpack.provides]` with a key `mixins = [ "<mixin name>" ]`
- Naming the `any` key `all`
- Naming the `any` key `provides-any`

jkutner:
I added `/proc` to the list of ignored dirs. In my prototyping, I found that it's pretty noisy, and I don't think this is stuff you'd want to apply to the stack. Alternatively, we can let people exclude it.

sclevine:
If we add `/proc`, we should probably add other directories that aren't backed by filesystem storage, like `/sys`.
In container-land, `/dev` is often a tmpfs, so probably worth adding it as well.

jkutner:
@sclevine maybe we should match Kaniko's list of ignored files dirs? https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting

jkutner:
List of outstanding things we need to resolve:
1. What key names to use in `buildpack.toml` schema (see https://github.com/buildpacks/rfcs/pull/87#issuecomment-666359457)
2. What are the default dirs to ignore in the snapshot (see https://github.com/buildpacks/rfcs/pull/87#issuecomment-667772632)
3. How to enable different behavior between build/run phases for a stack buildpack (see https://github.com/buildpacks/rfcs/pull/87#discussion_r462537431)
4. What should we do with the `launch.toml` during the build phase of a stack buildpack (proposal: just ignore it)
5. What should we do if the stack buildpack writes to `layers/` (ignore it, or error out because its read-only)

text/0000-app-mixins.md
```diff
@@ -0,0 +1,335 @@
...
+
+A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
+
+```
+[[stack.order]]
+  [[stack.order.group]]
+    id = "<stackpack id>"
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, it's detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. FOr example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
```
jabrown85, line 81:
```suggestion
A stackpack will only execute if it passes detection. When the stackpack is executed, it's detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
```

text/0000-app-mixins.md
ekcasey: > However, I'm reluctant to allow mixins to be specified dynamically during detect, given that the app directory is not available.

I agree that the stackpacks shouldn't dynamically decide which mixins to provide during detect, but regular buildpacks (which have access to the app dir) should probably be able to require specific mixins during detect.

text/0000-app-mixins-stack-buildpacks.md
```diff
@@ -0,0 +1,335 @@
...
+
+```
+[[stack.order]]
+  [[stack.order.group]]
+    id = "<stackpack id>"
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, it's detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
```
ekcasey, line 83: Where should the buildpack write `stack.toml`? If it is analogous to `launch.toml` it should be in the layers dir, but we have said the layers dir is readonly for stackpacks.

Also we already have a `stack.toml` file, maybe we should just use `launcher.toml` and have a special section for stackpacks? Having a `launch.toml` for stackpacks would allow stackpacks to modify the BOM and access other buildpack features if/when we allow that.

text/0000-app-mixins-stack-buildpacks.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
```
ekcasey, line 87: > the platform must re-run the any stackpacks that were used to build the launch image against the new run-image

Is this really a "must"? While this would produce that most comprehensive upgrade, a regular rebase should work without a full upgrade? I could see a regular rebase being preferable in some situations where performance is critical.

text/0000-app-mixins-stack-buildpacks.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Example: Apt Buildpack
+
+A buildpack that can install an arbitrary list of mixins would have a `buildpack.toml` like this:
+
+ ```toml
+[buildpack]
+id = "example/apt"
+privileged = true
```
ekcasey, line 96: I wonder if `privileged` is more of an implementation-detail/requirement than the primary difference/intent here? Maybe we want something like `scope = "stack"` (where scope defaults to "app"). This would leave open the option of privileged "regular" buildpacks or unprivileged stackpacks in the future (not sure we actually want that but it might be nice to leave our options open...).

Also, `scope = "stack"` might make it easier for folks to understand how these buildpacks relate to ` stack.order` in a `builder.toml` or ` stack-group.toml` in the lifecycle.

text/0000-app-mixins-stack-buildpacks.md
```diff
@@ -0,0 +1,335 @@
...
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
+* Does not have access to the application source code
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not write to the `/layers` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
```
ekcasey, line 67: Generally the working dir is the app dir for buildpacks? What do we expect the working dir and the layers dir input to be for stackpacks if we are keeping the same interface?

jkutner: I'm using `/` as the working dir in my POC. Not married to it though. I'm starting to think we should pass the layers dir, and let it be writeable, but ignore anything written to it. This would make development easy if, for example, someone is using a library in their buildpack implementation that initializes layer dirs or launch.toml. The lifecycle would just ignore these.

text/0000-app-mixins-stack-buildpacks.md
jkutner: Yea, I like the idea of stackpacks modifying the BOM. I don't feel strongly about the filename or if it should be a separate file.

text/0000-app-mixins-stack-buildpacks.md
jkutner: I think this would introduce the risk of a stackpack appending to some file, but then the rebase also changes that files, and those rebase changes are lost.

jkutner:
I'm moving this to draft status with the hope that #111 and #112 make the proposal(s) easier to understand (especially when we look back on these documents in the future).

jkutner:
Closing this in favor of #111 and #112

PR #89, opened 2020-06-26
Cache scope

jromero:
[Readable](https://github.com/buildpacks/rfcs/blob/cache-types/text/0000-cache-types.md)

### DIFF
--- a/text/0000-cache-types.md
+++ b/text/0000-cache-types.md
@@ -44,17 +44,17 @@ There will be two defined scopes:
 
 #### Platform Implementation
 
-Platforms can deside how to _share_ the "shared" cached layers based on their specific use cases.
+Platforms can decide how to _share_ the "shared" cached layers based on their specific use cases.
 
 In `pack`, it being a local development tool, shared cached layers could be implemented as a single global cache that is shared across all applications. Additional options can be provided to restrict or otherwise configure this sharing.
 
-In a different platform, one can forsee that the cache is shared only within a "project" or an "organization".
+In a different platform, one can foresee that the cache is shared only within a "project" or an "organization".
 
 #### Lifecycle Implementation
 
-1. The lifecycle will provide an additional `-shared-cache-dir` and `-shared-cache-image` options where `-cache-dir` and `-cache-image` could be provided.
+1. The lifecycle will provide additional `-shared-cache-dir` and `-shared-cache-image` options where `-cache-dir` and `-cache-image` could be provided.
 2. Layers marked as `cache = true` and `cache-scope = "shared"` will be stored in either `-shared-cached-dir` or `-shared-cache-image`.
-3. If `-share-cached-dir` or `-share-cached-image` are not provide and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standart cache location (`-cache-dir`/`-cache-image`).
+3. If `-share-cached-dir` or `-share-cached-image` are not provide and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standard cache location (`-cache-dir`/`-cache-image`).
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -66,8 +66,8 @@ In a different platform, one can forsee that the cache is shared only within a "
 
 - Generic Read/Write volume mounts (RFC PR [#85](https://github.com/buildpacks/rfcs/pull/85))
     - The buildpack user must know what and where to mount things.
-    - Has safety concerns when writting to Linux mounts.
-    - Solution violates the spec dependant on use.
+    - Has safety concerns when writing to Linux mounts.
+    - Solution violates the spec dependent on use.
     - Solution is specific to `pack`.
     
 # Prior Art
--- a/text/0000-cache-types.md
+++ b/text/0000-cache-types.md
@@ -1,6 +1,6 @@
 # Meta
 [meta]: #meta
-- Name: Cache Types
+- Name: Cache Scope
 - Start Date: 2020-06-26
 - Author(s): [Javier Romero](https://github.com/jromero)
 - RFC Pull Request: (leave blank)
@@ -30,33 +30,36 @@ In order to solve this use case in a safe manner, the platform needs to know wha
 # What it is
 [what-it-is]: #what-it-is
 
-Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by associating a "type" to the cache.
+Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by associating specifing a scope of "shared".
 
 - Target persona: buildpack author
 
-There will be two types: 
+There will be two defined scopes: 
 
-- `app-specific` - This cache is specific to the application and SHOULD NOT be shared with any other applications.
-- `buildpack-specific` - This cache is specific to the buildpack and MAY be shared across multiple applications.
+- `app` - This cached layer is specific to the application and SHOULD NOT be shared with any other applications.
+- `shared` - This cached layer is specific to the buildpack and MAY be shared across multiple applications.
 
 # How it Works
 [how-it-works]: #how-it-works
 
 #### Platform Implementation
 
-Provided that there is something that denotes what type of cached layers are available the platform can choose to which scope, if any, they'd like to share the `buildpack-specific` caches.
+Platforms can deside how to _share_ the "shared" cached layers based on their specific use cases.
 
-In `pack`, this could be implemented to limit cache sharing per builder but across applications.
+In `pack`, it being a local development tool, shared cached layers could be implemented as a single global cache that is shared across all applications. Additional options can be provided to restrict or otherwise configure this sharing.
+
+In a different platform, one can forsee that the cache is shared only within a "project" or an "organization".
 
 #### Lifecycle Implementation
 
-TODO
+1. The lifecycle will provide an additional `-shared-cache-dir` and `-shared-cache-image` options where `-cache-dir` and `-cache-image` could be provided.
+2. Layers marked as `cache = true` and `cache-scope = "shared"` will be stored in either `-shared-cached-dir` or `-shared-cache-image`.
+3. If `-share-cached-dir` or `-share-cached-image` are not provide and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standart cache location (`-cache-dir`/`-cache-image`).
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Doesn't solve for the possibility of "seeding" a cached layer with data on first run. 
-    - Use case: I have `~/.m2` locally and would like to provide that to the build process so that it doesn't have to download all dependencies on first run. 
+- Requires that the buildpack author be continuous about providing this information for optimization purposes.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -81,7 +84,7 @@ None
 
 ### Buildpacks Spec
 
-##### [https://github.com/buildpacks/spec/blob/main/buildpack.md#layer-content-metadata-toml](Layer Content Metadata TOML)
+##### [Layer Content Metadata TOML](https://github.com/buildpacks/spec/blob/main/buildpack.md#layer-content-metadata-toml)
 ```toml
 # whether this layer should be available to at launch (run-time)
 launch = false
@@ -97,9 +100,9 @@ cache = false
 
 # type of cache (if `cache = true`)
 # type: string
-# values: app-specific, buildpack-specific
-# default: app-specific
-cache-type = "app-specific" # <"app-specific" | "buildpack-specific" [default: ]>
+# values: app, shared
+# default: app
+cache-scope = "app"
 
 [metadata]
 # buildpack-specific data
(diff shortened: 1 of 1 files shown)


nebhale:
I don't think this addresses the same target audience as #85.  Specifically this require premeditation on the part of buildpack authors.  #85 puts agency with the application developer giving them as much flexibility as we can, allowing them to do things that buildpack developers have not thought of.

text/0000-cache-types.md
```diff
@@ -0,0 +1,82 @@
...
+When I build a Java app B
+Then I expect similar maven dependencies not to have to be downloaded again
+```
+
+In order to solve this use case in a safe manner, the platform needs to know what cache it can share with other applications versus what cache is specific to a single application. This information is primarily known by the buildpacks.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by associating a "type" to the cache.
```
yaelharel, line 33: If I understand correctly, the buildpack author will decide what type of cache is going to be used, right?
I have a few questions about it:
1. Will the author be able to use both types of cache (each one for different purpose)?
2. If the buildpack author is using cache type x, does it make sense that the buildpack user won't want to use this type of cache? If yes, what will happen in this scenario? (I can't think about such real world scenario but I'm not familiar enough with our use cases)
3. What will happen if the buildpack author doesn't specify the cache type? I assume the default cache is `app-specific`, right?

jromero: 1. Yes, the buildpack author would be able to set a cache type per cache layer.
2. This is an interesting question. This would ultimately be up to the platforms to determine what level of control they want to provide to the end user. As you mention, it's hard to think of a case where this would be necessary since this is ultimately just creating a distinction and not so much changing any functionality.
3. `app-specific` would be the default (I'll update the doc).

ekcasey:
I am +1 on this idea, it solves a real need. But I want it _in addition to_ rather than _instead of_ #85 .

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+# whether this layer should be cached
+# boolean
+# type: boolean
+cache = false
+
+# type of cache (if `cache = true`)
+# type: string
+# values: app, shared
+# default: app
+cache-scope = "app"
```
jkutner, line 105: Maybe `local` instead of `app`? I feel like we should avoid "app"

Do we anticipate other values in the future? Why not maybe this a boolean `cache-shared`?

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+
+Platforms can decide how to _share_ the "shared" cached layers based on their specific use cases.
+
+In `pack`, it being a local development tool, shared cached layers could be implemented as a single global cache that is shared across all applications. Additional options can be provided to restrict or otherwise configure this sharing.
+
+In a different platform, one can foresee that the cache is shared only within a "project" or an "organization".
+
+#### Lifecycle Implementation
+
+1. The lifecycle will provide additional `-shared-cache-dir` and `-shared-cache-image` options where `-cache-dir` and `-cache-image` could be provided.
```
jkutner, line 55: I feel like "global" is better than "shared". But this might also be my bikeshed.

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+2. Layers marked as `cache = true` and `cache-scope = "shared"` will be stored in either `-shared-cached-dir` or `-shared-cache-image`.
+3. If `-share-cached-dir` or `-share-cached-image` are not provide and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standard cache location (`-cache-dir`/`-cache-image`).
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
jkutner, line 60: Are there any security concerns? Like, could there be secrets meant to be used on one buildpack (like a Maven `settings.xml`) that accidentally get shared with another less-trusted buildpack? It seems not that big of a problem since a buildpack can already do just about any thing (you shouldn't be using untrusted buildpacks).

Are there any debugging concerns? Could a buildpack with a bug that creates a malformed `.m2` buildpack cause an error in another buildpack? What if two buildpacks intend to share a cache layer, but they are using two different versions of some shared schema?

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+
+- Requires that the buildpack author be continuous about providing this information for optimization purposes.
```
jkutner, line 62: I don't understand this sentence

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+    - Solution violates the spec dependent on use.
+    - Solution is specific to `pack`.
+    
+# Prior Art
+[prior-art]: #prior-art
+
+None
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
```
jkutner, line 79: If one buildpack need to wipe (`rm -rf`) a cache layer, does that mean other buildpacks will lose it too?

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+When I build a Java app B
+Then I expect similar maven dependencies not to have to be downloaded again
+```
+
+In order to solve this use case in a safe manner, the platform needs to know what cache it can share with other applications versus what cache is specific to a single application. This information is primarily known by the buildpacks.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by associating specifing a scope of "shared".
```
jkutner, line 33:
```suggestion
Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by specifying a scope of "shared".
```

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+
+- `app` - This cached layer is specific to the application and SHOULD NOT be shared with any other applications.
+- `shared` - This cached layer is specific to the buildpack and MAY be shared across multiple applications.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+#### Platform Implementation
+
+Platforms can decide how to _share_ the "shared" cached layers based on their specific use cases.
```
jkutner, line 47: Could a platform opt to ignore shared caching?

text/0000-cache-types.md
edmorley: > What if two buildpacks intend to share a cache layer, but they are using two different versions of some shared schema?

By "two buildpacks" do you mean buildpacks with different `id`s? I'd interpreted the `This cached layer is specific to the buildpack` in the "what it is" section as meaning the cache is still buildpack-specific, in which case this scenario can't occur?

That said, I can see a few other problematic shared-cache scenarios:
1. The shared-cache is used by two different major versions of the same buildpack, where the cache contents is not compatible between them (for example a new version of the yarn buildpack defaulting to newer yarn, which has a different yarn cache schema)
2. The shared-cache contents is dependant on app source input (for example yarn version specified in `package.json`) meaning the shared-cache can be incompatible even with the same version of buildpack used across all apps

To help with (1), shared-caches could be made specific not only to the buildpack ID, but the buildpack version too. 

However:
* this wouldn't help with (2), so buildpack authors will still need to think about cache versioning themselves
* non-shared caches (presumably) aren't tied to buildpack version either, so doing this would be inconsistent

...so it making the cache buildpack-version specific doesn't seem worthwhile to me.

Though it does seem like it would be worth warning buildpack authors about general cache risks in the spec? (eg "Buildpacks SHOULD ensure that use of cached layers is forwards and backwards compatible across buildpack and tooling versions".)

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+
+In `pack`, it being a local development tool, shared cached layers could be implemented as a single global cache that is shared across all applications. Additional options can be provided to restrict or otherwise configure this sharing.
```
dfreilich, line 49:
```suggestion
In `pack`, being a local development tool, shared cached layers could be implemented as a single global cache that is shared across all applications. Additional options can be provided to restrict or otherwise configure this sharing.
```

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+
+In a different platform, one can foresee that the cache is shared only within a "project" or an "organization".
+
+#### Lifecycle Implementation
+
+1. The lifecycle will provide additional `-shared-cache-dir` and `-shared-cache-image` options where `-cache-dir` and `-cache-image` could be provided.
+2. Layers marked as `cache = true` and `cache-scope = "shared"` will be stored in either `-shared-cached-dir` or `-shared-cache-image`.
+3. If `-share-cached-dir` or `-share-cached-image` are not provide and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standard cache location (`-cache-dir`/`-cache-image`).
```
dfreilich, line 57:
```suggestion
3. If `-share-cached-dir` or `-share-cached-image` are not provided and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standard cache location (`-cache-dir`/`-cache-image`).
```

text/0000-cache-types.md
dfreilich: Agreed; did you mean consistent?

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+[what-it-is]: #what-it-is
+
+Buildpacks will provide a mechanism to denote a cached layer as something that can be shared by associating specifing a scope of "shared".
+
+- Target persona: buildpack author
+
+There will be two defined scopes: 
+
+- `app` - This cached layer is specific to the application and SHOULD NOT be shared with any other applications.
+- `shared` - This cached layer is specific to the buildpack and MAY be shared across multiple applications.
```
ekcasey, line 40: If there are only two allowed scopes should we just have a `shared` bool?

text/0000-cache-types.md
ekcasey: I think `global` implies it is shared will all builds but that may not be true depending on how the platform to implement sharing. It could be shared only within a subset of builds owned by a given user, for example. `pack` on the other hand could use the word `global` if that was truly the only scope for sharing.

text/0000-cache-types.md
```diff
@@ -0,0 +1,109 @@
...
+# type: boolean
+cache = false
+
+# type of cache (if `cache = true`)
+# type: string
+# values: app, shared
+# default: app
+cache-scope = "app"
+
+[metadata]
```
ekcasey, line 107: I wonder if `metadata` gets problematic in this situation. The buildpack would need to be careful not to fight with itself about the contents of the metadata. 

E.G it would append new dependencies to the metadata if it adds deps to a shared layers (if it is recording that type of thing) instead of replacing the metadata wholesale with the list of deps relevant to a given build.

text/0000-cache-types.md
ekcasey, line 57:
```suggestion
3. If `-shared-cache-dir` or `-shared-cache-image` are not provided and a buildpack sets a layer to cache with scope `shared`, then it would be cached to the standard cache location (`-cache-dir`/`-cache-image`).
```

text/0000-cache-types.md
jromero: 🤦 `conscious`

jromero:
```suggestion
- Requires that the buildpack author be conscious about providing this information for optimization purposes.
```

text/0000-cache-types.md
jromero: Ditto @ekcasey. In my mind, "global" explicitly means shared with everyone where "shared" leaves it up to interpretation.

text/0000-cache-types.md
jromero: Yes, this goes into how they use the `lifecycle`. 

1. They can omit providing a `-shared-cache-{dir.image}` and the lifecycle would store the cache data in the standard cache.
2. They can `/dev/null` equivalent the `-shared-cache-{dir}`.

jromero:
Some feedback based on the conversations of the [working group](https://youtu.be/tiic9gT3BP8?t=1074):

Minor:
* There is a minor concern about indefinite shared cache size. This can be mediated with `pack` volume management subcommands.

Major:
* Cache poisoning. Although this is a pre-existing risk on something like local development (eg. `.m2`), the concerns are still present and strong. For example, if build for app A provides configuration that then populates the cache with sensitive or compromised resources then all other apps could also be compromised (or leaking sensitive resources).

Alternatives:
* HTTP proxy caching - can `pack` instantiate an HTTP proxy that then caches all HTTP requests? Similar to an enterprise providing a squid proxy for caching resources.
* Named cache volumes - can `pack` provide an option where a user can provide a named cache volume that _they_ can then use in multiple build runs with the understanding that the multiple build runs are related.
* Offline buildpacks - this would extensively resolve caching but given that [Offline Buildpacks RFC](https://github.com/buildpacks/rfcs/pull/81) provides a manifest of dependencies, could `pack` use that manifest to cache those resource separately and provided through other means.

PR #91, opened 2020-07-07
Experimental Features

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/experimental/text/0000-experimental-features.md)

### DIFF
--- a/text/0000-experimental-features.md
+++ b/text/0000-experimental-features.md
@@ -32,7 +32,9 @@ We seek to change the current process such that experimental features do not req
 # How it Works
 [how-it-works]: #how-it-works
 
-Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind a  mechanism that enables or disables these features.
+Experimental Features should have a tracking GitHub issue with an appropriate "experimental" label in the relevant repo. That issue should describe what it is, what it's trying to prove, and a proposed end date of the experiment. The end dates can be extended if needed. When the experiment has ended, it should be removed (if deemed a failure) or a RFC/implementation should be done depending on the feature.
+
+The sub-team that maintains the project can decide with a majority vote whether or not to introduce the feature as long as it is behind a mechanism that enables or disables these features. Where possible when a feature is used it should log a warning after it's used with a message that you used this experimental feature and a link to the tracking issue to collect feedback.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-experimental-features.md
+++ b/text/0000-experimental-features.md
@@ -32,7 +32,7 @@ We seek to change the current process such that experimental features do not req
 # How it Works
 [how-it-works]: #how-it-works
 
-Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind an `--experimental` flag.
+Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind a  mechanism that enables or disables these features.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -75,4 +75,3 @@ Such a plugin mechanism would require a well defined plugin-interface, and versi
 [unresolved-questions]: #unresolved-questions
 
 - How does an experimental features become a real feature (does it use the same flag/command minus the `--experimental` flag)?
-
--- a/text/0000-experimental-features.md
+++ b/text/0000-experimental-features.md
@@ -32,9 +32,15 @@ We seek to change the current process such that experimental features do not req
 # How it Works
 [how-it-works]: #how-it-works
 
-Experimental Features should have a tracking GitHub issue with an appropriate "experimental" label in the relevant repo. That issue should describe what it is, what it's trying to prove, and a proposed end date of the experiment. The end dates can be extended if needed. When the experiment has ended, it should be removed (if deemed a failure) or a RFC/implementation should be done depending on the feature.
+In order to suggest an experimental feature, a tracking GitHub issue is created with the "experimental" label in the relevant repo. The issue should describe what it is, what it's trying to prove, and a proposed end date of the experiment. The end dates can be extended if needed.
 
 The sub-team that maintains the project can decide with a majority vote whether or not to introduce the feature as long as it is behind a mechanism that enables or disables these features. Where possible when a feature is used it should log a warning after it's used with a message that you used this experimental feature and a link to the tracking issue to collect feedback.
+
+When the experiment has ended, there are two possible outcomes:
+- if deemed not successful, the feature should be removed
+- if deemed successful a RFC/implementation should be done depending on the feature
+
+If not much feedback is collected, that will be taken into account but ultimately up to the subteam in how to proceed.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-experimental-features.md
+++ b/text/0000-experimental-features.md
@@ -1,0 +1,78 @@
+# Meta
+[meta]: #meta
+- Name: Space and Process for Experimental Features
+- Start Date: 2020-07-02
+- Author(s): [@jkutner](https://github.com/jkutner)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal for allowing experimental features to enter Pack and Lifecycle without requiring a specification, core team review, or RFC.
+
+# Motivation
+[motivation]: #motivation
+
+End-users and early adopters of CNB have real problems that need to be solved by the CNB project. Sometimes these problems prevent them from adopting CNB in their organizations. Our RFC process, specification guidelines, and governance structure provide a strong mechanism through which we can enact change and solve problems; but sometimes it's too heavy and inhibits experimentation.
+
+We need a way to introduce experiments before writing an RFC. Such a mechanism would allow us to deliver new capabilities faster, improve our means of collecting real end-user feedback, and improve the quality of our RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *experimental feature* - any behavior or options that are likely to change before being added to the specification
+
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features into Pack or Lifecycle. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
+
+We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of Platform Maintainers.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind an `--experimental` flag.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- End users may become tied to features that eventually go away or significantly change
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Plugins
+
+A plugin system for Pack and/or Lifecycle would allow experimental features to be implemented completely outside of the CNB project. For example, a plugin could be installed to Pack with a command like:
+
+```
+$ pack install-plugin skele-gro
+```
+
+Then, a user can execute new commands provided by the plugin:
+
+```
+$ pack grow-bones
+```
+
+Or use new flags on existing commands:
+
+```
+$ pack build --with-bone-growth
+```
+
+Such a plugin mechanism would require a well defined plugin-interface, and versioning of that interface. This is a significantly larger project that the current proposal. However, the current proposal does not prevent a plugin system from being introduced later on.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- Rust's [Unstable features](https://doc.rust-lang.org/rustdoc/unstable-features.html)
+- Java's [`-XX:+UnlockExperimentalVMOptions` flag](https://bugs.openjdk.java.net/browse/JDK-6618726)
+- [oclif Plugins](https://oclif.io/docs/plugins)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How does an experimental features become a real feature (does it use the same flag/command minus the `--experimental` flag)?
+
--- a/text/0000-experimental-features.md
+++ b/text/0000-experimental-features.md
@@ -11,7 +11,7 @@
 # Summary
 [summary]: #summary
 
-This is a proposal for allowing experimental features to enter Pack and Lifecycle without requiring a specification, core team review, or RFC.
+This is a proposal for allowing experimental features without requiring a specification, core team review, or RFC.
 
 # Motivation
 [motivation]: #motivation
@@ -25,7 +25,7 @@ We need a way to introduce experiments before writing an RFC. Such a mechanism w
 
 - *experimental feature* - any behavior or options that are likely to change before being added to the specification
 
-This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features into Pack or Lifecycle. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
 
 We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of the relevant sub-teams maintainers.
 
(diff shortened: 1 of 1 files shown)


text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+[drawbacks]: #drawbacks
+
+- End users may become tied to features that eventually go away or significantly change
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Plugins
+
+A plugin system for Pack and/or Lifecycle would allow experimental features to be implemented completely outside of the CNB project. For example, a plugin could be installed to Pack with a command like:
```
elbandito, line 47: @jkutner is this suggesting we support experimental 3rd party plugins, or do you envision a separate CNB owned repo where these experimental plugins could live?

jkutner: Yea, I was thinking 3rd party (where the most likely 3rd parties would be Salesforce, Paketo, etc). But not a CNB owned repo.

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+End-users and early adopters of CNB have real problems that need to be solved by the CNB project. Sometimes these problems prevent them from adopting CNB in their organizations. Our RFC process, specification guidelines, and governance structure provide a strong mechanism through which we can enact change and solve problems; but sometimes it's too heavy and inhibits experimentation.
+
+We need a way to introduce experiments before writing an RFC. Such a mechanism would allow us to deliver new capabilities faster, improve our means of collecting real end-user feedback, and improve the quality of our RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *experimental feature* - any behavior or options that are likely to change before being added to the specification
+
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features into Pack or Lifecycle. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
```
ekcasey, line 28: I wonder if this particular mechanism for enabling experimental features is better suited to `pack` than the lifecycle. Not all lifecycle features can or should be enabled with flags (think buildpack API features).

I am wondering if experimental API versions as proposed in https://github.com/buildpacks/rfcs/pull/92 would better support this goal at the lifecycle level

text/0000-experimental-features.md
hone: I think a plugin system would be great independent of this RFC.

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+[prior-art]: #prior-art
+
+- Rust's [Unstable features](https://doc.rust-lang.org/rustdoc/unstable-features.html)
+- Java's [`-XX:+UnlockExperimentalVMOptions` flag](https://bugs.openjdk.java.net/browse/JDK-6618726)
+- [oclif Plugins](https://oclif.io/docs/plugins)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How does an experimental features become a real feature (does it use the same flag/command minus the `--experimental` flag)?
```
hone, line 77: Wouldn't it go through a RFC?

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind an `--experimental` flag.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- End users may become tied to features that eventually go away or significantly change
+
```
hone, line 41: The big concern I have with this is creating a set of "hidden" APIs that people depend on if we do decide to take it away.

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+# Alternatives
```
hone, line 42: What about experimental features that use RFCs but just for the subteam?

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+```
+$ pack build --with-bone-growth
+```
+
+Such a plugin mechanism would require a well defined plugin-interface, and versioning of that interface. This is a significantly larger project that the current proposal. However, the current proposal does not prevent a plugin system from being introduced later on.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- Rust's [Unstable features](https://doc.rust-lang.org/rustdoc/unstable-features.html)
```
hone, line 70: FWIW, unstable features in rust (mostly) go through RFCs

text/0000-experimental-features.md
jkutner: i guess i don't mean the process, but the mechanics. Like, do we need to enforce that they are the same flag?

text/0000-experimental-features.md
ekcasey: Personally I think experimental features in the lifecycle should typically (but not always) have RFCs b/c of the lifecycle's place in the overall architecture. However I don't RFCs should be necessary (most of the time) for experimental pack features.

ekcasey: But generally I think we should trust the judgement of maintainers when deciding whether an RFC is a good idea for an experimental feature.

text/0000-experimental-features.md
hone: Why would it need to be the same flag? If an experimental feature "graduates" it could have changes etc. from the feedback/things we learn.

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+We need a way to introduce experiments before writing an RFC. Such a mechanism would allow us to deliver new capabilities faster, improve our means of collecting real end-user feedback, and improve the quality of our RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *experimental feature* - any behavior or options that are likely to change before being added to the specification
+
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features into Pack or Lifecycle. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
+
+We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of Platform Maintainers.
```
dfreilich, line 30:
```suggestion
We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of the relevant sub-teams maintainers.
```

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Plugins
+
+A plugin system for Pack and/or Lifecycle would allow experimental features to be implemented completely outside of the CNB project. For example, a plugin could be installed to Pack with a command like:
+
+```
+$ pack install-plugin skele-gro
```
dfreilich, line 50: 👍  to the HP reference 🧹

text/0000-experimental-features.md
```diff
@@ -0,0 +1,78 @@
...
+- *experimental feature* - any behavior or options that are likely to change before being added to the specification
+
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features into Pack or Lifecycle. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
+
+We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of Platform Maintainers.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind an `--experimental` flag.
```
jromero, line 35: Is this intended to mention our existing mechanism for gating experimental features? I know we've referenced it as an "experimental flag" in conversation but it's currently implemented in `pack` as a config property.

If so I would propose updating the wording to:

```suggestion
Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind a  mechanism that enables or disables these features.
```

nebhale:
@ekcasey Can you please weigh in on the new changes?

text/0000-experimental-features.md
```diff
@@ -0,0 +1,77 @@
...
+- *experimental feature* - any behavior or options that are likely to change before being added to the specification
+
+This proposal affects buildpack users, platform implementors, and project contributors who want to introduce experimental features. Today, any significant change that impacts end-users is introduced through an RFC, which requires a successful vote by the core team before being introduced into the specification (or the Pack platform).
+
+We seek to change the current process such that experimental features do not require an RFC or specification. Instead, an experimental feature would only require approval by a simple majority of the relevant sub-teams maintainers.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Experimental features can be submitted as a PR or discussed in Github Issues on the relevant project. The team that maintains that project can decide whether or not to introduce the feature as long as it is behind a  mechanism that enables or disables these features.
```
hone, line 35: I would also add that this mechanism needs to make it clear this is experimental, can change, etc.

jromero (approved):

dfreilich (approved):

dfreilich:
Definitely agree with @hone's points. Some suggestions to make this a bit more rigorous/help with it:
* We should have a tracking issue in the relevant repo for any given experiment. Something that says what it is, what we are trying to prove, when we hope to know it by, and which allows for feedback. (@hone thought it should be fairly short, like 3 sentences total)
* It could be useful, as such, to have an expiration/suggested expiration date for an experiment. Something like 2 releases in the future, though we could adjust it on a case-by-case basis
* Currently, we don't really have explicit logs if you are using an experimental feature. Perhaps we should have a `warning` log, after a command concludes, with a message that you used **this** experimental feature
*  If we do have more explicit logging, and a tracking issue, we should print the issue as part of the logs, and invite users to give feedback on the issue, or at least tell us whether they used it, and whether they liked it.

As a side note, as part of this we should consider:
*  Make it easier to flip the `experimental` switch in your config (i.e., have a command `pack config experimental`/`pack config experimental true` and `pack config experimental false`), to make this more discoverable for users
* Document experimental features (https://github.com/buildpacks/docs/issues/141), and probably have a tracking document in pack with a list of running experiments

hone:
From conversations with @dfreilich and @jkutner, I'm going to take this over to do some edits. I'm going to move this into draft while I re-work it.

text/0000-experimental-features.md
elbandito: I agree.  This seems like an unsustainable path of where we're having to constantly consider the ramification of making breaking changes to experimental features while developing mainstream features.  After all, we're not exactly a stable project per se.  Maybe once we achieve `1.0.0` is won't be as much of a concern.

text/0000-experimental-features.md
elbandito: I wonder if we need an experimental flag?   I don't know how golang supports external plugins, but I can envision the command just being added to the existing list of commands, and the user would use it as any other command.   We could then mark experimental commands in the help menu or introduce an additional command to list all the experimental commands that have been installed (probably might want to do this anyway so users can quickly identify and remove any experimental commands).  

```
Available Commands:
  build                     Generate app image from source code
  grow-bones (experimental) Some custom experimental thingy
  rebase                    Rebase app image with latest run image
  inspect-image             Show information about a built image
  set-run-image-mirrors     Set mirrors to other repositories for a given run image
  set-default-builder       Set default builder used by other commands
```

text/0000-experimental-features.md
elbandito: Are you saying that the creator of an experimental feature wouldn't have to author an RFC, but the subteams would, as a process by which to accept/bless an experimental feature?

text/0000-experimental-features.md
```diff
@@ -0,0 +1,77 @@
...
+$ pack grow-bones
+```
+
+Or use new flags on existing commands:
+
+```
+$ pack build --with-bone-growth
+```
+
+Such a plugin mechanism would require a well defined plugin-interface, and versioning of that interface. This is a significantly larger project that the current proposal. However, the current proposal does not prevent a plugin system from being introduced later on.
```
elbandito, line 65: This is definitely something to consider.  Especially the work to maintain the integrity of the interface and how much effort we're willing to put into make sure we don't break the interface and any plugins.

elbandito:
> Circling back to this. I'm behind the motivation, but I still have some concerns.
> 
> * Experiments that are experiments forever. At Heroku, we've had features that never shipped for GA, but essentially were GA since so many people ended up using it. Experiments should have an end goal to graduate/sunset.
> * The "mechanism" needs to make it clear to users that this is experimental and unsupported.

I like the idea of having an expiration date associated with an experiment to hold us to either adopting or sunsetting it.   I just wonder what criteria would we follow to determine the end result of an experimental feature?   Do we somehow poll how much it's being used in the community or by how valuable the subteams find it to be?

hone:
> > Circling back to this. I'm behind the motivation, but I still have some concerns.
> > 
> > * Experiments that are experiments forever. At Heroku, we've had features that never shipped for GA, but essentially were GA since so many people ended up using it. Experiments should have an end goal to graduate/sunset.
> > * The "mechanism" needs to make it clear to users that this is experimental and unsupported.
> 
> I like the idea of having an expiration date associated with an experiment to hold us to either adopting or sunsetting it. I just wonder what criteria would we follow to determine the end result of an experimental feature? Do we somehow poll how much it's being used in the community or by how valuable the subteams find it to be?

That's definitely the hard question. I don't want to codify it too much since I'm not on the subteam, and I trust the team will figure out what will work over time vs. my ivory tower.

dfreilich (approved):
🔥 

I like @elbandito's suggestion of explicitly calling out `experimental` in the help test for an experimental command/flag.

hone:
> fire
> 
> I like @elbandito's suggestion of explicitly calling out `experimental` in the help test for an experimental command/flag.

yeah that's great, would you want me to add this to the RFC?

hone:
Would love a re-review since my changes from @ekcasey @nebhale @jromero and not sure but @jkutner should vote somehow.

jkutner:
I vote ✅

text/0000-experimental-features.md
```diff
@@ -0,0 +1,85 @@
...
+[summary]: #summary
+
+This is a proposal for allowing experimental features without requiring a specification, core team review, or RFC.
+
+# Motivation
+[motivation]: #motivation
+
+End-users and early adopters of CNB have real problems that need to be solved by the CNB project. Sometimes these problems prevent them from adopting CNB in their organizations. Our RFC process, specification guidelines, and governance structure provide a strong mechanism through which we can enact change and solve problems; but sometimes it's too heavy and inhibits experimentation.
+
+We need a way to introduce experiments before writing an RFC. Such a mechanism would allow us to deliver new capabilities faster, improve our means of collecting real end-user feedback, and improve the quality of our RFCs.
```
ekcasey, line 21: Doesn't `pack` already have experimental features? I think the UX around pack experimental features could maybe use some improvement, but they do exist.

elbandito (approved):

jromero:
> TBH I am not convinced we need an RFC for this. `pack` already has experimental features. I would be happy to see improvements like:
> 
> 1. label experimental features clearly in the help text
> 2. warning text when experimental features are used
> 3. easier and more discoverable UX to toggle on/off experimental features
> 4. a process for checking in on existing experimental features and ensure we are removing and or promoting them when appropriate
> 
> However, I think we could address the above through a series of pack issues. I am leery about introducing a new/different formal voting process for experimental features. It would add logistical hurdles that feel potentially unnecessary.

I share @ekcasey's sentiment, initially this RFC included lifecycle experimental features which in that vein would require additional discussion. Now that it's been removed I'm failing to see the purpose of this RFC given that it's essentially what we are currently doing. I also agree with the suggested improvements but as mentioned could be tackled as individual pack issues.

The only specific call out I see in this RFC is the formal acknowledgement allowing for "experimental features without requiring a specification, core team review, or RFC".

Another benefit to this RFC is recognizing and discussing alternatives. In which I'd like to emphasize @elbandito's concerns about the added complexity of maintaining an additional interface and distributing the plugins themselves. 

> Especially the work to maintain the integrity of the interface and how much effort we're willing to put into make sure we don't break the interface and any plugins.

We would need to be very methodical on how that would be implemented that it's probably not worth the effort for this specific use case.

hone:
> TBH I am not convinced we need an RFC for this. `pack` already has experimental features. I would be happy to see improvements like:
> 
>     1. label experimental features clearly in the help text
> 
>     2. warning text when experimental features are used
> 
>     3. easier and more discoverable UX to toggle on/off experimental features
> 
>     4. a process for checking in on existing experimental features and ensure we are removing and or promoting them when appropriate
> 
> 
> However, I think we could address the above through a series of pack issues. I am leery about introducing a new/different formal voting process for experimental features. It would add logistical hurdles that feel potentially unnecessary.

@ekcasey I'm happy to cut out any "voting"/formality. The suggestions listed above make up the bulk of my proposal of providing a bit more rigor around "experiments" to ensure both clarity and they are indeed experiments and not secret features that live on forever. These are the things I care the most about.

hone:
@jromero I'm not entirely sure what you're suggesting. I do think the callout of not requiring experimental features to go through the RFC process is a change and the point of this RFC. As far as plugins, I do like that idea, but think they're out of scope of the intent of this RFC.

text/0000-experimental-features.md
hone: I would mark plugins as out of scope.

hone:
@jromero after talking with @ekcasey, let's find time to chat about this RFC. I'm happy to close this if we think it's unnecessary, but my biggest concern is making sure experimental features don't languish/have an end date.

hone:
I met with @jromero and we decided to park/close this RFC for now. While we're on the same page of what I wanted to achieve with this proposal, but it's not something that is needed right now. Instead, we believe this need is worth revisiting after `pack` hits `1.0` where backwards breaking changes will become harder.

PR #92, opened 2020-07-08
Multi API lifecycle descriptor

ekcasey:
[Rendered](https://github.com/buildpacks/rfcs/blob/multi-api-lifecycle-descriptor/text/0000-multi-api-lifecycle-descriptor.md)

### DIFF
--- a/text/0000-multi-api-lifecycle-descriptor.md
+++ b/text/0000-multi-api-lifecycle-descriptor.md
@@ -105,6 +105,8 @@ Experimental API versions may or may not correspond to a spec release.
 
 The behavior of experimental APIs may change between lifecycle versions without notice.
 
+When `CNB_PLATFORM_API` or the `api` field in a `buildpack.toml` file is set to an experimental API the lifecycle will print a warning.
+
 # How it Works
 [how-it-works]: #how-it-works
 
--- a/text/0000-multi-api-lifecycle-descriptor.md
+++ b/text/0000-multi-api-lifecycle-descriptor.md
@@ -150,6 +150,7 @@ Understanding buildpack/platform API support will require more documentation and
 # Alternatives
 [alternatives]: #alternatives
 
+- We could have `deprected` and `supported` lists instead of using `min` and `max` to define ranges.
 - We could stick with a single `api.buildpack` and `api.platform` in the lifecycle descriptor and let platforms infer support ranges based on the guidelines outlined in [RFC-0041](https://github.com/buildpacks/rfcs/blob/main/text/0041-api-version-compat.md#non-breaking-0x-api-versions)
 - Instead of introducing deprecated APIs we could commit to never deprecating or removing support for older APIs
 
--- a/text/0000-multi-api-lifecycle-descriptor.md
+++ b/text/0000-multi-api-lifecycle-descriptor.md
@@ -34,21 +34,24 @@ The following is an example `lifecycle.toml` following the proposed schema:
 [apis.buildpack]
   deprecated = ["1"]
   supported = ["1.2", "2.1"]
-  experimental = "3.0"
+  experimental = ["3.0"]
 [apis.platform]
   deprecated = ["0.4"]
   supported = ["0.4", "0.5", "1.3"]
-  experimental = "2.0"
+  experimental = ["1.4", "2.0"]
 
 [lifecycle]
   version = "0.9.0"
 ```
-* `supproted` contain an array of support API versions:
-  * for version `1.0+`, version `x.n` implies support for `x.0 - x.n`
-  * `supported` should contain all deprecated and non-deprecated versions that the lifecycle supports
+* `supproted` contains an array of support API versions:
+  * for version `1.0+`, version `x.n` implies support for [`x.0`,`x.n`]
+  * `supported` should be a superset of `deprecated`
+  * `supported` should not contain experimental apis
 * `deprecated` contain an array of deprecated APIs:
   * `deprecated` apis should only contain `0.x` or major versions
-* `experimental` describes a single experimental API
+* `experimental` contains an array of experimental API versions
+ * for version `1.0+`, version `x.n` implies experimental support for (`x.y`,`x.n`], where `x.y` is the highest supported minor
+ * if no minor of the given major is supported `x.n` implies experimental support for [`x.0`, `x.n`]
 
 Given lifecycle a lifecycle with the above descriptor file:
 - buildpack API versions `1.0`, `1.1` `1.2`, `2.0`, and `2.1` are supported
@@ -56,7 +59,7 @@ Given lifecycle a lifecycle with the above descriptor file:
 - buildpack API version `3.0` is experimental
 - platform API versions `0.4`, `0.5`, `1.0`, `1.1`, `1.2` and `1.3` are supported
 - platform API version `0.4` is deprecated
-- platform API version `2.0` is experimental
+- platform API versions `1.4` and `2.0` are experimental
 
 ## Lifecycle Labels 
 A builder or lifecycle image with the above descriptor file should have the following labels
@@ -67,12 +70,12 @@ A builder or lifecycle image with the above descriptor file should have the foll
   "buildpack": {
     "deprecated": ["1"],
     "supported": ["1.2", "2.4"],
-    "experimental": "3.0"
+    "experimental": ["3.0"]
   },
   "platform": {
     "deprecated": ["0.4"],
     "supported": ["0.4", "0.5", "1.3"],
-    "experimental": "2.0"
+    "experimental": ["1.4", "2.0"]
   }
 }
 ```
@@ -154,7 +157,6 @@ Understanding buildpack/platform API support will require more documentation and
 # Alternatives
 [alternatives]: #alternatives
 
-- We could have `deprected` and `supported` lists instead of using `min` and `max` to define ranges.
 - We could stick with a single `api.buildpack` and `api.platform` in the lifecycle descriptor and let platforms infer support ranges based on the guidelines outlined in [RFC-0041](https://github.com/buildpacks/rfcs/blob/main/text/0041-api-version-compat.md#non-breaking-0x-api-versions)
 - Instead of introducing deprecated APIs we could commit to never deprecating or removing support for older APIs
 
@@ -167,8 +169,6 @@ Understanding buildpack/platform API support will require more documentation and
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Should the lifecycle warn when an experimental API is used?
-- Should we support a range of experimental APIs instead of a single API?
 - Should we require platforms to turn on support for experimental APIs with an environment variable?
 
 # Spec. Changes (OPTIONAL)
--- a/text/0000-multi-api-lifecycle-descriptor.md
+++ b/text/0000-multi-api-lifecycle-descriptor.md
@@ -11,7 +11,7 @@
 # Summary
 [summary]: #summary
 
-This RFC proposed three levels of API support: `deprecated`, `supported`, and `experimental`. Supported APIs will be expressed in ranges.
+This RFC proposed three levels of API support: `deprecated`, `supported`, and `experimental`. Supported APIs will be expressed as arrays.
 
 This RFC introduces a new lifecycle descriptor file format, and an analogous label that should be applied to builder and lifecycle images.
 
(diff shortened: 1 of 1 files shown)


text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,172 @@
...
+    }
+  },
+  "stack": {},
+}
+```
+For backwards compatibility `pack create-builder` should continue to populate the above label, setting `lifecycle.api.buildpack`, and `lifecycle.api.platform` to the corresponding supported `min`.
+
+This will allow platforms the read the metadata in it's existing format to update gracefully. Eventually this compatibility layer will be removed.
+
+The lifecycle descriptor file itself will not preserve backwards compatibility. Most platforms consume builders rather than directly consuming lifecycle artifacts, and therefore the builder label is the important compatibility concern. Those platforms that do consume lifecycle artifacts control which versions of the lifecycle they bring in and can update their behavior to parse the new format for newer lifecycles.
```
hone, line 143: afaik `lifecycle.toml` isn't part of the spec. Do we actively tell platforms to not consume this file?

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,172 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Understanding buildpack/platform API support will require more documentation and education.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- We could have `deprected` and `supported` lists instead of using `min` and `max` to define ranges.
```
hone, line 153: Just thinking out loud, but I assume there's never a point where we would deprecate a API in the middle of a range b/c we messed up badly.

text/0000-multi-api-lifecycle-descriptor.md
hone, line 143: Do we actively discourage platforms from using `lifecycle.toml`?

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,172 @@
...
+[prior-art]: #prior-art
+
+* Experimental APIs are similar to [alpha level APIs](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning) in Kubernetes.
+* https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the lifecycle warn when an experimental API is used?
+- Should we support a range of experimental APIs instead of a single API?
```
hone, line 167: I'm not sure we'd have multiple versions. Is experimental a version or a mode?

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,172 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+* Experimental APIs are similar to [alpha level APIs](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning) in Kubernetes.
+* https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the lifecycle warn when an experimental API is used?
```
hone, line 166: I lean towards yes, just to be extra clear that we don't guarantee it won't change in the next version of the lifecycle.

hone:
Overall, :+1: on this idea.

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,172 @@
...
+
+In addition, we should have a mechanism for deprecating an entire API, so that buildpack and platform authors are not caught off guard when their platform or buildpack no longer works with a newer lifecycle. We likely will not choose to deprecate any APIs in the near future but building API deprecation into any new design prepares us well for the future.
+
+Experimental APIs will give platform and buildpack authors an opportunity to try out unreleased APIs and give us feedback. This will be valuable as we prepare for the `1.0` versions of the APIs. We want to get the design of the `1.0` APIs as close to "right" as possible because `1.n` APIs will not permit breaking changes. Feedback will help us improve the design before we finalize it. The quality of feedback we get from user about their real life usage of experimental APIs will help us avoid problems we may miss when approving RFCs.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Lifecycle Descriptor
+The following is an example `lifecycle.toml` following the proposed schema:
```
jkutner, line 31: I'm confused about how/where this file is used.

ekcasey: `pack create-builder` uses it to set metadata on the builder and to validate that a builder has compatible lifecycle/buildpacks.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: `lifecycle.toml` is only relevant if a platform is creating a builder image. Given that we haven't yet specified builders or builder creation I think you could call that discouragement.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: versions are modes right? Not sure I follow the question.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: Yeah, I thought about this and I feel fairly confident we would never want to deprecate a API in the middle of a range. I think that would cause a major headache for end users and I can't foresee us wanting to anyways.

hone: I agree, I just wanted to throw the idea out there.

text/0000-multi-api-lifecycle-descriptor.md
hone: I just mean a generic `experiment` version vs an explicit number.

text/0000-multi-api-lifecycle-descriptor.md
hone: @ekcasey where do these "warnings" go? This is more interesting for the buildpack API since either the user running the buildpack would see it or the builder creator. Can you elaborate more how you see the warnings working? (I might just be missing something obvious).

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: I like an explicit number better so that when a new lifecycle lands and the experiment becomes supported you don't need to change the value of `CNB_PLATFORM_API` to avoid weird behavior (i.e. when there is no experiment or it's a totally new API).

hone: is that worse b/c if that new number isn't an experiment we don't promise that it will behave the same?

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: I think that the lifecycle, before using a group that contains an experimental buildpack, should check the APIs and print warning if necessary. I am inclined to say we should nor print warnings unless lifecycle actually interacts with the buildpack in question (e.g. if the first group detects, no need to warn about the second group).

I think whether a given platform wants to print a warning when creating a builder is outside the scope of this RFC.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: I think it is better. If we are promoting an experimental API to a supported API the diff between the behavior of the final iteration of experimental version and the supported version is likely to small (certainly smaller than the diff between that and the next experimental version, if there even is one).

hone: But based on our guarantees we're still implying there could be a difference and that's just the cost of using the experimental API even from `1.0-alpha1` vs `1.0-alpha2`. My thoughts that if someone doesn't even know that upgrading where they're now suddenly on a stable API could be pretty surprising.

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,176 @@
...
+  experimental = ["3.0"]
+[apis.platform]
+  deprecated = ["0.4"]
+  supported = ["0.4", "0.5", "1.3"]
+  experimental = ["1.4", "2.0"]
+
+[lifecycle]
+  version = "0.9.0"
+```
+* `supproted` contains an array of support API versions:
```
hone, line 46:
```suggestion
* `supported` contains an array of support API versions:
```

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,176 @@
...
+The following is an example `lifecycle.toml` following the proposed schema:
+```toml
+[apis]
+[apis.buildpack]
+  deprecated = ["1"]
+  supported = ["1.2", "2.1"]
+  experimental = ["3.0"]
+[apis.platform]
+  deprecated = ["0.4"]
+  supported = ["0.4", "0.5", "1.3"]
```
yaelharel, line 40: I'm confused. Is it a mistake that `0.4` is both "deprecated" and "supported"?

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,176 @@
...
+  experimental = ["1.4", "2.0"]
+
+[lifecycle]
+  version = "0.9.0"
+```
+* `supproted` contains an array of support API versions:
+  * for version `1.0+`, version `x.n` implies support for [`x.0`,`x.n`]
+  * `supported` should be a superset of `deprecated`
```
yaelharel, line 48: It is because `We likely will not choose to deprecate any APIs in the near future but building API deprecation into any new design prepares us well for the future.`, right?

Moreover, according to your example, under `apis.buildpack`, if `1` is deprecated, `1` should be also `supported` (and not only `1.2`), right?

text/0000-multi-api-lifecycle-descriptor.md
jkutner: does that mean `pack create-builder` would have two files as input (with `builder.toml` being the second)? Could those two files be merged?

ekcasey:
Final Comment Period with merge disposition, closing on 30 July, 2020.

text/0000-multi-api-lifecycle-descriptor.md
```diff
@@ -0,0 +1,176 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+* Experimental APIs are similar to [alpha level APIs](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning) in Kubernetes.
+* https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we require platforms to turn on support for experimental APIs with an environment variable?
```
yaelharel, line 172: Why would we need an environment variable for that?

hone: I was thinking that or have experimental versions not just be numbers, i.e. `1.0-alpha`

text/0000-multi-api-lifecycle-descriptor.md
jkutner: >  the diff between the behavior of the final iteration of experimental version and the supported version is likely to small

I don't want to hold ourselves to this. We should be able to make dramatic changes in behavior b/w experiment and official feature

text/0000-multi-api-lifecycle-descriptor.md
natalieparellano: `pack` when creating a builder expects to find the `lifecycle.toml` inside the lifecycle blob. The user doesn't need to specify it.

text/0000-multi-api-lifecycle-descriptor.md
hone: Following up on this from the WG meeting today. We settled on allowing both "less" experimental APIs like platform `0.4` where the feature set is much more well known as well as a `1.0` but can designate with alphanumeric characters like "1.0-alpha1". This means extending the API versioning beyond `X.Y` to `X.Y-<alphanumeric>` but only for experimental versions.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: In this example the deprecated `1` refers to the entire `1.x` line:
- `1.2` in `supported` means `1.0-1.2` are supported
- `1` in `deprecated` means all supported `1.x` APIs (i.e. `1.0-1.2`) are deprecated

Therefore `supported` is a superset of deprecated semantically, even though the strings are not literally the same.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: I removed experimental APIs entirely. I think we should punt this discussion to a future RFC.

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: Resolving because experimental has been removed entirely

text/0000-multi-api-lifecycle-descriptor.md
ekcasey: Resolving because experimental has been removed entirely

text/0000-multi-api-lifecycle-descriptor.md
yaelharel: I know that `1` refers to the entire `1.x` but I didn't know that if `1.2` is `supported` it means that `1.0-1.2` are all `supported`.
A few questions about it:
1. Did you write that if `x.y` is `supported`, it means that `x.0-x.y` are `supported`?
2. I'm afraid that there might be a new version `1.3` and then it will be part of the `deprecated` and not part of the `supported`. Can it happen?
3. Does `supported` have to be a superset of `deprecated` because `We likely will not choose to deprecate any APIs in the near future`? Otherwise I don't understand why are we doing that.

ekcasey: > 1. Did you write that if x.y is supported, it means that x.0-x.y are supported?

Yes. That is what  
>  * for version `1.0+`, version `x.n` implies support for [`x.0`,`x.n`]`

is attempting to convey.

> 2. I'm afraid that there might be a new version 1.3 and then it will be part of the deprecated and not part of the supported. Can it happen?

Because `1.2` is the max supported and deprecated must be a subset of supported, `1` in deprecated implies `1.0 - 1.2`.

> 3. Does supported have to be a superset of deprecated because We likely will not choose to deprecate any APIs in the near future? Otherwise I don't understand why are we doing that.

In the original RFC they were disjoint sets, but @sclevine made the valid point that deprecated APIs are in fact "supported" in that they will run successfully. Now `deprecated` is a special type of support, which I think makes sense.

PR #93, opened 2020-07-08
pack subcommands

jromero:
[Readable](https://github.com/buildpacks/rfcs/blob/pack-subcommands/text/0000-pack-subcommands.md)

### DIFF
--- a/text/0000-pack-subcommands.md
+++ b/text/0000-pack-subcommands.md
@@ -96,10 +96,7 @@ The `config` subcommand would work with the notion of two different type of reso
 
 ##### Singular
 
-Singular resources would have the following subcommands:
-
-* set
-* (implied get)
+Singular resources would have no subcommands but accept possitional arguments and flags with an implied `get` operation and a common `--unset` flag.
 
 ###### Example
 
--- a/text/0000-pack-subcommands.md
+++ b/text/0000-pack-subcommands.md
@@ -1,0 +1,186 @@
+# Meta
+[meta]: #meta
+- Name: `pack` subcommands
+- Start Date: 2020-07-08
+- Author(s): @jromero
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+Update the `pack` CLI to use more common subcommand patterns.
+
+# Motivation
+[motivation]: #motivation
+
+As the number of subcommands for `pack` grow, it's becoming more difficult to identify interworking subcommands at first glance. See the current list of subcommands below.
+
+```
+pack
+
+  build                 Generate app image from source code
+  rebase                Rebase app image with latest run image
+  inspect-image         Show information about a built image
+  set-run-image-mirrors Set mirrors to other repositories for a given run image
+  set-default-builder   Set default builder used by other commands
+  inspect-builder       Show information about a builder
+  suggest-builders      Display list of recommended builders
+  trust-builder         Trust builder
+  untrust-builder       Stop trusting builder
+  list-trusted-builders List Trusted Builders
+  create-builder        Create builder image
+  package-buildpack     Package buildpack in OCI format.
+  suggest-stacks        Display list of recommended stacks
+  version               Show current 'pack' version
+  report                Display useful information for reporting an issue
+  completion            Outputs completion script location
+  help                  Help about any command
+```
+
+# What it is
+[what-it-is]: #what-it-is
+
+
+### Proposed structure
+
+```
+pack
+
+  app
+    build
+    inspect
+    rebase
+  builder
+    create
+    inspect
+    suggest
+  buildpacks
+    package
+    register
+  completion
+  config
+    default-builder
+      set
+      unset
+      show
+    run-image-mirror
+      add
+      remove
+      list
+    registry
+      add
+      remove
+      list
+    trusted-builder
+      add
+      remove
+      list
+  help
+  report
+  stacks
+    suggest
+  version
+```
+
+### Top-Level Aliases
+
+There would also be some alias subcommands for the `app` set of subcommands for easier usage and compatibility.
+
+* `pack build` -> `pack app build`
+* `pack inspect` -> `pack app inspect`
+* `pack rebase` -> `pack app rebase`
+
+### Config
+
+The `config` subcommand would work with the notion of two different type of resources:
+
+##### Singular
+
+Singular resources would have the following subcommands:
+
+* set
+* unset
+* show (aliased to subcommand root)
+
+###### Example
+
+```
+# sets the default builder
+pack config default-builder set gcr.io/my-org/my-builder:latest
+
+# unset the default builder
+pack config default-builder unset
+
+# show the default builder
+pack config default-builder show
+
+# show the default builder (w/o subcommand)
+pack config default-builder
+```
+
+##### List
+
+List resources would have the following subcommands:
+
+* add
+* remove
+* list
+
+The explicit additional parameters would be based on the specific implementation needs.
+
+###### Example
+
+```
+# adds `gcr.io/my-org/my-image` as a mirror for `my-image`  
+pack config run-image-mirrors add my-image gcr.io/my-org/my-image
+
+# removes `gcr.io/my-org/my-image` as a mirror for `my-image`
+pack config run-image-mirrors remove my-image gcr.io/my-org/my-image
+
+# removes all mirrors for `my-image`
+pack config run-image-mirrors remove my-image
+
+# list all mirrors
+pack config run-image-mirrors list
+
+# list all mirrors for `my-image`
+pack config run-image-mirrors list my-image
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+
+## Backwards Compatibility
+
+1. There would be a period in which all pre-existing subcommands would work.
+2. [Top-level aliases](#top-level-aliases) will provide continued compatibility for common operations.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Change is bad... mkay.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep it as it is.
+- Some [Config](#config) subcommands could be more intertwined with their respective resource. For example, `pack builder trust ...` instead of `pack config trust-builder ...`.
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Are there better alternatives to the name "app"?
+
+# Spec. Changes
+[spec-changes]: #spec-changes
+
+None
--- a/text/0000-pack-subcommands.md
+++ b/text/0000-pack-subcommands.md
@@ -58,7 +58,6 @@ pack
   build
   builder
     create
-    inspect
     suggest
   buildpack
     package
--- a/text/0000-pack-subcommands.md
+++ b/text/0000-pack-subcommands.md
@@ -46,13 +46,16 @@ pack
 
 ### Proposed structure
 
+The pack commands will be reorganized in a way where they are grouped by resource type where possible. Additionally, a new
+command `config` will group subcommands that solely operate on mutating the config file.
+
+
+An overview of the (sub)commands:
+
 ```
 pack
 
-  app
-    build
-    inspect
-    rebase
+  build
   builder
     create
     inspect
@@ -79,19 +82,13 @@ pack
       remove
       list
   help
+  inspect
   report
+  rebase
   stack
     suggest
   version
 ```
-
-### Top-Level Aliases
-
-There would also be some alias subcommands for the `app` set of subcommands for easier usage and compatibility.
-
-* `pack build` -> `pack app build`
-* `pack inspect` -> `pack app inspect`
-* `pack rebase` -> `pack app rebase`
 
 ### Config
 
@@ -167,13 +164,36 @@ pack config run-image-mirrors list my-image
 # Alternatives
 [alternatives]: #alternatives
 
-- Keep it as it is.
+### Config alternatives
+
+- Only allow manual edits of config file.
+    - A `config` command would allow providing `help` for possible operations and argument values.
 - Some [Config](#config) subcommands could be more intertwined with their respective resource. For example, `pack builder trust ...` instead of `pack config trust-builder ...`.
+    - Other commands are operations on the resource whereas having `config` subcommands would make it more obvious that
+    `pack` is solely mutating local configuration.
+
+### App / Project / Image images commands
+
+- Instead of having top-level commands `build`, `inspect`, `rebase` they can be associated with an `app`/`project`/`image` resource.
+    - There would also be some alias subcommands for the `app` set of subcommands for easier usage and compatibility.
+        - `pack build` -> `pack app build`
+        - `pack inspect` -> `pack app inspect`
+        - `pack rebase` -> `pack app rebase`
+    - The consensus appears to be NOT to provide a resource name based on the confusion it can impose depending on usage.
+        - A few examples: "Is a function image an _app_?", "A _project_ may be composed of multiple images, am I operating on a single or multiple images?", "What type of _image_ am I building?".
 
 # Prior Art
 [prior-art]: #prior-art
 
-Discuss prior art, both the good and bad.
+* colon (`:`) command delimiters
+    * https://oclif.io/
+* nested subcommands
+    * http://docopt.org/
+    * https://picocli.info/#_nested_sub_subcommands
+    * https://oclif.io/
+* config
+    * https://git-scm.com/docs/git-config
+    * https://cloud.google.com/sdk/gcloud/reference/config
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-pack-subcommands.md
+++ b/text/0000-pack-subcommands.md
@@ -154,7 +154,7 @@ pack config run-image-mirrors list my-image
 ## Backwards Compatibility
 
 1. There would be a period in which all pre-existing subcommands would work.
-2. [Top-level aliases](#top-level-aliases) will provide continued compatibility for common operations.
+2. Critical operations (`build`, `rebase`) will remain unchanged.
 
 # Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+
+  app
+    build
+    inspect
+    rebase
+  builder
+    create
+    inspect
+    suggest
+  buildpacks
```
ryanmoran, line 60: This noun is plural when the others are singular. Is that intentional?

jromero:
```suggestion
  buildpack
```

It's not. I kept waffling. I forgot which one @ekcasey prefers.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+      add
+      remove
+      list
+    trusted-builder
+      add
+      remove
+      list
+  help
+  report
+  stacks
```
jromero, line 83:
```suggestion
  stack
```

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+      remove
+      list
+  help
+  report
+  stack
+    suggest
+  version
+```
+
+### Top-Level Aliases
```
jkutner, line 88: I'm used to calling these "topics" like in [oclif](https://oclif.io/docs/topics)

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep it as it is.
+- Some [Config](#config) subcommands could be more intertwined with their respective resource. For example, `pack builder trust ...` instead of `pack config trust-builder ...`.
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
```
jkutner, line 176: https://oclif.io/

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+1. There would be a period in which all pre-existing subcommands would work.
+2. [Top-level aliases](#top-level-aliases) will provide continued compatibility for common operations.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Change is bad... mkay.
+
+# Alternatives
+[alternatives]: #alternatives
```
jkutner, line 168: We could use a `pack <topic>:<command>` pattern instead of space separated.

hone: I do think `:` makes it clearer vs an arg.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+
```
hone, line 169: Another alternative is to focus "pack" more for specific audiences which would reduce the number of commands that it would have to know.

text/0000-pack-subcommands.md
hone: docker does commands and "management" commands

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Are there better alternatives to the name "app"?
```
hone, line 181: image?

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+  suggest-stacks        Display list of recommended stacks
+  version               Show current 'pack' version
+  report                Display useful information for reporting an issue
+  completion            Outputs completion script location
+  help                  Help about any command
+```
+
+# What it is
+[what-it-is]: #what-it-is
+
```
hone, line 45:
```suggestion
This RFC proposes re-organizing the `pack` CLI to use the subcommand pattern.
```

It'd be good to have some text before proposing the structure.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Change is bad... mkay.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep it as it is.
+- Some [Config](#config) subcommands could be more intertwined with their respective resource. For example, `pack builder trust ...` instead of `pack config trust-builder ...`.
```
hone, line 171: Another alternative is that `pack config` could be a command where the argument is the first resource.

hone:
I do agree with the motivations behind this RFC. Probably due to using the Heroku CLI, I have a preference for not using space and it's clearer to me.

text/0000-pack-subcommands.md
jromero: I'm not keen on image because on the technical side almost every artifact we use is an image. Additionally, some args are named "image-name" which refer to OCI images in general and not the "resulting image produced from a pack build".

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+  trust-builder         Trust builder
+  untrust-builder       Stop trusting builder
+  list-trusted-builders List Trusted Builders
+  create-builder        Create builder image
+  package-buildpack     Package buildpack in OCI format.
+  suggest-stacks        Display list of recommended stacks
+  version               Show current 'pack' version
+  report                Display useful information for reporting an issue
+  completion            Outputs completion script location
+  help                  Help about any command
```
ekcasey, line 40: Aside: even w/o subcommand the current list would be a lot easier to navigate if it was alphabetized.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+    suggest
+  version
+```
+
+### Top-Level Aliases
+
+There would also be some alias subcommands for the `app` set of subcommands for easier usage and compatibility.
+
+* `pack build` -> `pack app build`
+* `pack inspect` -> `pack app inspect`
```
ekcasey, line 93: I like the `pack build` and `pack rebase` aliases but I think having a `pack inspect` that can't inspect builders would be confusing. I instead of making it smarter (which might fall apart if someday we can inspect things that aren't images) I think we should just remove this alias.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+    suggest
+  buildpack
+    package
+    register
+  completion
+  config
+    default-builder
+      set
+      unset
+      show
```
ekcasey, line 68: Feel free to ignore this bikeshed: I prefer `get` to `show` because it makes a natural pairing with `set`

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,186 @@
...
+pack config default-builder
+```
+
+##### List
+
+List resources would have the following subcommands:
+
+* add
+* remove
+* list
```
ekcasey, line 130: Do we want to alias things like `pack config trusted-builders` to `pack config trusted-builder list`?

text/0000-pack-subcommands.md
ekcasey: I disagree, I feel like the vast majority of CLI tools that I like use space separate subcommands and most users expect it. Also, just from a practical perspective `pack` uses [cobra](https://github.com/spf13/cobra) which AFAIK only supports space separated subcommands.

text/0000-pack-subcommands.md
ekcasey: That might get tricker b/c some of the config options have context specific usage. Example: setting a run-image-mirror requires and extra arg because you need to know what run-image you are mirroring.

text/0000-pack-subcommands.md
ekcasey: I am +1 on "app"

In the spec we ended up going with "app image" everywhere for clarity. I know at one point we wanted to steer clear of "app" but I am now of the opinion that not using `app` just makes everything more confusing (even if some images aren't strictly speaking "apps").

text/0000-pack-subcommands.md
jromero: Added oclif as prior art with mention to the colon seperator.

text/0000-pack-subcommands.md
jromero: What would that look like?

Assuming `pack` remains the platform I could foresee the following commands:

```
pack

  build
  config
     ...
  inspect
  run
```

The alternative utility CLI would be everything else:

```
bputils

  builder
    create
    inspect
    suggest
  buildpack
    create
    package
    register
  stack
    suggest
```

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,206 @@
...
+    create
+    inspect
+    suggest
+  buildpack
+    package
+    register
+  completion
+  config
+    default-builder
+      set
```
jkutner, line 69: so for this one, the command would be `pack config default-builder set <value>`?

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,206 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+* colon (`:`) command delimiters
+    * https://oclif.io/
+* nested subcommands
+    * http://docopt.org/
+    * https://picocli.info/#_nested_sub_subcommands
+    * https://oclif.io/
```
dfreilich, line 193:
```suggestion
    * https://oclif.io/
    * kubectl
```

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,206 @@
...
+pack config run-image-mirrors list my-image
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+
+## Backwards Compatibility
+
+1. There would be a period in which all pre-existing subcommands would work.
```
yaelharel, line 156: Maybe it's worth adding that a warning should appear once running the pre-existing subcommands.

text/0000-pack-subcommands.md
jromero: Based on some discussions I think we want to move to where `pack inspect` can auto-detect the image resource type and display the appropriate information. For example, `pack inspect <builder>` will display builder info, `pack inspect <app-image>` will display app image information, etc.

text/0000-pack-subcommands.md
jromero:
```suggestion
1. There would be a period in which all pre-existing subcommands would work (with a deprecation warning).
```

text/0000-pack-subcommands.md
jromero:
```suggestion
[alternatives]: #alternatives

### Colon separated commands

> We could use a `pack <topic>:<command>` pattern instead of space separated.
```

text/0000-pack-subcommands.md
jromero:
```suggestion

### Alternative Utility

> Another alternative is to focus "pack" more for specific audiences which would reduce the number of commands that it would have to know.

<code>
pack

  build
  config
     ...
  inspect
  run
</code>

The alternative utility CLI would be everything else:

<code>
bputils

  builder
    create
    inspect
    suggest
  buildpack
    create
    package
    register
  stack
    suggest
</code>

```

text/0000-pack-subcommands.md
jromero:
```suggestion
      get
```

text/0000-pack-subcommands.md
jromero: Yes, that's correct.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,206 @@
...
+
+The `config` subcommand would work with the notion of two different type of resources:
+
+##### Singular
+
+Singular resources would have the following subcommands:
+
+* set
+* unset
+* show (aliased to subcommand root)
```
jromero, line 103:
```suggestion
* get (aliased to subcommand root)
```

text/0000-pack-subcommands.md
jromero:
```suggestion
* list (aliased to subcommand root)
```

text/0000-pack-subcommands.md
jkutner: Can we consider `pack config default-builder=<value>`?

jromero:
@buildpacks/core-team Given that this only applies to pack and has 2 out of 3 positive votes from the Platform Maintainers can we motion for FCP?

text/0000-pack-subcommands.md
jromero: Is there a proposed similar solution to unsetting? Alternative to `pack config default-builder unset`.

Maybe, since it's an uncommon operation, it can simply be a flag?

```bash
# set resource
pack config default-builder=<builder-name>

# unset resource
pack config default-builder --unset

# get
pack config default-builder
```

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,242 @@
...
+[drawbacks]: #drawbacks
+
+- Change is bad... mkay.
+
+# Alternatives
+[alternatives]: #alternatives
+
+### Colon separated commands
+
+> We could use a `pack <topic>:<command>` pattern instead of space separated.
```
elbandito, line 169: I personally find space delimited commands to be less intuitive, and thus harder for the user to learn without having to constantly use `--help`.  This was something we also encountered at Heroku.  However, I understand that most of the golang CLI community uses space delimited sub-commands.  

See [11. Be clear about subcommands](https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46)

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,242 @@
...
+  buildpack
+    package
+    register
+  completion
+  config
+    default-builder
+      set
+      unset
+      get
+    run-image-mirror
```
elbandito, line 72: Some off the cut thought, as something else to consider.  Make the `List` resource types plural like:

`registries`
`trusted-builders`
`run-image-mirrors`

Then remove having to explicitly type `list` and have that be the default behavior when no args are passed.

For example, the following would by default list all the registries and trusted builders:
`pack registries` 
`pack trusted-builders`

`add` and `remove` would behave the same.   Something to think about.

text/0000-pack-subcommands.md
hone: @jromero is this a change you're hoping to make in the RFC?

ekcasey: I personally prefer `pack config default-builder set <value>` to `pack config default-builder=<builder-name>`. Using the verb allow for more natural/consistent extension if we ever want to  include other verbs. For example imagine a future where we want additional operations on set (in the other sense) values like `pack config trusted-builders add <value>`.

ekcasey: If other folks _really_ don't like the verbs I will concede but even in that case I would prefer `pack config default-builder <builder-name>` to `pack config default-builder=<builder-name>`. Having an `=` in a positional argument bugs me for some reason.

hone: @ekcasey I think that's b/c in unix tooling you usually see it on flags and not position args. Take `ls` for instance:

```
--format=WORD
```

Malax (approved):

text/0000-pack-subcommands.md
jromero: This was discussed during WG and there still seems to be a majority leaning towards it being space delimited.

text/0000-pack-subcommands.md
jromero: Suggestions applied.

text/0000-pack-subcommands.md
jromero: Changes in this regard have been made to the RFC. Please re-review.

text/0000-pack-subcommands.md
jromero: Made my best attempt at trying to prototype what this could look like [here](https://github.com/buildpacks/rfcs/blob/pack-subcommands/text/0000-pack-subcommands.md#colon-separated-commands). Wasn't able to come up with a favorable format IMO.

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,275 @@
...
+    registries
+      add
+      remove
+      list
+    trusted-builders
+      add
+      remove
+      list
+  help
+  inspect
```
dfreilich, line 81: We currently have both `pack inspect-image` and `pack inspect-builder`. What would you envision with this `inspect` command?

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,275 @@
...
+  version
+```
+
+### Config
+
+The `config` subcommand would work with the notion of two different type of resources:
+
+##### Singular
+
+Singular resources would have no subcommands but accept possitional arguments and flags with an implied `get` operation and a common `--unset` flag.
```
dfreilich, line 95:
```suggestion
Singular resources would have no subcommands but accept positional arguments and flags with an implied `get` operation and a common `--unset` flag.
```

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,275 @@
...
+
+```
+# sets the default builder
+pack config default-builder gcr.io/my-org/my-builder:latest
+
+# unset the default builder
+pack config default-builder --unset
+
+# gets the default builder
+pack config default-builder
```
dfreilich, line 107: If there is none set, what should the default behavior be? Should it run `pack suggest-builders`?

text/0000-pack-subcommands.md
```diff
@@ -0,0 +1,275 @@
...
+  stack
+    suggest
+```
+
+### Config alternatives
+
+- Only allow manual edits of config file.
+    - A `config` command would allow providing `help` for possible operations and argument values.
+- Some [Config](#config) subcommands could be more intertwined with their respective resource. For example, `pack builder trust ...` instead of `pack config trust-builder ...`.
+    - Other commands are operations on the resource whereas having `config` subcommands would make it more obvious that
```
dfreilich, line 240: I like this point a lot. I think it makes the UX a bit more sensible, though sometimes wordy.

dfreilich (approved):
Overall, definitely a fan. I think there was a bit of a regression with `inspect`, so I'd appreciate a clarification there, but 🥳

sclevine (approved):

hone:
Moving this to Final Comment Period to close on Sept. 9, 2020.

PR #94, opened 2020-07-10
Layer origin metadata

pclalv:
[readable](https://github.com/pclalv/rfcs/blob/layer-origin-metadata/text/0000-layer-origin-metadata.md)

obviates https://github.com/buildpacks/rfcs/pull/63


text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+<!-- One paragraph explanation of the feature. -->
+
+Exposing layer origin metadata in `report.toml` will improve layer
```
ekcasey, line 16: Maybe we could use the term "traceability" instead of "reproducibility" throughout this document to avoid potential confusion with the other (related but subtly distinct) meaning of "build reproducibility"? https://github.com/buildpacks/spec/pull/111

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+previous-image-digest = "97e47fb9e0a6"
+
+[[layers]]
+buildpack = "yabba"
+name = "build"
+on-export = "reused"
+digest = "1742affe03b5"
+modification-date = "2020-07-09 12:41:35-04:00"
+size = 17813
+previous-layer-digest = "c54bba046158"
```
ekcasey, line 95: Am I correct in interpreting `previous-layer-digest` as the digest of the restored version of this layer?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+
+cached = true
```
ekcasey, line 97:
```suggestion
cache = true
```
For consistency with `layer.toml`

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+launch = false
+build = true
+
```
ekcasey, line 100: Do we want to include the regular old `[layers.metadata]` in `report.toml` as well?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+[[layers.application-source]]
+type = "git"
+
+[[layers.application-source.version]]
+sha = "26e8c05b151f57161d64b7a5f844bb04caed0ba3"
```
ekcasey, line 105: In order to provide correct values for fields like this in the case of reused cached layers we will need to store layer origin metadata in the cache. I am not sure we need to describe that mechanism in this RFC but it seems worth pointing out.

ekcasey: What concerns me more is how we provide this information for `cache=false` layers. We could still store origin metadata about those layers in the cache (even if we don't store the entire layer) but depending on cache availability there may be cases where we wish to reuse a layer from the previous app image and we don't have cache origin metadata available.

Maybe we should simply avoid setting these fields when the data is unavailable?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+[how-it-works]: #how-it-works
+
+<!-- This is the technical portion of the RFC, where you explain the design -->
+<!-- in sufficient detail. -->
+
+<!-- The section should return to the examples given in the previous -->
+<!-- section, and explain more fully how the detailed proposal makes those -->
+<!-- examples work. -->
+
+The **builder lifecycle phase should record that it's invoking a
```
ekcasey, line 128: I think it is actually the `exporter` that will need to populate this data. The exporter has access to all of the relevant information (the cache, previous image metadata, `project-metadata.toml`) and is responsible for calculating layer digests.

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,234 @@
...
+See [above][expose-all-layer-metadata-recap].
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+<!-- - What parts of the design do you expect to be resolved before this -->
+<!--   gets merged? -->
+
+* Does this affect how `report.toml` should be spec'd?
+* How can I use `image.previous-image-digest` to find that image's
```
ekcasey, line 207: I think storing `report.toml` files and associating them with images will have to be a platform concern given that we want to avoid embedding that information in the image itself.

text/0000-layer-origin-metadata.md
pclalv: regarding how cache availability affects builds or not, i'm not sure what the context is. are you talking about the case where the layer is there but its origin metadata, for whatever reason, isn't there? or maybe about the case where we have to retrieve those via two separate requests, the request for the layer succeeding and the request for its origin metadata failing? 

generally, i can imagine some people insisting on having the data about the previous build, for examples when building images bound for prod - maybe such a build is only valid if all origin metadata is present. i can also imagine some people caring more about build availability, not wanting to sacrifice that for lack of traceability. 

in the case of a build that didn't use the cache, these fields would certainly be absent.

text/0000-layer-origin-metadata.md
pclalv: oops, yeah, absolutely.

text/0000-layer-origin-metadata.md
ekcasey: > are you talking about the case where the layer is there but its origin metadata, for whatever reason, isn't there?

Yes

In some ways the problem I am describing is the inverse of the problem statement from your original RFC. For `cache=false` layers we always have the existing `layer.metadata` but additional layer origin metadata won't exist on the image and needs to come from the cache (or be supplied in some other fashion). However, the cache from the previous build is not guaranteed to be always available (For example if you are using a volume cache and rebuilding an image with `pack` in a different docker daemon. Or if you deleted the cache volume).

The availability depends on the caching strategy. An image cache is always available but less performant than a volume cache. That is the tradeoff. I suspect we can probably just populate these fields when the data is available and let folks pick the caching strategy that meets their availability or performance requirements.

sclevine:
@natalieparellano @jabrown85 to assist

text/0000-layer-origin-metadata.md
jabrown85: I would agree with your distinction @pclalv.

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,225 @@
...
+modification-date = "2020-07-09 12:41:35-04:00"
+size = 17813
+previous-layer-digest = "c54bba046158"
+
+cache = true
+launch = false
+build = true
+
+[[layers.application-source]]
+type = "git"
```
jabrown85, line 102: Is this RFC proposing lifecycle to inspect the `/workspace` to figure out if it is using `git`/ `mercurial` / other?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,225 @@
...
+Now consider that I have access to each build's `report.toml`, and
+that the second build's `report.toml` looks something like this:
+
+```toml
+[image]
+tags = ["index.docker.io/image/name:latest", "index.docker.io/image/name:other-tag"]
+digest = "26cca5b0c787"
+previous-image-digest = "97e47fb9e0a6"
+
+[[layers]]
```
jabrown85, line 88: I think this RFC could be use a schema for these new tables to define the fields needed to accomplish. I can make some educated guesses on the what these mean, but I think it makes sense to define them.

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,225 @@
...
+cache = true
+launch = false
+build = true
+
+[[layers.application-source]]
+type = "git"
+
+[[layers.application-source.version]]
+sha = "26e8c05b151f57161d64b7a5f844bb04caed0ba3"
+
```
jabrown85, line 106: Do you think we need to include things like lifecycle binary versions/builds?

What about build/run image digest? I know you said you can get at that data today, should it be present in `report.toml` as well?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,225 @@
...
+Exposing layer origin metadata in `report.toml` will improve layer
+reproducibility and, in turn, build reproducibility.
+
+# Motivation
+[motivation]: #motivation
+
+- Layer generation can be influenced by the state of the file system,
+  namely by the state of the application directory, other layers, and
+  the stack's build image
+- Layers aren't reproducible, because we can't know enough about the
```
jabrown85, line 25: I think this gets more complicated with things mounted by a platform like `kpack` or `pack`. Imagine the https://github.com/buildpacks/rfcs/pull/87's example of a cert buildpack. Does the platform have to document things about the cert volume it mounted? 

Similarly, what about `<platform>/env/` values? Those changing will change the resulting build as well 🤔

text/0000-layer-origin-metadata.md
jabrown85: Yeah, I think I would vote to keep any SCM info out of `lifecycle` itself. But maybe others would like it.

text/0000-layer-origin-metadata.md
jabrown85: Sure

text/0000-layer-origin-metadata.md
jabrown85: Sorry, I think at one point we talked about the [cert provider](https://github.com/buildpacks/rfcs/blob/app-mixins/text/0000-app-mixins-stack-buildpacks.md#ca-cert-provider-buildpack) may do something like read from `/<platform>/certs`. If you imagine a platform where you upload certs for all your apps to use and that buildpack pulls them in, there maybe should be a representation of that in this report so you can tell something change. The same could be true for `/platform/secrets` mounts.

I like the immutable representation idea for `<platform>/env` (and others I suppose?).

pclalv: it'd be nice if the platform could reference `env`, `certs` and whatever else as layers. you put those platform auxiliary layers existed on a registry, and then you've also solved the problem of durability.

unfortunately, that seems pretty out-of-scope - the ideas in this RFC depend on those ideas being implemented first, because yeah, otherwise we've got mystery files floating around.

pclalv:
@jabrown85 not to be a bother, but i just wanted to ping you on this PR, and make clear that it's ready for another look. i pushed up some changes and responded to some comments. i'll try to be more explicit about that going forwards.

jabrown85:
@pclalv no bother at all. I did take a peek and it looked good to me to bring up again in WG 👍  - sorry for not responding earlier.

pclalv:
ah, i see. should i come to a WG meeting sometime so that we can talk about it there?

jabrown85:
> ah, i see. should i come to a WG meeting sometime so that we can talk about it there?

I believe so, yeah. This process is new to me as well. I think the RFC looks great and is ready for a wider audience with the modifications.

pclalv:
sounds good. if it's not too last minute, i can show up at this afternoon's (9/9) WG meeting or, even tomorrow (9/10). what do you think?

jabrown85:
It's never too late AFAIK! Show up and add your RFC to the agenda for discussion when you arrive.

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,276 @@
...
+second build:
+
+1. the version of the `foo-api` application
+2. the version of the `yabba` buildpack
+3. the version of the stack's build image
+4. the version of the `yabba/build` layer that was restored from the
+   cache
+
+I can account for all but the last - I neither know that layer's
+identity, nor how it was created (the version of the `foo-api`, the
```
natalieparellano, line 78: Currently when I run a `pack build` that uses cached layers, I see something like 
```
Restoring data for "google.java.runtime:java" from cache
Retrieving data for "sha256:b12e5dfa0c2c4c4a384c03c0506635e50293a12624ad5abb0a3cee7c525efc9a"
```
is that sha256 not akin to an ID? Can you clarify how what you are proposing is different?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,276 @@
...
+digest = "26cca5b0c787"
+image-id = "sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9"
+
+[restored-image]
+# Same schema as [image] above, but without tags
+digest = "97e47fb9e0a6"
+image-id = "sha256:a8d2b367f43398fbabe25d45291451003e529fdd5d50f27437291fe61d637ba1"
+
+[[layers]]
+buildpack = "yabba"
```
natalieparellano, line 98: Would it be necessary to capture the buildpack version as well?

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,276 @@
...
+
+1. the version of the `foo-api` application
+2. the version of the `yabba` buildpack
+3. the version of the stack's build image
+4. the version of the `yabba/build` layer that was restored from the
+   cache
+
+I can account for all but the last - I neither know that layer's
+identity, nor how it was created (the version of the `foo-api`, the
+version of the `yabba` buildpack, whatever layers existed at that
```
natalieparellano, line 79: I believe the version of the buildpack is discoverable if you inspect the `io.buildpacks.lifecycle.cache.metadata` label and/or file, depending on the type of cache.

natalieparellano (commented):
@pclalv thanks for putting this together! I learned a lot from reading through this. Added a couple of questions. I'm looking forward to seeing this discussed further!

text/0000-layer-origin-metadata.md
```diff
@@ -0,0 +1,276 @@
...
+time). I cannot identify the second image's inputs - in other words, I
+cannot trace the second image.
+
+Now consider that I have access to each image's `report.toml`, and
+that the second image's `report.toml` looks something like this:
+
+```toml
```
jkutner, line 86:
```suggestion
```
```

jkutner, line 86: I suggest removing the `toml` here because the conents aren't valid toml, and it makes the diff hard to read. But I can't figure out how to put just three backticks inside of a code block :)

jkutner (approved):

jooshbzm:
We (greenhouse) had proposed this, but don't actively need it anymore. Feel free to close it and we can reopen if we end up needing it in the future.

PR #95, opened 2020-07-21
RFC for Project Descriptor Flexibility

hone:
[readable](https://github.com/buildpacks/rfcs/blob/project-descriptor-flexibility/text/0000-project-descriptor-flexibility.md)

### DIFF
--- a/text/0000-project-descriptor-flexibility.md
+++ b/text/0000-project-descriptor-flexibility.md
@@ -1,0 +1,91 @@
+# Meta
+[meta]: #meta
+- Name: Project Descriptor Flexibility
+- Start Date: 2020-07-21
+- Author(s): hone
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The Project Descriptor will become more flexible by making two changes:
+
+1. Derive the top-level "primary" table to match the filename.
+1. The Cloud Native Buildpacks project only reserves the `[build]` and filename keys at the "top level".
+
+# Motivation
+[motivation]: #motivation
+
+The current spec of the Project Descriptor is simple but restrictive for companies and other projects to adopt it for their own use case. These projects may want to use a different filename other than the default of `project.toml` which is currently supported, but then they're required to use the `[project]` key which does not align.
+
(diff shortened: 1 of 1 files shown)


nebhale:
@hone What is the compatibility store for the future if the CNB project does want to reserve a new top-level key beyond `[<filename>]` and `[build]`?  What's the story for preventing collisions in the top-level namespace between different platforms and projects?

jkutner:
@nebhale in working group we discussed moving `[build]` (and any keys we need to define) under the top-level table, yielding:

```
[project]
id = ....

[[project.build.buildpacks]]
id = ...
```

We'd then commit to not using anything outside of the project/top-level table

nebhale:
@jkutner I'm not sure that we can make that commitment though.  And if we did, it could significantly constrain new features we might want to introduce, that we haven't thought of before.  This is why putting user configuration into a single key that we _can_ keep our hands off of, is so compelling.  It gives us a story where we can add new features without conflict with user configuration.

jkutner:
@nebhale on the flip side: platforms won't use the `[metadata]` table if they don't adopt the file format. I think we are more likely to see `project.toml` adoption if we force ourselves under one key, rather than forcing our adopters under one key.

dfreilich:
Is there a request for this? Have other projects expressed interest in also using `project.toml`, such that we're focusing on making it compatible for that?

dfreilich:
Also, I'm a bit confused by what you describe the current behavior is. Is that to say that if the descriptor is `app.toml`, it will expect an `[app]` key, and if the descriptor is `build.toml`, it will expect a `[build]` key?

jkutner:
@dfreilich yes, that's what I'm proposing

hone:
> Is there a request for this? Have other projects expressed interest in also using `project.toml`, such that we're focusing on making it compatible for that?

This came out of a request internally at Salesforce/Heroku where the issue with the current `project.toml` requiring Salesforce specific config being forced under `[metadata]` made it feel like a second class citizen for a config for specific to the platform.

briandealwis:
And if I call it `build.toml` then we… have a conflct?

Would it make sense to allow providing a "configurator" that could synthesize a CNB project descriptor from the workspace?  That could be useful for downstream projects that already use a different format than toml.

hone:
Based on feedback from the group #140 supersedes this RFC, so I'm going to close this.

PR #98, opened 2020-07-27
RFC: Override Env Vars By Default

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/override-env-default/text/0000-override-env-by-default.md)

### DIFF
--- a/text/0000-override-env-by-default.md
+++ b/text/0000-override-env-by-default.md
@@ -27,7 +27,7 @@ Buildpack authors can create environment variable files as `<layer>/env(.launch|
 
 These environment variable files are applied dynamically, either before subsequent buildpacks or when the app container starts. Their application may be modulated by file extensions such as `.append`, `.prepend`, `.default`, and `.override`. For `.append` and `.prepend`, a special `NAME.delim` file may be used to specify the delimiter.
 
-The current behavior is to assume `.append` and `.delim` (with contents `:`) when no extension is specify. This is confusing (for reasons stated above) and does not match the behavior of `NAME=value` when environment variables are set in other contexts. This RFC proposed we change the behavior to be identical to `.override`, with overrides all previous values (such that the last buildpack "wins").
+The current behavior is to assume `.append` and `.delim` (with contents `:`) when no extension is specified. This is confusing (for reasons stated above) and does not match the behavior of `NAME=value` when environment variables are set in other contexts. This RFC proposes we change the behavior to be identical to `.override`, which overrides all previous values (such that the last buildpack "wins").
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-override-env-by-default.md
+++ b/text/0000-override-env-by-default.md
@@ -1,0 +1,54 @@
+# Meta
+[meta]: #meta
+- Name: Override Env Vars by Default
+- Start Date: 2020-07-26
+- Author(s): @sclevine
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes that we change the behavior of `/env/`, `/env.build/`, and `/env.launch/` directories such that the default, suffix-less behavior is equivalent to `VAR.override` instead of `VAR.append`+`VAR.delim=:`.
+
+# Motivation
+[motivation]: #motivation
+
+The current behavior is surprising, especially because adding `.append` silently removes the delimiter. Changing to `.override` and requiring a `.delim` file to add a delimiter would be more clear.
+
+Additionally, not all environment variables are paths, so assuming `:` may also surprise users when two unrelated buildpacks set the same env var without an extension.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors can create environment variable files as `<layer>/env(.launch|.build|)/NAME` such the an environment variable with name `NAME` is exported into the runtime, build-time (for subsequent buildpacks), or both environments. The contents of the file are set as the value.
+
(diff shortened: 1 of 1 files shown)


text/0000-override-env-by-default.md
```diff
@@ -0,0 +1,54 @@
...
+[summary]: #summary
+
+This RFC proposes that we change the behavior of `/env/`, `/env.build/`, and `/env.launch/` directories such that the default, suffix-less behavior is equivalent to `VAR.override` instead of `VAR.append`+`VAR.delim=:`.
+
+# Motivation
+[motivation]: #motivation
+
+The current behavior is surprising, especially because adding `.append` silently removes the delimiter. Changing to `.override` and requiring a `.delim` file to add a delimiter would be more clear.
+
+Additionally, not all environment variables are paths, so assuming `:` may also surprise users when two unrelated buildpacks set the same env var without an extension.
```
edmorley, line 21: Additional reasons for this change... 

- Even when an environment variable contains a path that needs to be joined, there's no guarantee that `.append` is the correct choice over `.prepend`. Making one the default hides the need to think about what is the most appropriate choice, and results in asymmetrical syntax for the two use-cases.
- Buildpack authors have to remember which of `.append` vs `.prepend` is the default, or else continually look in the docs. This seems less likely to occur if the default were `override`.
- Whilst many env vars in classic buildpacks are paths, my understanding is that for CNBs the platform will automatically set env vars for many common directories (https://github.com/buildpacks/spec/blob/18d50ea3d637d4fdcfe77b4a88a17e8a19b80799/platform.md#buildpack-environment) reducing the amount of path handling that is required by the buildpack

text/0000-override-env-by-default.md
```diff
@@ -0,0 +1,54 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors can create environment variable files as `<layer>/env(.launch|.build|)/NAME` such the an environment variable with name `NAME` is exported into the runtime, build-time (for subsequent buildpacks), or both environments. The contents of the file are set as the value.
+
+These environment variable files are applied dynamically, either before subsequent buildpacks or when the app container starts. Their application may be modulated by file extensions such as `.append`, `.prepend`, `.default`, and `.override`. For `.append` and `.prepend`, a special `NAME.delim` file may be used to specify the delimiter.
+
+The current behavior is to assume `.append` and `.delim` (with contents `:`) when no extension is specify. This is confusing (for reasons stated above) and does not match the behavior of `NAME=value` when environment variables are set in other contexts. This RFC proposed we change the behavior to be identical to `.override`, with overrides all previous values (such that the last buildpack "wins").
```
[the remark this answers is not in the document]
sclevine: Thanks! (This is what I get for writing four RFCs as fast as possible at 1 am.)

text/0000-override-env-by-default.md
sclevine: Mind proposing these as a change via the github UI?

edmorley, line 21: Sure :-)

```suggestion
Additionally:

- Not all environment variables are paths, so assuming `:` may also surprise users when two unrelated buildpacks set the same env var without an extension.
- Even when an environment variable contains a path that needs to be joined, there is no guarantee that `.append` is the correct choice over `.prepend`. Making one the default hides the need to think about what is the most appropriate choice, results in asymmetrical syntax for the two use-cases, and means buildpack authors need to remember which of them is actually the default.
```

(I omitted the CNB setting path env vars part, since I'm fuzzy on how that actually works)

nebhale:
Final Comment Period with merge disposition, closing on 10 August, 2020.

text/0000-override-env-by-default.md
```diff
@@ -0,0 +1,57 @@
...
+
+Additionally:
+
+- Not all environment variables are paths, so assuming `:` may also surprise users when two unrelated buildpacks set the same env var without an extension.
+- Even when an environment variable contains a path that needs to be joined, there is no guarantee that `.append` is the correct choice over `.prepend`. Making one the default hides the need to think about what is the most appropriate choice, results in asymmetrical syntax for the two use-cases, and means buildpack authors need to remember which of them is actually the default.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors can create environment variable files as `<layer>/env(.launch|.build|)/NAME` such the an environment variable with name `NAME` is exported into the runtime, build-time (for subsequent buildpacks), or both environments. The contents of the file are set as the value.
```
dfreilich, line 29:
```suggestion
Buildpack authors can create environment variable files as `<layer>/env(.launch|.build|)/NAME` such that an environment variable with name `NAME` is exported into the runtime, build-time (for subsequent buildpacks), or both environments. The contents of the file are set as the value.
```

text/0000-override-env-by-default.md
```diff
@@ -0,0 +1,57 @@
...
+
+These environment variable files are applied dynamically, either before subsequent buildpacks or when the app container starts. Their application may be modulated by file extensions such as `.append`, `.prepend`, `.default`, and `.override`. For `.append` and `.prepend`, a special `NAME.delim` file may be used to specify the delimiter.
```
dfreilich, line 31: To clarify – if you wanted to append to `PATH`, you would create a file `<layer>/env(.launch|.build|)/NAME.append`?

hone:
Overall, we should make this change but the devil is in handling the backwards breaking changes. I assume this is slated for 1.0. Do we plan to handle the current behavior and this behavior in the same group?

nebhale:
@hone If we can find a comfortable way to do it, I would say yes.  Some options might be:

1.  Include Buildpack API Version information in the built image metadata such that the lifecycle can behave differently as the environment is bootstrapped.
1.  Rewrite the filenames during export to match the "canonical" names for the behavior based on the inputs at build time.

I think I like 2 best, but it feels like we've got some options to play around with that should give us the compatibility we desire.

sclevine:
> Include Buildpack API Version information in the built image metadata such that the lifecycle can behave differently as the environment is bootstrapped.

I think @ekcasey had plans to do this already.

PR #100, opened 2020-07-27
RFC: Decouple Buildpack Plan and Bill-of-Materials

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/decouple-plan-and-bom/text/0000-decouple-buildpack-plan-and-bom.md)

### DIFF
--- a/text/0000-decouple-buildpack-plan-and-bom.md
+++ b/text/0000-decouple-buildpack-plan-and-bom.md
@@ -28,7 +28,7 @@ Currently, there is a single file provided to `/bin/build` that contains the Bui
 This RFC proposes that we replace the current Buildpack Plan argument with:
 1. The existing Buildpack Plan argument (read-only).
 2. A new `[[bom]]` section in `<layers>/launch.toml` for runtime Bill-of-Materials entries. This would be used for stuff that's actually in the image, and added to image labels in the same format as the current Bill-of-Materials
-3. A new `[[bom]]` section in a new file called `<layers>/build.toml` for build-time Bill-of-Materials entries. This would be used for build-time dependencies, and would not be added to `report.toml` instead of the image labels.
+3. A new `[[bom]]` section in a new file called `<layers>/build.toml` for build-time Bill-of-Materials entries. This would be used for build-time dependencies, and would be added to `report.toml` instead of the image labels.
 4. A new `[[entries]]` section in `<layers>/build.toml` for Buildpack Plan entries that should be passed to subsequent buildpacks that may provide the dependencies. 
 
 # Drawbacks
--- a/text/0000-decouple-buildpack-plan-and-bom.md
+++ b/text/0000-decouple-buildpack-plan-and-bom.md
@@ -25,29 +25,28 @@ Additionally, it's currently impossible to remove build-time dependency entries 
 
 Currently, there is a single file provided to `/bin/build` that contains the Buildpack Plan entries for the buildpack. Entries may be removed from the Buildpack Plan by the buildpack in order to pass the entries to subsequent buildpacks that offered to provide the same entry. Entries may instead be modified, so that the Bill-of-Materials contains more detailed metadata.
 
-This RFC proposes that we replace the current Buildpack Plan argument with four arguments:
-1. The Buildpack Plan (read-only)
-2. Bill-of-Materials entries for metadata that is added to the image's Bill-of-Materials label. (Would starts empty, be r/w, used for stuff that's actually in the image.)
-3. Bill-of-Materials entries for metadata that is not added to the image labels. (Would start empty, be r/w, used for build-time dependencies.)
-4. Buildpack Plan entries that should be passed to subsequent buildpacks. (Would start empty, be r/w.)
+This RFC proposes that we replace the current Buildpack Plan argument with:
+1. The existing Buildpack Plan argument (read-only).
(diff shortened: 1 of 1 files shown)


text/0000-decouple-buildpack-plan-and-bom.md
```diff
@@ -0,0 +1,54 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently, there is a single file provided to `/bin/build` that contains the Buildpack Plan entries for the buildpack. Entries may be removed from the Buildpack Plan by the buildpack in order to pass the entries to subsequent buildpacks that offered to provide the same entry. Entries may instead be modified, so that the Bill-of-Materials contains more detailed metadata.
+
+This RFC proposes that we replace the current Buildpack Plan argument with four arguments:
+1. The Buildpack Plan (read-only)
+2. Bill-of-Materials entries for metadata that is added to the image's Bill-of-Materials label. (Would starts empty, be r/w, used for stuff that's actually in the image.)
+3. Bill-of-Materials entries for metadata that is not added to the image labels. (Would start empty, be r/w, used for build-time dependencies.)
```
ekcasey, line 31: Instead of adding two more files here I think we should put the launch `[bom]` in `launch.toml` (it fits nicely with other non-layer launch-specific contributions).

We could add an analogous `build.toml`. We still need a new file here but it has nice symmetry with `launch.toml`.

ekcasey: We could maybe put buildpack plan entries that should be passed to subsequent buildpacksh in `build.toml` as well (not as confident about this suggestion as I am about the above suggestion).

sclevine: I love this idea, but I'm also not sure what to do with the push-entries.

sclevine: Used `build.toml` as suggested.

sclevine:
@ekcasey changes made as suggested

text/0000-decouple-buildpack-plan-and-bom.md
```diff
@@ -0,0 +1,53 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently, there is a single file provided to `/bin/build` that contains the Buildpack Plan entries for the buildpack. Entries may be removed from the Buildpack Plan by the buildpack in order to pass the entries to subsequent buildpacks that offered to provide the same entry. Entries may instead be modified, so that the Bill-of-Materials contains more detailed metadata.
+
+This RFC proposes that we replace the current Buildpack Plan argument with:
+1. The existing Buildpack Plan argument (read-only).
+2. A new `[[bom]]` section in `<layers>/launch.toml` for runtime Bill-of-Materials entries. This would be used for stuff that's actually in the image, and added to image labels in the same format as the current Bill-of-Materials
+3. A new `[[bom]]` section in a new file called `<layers>/build.toml` for build-time Bill-of-Materials entries. This would be used for build-time dependencies, and would not be added to `report.toml` instead of the image labels.
```
ekcasey, line 31:
```suggestion
3. A new `[[bom]]` section in a new file called `<layers>/build.toml` for build-time Bill-of-Materials entries. This would be used for build-time dependencies, and would be added to `report.toml` instead of the image labels.
```

text/0000-decouple-buildpack-plan-and-bom.md
```diff
@@ -0,0 +1,53 @@
...
+[alternatives]: #alternatives
+
+- Keep current behavior as-is.
+- Keep current behavior around removing entries from the build plan, but switch to new files for the Bill-of-Materials.
+- Keep current behavior around removing entries from the build plan, but switch to a single file for the Bill-of-Materials (just runtime dependencies).
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Does `[[entries]]` make sense in `<layers>/build.toml`?
```
ekcasey, line 49: I was thinking about this some more, do we even need a mechanism for pushing entries to subsequent buildpacks? The buildpack doesn't have a way of knowing if any other buidpack can even provide the entry. Do we have any real-world examples where a buildpack receives an entry and doesn't "claim" it?

nebhale: None in our 50.

sclevine: To me, the major use case is adding a custom provider buildpack to the beginning of a group that contains a meta-buildpack. The custom provider buildpack can conditionally override a provider buildpack inside of the meta-buildpack based on app contents / require metadata.

If we want to take this out, I suggest that it be a separate RFC. (That said, I'm strongly in favor of keeping it.)

nebhale:
Final Comment Period with merge disposition, closing on 10 August, 2020.

dfreilich:
Would you mind clarifying the motivation behind this a bit more for me? Perhaps I don't understand the BOM usage as much as I thought I did, but why would you want to remove entries from it? Isn't it meant to provide an accurate assessment of everything that happened to the image?

nebhale:
@dfreilich There's a difference between "what's in an image" and "what was contributed by a buildpack".  In some cases these overlap (e.g. the BellSoft Liberica Buildpack contributes a version of Java) and in others they don't (e.g. the user's application has these dependencies at these versions).  As it stands now, if a buildpack wants to contribute to the BOM, it direct effects **the execution of later buildpacks** via the build plan.  In addition, if a buildpack wants to affect other the execution of later buildpacks, it **must modify the BOM**.  So the goal of this RFC is to say, these two things are separate from one another and should be populated as such.

PR #103, opened 2020-07-29
Project Descriptor Schema

joshwlewis:
[Readable](https://github.com/joshwlewis/rfcs/blob/project-descriptor-schema/text/0000-project-descriptor-schema.md)

### DIFF
--- a/text/0000-project-descriptor-schema.md
+++ b/text/0000-project-descriptor-schema.md
@@ -1,0 +1,200 @@
+# Meta
+[meta]: #meta
+- Name: Project Descriptor (project.toml) Schema
+- Start Date: 2020-28-07
+- Author(s): @joshwlewis
+- RFC Pull Request:
+- CNB Pull Request:
+- CNB Issue:
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal to introduce a schema for the project descriptor
+(project.toml) and leverage this schema within `pack` to validate user-supplied
+project descriptors for compliance with the schema.
+
+# Definitions
+[definitions]: #definitions
+
+- Project Descriptor: Otherwise known as project.toml. It is defined
+  in [RFC 19](./0019-project-descriptor.md).
+- Schema: A consumable representation of a desired data structure.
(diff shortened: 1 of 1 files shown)


text/0000-project-descriptor-schema.md
```diff
@@ -0,0 +1,200 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [The project descriptor RFC](./0019-project-descriptor.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+  - Where should this schema live?
```
jkutner, line 187: I'd propose `github.com/buildpacks/project-toml`. But if we want to aim high, we could create a new Github org, and do something like `github.com/project-toml/schema`

text/0000-project-descriptor-schema.md
```diff
@@ -0,0 +1,200 @@
...
+  - Should it provide generated types for consumption in `pack`?
+  - Should it provide a generalized binary for validating a `project.toml`?
+  - Is this an extension point for other CNB `toml` files, or directly targeted
```
jkutner, line 190: hot take: this would be really valuable in generating better error messages

text/0000-project-descriptor-schema.md
jkutner: also `github.com/buildpacks/schemas` if we want it to include schemas for all our file formats

ekcasey: What if, for all specified files, we put schemas, and go bindings directly in the `buildpacks/spec`?

Prior art: 
https://github.com/opencontainers/image-spec contains `schema` and `specs-go` directories

ekcasey: This wouldn't be an appropriate place for unspecified schemas like `builder.toml`, but it seems appropriate for schemas describing `pack` specific files to live in the `pack` repo

nebhale:
Final Comment Period with merge disposition, closing on 12 August, 2020.

PR #104, opened 2020-07-29
RFC: Shell-Free Profile.d

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/shell-free-profile-d/text/0000-shell-free-profile-d.md).

### DIFF
--- a/text/0000-shell-free-profile-d.md
+++ b/text/0000-shell-free-profile-d.md
@@ -1,0 +1,58 @@
+# Meta
+[meta]: #meta
+- Name: Shell-free Profile.d
+- Start Date: 2020-07-26
+- Author(s): @sclevine
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Currently, `<layer>/profile.d/` scripts must be text files containing Bash 3+ scripts. This RFC proposes that we support arbitrary executables in `<layer>/profile.d/` that do not depend on a shell, but can still modify the environment of the app process.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, stacks must have Bash 3+ to support fully-dynamic modification of the environment at launch (via profile.d scripts).
+This means makes it difficult to modify the environment of app images that use minimal run base images.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Executable files in `<layer>/profile.d/` should be executed directly instead of sourced by Bash 3+.
+- Executable files must only output an environment variable set to `stdout` in the form:
+  ```
+  KEY1=VALUE1
+  KEY2=VALUE2
+  ```
+- The launcher will manually add those key/value pairs to the environment of the chosen process type.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Might encourage adding additional large binaries to app images
+- Doubles-down on the launcher being a necessary component of CNB images
+- Some existing profile scripts will break if they are currently made executable by mistake
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep Bash 3+ requirement as-is
+- Allow stacks to configure the shell (wouldn't solve shell-less case though)
+
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- If a text file in `<layer>/profile.d/` is executable but doesn't have a shebang, should we still attempt to execute it directly? Some existing profile scripts (that are executable by mistake) will break if we do that. (Even if we detect this case, some existing profile scripts likely have a shebang by mistake.)
+- Should we use a named file instead of `stdout` for environment variable sets?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+This will require changes to the Buildpack API spec.


sclevine:
@nebhale

text/0000-shell-free-profile-d.md
```diff
@@ -0,0 +1,58 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Executable files in `<layer>/profile.d/` should be executed directly instead of sourced by Bash 3+.
+- Executable files must only output an environment variable set to `stdout` in the form:
+  ```
+  KEY1=VALUE1
+  KEY2=VALUE2
+  ```
```
micahyoung, line 30: Should we spec out the format, perhaps even just as conforming to `env` output format for POSIX and `set` output format for CMD?  I'm pretty sure those two formats are identical. That would implicitly mean (I think):
* newline separated (with newline at EOF)
* all chars besides the first `=` are literal and are not special (`"` or`'` or `/` or `\` or `;` or `:`), just embedded in the key or value.
* probably UTF-8 only

I feel like standardizing on those `env`/`set` commands could give users a default way to for their shell-less process to print out their changes (tweak env vars in your process, then subshell/exec to `env`/`set` to print it out)

Subtle formatting was an issue for us on a [similar implementation](https://github.com/cloudfoundry/buildpackapplifecycle/tree/8f2039f5b509d2945597f38d6659f884f396e5ef/getenv) of this for Cloud Foundry. In that case, we controlled the format on both the writer and reader but I worry user binaries could generate just about anything.

text/0000-shell-free-profile-d.md
```diff
@@ -0,0 +1,58 @@
...
+
+- Keep Bash 3+ requirement as-is
+- Allow stacks to configure the shell (wouldn't solve shell-less case though)
+
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- If a text file in `<layer>/profile.d/` is executable but doesn't have a shebang, should we still attempt to execute it directly? Some existing profile scripts (that are executable by mistake) will break if we do that. (Even if we detect this case, some existing profile scripts likely have a shebang by mistake.)
+- Should we use a named file instead of `stdout` for environment variable sets?
```
jabrown85, line 53: I'm imagining an existing binary that does something the user wants (say logging into a pre-installed `aws` cli with runtime secrets) that outputs like `successfully logged into acct:234343`...do we expect users to wrap that with a shell script or another binary? Otherwise the launcher would execute it and attempt to parse the output of the `aws` binary. 

I think my vote would be to enforce writing files to prevent accidentally parsing output.

text/0000-shell-free-profile-d.md
```diff
@@ -0,0 +1,58 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, stacks must have Bash 3+ to support fully-dynamic modification of the environment at launch (via profile.d scripts).
+This means makes it difficult to modify the environment of app images that use minimal run base images.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Executable files in `<layer>/profile.d/` should be executed directly instead of sourced by Bash 3+.
```
jabrown85, line 25: I vote for `/exec.d` or similar. I think the open question around shebang-less executable files pulls me in favor of this being more explicit.

text/0000-shell-free-profile-d.md
micahyoung: Here's some real examples of those for discussion.
## bash on bionic
```sh
root@ca8495ca2c6a:/# export FOO=\\bar\=baz
root@ca8495ca2c6a:/# env
FOO=\bar=baz
LS_COLORS=rs=0:di=01;34:ln=01;36:mh=00:pi=40;33:so=01;35:do=01;35:bd=40;33;01:cd=40;33;01:or=40;31;01:mi=00:su=37;41:sg=30;43:ca=30;41:tw=30;42:ow=34;42:st=37;44:ex=01;32:*.tar=01;31:*.tgz=01;31:*.arc=01;31:*.arj=01;31:*.taz=01;31:*.lha=01;31:*.lz4=01;31:*.lzh=01;31:*.lzma=01;31:*.tlz=01;31:*.txz=01;31:*.tzo=01;31:*.t7z=01;31:*.zip=01;31:*.z=01;31:*.Z=01;31:*.dz=01;31:*.gz=01;31:*.lrz=01;31:*.lz=01;31:*.lzo=01;31:*.xz=01;31:*.zst=01;31:*.tzst=01;31:*.bz2=01;31:*.bz=01;31:*.tbz=01;31:*.tbz2=01;31:*.tz=01;31:*.deb=01;31:*.rpm=01;31:*.jar=01;31:*.war=01;31:*.ear=01;31:*.sar=01;31:*.rar=01;31:*.alz=01;31:*.ace=01;31:*.zoo=01;31:*.cpio=01;31:*.7z=01;31:*.rz=01;31:*.cab=01;31:*.wim=01;31:*.swm=01;31:*.dwm=01;31:*.esd=01;31:*.jpg=01;35:*.jpeg=01;35:*.mjpg=01;35:*.mjpeg=01;35:*.gif=01;35:*.bmp=01;35:*.pbm=01;35:*.pgm=01;35:*.ppm=01;35:*.tga=01;35:*.xbm=01;35:*.xpm=01;35:*.tif=01;35:*.tiff=01;35:*.png=01;35:*.svg=01;35:*.svgz=01;35:*.mng=01;35:*.pcx=01;35:*.mov=01;35:*.mpg=01;35:*.mpeg=01;35:*.m2v=01;35:*.mkv=01;35:*.webm=01;35:*.ogm=01;35:*.mp4=01;35:*.m4v=01;35:*.mp4v=01;35:*.vob=01;35:*.qt=01;35:*.nuv=01;35:*.wmv=01;35:*.asf=01;35:*.rm=01;35:*.rmvb=01;35:*.flc=01;35:*.avi=01;35:*.fli=01;35:*.flv=01;35:*.gl=01;35:*.dl=01;35:*.xcf=01;35:*.xwd=01;35:*.yuv=01;35:*.cgm=01;35:*.emf=01;35:*.ogv=01;35:*.ogx=01;35:*.aac=00;36:*.au=00;36:*.flac=00;36:*.m4a=00;36:*.mid=00;36:*.midi=00;36:*.mka=00;36:*.mp3=00;36:*.mpc=00;36:*.ogg=00;36:*.ra=00;36:*.wav=00;36:*.oga=00;36:*.opus=00;36:*.spx=00;36:*.xspf=00;36:
HOSTNAME=ca8495ca2c6a
PWD=/
HOME=/root
TERM=xterm
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

## cmd on servercore
```
C:\> set FOO=\bar=baz                                      
C:\> set 
FOO=\bar=baz
ALLUSERSPROFILE=C:\ProgramData
APPDATA=C:\Users\ContainerAdministrator\AppData\Roaming
CommonProgramFiles=C:\Program Files\Common Files
CommonProgramFiles(x86)=C:\Program Files (x86)\Common Files
CommonProgramW6432=C:\Program Files\Common Files
COMPUTERNAME=6D0353F3A904
ComSpec=C:\Windows\system32\cmd.exe
DriverData=C:\Windows\System32\Drivers\DriverData
LOCALAPPDATA=C:\Users\ContainerAdministrator\AppData\Local
NUMBER_OF_PROCESSORS=2
OS=Windows_NT
Path=C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\Open
SSH\;C:\Users\ContainerAdministrator\AppData\Local\Microsoft\WindowsApps
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_IDENTIFIER=Intel64 Family 6 Model 94 Stepping 3, GenuineIntel
PROCESSOR_LEVEL=6
PROCESSOR_REVISION=5e03
ProgramData=C:\ProgramData
ProgramFiles=C:\Program Files                                                                            
ProgramFiles(x86)=C:\Program Files (x86)                                                                 
ProgramW6432=C:\Program Files                                                                            
PROMPT=$P$G                                                                                              
PSModulePath=%ProgramFiles%\WindowsPowerShell\Modules;C:\Windows\system32\WindowsPowerShell\v1.0\Modules 
PUBLIC=C:\Users\Public                                                                                   
SystemDrive=C:                                                                                           
SystemRoot=C:\Windows                                                                                    
TEMP=C:\Users\ContainerAdministrator\AppData\Local\Temp                                                  
```

text/0000-shell-free-profile-d.md
sclevine: If we use `stdout`, exec.d binaries would need to run subprocesses with `stdout` redirected to `stderr`. I agree that it's easy to forget to do that.

> I think my vote would be to enforce writing files to prevent accidentally parsing output.

I don't think we should create and open actual files (and require disk I/O) to start the app process. Security is also a concern, given that this would allow arbitrary env vars to be set at launch, so a named pipe is still less than ideal. I updated the proposal to use fd 3, which is secure (the fd can be passed directly) and not easy to write to accidentally (can't be accessed by subprocesses at all).

text/0000-shell-free-profile-d.md
sclevine: Done!

text/0000-shell-free-profile-d.md
sclevine: Definitely agree with the idea of standardizing on the behavior of `env` / `set`.

sclevine: Added that it must match `env`/`set` to the RFC. Is that sufficient?

sclevine:
Changed to `exec.d` + fd 3, with defined format for the env var list.

text/0000-shell-free-profile-d.md
vindvaki: What about variables with newlines and `=` beyond the first line? I think `env` and `set` do not really handle these well or consistently, and parsing the output of `env` can give unexpected results.

For example, if I run
```shell
export x="a
b=c
d=e"
```
then the output of `env` will have lines of the form
```
x=a
b=c
d=e
```
Consuming this output according to the above specification would result in an incorrect value for `x` and the extra variables `b` and `d`.

sclevine: Yikes, I assumed there was a solution for encoding edge cases like this. Docker uses the same format: 
https://docs.docker.com/compose/env-file/

Turns out it doesn't work:
https://github.com/moby/moby/issues/12997

I'd suggest JSON, but JSON doesn't allow tabs (which are valid in env vars).
TOML would match other parts of the API, but seems like an odd choice in this context.
I'm also hesitant to introduce a format that would require significant parsing before the app starts.

I think the options are:
1. Use the current format, but with a special encoding for newlines.
2. Use TOML. (Perhaps with `KEY = "value"` syntax instead of the `{ key = "KEY", value = "value" }` syntax we use elsewhere, due to the automated context?)

sclevine:
Holding this RFC on a resolution to: https://github.com/buildpacks/rfcs/pull/104#discussion_r464164626

(Thanks @vindvaki!)

sclevine:
After more research, here's an expanded list of options:
1. Use the current format and fail to support environment variables with newlines.
2. Use the current format but with a special encoding for newlines.
3. Use TOML. (Perhaps with KEY = "value" syntax instead of the { key = "KEY", value = "value" } syntax we use elsewhere, due to the automated context?)
4. Use the format returned by `bash -o posix -c "export -p"`:
   ```
   export C="A
   B=3"
   export HOME="/root"
   export HOSTNAME="574cc2a3302c"
   ```
5. Use the format returned by `export -p` (with Bash 5):
   ```
   declare -x C="A
   B=3"
   declare -x HOME="/root"
   declare -x HOSTNAME="574cc2a3302c"
   ```

Both (4) and (5) are only safe with Bash 4.4+. Use cases for exec.d generally involve images without shells, and I'm not sure how hard it is to reliably emit/parse that format without a shell. I lean towards option (3).

ekcasey:
> Use TOML. (Perhaps with KEY = "value" syntax instead of the { key = "KEY", value = "value" } syntax we use elsewhere, due to the automated context?)

This is my preference ^

nebhale:
I can get on board with 3 as well.

hone:
I'm also on board with option 3 based on the WG meeting today.

sclevine:
Ready for review again.

nebhale:
Final Comment Period with merge disposition, closing on 19 August, 2020.

PR #105, opened 2020-08-06
Deprecate Service Bindings

nebhale:
Given the existence of the Service Bindings Specification for Kubernetes, the Cloud Native Buildpacks Bindings Extension Specification should be deprecated.

### DIFF
--- a/text/0000-deprecate-service-bindings.md
+++ b/text/0000-deprecate-service-bindings.md
@@ -46,7 +46,7 @@ N/A
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Existing users of bindings will be affected.  The lifecycle and platforms should support both styles until the end of the deprecation period.
+- Existing users of bindings will be affected.  Platforms and buildpacks should support both styles until the end of the deprecation period.
 
 # Alternatives
 [alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


text/0000-deprecate-service-bindings.md
```diff
@@ -0,0 +1,69 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Existing users of bindings will be affected.  The lifecycle and platforms should support both styles until the end of the deprecation period.
```
hone, line 49: Should we just set the removal for the fated "1.0"?

nebhale:
Final Comment Period with merge disposition, closing on 19 August, 2020.

text/0000-deprecate-service-bindings.md
ekcasey, line 49:
```suggestion
- Existing users of bindings will be affected.  Platforms and buildpacks should support both styles until the end of the deprecation period.
```

The lifecycle is actually totally unaware of bindings.

I think it is worth linking to the deprecated binding spec in the "Buildpack Additional Guidance" Section mentioned above. We can suggest buildpacks that care about compatibility with older platforms should support both specifications.

PR #108, opened 2020-08-14
RFC: Label RFCs With Spec

ForestEckhardt:
[Readable](https://github.com/ForestEckhardt/rfcs/blob/rfc-tags/text/0000-label-rfcs.md)


kvedurmu:
This would be a great change! With the number of RFCs flowing in here, it's hard to keep track of what's applicable to my team. It would be even better if accepted RFCs could be grouped accordingly as well so users can easily go back and see all relevant RFCs for the Buildpack Interface, Platform, etc.

nebhale:
Well, @kvedurmu _you've_ got a [spec change](https://github.com/buildpacks/spec/pull/89) that needs some attention 😜

yaelharel:
This is a great idea! It's really hard (especially for new contributors) to track all of the RFCs. Applying labels will help a lot!
I hope we'll start do it soon!

hone:
I like this idea a lot and something that has come up when talking to our languages team at Heroku to make it easier to know what they should be reviewing. I believe there's some more detail that is missing. Who is responsible of labeling? Is that something we just do during the weekly WG on Wed? Also as an alternative or an addition, is it useful to label by "target audience":
* Buildpack Author
* CNB Contributor
* Platform Operator

ForestEckhardt:
@hone I love the idea of adding the target audience in addition to API labeling. Let me write up something quick to include this as well as a process around labeling the RFCs.

text/0000-label-rfcs.md
```diff
@@ -0,0 +1,35 @@
...
+focus in on certain RFCs that effect them. This means that potentially effected
+parties have to spend less time wading through all of the RFCs in the
+repository to find the handful that will affect them. Overall, this change will
+make reviewing RFCs a much more manageable process for new members and people
+that are concerned with specific parts of the specification.
+
+# Implementation
+[Implementation]: #implementation
+
+During any Working Group meeting, when a new RFC is brought up it should at
```
hone, line 33: Should we try to do this during the weekly review of RFCs or would that bog that process down too much? Thoughts @sclevine?

hone: From the WG: we should be able to handle this as we go through the RFCs every week if you want to update @ForestEckhardt

ForestEckhardt: @hone Sounds good to me. I will update this when I get the change.

nebhale:
Final Comment Period with merge disposition, closing on 7 September, 2020

text/0000-label-rfcs.md
```diff
@@ -0,0 +1,35 @@
...
+of the specification (i.e. Buildpack Interface, Distribution, Platform
+Interface) is being affected as well as a label to indicate the target audience
+(i.e. Buildpack Authors, CNB Contributors, Platform Operators) of the change.
+
+# Motivation
+[motivation]: #motivation
+
+Indicating which part of the specification the RFC is purposing to update or
+modify will make the RFC more discoverable because it will allow people to
+focus in on certain RFCs that effect them. This means that potentially effected
```
dfreilich, line 24:
```suggestion
focus in on certain RFCs that affect them. This means that potentially affected
```
https://www.grammarly.com/blog/affect-vs-effect/#:~:text=Here's%20the%20short%20version%20of,the%20result%20of%20a%20change.

dfreilich (approved):

yaelharel (approved):

PR #109, opened 2020-08-17
RFC: Relax Mixin Contract

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/relax-mixin-contract/text/0000-relax-mixin-contract.md).

### DIFF
--- a/text/0000-relax-mixin-contract.md
+++ b/text/0000-relax-mixin-contract.md
@@ -1,0 +1,69 @@
+# Meta
+[meta]: #meta
+- Name: Relax Mixin Contract
+- Start Date: 2020-08-12
+- Author(s): @sclevine
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Mixins are used to imply multiple contracts:
+
+1. A contract between a buildpack and a stack, where the buildpack may state that it can only be used with stack images that have certain dependencies.
+
+2. A contract between a stack run image and stack build image, where apps built on the build image must have a certain dependency present on the run image to start successfully on the run image.
+
+Currently, if a stack build image has the mixin `build:mypkg` and a stack run image has the mixin `run:mypkg`, a buildpack cannot require the mixin as `mypkg` (either statically in buildpack.toml, or dynamically via the stackpack mechanism described in #87). A stack author may choose to use this configuration because the same dependency is used in different, unrelated ways in both images.
+
+Similarly, if a stack build image shares the mixin `mypkg` with the stack run image, a buildpack cannot require the mixin `build:mypkg` or `run:mypkg`, even though both images contain `mypkg`.
+
+This RFC proposed that we update the buildpack specification to create a new relationship between `mypkg`, `run:mypkg`, and `build:mypkg`.
+
+- While `run:mypkg` may imply a different set of changes to the run image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `run:mypkg` at runtime (and vice versa).
+
+- While `build:mypkg` may imply a different set of changes to the build image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `build:mypkg` at build-time (and vice versa).
+
+- A runtime stackpack would still be passed either `run:mypkg` or `mypkg` depending on what the buildpack required, so that the stackpack understands whether or not it can expect the build image to have the corresponding mixin as well.
+
+- A build-time stackpack would still be passed either `build:mypkg` or `mypkg` depending on what the buildpack required, so that the stackpack understands whether or not it can expect the run image to have the corresponding mixin as well.
+
+# Definitions
+[definitions]: #definitions
+
+- [**mixins** and **stage specifiers**](https://github.com/buildpacks/spec/blob/main/platform.md#mixins)
+- [**contractual compatibility** driven by ABI](https://github.com/buildpacks/spec/blob/main/platform.md#compatibility-guarantees)
+
+
+# Motivation
+[motivation]: #motivation
+
+1. Stack authors must currently provide both stage-specific and non-stage-specific versions of mixin names on each image to ensure that buildpacks can require mixins with or without the stage specifier. This is unfriendly UX for stack authors, and difficult to explain.
+
+2. Stack authors are unable to relax the contract between the build and run image (by only using the stage-specifier version of the mixins) without requiring the buildpacks manually list the mixin twice with the different stage specifiers.
+
+3. Under the pre-RFC mixin definition, stackpacks (specified in #87) will attempt to re-install packages when stage-specifier is mismatched between the buildpack and stack images. For instance, a buildpack requesting `mypkg` would result in the stackpack attempting to install `mypkg` on a run image with `run:mypkg`.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Pack and lifecycle will relax contractual restrictions when initiating builds, creating builders, and exporting images.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This change reduces the power of the mixin model for stack authors. Previously, stack authors could reliably use `run:mypkg` to indicate that a run image strictly contains the runtime version of mypkg (which may provide different functionality than `mypkg`).
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Refactor mixins entirely, e.g., enforce stack<->stack and stack<->buildpack mixin interfaces separately.
+
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+This changes how mixin contracts are enforced in the Platform API spec.
--- a/text/0000-relax-mixin-contract.md
+++ b/text/0000-relax-mixin-contract.md
@@ -23,13 +23,14 @@ Similarly, if a stack build image shares the mixin `mypkg` with the stack run im
 
 This RFC proposed that we update the buildpack specification to create a new relationship between `mypkg`, `run:mypkg`, and `build:mypkg`.
 
-- While `run:mypkg` may imply a different set of changes to the run image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `run:mypkg` at runtime (and vice versa).
+- `run:mypkg` and `mypkg` imply the same set of changes to the run image, such that buildpacks requiring `mypkg` would be satisfied with `run:mypkg` at runtime (and vice versa).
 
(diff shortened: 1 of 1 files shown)


sclevine:
@jkutner this supports #87 by allowing package names to be used as mixins in all cases

text/0000-relax-mixin-contract.md
```diff
@@ -0,0 +1,69 @@
...
+
+Similarly, if a stack build image shares the mixin `mypkg` with the stack run image, a buildpack cannot require the mixin `build:mypkg` or `run:mypkg`, even though both images contain `mypkg`.
+
+This RFC proposed that we update the buildpack specification to create a new relationship between `mypkg`, `run:mypkg`, and `build:mypkg`.
+
+- While `run:mypkg` may imply a different set of changes to the run image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `run:mypkg` at runtime (and vice versa).
+
+- While `build:mypkg` may imply a different set of changes to the build image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `build:mypkg` at build-time (and vice versa).
+
+- A runtime stackpack would still be passed either `run:mypkg` or `mypkg` depending on what the buildpack required, so that the stackpack understands whether or not it can expect the build image to have the corresponding mixin as well.
```
ekcasey, line 30: Can you give an example of a case where the difference between `run:mypkg` and `mypkg` matters to the stackpack (while it is installing packages  on the run-image)?

text/0000-relax-mixin-contract.md
```diff
@@ -0,0 +1,69 @@
...
+
+2. A contract between a stack run image and stack build image, where apps built on the build image must have a certain dependency present on the run image to start successfully on the run image.
+
+Currently, if a stack build image has the mixin `build:mypkg` and a stack run image has the mixin `run:mypkg`, a buildpack cannot require the mixin as `mypkg` (either statically in buildpack.toml, or dynamically via the stackpack mechanism described in #87). A stack author may choose to use this configuration because the same dependency is used in different, unrelated ways in both images.
+
+Similarly, if a stack build image shares the mixin `mypkg` with the stack run image, a buildpack cannot require the mixin `build:mypkg` or `run:mypkg`, even though both images contain `mypkg`.
+
+This RFC proposed that we update the buildpack specification to create a new relationship between `mypkg`, `run:mypkg`, and `build:mypkg`.
+
+- While `run:mypkg` may imply a different set of changes to the run image compared to `mypkg`, buildpacks requiring `mypkg` should be satisfied with `run:mypkg` at runtime (and vice versa).
```
ekcasey, line 26: I understand the differences between the two contracts conceptually but it would be helpful to have a realistic example where the run-image changeset described by `run:<pk>` vs `<pkg>` would differ.

text/0000-relax-mixin-contract.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This change reduces the power of the mixin model for stack authors. Previously, stack authors could reliably use `run:mypkg` to indicate that a run image strictly contains the runtime version of mypkg (which may provide different functionality than `mypkg`).
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Refactor mixins entirely, e.g., enforce stack<->stack and stack<->buildpack mixin interfaces separately.
```
ekcasey, line 64: I am inclined to say we should either go this route or remove phase specific prefixes from `io.buildpacks.stack.mixins` entirely (keeping them in buildpack requirements to indicate whether they are required on one or both image).

The way I see it, the distinction between the changsets described by`build:x` and `x` either truly matter (in which case we should be enforcing the difference and refactoring the interface to make it easier to understand the difference) or they don't really matter, in which case we should remove the distinction between the contracts entirely.

If we are in the latter case, while it is true that some build image changes have an implied requirement for some run image changes, I think it think it is sufficient for a buildpack to use mixin `mypkg` to indicate both halves of that contract. The existance of build image mixins shouldn't introduce a requirement for complementary run-image mixins unless at least one buildpack actually depends on the mixin.

text/0000-relax-mixin-contract.md
sclevine: Example: requesting `build:libfoo` could mean adding just libfoo.so at build time for use by build tools, but requesting `libfoo` might mean adding both libfoo.so and libfoo.h at build time, so that apps can link against libfoo.so during build and expect it to be available at launch

text/0000-relax-mixin-contract.md
sclevine: Is coming up with this new interface worth blocking #87, vs. relaxing the restriction for now?

text/0000-relax-mixin-contract.md
ekcasey: Do we think users want that level of granularity in mixins? I.E. is there a good reason a user would definitely NOT want `libfoo.h` at build time? Sometimes getting headers when you don't need them might be worth it for the sake of simplicity in the contract? I would be less willing to make this argument given a realistic runtime example. Accidentally requiring anything extra at runtime seems worse than extras at build time.

Also, b/c stack authors can define the meaning of a mixin name, they could define two mixins: one that means `libfoo.so` and another that includes the headers (like the bionic mixin names). I think we should let folks build these distinctions into the mixin definitions themselves without specified phase prefixes.

text/0000-relax-mixin-contract.md
ekcasey: I think for now we should remove `build:` and `run:` entirely from `io.buildpacks.stack.mixins` (which is functionally similar to relaxing the requirement) and then reintroduce it later in a more thoughtful manner if we feel the need for it. I think the phase specific mixin UX is currently confusing to users and relaxing the requirement doesn't go far enough.

sclevine: > I think for now we should remove build: and run: entirely from io.buildpacks.stack.mixins

Not sure I understand. Wouldn't this be a huge breaking change that also forces stack image authors to include many additional, unnecessary packages in run images?

I'm very opposed to losing functionality we already have. This proposal just iterates what we already have into something that isn't more complex. While I totally believe we can find a more elegant solution to the problem, I don't see a disadvantage to making the change proposed.

text/0000-relax-mixin-contract.md
sclevine: I don't think this matters very much, practically. But it would be a more drastic change if `run:mypkg` and `mypkg` were considered the same mixin, and `run:` was just a property the buildpack/stack used to imply where the mixin must be installed. I'm worried that there are cases where the stack<->stack contract and buildpack<->stack contract are interwoven, so that the stackpack needs all of the information about where the mixin will be installed.

But I could probably be convinced to switch to this approach if you felt strongly that the iterative changes described are worse than the current state.

text/0000-relax-mixin-contract.md
ekcasey: > Wouldn't this be a huge breaking change that also forces stack image authors to include many additional, unnecessary packages in run images?

I am not suggesting that we get rid of `run:pkg` declarations in buildpacks stack requirements, but merely get rid of them in the stack image labels (i.e.`run:pkg` means `pkg` is required but only on the run image).

sclevine: That would break validating compatibility between the build image and run image.

text/0000-relax-mixin-contract.md
```diff
@@ -0,0 +1,69 @@
...
+# Summary
+[summary]: #summary
+
+Mixins are used to imply multiple contracts:
+
+1. A contract between a buildpack and a stack, where the buildpack may state that it can only be used with stack images that have certain dependencies.
+
+2. A contract between a stack run image and stack build image, where apps built on the build image must have a certain dependency present on the run image to start successfully on the run image.
+
+Currently, if a stack build image has the mixin `build:mypkg` and a stack run image has the mixin `run:mypkg`, a buildpack cannot require the mixin as `mypkg` (either statically in buildpack.toml, or dynamically via the stackpack mechanism described in #87). A stack author may choose to use this configuration because the same dependency is used in different, unrelated ways in both images.
```
hone, line 20:
```suggestion
Currently, if a stack build image has the mixin `build:mypkg` and a stack run image has the mixin `run:mypkg`, a buildpack cannot require the mixin as `mypkg` (either statically in buildpack.toml, or dynamically via the stackpack mechanism described in [#87](https://github.com/buildpacks/rfcs/issues/87)). A stack author may choose to use this configuration because the same dependency is used in different, unrelated ways in both images.
```

text/0000-relax-mixin-contract.md
sclevine: Switched as discussed.

text/0000-relax-mixin-contract.md
sclevine: Removed differences instead of providing an example, as discussed above.

text/0000-relax-mixin-contract.md
sclevine: Keeping intra-stack contracts for now, as discussed.

sclevine:
Should be ready for votes again.

nebhale (approved):

ekcasey (approved):

nebhale:
Final Comment Period with merge disposition, closing on 17 September, 2020.

text/0000-relax-mixin-contract.md
```diff
@@ -0,0 +1,69 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Pack and lifecycle will relax contractual restrictions when initiating builds, creating builders, and exporting images.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This change reduces the power of the mixin model for stack authors. Previously, stack authors could reliably use `run:mypkg` to indicate that a run image strictly contains the runtime version of mypkg (which may provide different functionality than `mypkg`). Now `run:mypkg` and `mypkg` imply the exact same thing for the run image.
```
dfreilich, line 60: Would you mind clarifying this point? Are `run:mypkg` and `mypkg` different packages? If yes, why do they satisfy each other, and if not, why would they provide different functionality? A concrete example would be fantastic!

hone: I believe they could be different packages. The apt contract we have applies only to the `io.buildpacks.stacks.bionic` stack as per this [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0031-bionic-mixins.md).

hone (approved):

hone (approved):

text/0000-relax-mixin-contract.md
hone: @sclevine mind chiming in here?

sclevine: The set of changes to the run image represented by `run:mypkg` and `mypkg` are the same, so if `mypkg` represents an OS package in the context of the run image, `run:mypkg` must represent that package as well.

PR #110, opened 2020-08-21
Buildpacks should be able to provide the default process for an app

natalieparellano:
[Readable](https://github.com/natalieparellano/rfcs/blob/buildpacks-default-process/text/000-buildpacks-contribute-default-process-type.md)


text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,113 @@
...
+- Why should we do this? Easier and more intuitive for end users.
+- What use cases does it support? Apps with one or more process type(s).
+- What is the expected outcome? The experience should be seamless for end users. Buildpacks should produce a sensible default process type without the user having to know or care that process types exist. The user should still be able to override the default process type by passing `-process-type` to the lifecycle.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Example: single process app with a buildpack declaring a `default-process` of type `web` -> the app image will have a default process type of `web`
+
+Example single process app with no buildpacks declaring a `default-process` -> the app image will have no default process. Users must specify the process to run at runtime. (This assumes https://github.com/buildpacks/spec/pull/137)
```
ekcasey, line 37:
```suggestion
Example: single process app with no buildpacks declaring a `default-process` -> the app image will have no default process. Users must specify the process to run at runtime. (This assumes https://github.com/buildpacks/spec/pull/137)
```

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,113 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The changes from this RFC should require a bump to the buildpack API.
+
+Currently, during the `build` phase, buildpacks may declare process types by writing entries to `<layers>/launch.toml`. This RFC proposes a new key in `<layers>/launch.toml`, `[default-process]`, that specifies what each buildpack would like to designate as the default process type for the app image.
+
+```toml
+[default-process]
```
ekcasey, line 63: I wonder if we should consider adding a boolean `override` key to the `default-process` table. Then buildpacks could indicate either "this should be the default if there is none" (`override = false`) or "make this the default regardless" ( `override = true`). This would add slightly more complexity but might be very nice to have in certain use cases.

If we added this key I suspect `override` should be `true` by default.

text/000-buildpacks-contribute-default-process-type.md
ekcasey: @natalieparellano I stopped proofreading b/c this is a draft. Let me know if/when you are ready for proofreading level feedback.

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Ah. So if a buildpack with `override = true` sets the default process, the user cannot override?

nebhale: @ekcasey Can you give an example of where this might be useful?  It seems like this effectively inverts the precedence chain (first wins instead of last wins) and could be achieved by a user doing the same but changing the order of the buildpacks.

nebhale: @natalieparellano No, I don't think it stops a user from contributing.  Envision this as all the buildpacks do their default process type contributions and if the user has specified one, theirs comes last and would override everything else.

This change is a suggestion such that right now a chain of A -> B -> C always results in C being the selected default where as `A:false -> B:true -> C:false` results in B being the default.  The flag is whether _that_ default will override one prior to it.

natalieparellano: Added the concept of `override` to the RFC.

jromero:
@natalieparellano this was discussed at some length today during working group. I will try to update this comment to the link to the recording (on YouTube) when it becomes available.

hone:
Here's the [recording](https://www.youtube.com/watch?v=eIKy0JrVJ6A).

Some notes from what I recalled from yesterday. @sclevine had some concerns about doing this at the buildpack level. He proposed if this should be up to the platform. One of the ideas that came up was trying to use something like Stackpacks or a Platform Pack as a generate mechanism to not just solve this problem. It would need to happen between the build and export phases, but probably doesn't need root access. @jromero expressed concerns that this would make images non portable depending on what platform they're built on given the same source, stack, and buildpacks.

nebhale (approved):

jabrown85 (approved):

ekcasey (approved):

jkutner (approved):

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Overrides
+
+Example: multi-process app with an earlier buildpack declaring a `default-process` of type `web` and a later buildpack declaring a `default-process` of type `worker` -> the app image will have a default process type of `worker`
+
+Example: multi-process app with an earlier buildpack declaring a `default-process` of type `web` and a later buildpack declaring a `default-process` of type `worker` and `override = false` -> the app image will have a default process type of `web`
+
+Example: multi-process app with a buildpack declaring a `default-process` of type `web` and the user passes `-process-type worker` -> the app image will have a default process type of `worker`
+
+#### Overrides - edge cases
```
natalieparellano, line 53: Should we warn when these occur?

@sclevine added this section and https://github.com/buildpacks/rfcs/pull/110/files#diff-419c1902fa5125d1c8e9a11131832217R97 to address some of your concerns from WG.

ekcasey: > Should we warn when these occur?

Just my two cents..

> Example: multi-process app with a buildpack declaring a process type worker and a default-process of type web -> the app image will have no default process. Users must specify the process to run at runtime.

I think this one ^ should be an error case, assuming the buildpack attempting to set the default did not declare a `web` process type at all.

> Example: multi-process app with an earlier buildpack declaring a default-process of type web and a later buildpack redefining what web means -> the app image will have no default process. Users must specify the process to run at runtime.

I think this one ^ should be a warning.

> Example: multi-process app with an earlier buildpack declaring a default-process of type web and a later buildpack redefining what web means while also declaring web as the default-process -> the app image will have a default process type of web

And this last one should succeed without warning.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,133 @@
...
+
+The lifecycle if not provided any `-process-type` will set the default process type to the only process type (if there is only one), otherwise it will warn and set the image entrypoint to `/cnb/lifecycle/launcher` as described above. This has the advantage of "just working" for single-process apps. However this requires the lifecycle to be "opinionated" and may be surprising for some users. See [spec PR](https://github.com/buildpacks/spec/pull/137) for the proposed removal of this behavior which should go hand-in-hand with this RFC.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should `pack` continue to pass `-process-type web`? It could inspect the buildpack API and only pass the flag if the API version is below 0.X. 
+- What about buildpacks implementing different buildpack APIs that need to work together?
```
hone, line 128: Thanks for marking this question. I'm interested in what people think should happen here. Since this functionality doesn't exist in previous APIs I believe there are only two options:
* we assume buildpacks in a previous API provide no default process types and this RFC would function as stated. This is clean but may come as a surprise
* we assume buildpacks that define a `web` process type intend for it to be default.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,133 @@
...
+
+The lifecycle if not provided any `-process-type` will set the default process type to the only process type (if there is only one), otherwise it will warn and set the image entrypoint to `/cnb/lifecycle/launcher` as described above. This has the advantage of "just working" for single-process apps. However this requires the lifecycle to be "opinionated" and may be surprising for some users. See [spec PR](https://github.com/buildpacks/spec/pull/137) for the proposed removal of this behavior which should go hand-in-hand with this RFC.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should `pack` continue to pass `-process-type web`? It could inspect the buildpack API and only pass the flag if the API version is below 0.X.
```
hone, line 127: If `pack` passes `-process-type web` does it look like the user is overriding the process type?

hone:
Thanks for all the hard work you've put in this RFC @natalieparellano! I like all the cases outlined here.

text/000-buildpacks-contribute-default-process-type.md
hone: Do we want to add this to the RFC or leave this out?

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,149 @@
...
+type = web
+command = bundle exec ruby app.rb
+  [[processes.default]]
+  override = true
+
+[[processes]]
+type = worker
+command = bundle exec ruby worker.rb
+override = false
+```
```
simonjjones, line 101: As discussed in the WG, a possible alternative to these combinations of `override` & `default` could be to provide a single `priority` parameter. This could either be a generic integer, where the process claiming the highest priority (probably with the lowest number) would become the default process for the app; or a predetermined list of priority strings such as "highest", "high", "medium", "low" & "background".

Not only would this approach provide an overriding highest priority process for the app, but this would also provide a simple way of determining the highest priority process for each process type.

The example above could be rewritten as

```toml
[[processes]]
type = web
command = bundle exec ruby app.rb
priority = 1 // highest

[[processes]]
type = worker
command = bundle exec ruby worker.rb
priority = 4 // low
```

natalieparellano:
Adding some notes from today's WG:
* It would be good to understand further what is the use case for a buildpack declaring "this process should be the default, but only if there isn't already a default" (override false); @ekcasey can you share your thoughts here?
* The currently proposed nested `[processes.default]` is a little confusing (empty table shouldn't change the behavior, etc.), an alternative in a similar vein might be a key such as `default = ["always", "if-not-present", "never"]` (or something)
* It might make sense to clarify if `override` means `overriding` vs. `overridable`

natalieparellano:
Adding some more notes:
* We concluded in WG that `override` is likely adding more confusion than value, propose taking it out
* `default` should still be a property of the process, so if an earlier buildpack defines `web` with `default = true` and a later buildpack defines `web` with `default = false` there will be no default process type
* Adding further refinements to the concept of default (such as "if not present") introduces confusion when later buildpacks' process types override earlier ones, we could wait to introduce this when (a) there's evidence that it's needed and (b) consider namespacing process types by buildpack so that overriding the process type definition is not a thing

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Yep, from the lifecycle's perspective it's the same thing.

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Removed the concept of `override` per most recent discussion in WG.

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Now that `default` is a property of the process, the first case (error case) isn't relevant anymore.

Noted in the RFC that the lifecycle should print a warning for the second case:

> Example: multi-process app with an earlier buildpack declaring a default-process of type web and a later buildpack redefining what web means -> the app image will have no default process. Users must specify the process to run at runtime.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,138 @@
...
+
+The lifecycle if not provided any `-process-type` will set the default process type to the only process type (if there is only one), otherwise it will warn and set the image entrypoint to `/cnb/lifecycle/launcher` as described above. This has the advantage of "just working" for single-process apps. However this requires the lifecycle to be "opinionated" and may be surprising for some users. See [spec PR](https://github.com/buildpacks/spec/pull/137) for the proposed removal of this behavior which should go hand-in-hand with this RFC.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should `pack` continue to pass `-process-type web`?
+  - It could inspect the buildpack API and only pass the flag if the API version is below 0.X.
+  - Builders could set an environment variable to communicate to `pack` that it doesn't need to pass -process-type to the lifecycle
```
natalieparellano, line 131: @jromero it would be good to get your thoughts here.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,138 @@
...
+- Buildpacks could stomp on each other, each trying to declare different default process types. This might be confusing.
+- Changing the way things work now could surprise some people.
+- One more thing for buildpack authors to know and care about.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Keep things as they are.
+  - Earlier iterations of this PR.
```
jkutner, line 113: could we capture a brief summary of some of those iterations? in particular override

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Relevant [Slack conversation](https://buildpacks.slack.com/archives/CSFE49VGR/p1602691248066000)

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,138 @@
...
+
+default process type - describes the process type that will be started by the launcher when the app image is run. It can be overriden by overriding the image entrypoint. Example: an image with default process type of web would have an entrypoint of `/cnb/process/web` which is a symlink that points to `/cnb/lifecycle/launcher`. The launcher uses argv0 to determine which process to start.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this? Easier and more intuitive for end users.
+- What use cases does it support? Apps with one or more process type(s).
+- What is the expected outcome? The experience should be seamless for end users. Buildpacks should produce a sensible default process type without the user having to know or care that process types exist. The platform should still be able to override the default process type by passing `-process-type` to the lifecycle.
```
jabrown85, line 30: If one day in the mid-future `pack` no longer passes `web`, the *hope* is that the buildpacks will produce a runnable image with `web` because the buildpacks were updated. But in truth, this will require buildpack authors to make a change before this future date.

Is there some world in which we might use the latest buildpack API to determine if `pack` should send `web` in? If all your buildpacks are on the old buildpack API, `pack` will always send in `web`. If any of your buildpacks are on this new spec version, `pack` no longer sends `web`. 

Maybe we could go further and perhaps make `default = true` for buildpacks that produce `web` processes that are running on the old buildpack spec. 

I'm still a bit worried about buildpacks not being updated and an impending deadline to update them that the authors may not be aware of.

natalieparellano: > Maybe we could go further and perhaps make default = true for buildpacks that produce web processes that are running on the old buildpack spec.

This seems like a clean way of solving the problem, obviating the need for `pack` to send `web` at all. It is pushing knowledge about `web` back into the lifecycle, but only for older buildpack APIs (and defaulting to `web` is already the behavior for older platform APIs).

text/000-buildpacks-contribute-default-process-type.md
jromero: For 3, what would happen if the buildpack is setting a default process type of "worker" but pack continues to pass "web"?

Am I right in assuming that the end result image would A. warn if "web" process type doesn't exist but continue to set "worker" as process type or B. set "web" process type if it exists.

jromero: Seems like this conversation is also relevant: https://github.com/buildpacks/rfcs/pull/110/files#r504925751

May be worth discussing more at length (yet again, I'm sure) during a WG meeting.

nebhale (approved):

text/000-buildpacks-contribute-default-process-type.md
ekcasey: > Maybe we could go further and perhaps make default = true for buildpacks that produce web processes that are running on the old buildpack spec.

Practically I think this ^ suggestion would produce behavior that is identical to this one [from the other thread](https://github.com/buildpacks/rfcs/pull/110#discussion_r504918609)

> Have lifecycle default to web when no default process is set. 

The latter framing seems easier to explain to me, because we aren't creating new rules about old APIs.

hone: Can we capture this migration plan in this RFC?

Do we need a way for the platform to pass a default process type that is separate from the users or is that just a platform feature ala `pack`?

hone (commented):
Overall, :+1: just awaiting the addition of @jabrown85's migration idea.

text/000-buildpacks-contribute-default-process-type.md
ekcasey: > Maybe we could go further and perhaps make default = true for buildpacks that produce web processes that are running on the old buildpack spec.

After another day of thinking about this I have decided that @jabrown85 is correct here and I was wrong above. Tying this to the buildpack API allows us to phase out this behavior as buildpacks upgrade.

text/000-buildpacks-contribute-default-process-type.md
jromero: IIRC, yesterday during working group we explored the idea of easing the migration from a platform's perspective by changing the behaviour of `-process-type`. It would now fail if the process type is not available as opposed to simply warning as it currently does.

By making this change in conjunction to making buildpacks own setting default process types platforms have a more specific piece of data to look at and determine whether they should be passing in a "default" process type.

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Added a section to describe the migration approaches discussed: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR139-R160

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Added here: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR113-R120

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Added a section to describe the migration approaches discussed: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR139-R160

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,168 @@
...
+    Pros:
+
+    * Usage of `CNB_PLATFORM_API` will help smooth the transition for other platforms like Tekton.
+
+    Cons:
+
+    * We can't assume that buildpacks will set `default = true`. This could still lead to a prevalence of app images where the start command must be specified by the user at runtime.
+    * The lifecycle's warnings could go unseen when `pack` is used in a CI system.
+
+  B. We introduce a new flag, `-platform-process-type` which will be the default process type IF neither the user (via `-process-type`) nor buildpacks specify a default process type. `-platform-process-type <unknown process type>` would be a warn case, and `-process-type <unknown process type>` could then become a failure case.
```
natalieparellano, line 152: It's worth noting, this would also require a platform API bump.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,168 @@
...
+The lifecycle if not provided any `-process-type` will set the default process type to the only process type (if there is only one), otherwise it will warn and set the image entrypoint to `/cnb/lifecycle/launcher` as described above. This has the advantage of "just working" for single-process apps. However this requires the lifecycle to be "opinionated" and may be surprising for some users. See [spec PR](https://github.com/buildpacks/spec/pull/137) for the proposed removal of this behavior which should go hand-in-hand with this RFC.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should `pack` continue to pass `-process-type web`?
+
+  Options discussed:
+
+  A. `pack` stops sending `-process-type web` when `CNB_PLATFORM_API` is greater than `0.X`. The lifecycle will assume `default = true` for `web` processes from buildpacks implementing buildpack APIs earlier than `0.Y`. The lifecycle could print a warning when buildpacks do not specify a default process type. `-process-type <unknown process type>` could then become a failure case (vs. warn, what it is today).
```
natalieparellano, line 141: It sounds like this is the preferred option. Unless there are any objections, I can update the RFC to state that this will be the migration path.

ekcasey: I agree that we should go forward with this one.

One small nitpick: Not sure if I think the lifecycle should print a _warning_ if buildpacks do not specify a default. I think we imagined "no default" may be an intentional case for some users of the new APIs and we try not to warn on things that could be intentional. We can certainly print an info statement though.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -127,8 +135,30 @@ The lifecycle if not provided any `-process-type` will set the default process t
 [unresolved-questions]: #unresolved-questions
 
 - Should `pack` continue to pass `-process-type web`?
-  - It could inspect the buildpack API and only pass the flag if the API version is below 0.X.
-  - Builders could set an environment variable to communicate to `pack` that it doesn't need to pass -process-type to the lifecycle
+
+  Options discussed:
+
+  A. `pack` stops sending `-process-type web` when `CNB_PLATFORM_API` is greater than `0.X`. The lifecycle will assume `default = true` for `web` processes from buildpacks implementing earlier buildpack APIs. The lifecycle could print a warning when buildpacks do not specify a process type. `-process-type <unknown process type>` could then become a failure case (vs. warn, what it is today).
```
jromero: I'm really in favor of this option. How can we carry it forward? Can we incorporate it into this PR? Add it to the next version of the spec? Or create a new RFC for just that change?

@ekcasey any preference?

jromero: It's also worth noting that this isn't solely for `pack` but would also be useful for any other platform.

ekcasey: I agree
https://github.com/buildpacks/rfcs/pull/110#discussion_r515111439 https://github.com/buildpacks/rfcs/pull/110#discussion_r510205574

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: I believe this is covered [here](https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR166) and [here](https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR105).

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Migration plan covered here: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR103-R114

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Migration plan covered here: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR103-R114

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Migration plan covered here: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR103-R114

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Updated the language from warn -> info message.

text/000-buildpacks-contribute-default-process-type.md
natalieparellano: Migration plan covered here: https://github.com/buildpacks/rfcs/pull/110/files#diff-00a2fb620563b974f876628cc19e746a4a96606f9dea0fd58dc1bb13f010b4abR103-R114

natalieparellano:
I want to say that with the `pack` migration plan, all the outstanding questions on this one have been addressed.

text/000-buildpacks-contribute-default-process-type.md
```diff
@@ -0,0 +1,180 @@
...
+
+If a buildpack attempts to define two processes with `default = true` specified, the lifecycle will fail.
+
+#### `pack` migration plan
+
+`pack build` has a `--default-process` flag. If this flag is not provided by the user, `pack` will provide `-process-type web` to the lifecycle. As currently architected, this will prevent any default process type contributed by buildpacks from ever taking hold, as the lifecycle will prefer the provided `-process-type` to any default process type contributed by buildpacks.
+
+Migration plan:
+
+`pack` stops sending `-process-type web` when `CNB_PLATFORM_API` is greater than `0.X`. The lifecycle will assume `default = true` for `web` processes from buildpacks implementing buildpack APIs earlier than `0.Y`. The lifecycle could print an info message when buildpacks do not specify a default process type. `-process-type <unknown process type>` could then become a failure case (vs. warn, what it is today).
```
hone, line 105: Do we define `0.X` and `0.Y` anywhere?

natalieparellano: Nope. I'm not sure that they need to be any particular value for the RFC to function as intended, but practically speaking it seems like they would both be 0.6.

hone (approved):

text/000-buildpacks-contribute-default-process-type.md
hone: well, from discussions I assume you mean by `0.X` you mean "a version of the Platform API that does not include this RFC" and `0.Y` is a version of the Buildlpack API that doesn't include this RFC.

jromero (approved):

sclevine:
FCP closing on 11/11

ryanmoran (approved):

PR #111, opened 2020-08-27
Stack buildpacks

jkutner:
This RFC has been split out from https://github.com/buildpacks/rfcs/pull/87 for the purpose of making the Stack buildpack concept easier to understand independently from App mixins. However these two RFCs are mutually dependent on one-another.

[Readable](https://github.com/buildpacks/rfcs/blob/stack-buildpacks/text/0000-stack-buildpacks.md)


text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+ A stack buildpack (a.k.a. stackpack) is a special case of buildpack that has the following properties:
+
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
+* Does not have access to the application source code
+* Is run before all regular buildpacks
```
hazemm96, line 50: I thought it ran during the extend phase (according to line 38) which occurs after all regular buildpacks

hazemm96: Or perhaps the extend phase is only for the run image being updated. Ok nvm

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+A new type of buildpack, called a Stack buildpack, may run against a stack (both build and run images) in order to extend it in ways that are only possible by running privileged commands. A stackpack may also define a list of mixins that it provides to the stack. In this way, a stack that is missing a mixin required by a buildpack may have that mixin provided by a stack buildpack.
+
+A stack provider may choose to include stack buildpacks with the stack they distribute. If a stack includes any stack buildpacks, the following will occur when the build phase starts:
+
+1. If any requested mixin is not provided by the stack, the lifecycle will compare the missing mixins to the static list of mixins provided by stack buildpacks. If any mixins are still not provided, the build will fail.
+1. The lifecycle will run the detect phase for all stackpacks defined in the builder.
+1. The lifecycle will execute the stack buildpack build phase for all passing stackpack(s). If no stack buildpacks pass detect, the build will continue the build phase as normal (running userspace buildpacks).
+1. After the lifecycle's build phase, the lifecycle will begin a new phase called _extend_.
```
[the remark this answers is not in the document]
jkutner: No, it could even be run in parallel to the build I think

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+A number of changes to the Platform Specification will be required to execute Stack Buildpacks. Those changes will be defined in a separate RFC.
+
+## Stack buildpacks
+
+Stack buildpacks are identical to other buildpacks, with the following exceptions:
+
+1. The `<layers>` directory is NOT writable.
```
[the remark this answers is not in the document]
jkutner: There may not be. And we may prefer to just ignore any writes to the `<layers>` dir anyways.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+[buildpack]
+privileged = <boolean (default=false)>
+non-idempotent = <boolean (default=false)>
+
+[buildpack.mixins]
+any = "<boolean (default=false)>"
+names = [ "<mixin name>" ]
+ ```
+
+* `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user.
```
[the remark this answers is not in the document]
jkutner: Yes, I've left it somewhat generic because in the future we could add a feature that would allow `root` access for buildpacks that are not stackpacks (but that would break `rebase`)

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, its detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
```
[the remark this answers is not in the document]
jkutner: Yea, this needs fleshing out. We've talked about a new `run-builder` concept, but haven't decided on anything yet.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+A buildpack that can install an arbitrary list of mixins would have a `buildpack.toml` like this:
+
+ ```toml
+[buildpack]
+id = "example/apt"
+privileged = true
+
+[buildpack.mixins]
+any = true
```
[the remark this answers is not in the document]
jkutner: Fixed in aae57b2

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not write to the `/layers` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase excluding the a few specific directories and files.
+
+Alternatively, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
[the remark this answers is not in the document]
jkutner: fixed in aae57b2

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: fixed in aae57b2

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: fixed in aae57b2

nebhale (approved):

sclevine (changes requested):

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+ This proposal adds new keys to the `[buildpack]` table in `buildpack.toml`, and a new `[[mixins]]` array of tables:
+
+ ```
+[buildpack]
+privileged = <boolean (default=false)>
+non-idempotent = <boolean (default=false)>
+
+[buildpack.mixins]
+any = "<boolean (default=false)>"
+names = [ "<mixin name>" ]
```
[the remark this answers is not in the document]
jkutner: yea, this is a mistake. We already agreed on what your describe

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not create layers using the `<layers>` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase excluding the a few specific directories and files.
+
+Additionally, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
```
ekcasey, line 59: > Additionally, a platform may store a new stack image to cache the changes.

What do we mean by this exactly? Maybe we should leave this out unless we want to describe how we expect platforms to do this.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory. By default, an implicit order will be created for all stackpacks included in a stack. The order can be overridden in the `builder.toml` with the following configuration:
```
ekcasey, line 61: Instead of having an implicit alphabetical order, I think we should just have a `/cnb/stack/order.toml` file that should be included on the `build-image` along with stackpacks.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+```
+[[stack.order]]
+  [[stack.order.group]]
+    id = "<stackpack id>"
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, its detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
```
ekcasey, line 69: I find this description of the input confusing.

Should we say that the `/bin/build` script will have the same arity as the regular buildpacks instead of the same parameters? Maybe we should just describe the expected inputs to the stack buildpack without comparing them to a regular buildpack? "The `<layers>` dir must not be the layers dir" (paraphrasing) is a confusing way to think about thing imho.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
```
ekcasey, line 71: > The `stack.toml` will define paths that will be restored even when the base image changes

I think we need a schema for `stack.toml` and a more thorough description of the behavior in this RFC.

What do we mean by "paths that will be restored even when the base image changes"? Here are the different ways I can interpret this:

1. We creating and caching a separate cache layer that is a subset of the snapshot layer and putting it in a directory or image cache? We restore this during the traditional "restore" phase.
    * In this case wouldn't we be restoring on every build just like a normal cache layer? "even when the base image changes" feel unnecessary. 
    * Should we actually be slicing "exactly one" layer into multiple layers so we can talk about their properties (launch/cache etc) in a straightforward way?
2. We potentially add a restore component to rebase? But that doesn't make sense b/c a normal rebase should work here and we haven't RFC'd or specified an hypothetical `upgrade` operation yet.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+Normally, buildpacks do not run as root. This is an intentional design decision that ensures operations like `rebase` will work on day-2.
+
+However, many applications and buildpacks require modifications to the stack they run on, such as adding system packages or custom certificates. For this reason, we need a mechanism that buildpack authors and buildpack users and leverage to extend their stacks.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *root buildpack* - a new type of buildpack that runs as the root user
+- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
```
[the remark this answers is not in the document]
jkutner: yea, this is just old

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+privileged = <boolean (default=false)>
+non-idempotent = <boolean (default=false)>
+
+[buildpack.mixins]
+any = "<boolean (default=false)>"
+names = [ "<mixin name>" ]
+ ```
+
+* `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user.
+* `non-idempotent` - when set to `true`, indicates that the buildpack is not idempotent. The lifecycle will provide a clean filesystem from the stack image(s) before each run (i.e. no cache).
```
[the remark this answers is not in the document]
jkutner: Maybe not. I think `excludes` in `launch.toml` might make it harder to handle cases where you don't know what files are modified on the fs. But if we support wildcards, that would suffice.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+ ```bash
+#!/usr/bin/env bash
+
+apt update -y
+
+for package in $(cat $3 | yj -t | jq -r ".entries | .[] | .name"); do
+  apt install $package
+done
+
+cat << EOF > launch.toml
```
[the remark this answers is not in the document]
jkutner: Yea, I think it should be `$1/launch.toml`. The `stack.toml` is a holdover from previous iterations of this RFC

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
+```
+
+Its `bin/detect` would have the following contents:
+
+ ```bash
+#!/usr/bin/env bash
+
+exit 0
```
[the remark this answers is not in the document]
jkutner: How would it know what `[[provides]]` to add to the build plan? The buildpack's detect phase doesn't receive any input. Can it do something like:

```
[[provides]]
any = true
mixin = true

File: text/0000-stack-buildpacks.md (hunk 12 above)
Comment (jkutner) [reply text/0000-stack-buildpacks.md]: this should be `launch.toml`. The schema is down below, but I can definitely add more detail like this in there too

File: text/0000-stack-buildpacks.md [hunk 17]
```diff
@@ -0,0 +1,299 @@
...
+- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
+- *userspace buildpack* - the traditional definition of a buildpack (i.e. does not run as root, and runs against an app)
+
+A new type of buildpack, called a Stack buildpack, may run against a stack (both build and run images) in order to extend it in ways that are only possible by running privileged commands. A stackpack may also define a list of mixins that it provides to the stack, or indicate that it will provide _any_ mixin. In this way, a stack that is missing a mixin required by a buildpack may have that mixin provided by a stack buildpack.
+
+A stack provider may choose to include stack buildpacks with the stack they distribute. If a stack includes any stack buildpacks, the following will occur when the build phase starts:
+
+1. If any requested mixin is not provided by the stack, the lifecycle will compare the missing mixins to the static list of mixins provided by stack buildpacks. If any mixins are still not provided, the build will fail.
```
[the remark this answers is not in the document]
jkutner: Do we still care about a step where the platform compares a static list of required mixins (from the buildpack.toml and potentially the project.toml) to a static list of provided mixins (from the stack and the stackpacks' buildpack.toml)?

/cc @sclevine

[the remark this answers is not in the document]
jkutner: fixed in 7742fcb

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+However, many applications and buildpacks require modifications to the stack they run on, such as adding system packages or custom certificates. For this reason, we need a mechanism that buildpack authors and buildpack users and leverage to extend their stacks.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *root buildpack* - a new type of buildpack that runs as the root user
+- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. It is distinguished by a static list of mixins it can provide.
+- *userspace buildpack* - the traditional definition of a buildpack (i.e. does not run as root, and runs against an app)
+
+A new type of buildpack, called a Stack buildpack, may run against a stack (both build and run images) in order to extend it in ways that are only possible by running privileged commands. A stackpack may also define a list of mixins that it provides to the stack, or indicate that it will provide _any_ mixin. In this way, a stack that is missing a mixin required by a buildpack may have that mixin provided by a stack buildpack.
```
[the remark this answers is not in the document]
jkutner: fixed in 3cba356

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+ A stack buildpack (a.k.a. stackpack) is a special case of buildpack that has the following properties:
+
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
```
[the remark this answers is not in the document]
jkutner: fixed in 26944bd

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+* Does not have access to the application source code
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not create layers using the `<layers>` directory
+
+The stackpack interface is identical to the buildpack interface (i.e. the same `bin/detect` and `bin/build` scripts are required). However, some of the context it is run in is different from regular buildpacks.
```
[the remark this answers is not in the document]
jkutner: What I'm trying to say with "interface" is that the positional arguments to the scripts, the env vars, etc are all the same. The values and the semantics might be different though. Is there a good word for that?

text/0000-stack-buildpacks.md
jkutner: i think someone (maybe @sclevine) suggested this as a performance optimization thing. I'm happy to omit it.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, its detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
```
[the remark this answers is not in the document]
jkutner: In the `rebase` scenario you describe, I think Stephen is worried about buildpacks that do things like append to `/etc/passwd`. Because there no concept of "merging" changes in the base-image with changes in the snapshot, the snapshot would essentially replace the new file in the base-image with the version it used during build.

We could potentially detect this and prevent the build from succeeding?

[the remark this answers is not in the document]
sclevine: > preserve the safety of the existing rebase functionality without requiring a rerun of the root buildpacks.

This is never safe. The ABI-compatibility restriction for stackpacks is to ensure that the app can still run on the run image **after the stackpack is re-run**, and has nothing to do with rebasing between ABI-compatible OS package layers.

There are many, entirely undetectable classes of errors that crop up when attempting to rebase between OS package layers. E.g., a package install script could read from something in the base layer that later changes. Even the undetectable errors that don't write overlapping files could introduce security vulnerabilities.

text/0000-stack-buildpacks.md
sclevine: Okay to omit from my perspective.

text/0000-stack-buildpacks.md
sclevine: Agree, implicit alphabetical order seems confusing. Also would be good to have entirely optional stackpacks that the builder must opt-in to.

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
sclevine: There's no `cache` flag for the "main" layer. I would support moving the property to `launch.toml` though.

text/0000-stack-buildpacks.md
sclevine: I think the intended behavior is that for stackpack builds in all three scenarios (build-time build, build-time run, and rebase run), globs of files may be marked as layers to be excluded the image when it is "used." Any excluded layer may also be marked as cached, so that those excluded layers are recovered when the image is "used." What "used" means is defined as:

1. **Build-time build:** A given layer is excluded at normal buildpack build-time, and recovered the next time the build image is extended with the stackpack.
2. **Build-time run:** A given layer is excluded from the final image, and recovered the next time the run image is extended with the stackpack (either rebase or rebuild).
2. **Rebase run:** A given layer is excluded from the rebased image, and recovered the next time the run image is extended with the stackpack (either rebase or rebuild).

jkutner: Addressed in 09316a1

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: fixed in d7d3ff3

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: fixed in f9fa626

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: fixed in 3cba356

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+- *userspace buildpack* - the traditional definition of a buildpack (i.e. does not run as root, and runs against an app)
+
+A new type of buildpack, called a Stack buildpack, may run against a stack (both build and run images) in order to extend it in ways that are only possible by running privileged commands. A stackpack may also define a list of mixins that it provides to the stack, or indicate that it will provide _any_ mixin. In this way, a stack that is missing a mixin required by a buildpack may have that mixin provided by a stack buildpack.
+
+A stack provider may choose to include stack buildpacks with the stack they distribute. If a stack includes any stack buildpacks, the following will occur when the build phase starts:
+
+1. If any requested mixin is not provided by the stack, the lifecycle will compare the missing mixins to the static list of mixins provided by stack buildpacks. If any mixins are still not provided, the build will fail.
+1. The lifecycle will run the detect phase for all stackpacks defined in the builder.
```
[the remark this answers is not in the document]
jkutner: fixed in 7742fcb1693293ca148507da7239ad44b7170630

text/0000-stack-buildpacks.md
ekcasey: > Build-time build: A given layer is excluded at normal buildpack build-time, and recovered the next time the build image is extended with the stackpack.

What do we mean by "excluded at normal buildpack build-time" exactly? That is not possible without undoing the changes right?

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
ekcasey: If the main layer cannot be cached, then what is the point of this flag. In that case, won't we always "provide a clean filesystem"?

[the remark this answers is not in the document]
jkutner: default is to cache the "main" layer. There's no flag because you can only opt-out by using the `[[excludes]]` paths

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+1. The lifecycle will run the detect phase for all stackpacks defined in the build-image.
+1. The lifecycle will execute the stack buildpack build phase for all passing stackpack(s). If no stack buildpacks pass detect, the build will continue the build phase as normal (running userspace buildpacks).
+1. After, during, or before the lifecycle's build phase, the lifecycle will begin a new phase called _extend_.
+1. During the extend phase, stack buildpacks that passed detection will run against the run images accordingly (see details below).
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+ A stack buildpack (a.k.a. stackpack) is a special case of buildpack that has the following properties:
```
yaelharel, line 47: I think it's worth moving the definition of `stackpack` to the "what it is" section because you're using it on line 32.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+- *root buildpack* - a new type of buildpack that runs as the root user
+- *stack buildpack* - a type of root buildpack that runs against the stack image(s) instead of an app. Stack buildpacks must not make changes to the build and run images that either violate stack [compatibility guarantees](https://github.com/buildpacks/spec/blob/main/platform.md#compatibility-guarantees) or violate the contract defined by that stack's author.
+- *userspace buildpack* - the traditional definition of a buildpack (i.e. does not run as root, and runs against an app)
+
+A new type of buildpack, called a Stack buildpack, may run against a stack (both build and run images) in order to extend it in ways that are only possible by running privileged commands. Unlike userspace buildpacks, stack buildpack can modify any path on the filesystem. Userspace buildpack can only create/modify disjoint layers (either by adding a dir to `<layers>` or modifying an app slice), which makes possible features like individual layer reuse that is independent or ordering.
```
yaelharel, line 30: The first part of the sentence looks like a redefinition of line 27. Maybe you can combine both definitions together?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+
+A stackpack may also define a list of mixins that it provides to the stack, or indicate that it will provide _any_ mixin. In this way, a stack that is missing a mixin required by a buildpack may have that mixin provided by a stack buildpack.
+
+A stack provider may choose to include stack buildpacks with the stack they distribute. If a stack includes any stack buildpacks, the following will occur when the build phase starts:
+
+1. The platform may compare the list of requested mixins with the static list of mixins provided by the stack buildpacks, and fail the build if it choses to do so.
```
yaelharel, line 36:
```suggestion
1. The platform may compare the list of requested mixins with the static list of mixins provided by the stack buildpacks, and fail the build if it chooses to do so.
```

Can you please elaborate why the platform may choose to fail the build? I guess it relates to some missing mixins.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+* Can only create one layer
+* May include, in the created layer, modifications to any part of the filesystem, excluding `<layers>`, `<platform>`, and `<cnb>` directories
+* Does not have access to the application source code
+* Is run before all regular buildpacks
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not create layers using the `<layers>` directory
+
+The stackpack interface is similar to the buildpack interface
+* the same `bin/detect` and `bin/build` scripts are required
```
yaelharel, line 60:
```suggestion
* The same `bin/detect` and `bin/build` scripts are required
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+* the `bin/detect` will have read-only access to the app
```
yaelharel, line 61:
```suggestion
* The `bin/detect` will have read-only access to the app
```

text/0000-stack-buildpacks.md
jkutner: fixed in 6e7f0cf

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+ A stack buildpack is a special case of buildpack that has the following properties:
+
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
+* May include, in the created layer, modifications to any part of the filesystem, excluding `<layers>`, `<platform>`, and `<cnb>` directories
+* Does not have access to the application source code
```
natalieparellano, line 53:
```suggestion
* Does not have access to the application source code during the `build` phase
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+
+A Stack Buildpack that needs to install mixins must select them from the build plan.
+
+## Caching and Restoring
+
+During the export phase, the lifecycle will store any snapshot layers created during the build phase in the cache.
+
+During the restore phase of the next build, the lifecycle will download the snapshot layer with the cache and store it as a tarball in the `<layers>` directory (i.e. it will not extract it). The `restorer` cannot extract the snapshot because it will not run with root privileges. In addition, the `restorer` may run in a different container than the build, which means changes made to the base image are not guaranteed to carry forward.
+
+During the build phase, the lifecycle will extract and apply each snapshot tarball before running stack buildpacks, but after a snapshot-baseline has been captured. This ensures that all changes from the previous snapshot are preserved even if the stack buildpack does not make any additional changes.
```
natalieparellano, line 163: Maybe it makes sense to call out here (assuming I understand it correctly) that the `builder` will run as root when running stack buildpacks, and then drops privileges to run userspace buildpacks. Because otherwise my first thought on reading this is "wait, but the builder also doesn't run as root"...

natalieparellano: > This ensures that all changes from the previous snapshot are preserved even if the stack buildpack does not make any additional changes.
Could you elaborate on this?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+* The environment variables and inputs for `bin/detect` and `bin/build` are the same (though the values may be different)
+
+However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase excluding the a few specific directories and files.
+
+Additionally, a platform may store a new stack image to cache the changes. All of the captured changes will be included in a single layer produced as output from the stackpack. The `/layers` dir MAY NOT be used to create arbitrary layers.
+
+A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory, and providing an `/cnb/stack/buildpacks/order.toml` (following the [`order.toml` schema](https://github.com/buildpacks/spec/blob/main/platform.md#ordertoml-toml)) to define their order of execution. The order can be overridden in the `builder.toml` with the following configuration:
```
natalieparellano, line 71: It looks like in the implementation this is `/cnb/stack/order.toml`

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+
+The stackpack interface is similar to the buildpack interface
+* The same `bin/detect` and `bin/build` scripts are required
+* The `bin/detect` will have read-only access to the app
+* The positional arguments for `bin/detect` and `bin/build` are the same (though the values may be different)
+* The environment variables and inputs for `bin/detect` and `bin/build` are the same (though the values may be different)
+
+However, some of the context it is run in is different from regular buildpacks.
+
+For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase excluding the a few specific directories and files.
```
yaelharel, line 67:
```suggestion
For each stackpack, the lifecycle will use [snapshotting](https://github.com/GoogleContainerTools/kaniko/blob/master/docs/designdoc.md#snapshotting-snapshotting) to capture changes made during the stackpack's build phase excluding a few specific directories and files.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+[[or.requires]]
+name = "jq"
+```
+
+## Future work
+
+In the future, we plan to enhance the stack buildpack interface with the following:
+
+* A `CNB_STACK_TYPE` env var that a stack buildpack can use to behave differently on each part of the stack
+* Support for `creator`. Because of the new extend phase, it is not possible to easily run the entire buildpack process with a single binary.
```
natalieparellano, line 335: Maybe it's worth calling out that `creator` can still run with stack buildpacks if there is no need to extend the run image. I believe this is true looking at https://github.com/buildpacks/lifecycle/pull/410 , but maybe I'm wrong.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,596 @@
...
+
+During the build phase, the lifecycle will extract and apply each snapshot tarball before running stack buildpacks, but after a snapshot-baseline has been captured. This ensures that all changes from the previous snapshot are preserved even if the stack buildpack does not make any additional changes.
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. The image containing the stack buildpacks and the builder binary must be provided to the rebaser operation as an argument. A platform may choose to provide the same stack buildpacks and builder binary used during the build that create the launch-image being rebased, or it may provide updates versions (which may increase the risk of something failing in the rebase process).
```
yaelharel, line 167: Can you please elaborate on what is the "launch image"?

text/0000-stack-buildpacks.md
jkutner: /cc @jabrown85 what did you decide for this this file?

text/0000-stack-buildpacks.md
jkutner: fixed in ca9b034

text/0000-stack-buildpacks.md
jkutner: @jabrown85 is it true that you can use `creator` if you don't extend the run image, or will `creator` just ignore stackpacks?

text/0000-stack-buildpacks.md
jkutner: @jabrown85 where did we land with `detector` having app access?

text/0000-stack-buildpacks.md
jkutner: addressed in c2c946f

text/0000-stack-buildpacks.md
jabrown85: `/cnb/stack/order.toml` is where it is at. I don't mind moving it if we want to. It made sense to me to be one level above the potentially mounted `stack/buildpacks` directory.

text/0000-stack-buildpacks.md
jabrown85: You can use `creator` as long as it doesn't detect you need to extend the run image.

jabrown85: But we don't have to keep that functionality if it makes things more confusing. I did that work prior to realizing `creator` wasn't going to work for extend images.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,597 @@
...
+[summary]: #summary
+
+This is a proposal for a new type of buildpack that runs against a stack in order to extend it in ways that are only possible by running privileged commands.
+
+# Motivation
+[motivation]: #motivation
+
+Normally, buildpacks do not run as root. This is an intentional design decision that ensures operations like `rebase` will work on day-2.
+
+However, many applications and buildpacks require modifications to the stack they run on, such as adding system packages or custom certificates. For this reason, we need a mechanism that buildpack authors and buildpack users and leverage to extend their stacks.
```
yaelharel, line 21:
```suggestion
However, many applications and buildpacks require modifications to the stack they run on, such as adding system packages or custom certificates. For this reason, we need a mechanism that buildpack authors and buildpack users can leverage to extend their stacks.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,299 @@
...
+
+The stackpack's snapshot layer may be modified by writing a `stack.toml` file. The `stack.toml` will define paths that will be restored even when the base image changes (ex. package indicies) and paths that will be excluded from the launch image (ex. `/var/cache`).
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
+
+## Example: Apt Buildpack
```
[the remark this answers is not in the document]
jkutner: fixed in 82eeb73

text/0000-stack-buildpacks.md
jkutner: fixed in 2148aea

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+#### `rebaser`
+Usage:
+```
+/cnb/lifecycle/rebaser \
+  [-build-image <build-image>] \
+  ...
+```
+
+##### Inputs
+| Input               | Environment Variable  | Default Value          | Description
```
ekcasey, line 558: Do we need a cache as an input to rebaser?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+
+A stack can provide stackpacks by including them in the `/cnb/stack/buildpacks` directory, and providing an `/cnb/stack/order.toml` (following the [`order.toml` schema](https://github.com/buildpacks/spec/blob/main/platform.md#ordertoml-toml)) to define their order of execution. The order can be overridden in the `builder.toml` with the following configuration:
+
+```
+[[stack.order]]
+  [[stack.order.group]]
+    id = "<stackpack id>"
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, its detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
```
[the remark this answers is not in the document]
jkutner: yea. i think the name "layers dir" should really be more like "output dir"

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+[[requires]]
+name = "<mixin name>"
+mixin = <boolean (default=false)>
+```
+
+A userspace buildpack MAY NOT provide mixins in the build plan.
+
+### Requiring Mixins
+
+Stack buildpacks MUST NOT require any entries in the build plan (neither mixins nor non-mixins). This ensures that we do not need to re-run the detect phase.
```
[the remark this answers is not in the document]
jkutner: I don't remember, but I think this had to do with something @jabrown85 ran into while prototyping the mixin resolution

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+ A stack buildpack is a special case of buildpack that has the following properties:
+
+* Is run as the `root` user
+* Configured with `privileged = true` in the `buildpack.toml`
+* Can only create one layer
+* May include, in the created layer, modifications to any part of the filesystem, excluding `<layers>`, `<platform>`, and `<cnb>` directories
+* Does not have access to the application source code during the `build` phase
```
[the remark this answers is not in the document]
jkutner: Changed to "Must not access" in dd0c7bd

text/0000-stack-buildpacks.md
jkutner: Fixed in dd0c7bd

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+
+```
+[[stack.order]]
+  [[stack.order.group]]
+    id = "<stackpack id>"
+```
+
+A stackpack will only execute if it passes detection. When the stackpack is executed, its detect and build scripts use the same parameters as the regular buildpacks. But the arguments may differ. For example, the first positional argument to the `bin/build` (the `<layers>` directory) MUST NOT be a standard layers directory adhering to the [Buildpacks Build specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#build).
+
+The stackpack's snapshot layer may be enriched by writing a `launch.toml` file. The `launch.toml` may define globs of files to be excluded from the image when it is _used_. Any excluded path may also be marked as _cached_, so that those excluded paths are recovered when the image is _used_. The term _used_ is defined as:
```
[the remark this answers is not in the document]
jkutner: Fixed in dd0c7bd

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+* Is run against both the build and run images
+* Is distributed with the stack run and/or build image
+* May not create layers using the `<layers>` directory
+
+The stackpack interface is similar to the buildpack interface
+* The same `bin/detect` and `bin/build` scripts are required
+* The `bin/detect` will have read-only access to the app
+* The positional arguments for `bin/detect` and `bin/build` are the same (though the values may be different)
+* The environment variables and inputs for `bin/detect` and `bin/build` are the same (though the values may be different)
+
```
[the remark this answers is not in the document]
jkutner: Addressed in dd0c7bde7d45dd51199a025a74ae99f71d612f3a

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. The image containing the stack buildpacks and the builder binary must be provided to the rebaser operation as an argument. A platform may choose to provide the same stack buildpacks and builder binary used during the build that create the launch-image being rebased, or it may provide updates versions (which may increase the risk of something failing in the rebase process).
+
+A platform may choose to store the stack buildpacks and builder binary in any of the following images:
+* A companion image of the launch-image
+* A builder image that is available when rebasing
+* The launch-image itself (store the stack buildpacks and builder binary alongside the application itself)
+
+Then, the rebase operation can be performed as normal, while including the stackpack layers as part of the stack. This will be made possible by including the stackpack in the run-image, but because the stackpack detect phase is not run, the operation does not need access to the application source.
```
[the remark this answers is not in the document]
jkutner: Good question. IMO it should either be:
* A `build-plan.toml` in the launch-image
* A `io.buildpacks.build.plan` label on the launch-image

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: Addressed in 0ee0f37a5437dadacbff26bf0bb5ddbb35984e94

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+```
+/cnb/lifecycle/exporter \
+  [-stack-group <stack-group>] \
+  ...
+```
+
+##### Inputs
+| Input               | Environment Variable  | Default Value       | Description
+|---------------------|-----------------------|---------------------|---------------------------------------
+| ...                 |                             |                           |
```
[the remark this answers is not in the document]
jkutner: Addressed in ffb4537

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+| `<stack-group>`     | `CNB_STACK_GROUP_PATH`      | `./stack-group.toml`      | Path to stack-group file (see [`group.toml`](#grouptoml-toml))
+
+
+#### `rebaser`
+Usage:
+```
+/cnb/lifecycle/rebaser \
+  [-build-image <build-image>] \
```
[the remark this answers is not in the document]
jkutner: Addressed in 43a7d1d

text/0000-stack-buildpacks.md
jkutner: Addressed in 43a7d1d

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+  ...
+```
+
+##### Inputs
+| Input               | Environment Variable  | Default Value          | Description
+|---------------------|-----------------------|------------------------|---------------------------------------
+| ...                 |                       |                        |
```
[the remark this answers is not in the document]
jkutner: Addressed in 43a7d1d

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+| `<stack-group>`     | `CNB_BUILD_IMAGE`     | derived from `<image>` | Image containing `builder` and stack buildpacks
```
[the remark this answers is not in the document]
jkutner: Addressed in 43a7d1d

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jkutner: Addressed in b33c1a9 and 43a7d1d

text/0000-stack-buildpacks.md
jkutner: i tried reworking this, but it was hard to avoid _some_ repetition between the definition and the problem statement. I'm gonna keep as-is, but feel free make an edit suggestion.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+For example, a stack buildpack may choose to exclude `/var/cache` from the final run image, but may want to mark it as _cached_ to have it restored during build-time and rebase.
+
+## Mixins
+
+### Providing Mixins
+
+A stack buildpack MAY define a set of mixins it provides in either of two ways:
+
+1. Statically in the `buildpack.toml`
+1. Dynamically in the Build Plan.
```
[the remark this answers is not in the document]
jkutner: Addressed in https://github.com/buildpacks/rfcs/pull/118

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+
+### Requiring Mixins
+
+Stack buildpacks MUST NOT require any entries in the build plan (neither mixins nor non-mixins). This ensures that we do not need to re-run the detect phase.
+
+### Resolving Mixins
+
+After the detect phase, the lifecycle will merge the list of provided mixins from the following sources:
+* `stack.toml`
+* `buildpack.toml` of any stack buildpacks
```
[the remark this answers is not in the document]
jkutner: Addressed in https://github.com/buildpacks/rfcs/pull/118

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,620 @@
...
+* `/layers`
+* `/workspace`
+* `/dev`
+* `/sys`
+* `/proc`
+* `/var/run/secrets`
+* `/etc/hostname`, `/etc/hosts`, `/etc/mtab`, `/etc/resolv.conf`
+* `/.dockerenv`
+
+## launch.toml (TOML)
```
sclevine, line 404: Suggestion after chatting with @ekcasey:

Could this be `build.toml` in the context of extending the build image, and `launch.toml` in the context of extending the run image?

text/0000-stack-buildpacks.md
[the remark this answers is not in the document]
jabrown85: I don't think we'd need to run detect twice, but we can't allow stackpacks to require `yj` from normal buildpacks. Otherwise you can't run `extend` and `build` in parallel and it means you have to run the stackpack after the buildpack that provides `yj`

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,606 @@
...
+
+During the build phase, the lifecycle will extract and apply each snapshot tarball before running stack buildpacks, but after a snapshot-baseline has been captured. This ensures that all changes from the previous snapshot are preserved even if the stack buildpack does not make any additional changes.
+
+## Rebasing an App
+
+Before a launch image is rebased, the platform must re-run the any stackpacks that were used to build the launch image against the new run-image. The image containing the stack buildpacks and the builder binary must be provided to the rebaser operation as an argument. A platform may choose to provide the same stack buildpacks and builder binary used during the build that create the launch-image being rebased, or it may provide updates versions (which may increase the risk of something failing in the rebase process).
+
+A platform may choose to store the stack buildpacks and builder binary in any of the following images:
+* A companion image of the launch-image
+* A builder image that is available when rebasing
```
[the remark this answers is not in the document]
jabrown85: We talked about this once before. IIRC the problem we are trying to solve is that a platform may wish to provide the _exact_ same stackpacks as the user used on first build for _all_ future rebases. If you always provide stackpacks in the new run image, you are upgrading the stackpacks with each rebase. This could be desirable for a tightly controlled platform, but it could be undesirable if your goal is to ensure the least amount of breakage without the user having to rebuild. The platform may wish to change the behavior of a stackpack to require new metadata from the requiring buildpacks, but this would not be a safe operation if the buildpacks are always updated. By allowing the stackpacks to be stored in a companion image, you could ensure the results of a rebase are consistent.

text/0000-stack-buildpacks.md
jkutner: yes

hone (approved):

text/0000-stack-buildpacks.md
jkutner: we discussed naming this `layer.toml`, but I'm going propose `stack-layer.toml` because it is less ambiguous.

jkutner:
@sclevine @ekcasey Question: Do we want to change the Buildpack Plan `[[entries]]` table to contain `mixin = "mixin name"` to match the Buildpack `[[requires]]` table?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+* `/etc/hostname`, `/etc/hosts`, `/etc/mtab`, `/etc/resolv.conf`
+* `/.dockerenv`
+
+## stack-layer.toml (TOML)
+
+```
+[[excludes]]
+paths = ["<sub-path glob>"]
+cache = false
+restore = false
```
ekcasey, line 392: Why did we introduce `restore`? Why would I ever set `cache = true` and `restore = false`?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+```
+
+Where:
+
+* `paths` = a list of paths to exclude from the layer
+* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in the cache layer.
```
ekcasey, line 398: > excluded from the launch image layer
Only makes sense for layers created during the `extend` phase. For layers created during the `build` phase we should remove these paths from the filesystem before executing any regular buildpacks.

ekcasey, line 398:
```suggestion
* `cache` = if true, the paths will be excluded from the launch image layer, but will be included in a cached layer.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+
+[stacks.provides]
+mixins = [ "mixin name or *" ]
+ ```
+
+* `privileged` - when set to `true`, the lifecycle will run this buildpack as the `root` user.
+
+Under the `[stacks.provides]` table:
+
+* `any` - a boolean that, when true, indicates that the buildpack can provide all mixins
```
ekcasey, line 427:
```suggestion
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+
+##### Inputs
+| Input                 | Environment Variable        | Default Value           | Description
+|-----------------------|-----------------------------|-------------------------|----------------------
+| ...                   |                             |                         |
+| `<stack-buildpacks>`  | `CNB_STACK_BUILDPACKS_DIR`  | `/cnb/stack/buildpacks` | Path to stack buildpacks directory (see [Buildpacks Directory Layout]
+| `<stack-group>`       | `CNB_STACK_GROUP_PATH`      | `./stack-group.toml`    | Path to output group definition(#buildpacks-directory-layout))
+
+##### Outputs
+| Output                                     | Description
```
ekcasey, line 517: We should add `<layers>/<buildpack ID>/stack-layer.toml` to the output list.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+|--------------------------------------------|----------------------------------------------
+| ...                                        |
+| `<layers>/<buildpack ID>.tgz`              | Layer snapshot (see [Buildpack Interface Specfication](buildpack.md)
```
ekcasey, line 520: Do we want this to be `<layers>/<buildpack ID>/stack-layer.tgz` instead?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+| `<stack-buildpacks>`  | `CNB_STACK_BUILDPACKS_DIR`  | `/cnb/stack/buildpacks` | Path to stack buildpacks directory (see [Buildpacks Directory Layout]
+| `<stack-group>`       | `CNB_STACK_GROUP_PATH`      | `./stack-group.toml`    | Path to output group definition(#buildpacks-directory-layout))
+
+##### Outputs
+| Output                                     | Description
+|--------------------------------------------|----------------------------------------------
+| [exit status]                              | (see Exit Code table below for values)
+| `/dev/stdout`                              | Logs (info)
+| `/dev/stderr`                              | Logs (warnings, errors)
+| `<layers>/<buildpack ID>.tgz`              | Layer snapshot (see [Buildpack Interface Specfication](buildpack.md)
```
ekcasey, line 585: Same comments as above:

* Do we want this to be `<layers>/<buildpack ID>/stack-layer.tgz` instead?
* We should add `<layers>/<buildpack ID>/stack-layer.toml` to the output list.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+| `<layers>/config/metadata.toml`            | Build metadata (see [`metadata.toml`](#metadatatoml-toml))
```
ekcasey, line 586: Are any of the fields in `metadata.toml` actually valid outputs for a stack buildpack?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,599 @@
...
+| `0`       | Success
+| `11`      | Platform API incompatibility error
+| `12`      | Buildpack API incompatibility error
+| `1-10`, `13-99` | Generic lifecycle errors
+| `401`     | Buildpack build error
+| `400`, `402-499`|  Build-specific lifecycle errors
+
+- The lifecycle SHALL execute all stack buildpacks in the order defined in `<stack-group>` according to the process outlined in the [Buildpack Interface Specification](buildpack.md).
+- The lifecycle SHALL add all invoked stack buildpacks to`<layers>/config/metadata.toml`.
+- The lifecycle SHALL aggregate all `processes` and BOM entries returned by buildpacks in `<layers>/config/metadata.toml`.
```
ekcasey, line 599: Can stack buildpacks define a process?

jkutner: not in this proposal

text/0000-stack-buildpacks.md
jkutner: because I thought that's what @sclevine wanted.

text/0000-stack-buildpacks.md
jkutner: Fixed in eb9cb5d

jkutner: Oh, but I think it should be `<layers>/stack-layer.toml` as `<buildpack ID>` is included in `<layers>`

text/0000-stack-buildpacks.md
jkutner: `stack-layer.tgz` seems fine. it's created by the lifecycle not the buildpack itself, so it doesn't really matter to the public interface.  I do think it should be `<layers>/stack-layer.tgz` though.

text/0000-stack-buildpacks.md
jkutner: The `[[buildpacks]]` and `[bom]`? I could see supporting `[[processes]]` in the future. That said, I don't mind removing this as an output.

text/0000-stack-buildpacks.md
jkutner: addressed in b9cf2b1

text/0000-stack-buildpacks.md
jkutner: removed `restore` in b9cf2b1

jabrown85 (approved):

sclevine (approved):

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+The image containing the stack buildpacks and the lifecycle binary must be provided to the rebaser operation as an argument. A platform may choose to provide the same stack buildpacks and lifecycle binary used during the build that create the app image being rebased, or it may provide updated versions (which may increase the risk of something failing in the rebase process).
+
+A platform may choose to store the stack buildpacks and extender binary in any of the following images:
+* A companion image of the app image
+* A builder image that is available when rebasing
+* The app image itself (store the stack buildpacks and builder binary alongside the application itself)
```
ekcasey, line 223: Maybe we can just omit the options here and leave it as an exercise for the platform. We can describe the requirements (extend must run on the new run image with the stackpacks etc. etc.) without describing how the platform should do this (until we have more/better guidance). I worry that some options here are confusing or imply lifecycle requirements that we don't actually want to meet (e.g putting the stackpacks into the exported image).

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+```toml
+[build]
+mixins = [ "libpq" ]
+```
+
+This would instruct the Apt Buildpack to install the libpq package.
+
+## Example: CA Certificate Buildpack
+
+Support for custom CA Certificates can be accomplished with two buildpacks: a stackpack that can install the cert, and a normal buildpack that can provide a cert in the build plan.
```
ekcasey, line 288:
```suggestion
Support for custom CA Certificates can be accomplished with two buildpacks: a stackpack that can install the cert, and a normal buildpack that can require a cert in the build plan.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+for file in $(cat $3 | yj -t | jq -r ".entries | .[] | select(.name==\"cacert\") | .metadata | .file"); do
+  cert="$(cat $3 | yj -t | jq -r ".entries | .[] | select(.name==\"cacert\") | .metadata | select(.file==\"$file\") | .content")"
+  echo $cert > /usr/share/ca-certificates/$file
+  chmod 644 /usr/share/ca-certificates/$file
+done
+
+update-ca-certificates
+```
+
+### CA Cert Provider Buildpack
```
ekcasey, line 331:
```suggestion
### CA Cert Requirer Buildpack
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+The stackpack must be used with a buildpack that provides a certificate(s) for it to install. That buildpack would have the following `buildpack.toml`:
```
ekcasey, line 333:
```suggestion
The stackpack must be used with a buildpack that requires a certificate(s) for it to install. That buildpack would have the following `buildpack.toml`:
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+[[or.requires]]
+name = "jq"
+```
+
+## Future work
+
+In the future, we plan to enhance the stack buildpack interface with the following:
+
+* A `CNB_STACK_TYPE` env var that a stack buildpack can use to behave differently on each part of the stack
```
ekcasey, line 395:
```suggestion
```
I would just take this out because I am not convinced we will want or need this eventually, but it's in future work so I don't care too much.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+[alternatives]: #alternatives
+
+- [Add root buildpack as interface for app image extension RFC](https://github.com/buildpacks/rfcs/pull/77)
+- [App Image Extensions (OS packages)](https://github.com/buildpacks/rfcs/pull/23)
+- [Root Buildpacks](https://github.com/buildpacks/rfcs/blob/root-buildpacks/text/0000-root-buildpacks.md): Allow application developers to use root buildpacks in `project.toml`. This would have significant performance implications, and creates a "foot gun" in which end users could build images they are not able to maintain. For this reason, we are holding off on a generic root buildpack feature.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- The term "non-idempotent" is used in section 9.1.2 of [Hypertext Transfer Protocol -- HTTP/1.1 (RFC 2616)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)
```
ekcasey, line 414:
```suggestion
```

I think we took out all references to `non-idempotent`.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- what about the bill of materials?
+    - extend phase artifacts go in the BOM
+    - build phase go in report.toml
+- how do we prevent problems on rebase if the mixins provided by the stack change?
+- how does detect know the run-image mixins?
+- If a mixin is required for build, but is not needed for extend, does the stackpack execute in both phases (and visa versa)?
```
ekcasey, line 424:
```suggestion
- If a mixin is required for build, but is not needed for extend, does the stackpack execute in both phases (and visa versa)?
- How is the buildpack plan provided to a stack buildpack during the extend phase of a rebase operation?
- Will `io.buildpacks.stack.mixins` on the app image be updated to include stack buildpack-provided mixins?
- During a rebase, if the new run image statically includes a mixin that was previously provided by a stack buildpack, should the providing stack buildpack rerun?
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+| `<stack-buildpacks>`  | `CNB_STACK_BUILDPACKS_DIR`  | `/cnb/stack/buildpacks` | Path to stack buildpacks directory (see [Buildpacks Directory Layout]
+| `<stack-group>`       | `CNB_STACK_GROUP_PATH`      | `./stack-group.toml`    | Path to output group definition(#buildpacks-directory-layout))
+| `<stack-platform>`   | `CNB_STACK_PLATFORM_DIR`    | `/stack-platform`       | Path to stack-specific platform directory
+
+##### Outputs
+| Output                                     | Description
+|--------------------------------------------|----------------------------------------------
+| ...                                        |
+| `<layers>/<buildpack ID>.tgz`              | Layer snapshot (see [Buildpack Interface Specfication](buildpack.md)
+| `<layers>/stack-layer.toml`                | Layer snaphot metadata
```
ekcasey, line 587:
```suggestion
| `<layers>/<buildpack-id>/stack-layer.toml`                | Layer snaphot metadata
```
These need to be scoped to the buildpack-specific layer dir, or else multiple stack buildpacks would be writing to the same path.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+| ...                   |                             |                         |
+| `<stack-buildpacks>`  | `CNB_STACK_BUILDPACKS_DIR`  | `/cnb/stack/buildpacks` | Path to stack buildpacks directory (see [Buildpacks Directory Layout]
+| `<stack-group>`       | `CNB_STACK_GROUP_PATH`      | `./stack-group.toml`    | Path to output group definition(#buildpacks-directory-layout))
+| `<stack-platform>`   | `CNB_STACK_PLATFORM_DIR`    | `/stack-platform`       | Path to stack-specific platform directory
+
+##### Outputs
+| Output                                     | Description
+|--------------------------------------------|----------------------------------------------
+| ...                                        |
+| `<layers>/<buildpack ID>.tgz`              | Layer snapshot (see [Buildpack Interface Specfication](buildpack.md)
```
ekcasey, line 586:
```suggestion
```
We don't actually need to take snapshots during build.

ekcasey, line 586: We need an output for tarballs from excludes, so they can be cached during export.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+##### Outputs
+| Output                                     | Description
+|--------------------------------------------|----------------------------------------------
+| [exit status]                              | (see Exit Code table below for values)
+| `/dev/stdout`                              | Logs (info)
+| `/dev/stderr`                              | Logs (warnings, errors)
+| `<layers>/<buildpack ID>.tgz`              | Layer snapshot (see [Buildpack Interface Specfication](buildpack.md)
+| `<layers>/stack-layer.toml`                | Layer snaphot metadata
+| `<layers>/config/metadata.toml`            | Build metadata (see [`metadata.toml`](#metadatatoml-toml))
```
ekcasey, line 655: Same comment as above. Where do the files from `[[excludes]]` end up?

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+| Exit Code | Result|
+|-----------|-------|
+| `0`       | Success
+| `11`      | Platform API incompatibility error
+| `12`      | Buildpack API incompatibility error
+| `1-10`, `13-99` | Generic lifecycle errors
+| `401`     | Buildpack build error
+| `400`, `402-499`|  Build-specific lifecycle errors
```
ekcasey, line 664: All `builder` phase errors are in the `400` range. I think we should pick a new range for `extender` errors.

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,668 @@
...
+
+- The lifecycle SHALL execute all stack buildpacks in the order defined in `<stack-group>` according to the process outlined in the [Buildpack Interface Specification](buildpack.md).
+- The lifecycle SHALL add all invoked stack buildpacks to`<layers>/config/metadata.toml`.
+- The lifecycle SHALL aggregate all `processes` and BOM entries returned by buildpacks in `<layers>/config/metadata.toml`.
```
ekcasey, line 668:
```suggestion
- The lifecycle SHALL aggregate all BOM entries returned by buildpacks in `<layers>/config/metadata.toml`.
```

ekcasey (approved):

text/0000-stack-buildpacks.md
jkutner: I'd like to leave it, because it's come up as a question a few times. So I'd like to at least call out that we _aren't_ doing it in this iteration (if at all)

jkutner:
Final Comment Period with merge disposition, closing on 24 December, 2020.

text/0000-stack-buildpacks.md
jkutner: Fixed in 848f348

text/0000-stack-buildpacks.md
jkutner: Fixed in 146d0d5

text/0000-stack-buildpacks.md
jkutner: Fixed in 146d0d5

text/0000-stack-buildpacks.md
jkutner: Fixed in a8ecc01

text/0000-stack-buildpacks.md
jkutner: Addressed in 439b4ec

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,671 @@
...
+```
+[[entries]]
+mixin = "libpq"
+```
+
+A stack buildpack that needs to install mixins must select them from the build plan.
+
+### Example: Any Mixin Buildpack
+
+As an example of these resolution rules, consider an Apt stackpack that provide any (`*`) mixin. During the detect phase, the stackpack will pass in the following cases:
```
edmorley, line 155:
```suggestion
As an example of these resolution rules, consider an Apt stackpack that provides any (`*`) mixin. During the detect phase, the stackpack will pass in the following cases:
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,671 @@
...
+* a mixin is required by a buildpack
+* a mixin is [required by an app](https://github.com/buildpacks/rfcs/pull/112)
+
+The example stackpack will be skipped after detection in the following cases:
+* no mixins are required by any component of the build
+* A mixin is required by a buildpack, but the stack already provides that mixin.
+* A mixin is required by an app, but the stack already provides that mixin.
```
edmorley, line 162:
```suggestion
* no mixins are required by any component of the build
* a mixin is required by a buildpack, but the stack already provides that mixin
* a mixin is required by an app, but the stack already provides that mixin
```
(for consistency with the other bulleted lists above)

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+If only the `chromedriver` is required, or the `google-chrome-stable` is already provided by the stack, the build phase would receive:
+
+```toml
+[[entries]]
+name = "chromedriver"
+```
+
+If the stack does not already provide `google-chrome-stable`, the stackpack may install it anyways. This prevents the case where an app may not work at runtime if that app requires only the `chromedriver` and the `google-chrome-stable` is not provided by the stack.
```
edmorley, line 203: I'm not sure I quite follow this paragraph. The intro above says:

> If only the `chromedriver` is required, or the `google-chrome-stable` is already provided by the stack,

...but I'm not clear which scenario of the two it's referring to.

Does it mean something like this?

```suggestion
If the stack does not already provide `google-chrome-stable`, the stackpack may install it anyways, even if the app didn't require it. This prevents the case where an app may not work at runtime if that app requires only the `chromedriver` and the `google-chrome-stable` is not provided by the stack.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,671 @@
...
+
+## Caching and Restoring
+
+During the export phase, the lifecycle will store any excluded cached stackpack layers created during the build or extend phase in the cache.
+
+During the restore phase of the next build, the lifecycle will download the excluded cached stackpack layers with the cache and store them as tarballs in the `<layers>` directory and the `<run-layers>` directory (i.e. they will not be extracted). The `restorer` cannot extract the these layers because it will not run with root privileges. In addition, the `restorer` may run in a different container than the build, which means changes made to the base image are not guaranteed to carry forward. The `builder` will run with root privileges and untar these layers and cache, but will drop privileges after executing stackpacks and before running application buildpacks. The `extender` will also run with root privileges and untar the cache.
```
edmorley, line 209:
```suggestion
During the restore phase of the next build, the lifecycle will download the excluded cached stackpack layers with the cache and store them as tarballs in the `<layers>` directory and the `<run-layers>` directory (i.e. they will not be extracted). The `restorer` cannot extract these layers because it will not run with root privileges. In addition, the `restorer` may run in a different container than the build, which means changes made to the base image are not guaranteed to carry forward. The `builder` will run with root privileges and untar these layers and cache, but will drop privileges after executing stackpacks and before running application buildpacks. The `extender` will also run with root privileges and untar the cache.
```

text/0000-stack-buildpacks.md
```diff
@@ -0,0 +1,671 @@
...
+
+During the build phase, after the stackpacks have run but before the application buildpacks have run, the lifecycle will forcefully delete everything in `[[excludes]]` after ALL stackpacks run, before buildpacks run (which implies that it does not need to do this for extend phase).
```
edmorley, line 211: I find this sentence hard to follow - it seems to be repeating itself, or else I'm misunderstanding the content?

jkutner:
```suggestion
During the build phase, after ALL stackpacks have run but before the application buildpacks have run, the lifecycle will forcefully delete everything in `[[excludes]]` (which implies that it does not need to do this for extend phase).
```

PR #113, opened 2020-09-03
Distribution Team

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/distribution-team/text/0000-distribution-team.md)

### DIFF
--- a/text/0000-distribution-team.md
+++ b/text/0000-distribution-team.md
@@ -27,7 +27,7 @@ We have two repositories that are not owned by a specific sub-team.
 # What it is
 [what-it-is]: #what-it-is
 
-The Distribution team is responsible for maintaining tools and services that support the distribution and discovery of buildpacks. It own the following repos:
+The Distribution team is responsible for maintaining tools and services that support the distribution and discovery of buildpacks. It owns the following repos:
 
 * [Buildpack Registry Index](https://github.com/buildpacks/registry-index)
 * [Buildpack Registry Namespace Owners](https://github.com/buildpacks/registry-namespaces)
--- a/text/0000-distribution-team.md
+++ b/text/0000-distribution-team.md
@@ -1,0 +1,71 @@
+# Meta
+[meta]: #meta
+- Name: Create a Distribution Team
+- Start Date: 2020-09-03
+- Author(s): [@jkutner](@jkutner)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal to create a Distribution sub-team in the project governance.
+
+# Definitions
+[definitions]: #definitions
+
+- *sub-team* - Sub-teams are responsible for narrower sets of concerns related to specific aspects of the project. Each sub-team will include at least one core team member to help align with the broader roadmap.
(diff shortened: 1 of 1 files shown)


text/0000-distribution-team.md
```diff
@@ -0,0 +1,71 @@
...
+[motivation]: #motivation
+
+We have two repositories that are not owned by a specific sub-team.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Create a sub-team in the [project governance]() called the "Distribution Team", which will own the following repos:
+
+* [CircleCI Pack Orb](https://github.com/buildpacks/pack-orb)
```
hone, line 32: Why the CircleCI orb and not the Tekton examples?

text/0000-distribution-team.md
```diff
@@ -0,0 +1,70 @@
...
+- *distribution* - relates to the delivery, discovery, and integration of buildpacks.
+
+# Motivation
+[motivation]: #motivation
+
+We have two repositories that are not owned by a specific sub-team.
+
+# What it is
+[what-it-is]: #what-it-is
+
```
hone, line 29: Can you also add the description/charter from the [community PR](https://github.com/buildpacks/community/pull/34/files):

> The Distribution team is responsible for maintaining tools and services that support the distribution, discovery, and integration of buildpacks.

text/0000-distribution-team.md
```diff
@@ -0,0 +1,70 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [Platform team](https://github.com/buildpacks/community/blob/main/TEAMS.md#Platform-Team)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Who are the maintainers?
```
hone, line 69: Travis + Joe?

dfreilich (commented):
In general, the idea is great. I'm not sure I love the name though - distribution sounds like it is in charge of distributing the Buildpacks products (either the lifecycle, or pack), but that seems to be more the responsibility of the platform team (as it focuses on controlling platforms on which you can run the lifecycle), and doesn't seem to accurately fit the core responsibility of the new team, which will focus more on the registration-yanking paths, and not really distributing buildpacks or the lifecycle.

jromero:
> In general, the idea is great. I'm not sure I love the name though - distribution sounds like it is in charge of distributing the Buildpacks products (either the lifecycle, or pack), but that seems to be more the responsibility of the platform team (as it focuses on controlling platforms on which you can run the lifecycle), and doesn't seem to accurately fit the core responsibility of the new team, which will focus more on the registration-yanking paths, and not really distributing buildpacks or the lifecycle.

The term distribution, like almost every other term, can have different meaning depending on context. In this context, we are talking about the project and thereby the spec (not the products). I think of it as similar terms to "OCI distribution spec". It centers around how buildpacks should be distributed. The term you seem to be describing I think of as delivery.

All that said, if there are alternative names that would further clear this potential confusion I think it's worth discussing.

hone (changes requested):
I'd like to see a description of the team.

sclevine (approved):

nebhale (approved):

jkutner:
@hone description added in 58b8f72

hone (approved):

ekcasey:
Final Comment Period with merge disposition, closing on 30 September, 2020.

PR #114, opened 2020-09-08
RFC: Build/Run image configs MUST contain env.PATH

micahyoung:
[Readable](https://github.com/micahyoung/buildpacks-rfcs/blob/rfcs-stack-images-config-env-path-must-exist/text/0000-stack-images-config-env-path-must-exist.md)

### DIFF
--- a/text/0000-stack-images-config-env-path-must-exist.md
+++ b/text/0000-stack-images-config-env-path-must-exist.md
@@ -1,0 +1,171 @@
+# Meta
+[meta]: #meta
+- Name: Build/Run image configs MUST contain env.PATH
+- Start Date: 2020-08-21
+- Author(s): @aemengo @ameyer-pivotal @mvalliath @TisVictress @micahyoung
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Windows build & run stack images need to set a `PATH` environment variable in the image config to avoid Docker's alternative mechanism for env var `PATH`-loading from the Windows registry, which typically contains the value of `PATH` for Windows images. Relying on the Windows registry creates inconsistency between Windows and Linux CNB mechanisms which causes Windows app images to be unrunnable in some circumstances [lifecycle#384](https://github.com/buildpacks/lifecycle/issues/384).
+
+I propose we change the spec to codify the current de-facto Linux image behavior -- to require a complete `config.Env.PATH` to be set on all build & run stack images, for Windows or Linux. When a `config.Env.PATH` is set on a Windows image, they will always behave like Linux images and avoid triggering Docker's Windows registry-based loading of the `PATH` environment variable. Linux stack images get additional validation but generally will never need to change because they already have a `PATH` set from the base image.
+
+See [spec-changes](#spec-changes-optional) for details.
+
+# Definitions
+[definitions]: #definitions
+* [Windows registry](https://en.wikipedia.org/wiki/Windows_Registry): Very common key-value store for Windows OS, used to store environment variable values among other things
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

nebhale (approved):

jkutner (approved):

dfreilich (approved):
Would it make sense for `pack` to set a default path, if one is not provided?

micahyoung:
@dfreilich I'm not sure if there's a great default, as even Microsoft's base images (which most builders would be based on) don't appear to have stable paths:

```
docker run mcr.microsoft.com/windows/nanoserver:1809 cmd /c echo %PATH%
C:\Windows\system32;C:\Windows;

docker run mcr.microsoft.com/windows/servercore:1809 cmd /c echo %PATH%
C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Users\ContainerAdministrator\AppData\Local\Microsoft\WindowsApps

docker run mcr.microsoft.com/windows:1809 cmd /c echo %PATH%
C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Users\ContainerAdministrator\AppData\Local\Microsoft\WindowsApps
```

Technically `c:\Windows;c:\Windows\System32` is a subset of all of them but would likely lead to surprising behavior for the later two images. 

The problem with `pack` being able to get that value itself, as mentioned in the RFC, is that it's only available at runtime.

But what if `pack` were to fail with an error message recommending a `docker run <image name> cmd /c echo %PATH%` command that can be used to capture the runtime variable? Or perhaps `pack` could run that relatively expensive command itself, to make a better failure message?

dfreilich:
Interesting. I wonder whether it may make sense to set for linux, at least, to lower the breaking changes necessary. 

Running `docker run...` is an interesting suggestion, but that wouldn't work for `--publish` situation, where we don't assume the presence of a docker daemon. We can definitely suggest that users run that if we're not able to make it work out.

micahyoung:
> Interesting. I wonder whether it may make sense to set for linux, at least, to lower the breaking changes necessary.

My current understand is that all common Linux base images will have it set (docker even adds a full path for `FROM scratch` Dockerfile-generated ones). But for ggcr `empty.Image`s, I'm not sure what a valid `PATH` should be in that situation.

Since it's potentially unclear what a default should be in Windows, and the rare Linux, cases where it's missing, I feel like it might be best to leave both implementations identical -- fail if the incoming run/build image is missing a `PATH`.

> Running `docker run...` is an interesting suggestion, but that wouldn't work for `--publish` situation, where we don't assume the presence of a docker daemon. We can definitely suggest that users run that if we're not able to make it work out.

Good call, yeah, maybe messaging describing a literal `docker run ... echo %PATH%` command the least worst option.

ekcasey (approved):

ekcasey:
Final Comment Period with merge disposition, closing on 30 September, 2020.

text/0000-stack-images-config-env-path-must-exist.md
```diff
@@ -0,0 +1,172 @@
...
+* Initial PR: https://github.com/moby/moby/pull/29048 
+* End of discussion: https://github.com/moby/buildkit/issues/74
+
+Another abandoned Dockerfile `ENV` approach: [`ENV --lazy-expand`](https://github.com/moby/moby/pull/31525)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+  - Should the spec require `PATH` to contain at least the shell?
```
hone, line 163: Maybe, but seems out of scope of this RFC.

hone (approved):

PR #115, opened 2020-09-10
RFC: Prelease Versions and Experimental Features for API

hone:


jromero:
Two general thoughts:

1. It would seem that these experimental versions should also be included in "supported" given prior notion that "supported" is all encompassing including deprecated, and in this case experimental.
2. I find it difficult to reason about how versions and the experimental features will be carried forward in practicality. For example, if there is an experimental feature "A" that is encompassed in platform api 0.6-beta and then 0.7 is slated to come out with feature "A" still experimental does that inherently force 0.7 to also be labeled experimental as a whole even if other features are "stable"? I prefer the alternative "Unstable Features" in order to solve this unless we can detail more explicitly how this could work.

hone:
> Two general thoughts:
> 
>     1. It would seem that these experimental versions should also be included in "supported" given prior notion that "supported" is all encompassing including deprecated, and in this case experimental.
One of the big differences is the guarantee and support we provide around experimental vs. both "supported" and "deprecated". If we want to include experimental in the supported list, I'm ok with that. We could even just drop the experimental moniker and have the version notation mark experimental.
> 
>     2. I find it difficult to reason about how versions and the experimental features will be carried forward in practicality. For example, if there is an experimental feature "A" that is encompassed in platform api 0.6-beta and then 0.7 is slated to come out with feature "A" still experimental does that inherently force 0.7 to also be labeled experimental as a whole even if other features are "stable"? Would we end up with to simultaneous yet difference spec versions `0.7` and `0.7-experimental`? I prefer the alternative "Unstable Features" in order to solve this unless we can detail more explicitly how this could work.
I imagine if a feature isn't stabilized in the API it wouldn't make it into `0.7` or `0.6`. What you're suggesting aligns much more with the "Unstable Features" alternative like you suggest. It separates features more from API versions themselves.

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+    "experimental": ["1.0-alpha1"],
+  }
+}
+```
+
+## API Types
+### Experimental APIs
+API versions aren't limited to those defined in a spec release, but can refer to upcoming spec work happening in a branch.
+
+New `CNB_PLATFORM_EXPERIMENTAL_MODE` and `CNB_BUILDPACK_EXPERIMENTAL` mode environment variables will control experimental mode with:
```
dfreilich, line 69:
```suggestion
New `CNB_PLATFORM_EXPERIMENTAL_MODE` and `CNB_BUILDPACK_EXPERIMENTAL_MODE` environment variables will control experimental mode with:
```

dfreilich (commented):
To clarify - this is different from rc releases of the lifecycle, and is rather rc releases of the APIs?

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+If we choose to do nothing, we run the risk of releasing less tested API/spec releases.
+
+### Not a Different Version Format
+The Experimental API version format could instead follow the same format as deprecated and supported: `<major>-<minor>`. By doing this, users could automatically get upgraded from an experimental to supported/deprecated API version which may be surprising to them.
+
+### Unstable Features
```
ekcasey, line 103: While there are things I like about this alternative I worry about combinatorial explosion with the API version. Can every unstable feature be used with every support API, are the interfaces the same in each case? etc. etc.

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+# Definitions
+[definitions]: #definitions
+
+experimental - a new API mode that is marked as experimental. Unlike supported and deprecated APIs, this may be for an unreleased API. The purpose is to gather feedback that will be used to improve the design before it's finalized. It will not carry the same promise of stability or compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+In order to have a smooth transition to 1.0 of the spec, there needs to be a way for the project to test out these ideas in the wild and gather feedback. Even beyond 1.0, this desire still stands for testing out concepts in an organized fashion where the feedback can be provided. This feedback will help us improve the design before we finalize it. The quality of feedback we get from users about their real life usage of experimental APIs will help us avoid problems we may miss when approving RFCs.
+/AP
```
jkutner, line 25:
```suggestion

```

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+```
+`experimental`:
+contains an array of experimental API versions
+the APIs will correspond to specs that may not be released and in branches that are still being worked on.
+the API version must be of the format: `<major>.<minor>-<alphanumeric>`, i.e. `1.0-alpha1`.
+
+Since experimental versions are defined with a different format, they will NOT automatically get upgraded to a `supported` release when the experimental API has finalized.
+
+## Lifecycle Labels
+This will extend the builder or lifecycle image with the above descriptor file will extend the label `io.buildpacks.lifecycle.apis` with the value:
```
jkutner, line 53:
```suggestion
This will extend the the `io.buildpacks.lifecycle.apis` label on the builder or lifecycle image with the values contained in the above descriptor file. For example:
```

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+This will add more complexity/overhead that most users will not use.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+If we choose to do nothing, we run the risk of releasing less tested API/spec releases.
+
+### Not a Different Version Format
+The Experimental API version format could instead follow the same format as deprecated and supported: `<major>-<minor>`. By doing this, users could automatically get upgraded from an experimental to supported/deprecated API version which may be surprising to them.
```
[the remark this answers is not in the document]
jkutner: @ekcasey is your question about this alternative, or did the alternative prompt you to ask this about the main proposal.

Where would we be removing the experimental version from?

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+## Rust Unstable Features
+Rust has the concept of [Unstable Features](https://doc.rust-lang.org/beta/unstable-book/the-unstable-book.html) where they have to be explicitly enabled in code. They're all documented in one place.
+
+## Heroku API Stability Attribute
+The Heroku API marks the status of every API endpoint with the [`stability' attribute](https://devcenter.heroku.com/articles/platform-api-reference#stability).
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+How does an experimental version map to the branch?
+Is there a better word than "experimental"?
```
jkutner, line 118: it's common to use `X` in place of "experimental if we want it to be shorter. But I don't think `CNB_PLATFORM_X_MODE` would be obvious to most people.

Also could use "preview"

jkutner (approved):

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,122 @@
...
+[apis.platform]
+  experimental = ["1.0-alpha1"]
+
+[lifecycle]
+  version = "0.10.0"
+```
+`experimental`:
+contains an array of experimental API versions
+the APIs will correspond to specs that may not be released and in branches that are still being worked on.
+the API version must be of the format: `<major>.<minor>-<alphanumeric>`, i.e. `1.0-alpha1`.
```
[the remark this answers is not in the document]
jkutner: Are the experimental versions meant to be "linear" (like, only one line of experiments per version). Or is it more a forking thing. That is, could we have both `0.1-stackpacks` and `0.1-newstuff` co-exist with neither experimental version containing the features of the other. And then merge them both into `0.1`?

text/0000-lifecycle-experimental-api-mode.md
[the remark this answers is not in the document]
ekcasey: > @ekcasey is your question about this alternative, or did the alternative prompt you to ask this about the main proposal?

The latter.

> Where would we be removing the experimental version from?

The `experimental` section of the next lifecycle release. I assume we wouldn't want to support every experimental API forever especially if we finalize and release the feature in a supported API?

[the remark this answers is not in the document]
jkutner: So you're saying when `lifecycle` releases 1.0, we would remove Platform/Buildpack API support for 1.0-alpha1?

I guess I was thinking it would be more like when `lifecycle` releases 1.0, we would remove Platform/Buildpack API support for 0.9-alpha1. That is:
* for any `<major>.<minor>` we would also support `<major>.<minor>-<experimental>`. 
* when we drop a given `<major>.<minor>` we would also drop the corresponding `<major>.<minor>-<experimental>`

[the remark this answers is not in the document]
ekcasey: I agree with your description of what would happen, but we still remove the experimental version eventually right? Do we deprecate the experimental version before we remove it?

I guess this drawback confuses me but it implies some things I am necessarily comfortable with.

In the original multi-api RFC the idea of "experimental support" was that the definition of a experimental API was subject to change. I think this is now implying that the definition of `1.0-alpha1` is never allow to change, instead we would need to add a `1.0-alpha2`? If that is true
1. Do we really want to do that? The lifecycle could end up supporting many more APIs concurrently, adding complexity
2. Do "experimental" APIs even make sense as a category of APIs? Wouldn't it make more sense then to say that `1.0-alpha1` is first "supported" and later "deprecated" in that case?

text/0000-lifecycle-experimental-api-mode.md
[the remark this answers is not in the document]
hone: It's intentionally open ended right now, but my hope is linear.

hone:
> To clarify - this is different from rc releases of the lifecycle, and is rather rc releases of the APIs?

yeah, a way to test out APIs like we would test out software changes.

text/0000-lifecycle-experimental-api-mode.md
hone: I do think it limits what is possible for experimental. There are definitely "safer" cross cutting features and some that aren't. I just listed it as an alternative b/c it's something we could consider.

text/0000-lifecycle-experimental-api-mode.md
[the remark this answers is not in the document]
hone: > Do "experimental" APIs even make sense as a category of APIs? Wouldn't it make more sense then to say that 1.0-alpha1 is first "supported" and later "deprecated" in that case?

Like I said to @jromero's questions above, I'm fine with experimental not being its own section, but I do believe unlike both of those it carries a different support guarantees/breaking policies. This just makes that clearer by making it separate.

> I think this is now implying that the definition of 1.0-alpha1 is never allow to change, instead we would need to add a 1.0-alpha2

My hope is that we could branch/tag a spec, but maybe that overhead isn't worth it.

> Do we really want to do that? The lifecycle could end up supporting many more APIs concurrently, adding complexity

The key is to have a good/fast way to sunset experimental modes IMO.

jromero:
> > Two general thoughts:
> > ```
> > 1. It would seem that these experimental versions should also be included in "supported" given prior notion that "supported" is all encompassing including deprecated, and in this case experimental.
> > ```
> 
> One of the big differences is the guarantee and support we provide around experimental vs. both "supported" and "deprecated". If we want to include experimental in the supported list, I'm ok with that. We could even just drop the experimental moniker and have the version notation mark experimental.
> 
> > ```
> > 2. I find it difficult to reason about how versions and the experimental features will be carried forward in practicality. For example, if there is an experimental feature "A" that is encompassed in platform api 0.6-beta and then 0.7 is slated to come out with feature "A" still experimental does that inherently force 0.7 to also be labeled experimental as a whole even if other features are "stable"? Would we end up with to simultaneous yet difference spec versions `0.7` and `0.7-experimental`? I prefer the alternative "Unstable Features" in order to solve this unless we can detail more explicitly how this could work.
> > ```
> 
> I imagine if a feature isn't stabilized in the API it wouldn't make it into `0.7` or `0.6`. What you're suggesting aligns much more with the "Unstable Features" alternative like you suggest. It separates features more from API versions themselves.

Because of the complexities I envision in this along with the "linear" question posted by @jkutner, I would much rather be in favor of feature flags via "unstable features" that can be enabled somehow. I would like to have a clearer idea on the strategy we would use for denoting what features are experimental in what version and a few examples on how we move from one version to the next.

hone:
I've updated this based on the discussion from the WG meeting 2 weeks ago.

jabrown85 (approved):

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Experimental API Mode
+
+This is a new mode for testing implementation of a API. The API can be a full release or a prerelease API.
+
+New `CNB_PLATFORM_EXPERIMENTAL_MODE` and `CNB_BUILDPACK_EXPERIMENTAL` mode environment variables will control experimental mode with:
+allowed values: `warn`, `error`, `silent`
+default value: `warn`
+
+**When** the `CNB_PLATFROM_API` environment variable is set to an API version in the deprecated platform API, the lifecycle shall:
```
natalieparellano, line 47:
```suggestion
**When** the `CNB_PLATFORM_API` environment variable is set to an API version in the experimental platform API, the lifecycle shall:
```

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+ - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE` is unset, **Then** print a warning and continue
+ - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=warn`, **Then** print a warning and continue
+ - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=error`, **Then** fail
+ - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=silent`, **Then** continue w/o warning
+
+**When** the `api` field in a `buildpack.toml` file is set to an API version in the deprecated buildpack API range the lifecycle shall:
```
natalieparellano, line 53:
```suggestion
**When** the `api` field in a `buildpack.toml` file is set to an API version in the experimental buildpack API range the lifecycle shall:
```

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+* Prerelease API versions which is used for testing API / spec changes.
+* `experimental` is a new API mode for testing implementation of the API similar to the modes defined in [RFC#49](https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md).
+
+## Prerelease API
+
+A prerelease API is a non-finalized API but is in a testable state. The the API version must be of the format: `<major>.<minor>-<alphanumeric>`, i.e. `1.0-alpha1`. Though these aren't final, there will be a tagged and release version of the appropriate spec for it. This way users can see what is supported when it's used. In most cases, these will be cut from the corresponding branch in `buildpacks/spec`, i.e. Buildpack API `0.5-alpha1` would be cut from the `buildpack/0.5` branch.
+
+When Lifecycle supports a prerelease API, it can be treated like any other API version and be used in any mode: `experimental`, `supported`, `deprecated`.
+
+## Experimental API Mode
```
hone, line 39: From the WG meeting today, @ekcasey was pushing to have this removed until it's needed.

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+This proposes two changes to Lifecycle:
+* Prerelease API versions which is used for testing API / spec changes.
+* `experimental` is a new API mode for testing implementation of the API similar to the modes defined in [RFC#49](https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md).
+
+## Prerelease API
+
+A prerelease API is a non-finalized API but is in a testable state. The the API version must be of the format: `<major>.<minor>-<alphanumeric>`, i.e. `1.0-alpha1`. Though these aren't final, there will be a tagged and release version of the appropriate spec for it. This way users can see what is supported when it's used. In most cases, these will be cut from the corresponding branch in `buildpacks/spec`, i.e. Buildpack API `0.5-alpha1` would be cut from the `buildpack/0.5` branch.
+
+When Lifecycle supports a prerelease API, it can be treated like any other API version and be used in any mode: `experimental`, `supported`, `deprecated`.
+
```
hone, line 38: @ekcasey asked a good question in the WG today about the support policy, i.e. if supporting `1.2` implies that `1.0-alpha1` that came before it. We can follow most implementations of semver range rules where alphanumeric versions fall outside of the normal operators.

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+## Do Nothing
+If we choose to do nothing, we run the risk of releasing less tested API/spec releases.
+
+### Not a Different Version Format
+The Experimental API version format could instead follow the same format as deprecated and supported: `<major>-<minor>`. By doing this, users could automatically get upgraded from an experimental to supported/deprecated API version which may be surprising to them.
+
+### Unstable Features
+Similar to [RFC#91](https://github.com/buildpacks/rfcs/pull/91), instead of tying features to a specific release, features can be tied behind flags for testing. There's a potential this could be complicated to do if the features are coupled to other changes going on in the API/spec. The benefit is folks can try out individual features in a more explicit way.
+
+### Experimental Section in the API
```
hone, line 120: From the WG today, @sclevine requested this be moved into the actual proposal.

ekcasey: If we pull in unstable features do we need something like `CNB_PLATFORM_UNSTABLE_MODE` with `error`, `warn` and `silent` options?

text/0000-lifecycle-experimental-api-mode.md
```diff
@@ -0,0 +1,139 @@
...
+
+A prerelease API is a non-finalized API but is in a testable state. The the API version must be of the format: `<major>.<minor>-<alphanumeric>`, i.e. `1.0-alpha1`. Though these aren't final, there will be a tagged and release version of the appropriate spec for it. This way users can see what is supported when it's used. In most cases, these will be cut from the corresponding branch in `buildpacks/spec`, i.e. Buildpack API `0.5-alpha1` would be cut from the `buildpack/0.5` branch.
+
+When Lifecycle supports a prerelease API, it can be treated like any other API version and be used in any mode: `experimental`, `supported`, `deprecated`.
+
+## Experimental API Mode
+
+This is a new mode for testing implementation of a API. The API can be a full release or a prerelease API.
+
+New `CNB_PLATFORM_EXPERIMENTAL_MODE` and `CNB_BUILDPACK_EXPERIMENTAL` mode environment variables will control experimental mode with:
```
ekcasey, line 43: Why two variables?

text/0000-lifecycle-experimental-api-mode.md
ekcasey: Assuming we go ahead with the unstable features, I think it is potentially confusing that "experimental" in `pack` is analogous to "unstable" in lifecycle, while "experimental" in lifecycle is an entirely different concept.

I would prefer to call the unstable features experimental features and leave the experimental API as documented in this RFC out for now. Having so many different mechanism for experimental support seems like it could be potentially confusing to end users imho.

nebhale (approved):

text/0000-lifecycle-experimental-api-mode.md
hone: copy / paste error :(

text/0000-lifecycle-experimental-api-mode.md
hone: @ekcasey I moved experimental features up into the proposal and moved experimental "mode" out into alternatives. Do you still have naming concerns?

text/0000-lifecycle-prelease-version-and-experimental-section.md
```diff
@@ -31,65 +31,41 @@ This proposes two changes to Lifecycle:
...
- - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=warn`, **Then** print a warning and continue
- - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=error`, **Then** fail
- - **If** `CNB_PLATFORM_EXPERIMENTAL_MODE=silent`, **Then** continue w/o warning
-
-**When** the `api` field in a `buildpack.toml` file is set to an API version in the deprecated buildpack API range the lifecycle shall:
- - **If** `CNB_BUILDPACK_EXPERIMENTAL_MODE` is unset, **Then** print a warning and continue
- - **If** `CNB_BUILDPACK_EXPERIMENTAL_MODE=warn`, **Then** print a warning and continue
- - **If** `CNB_BUILDPACK_EXPERIMENTAL_MODE=error`, **Then** fail
- - **If** `CNB_BUILDPACK_EXPERIMENTAL_MODE=silent`, **Then** continue w/o warning
+### Experimental Section in the API
```
ekcasey, line 38: I think we probably want to require users to explicitly opt into experimental features

We could do this with a boolean `CNB_EXPERIMENTAL` env var which defaults to false. If a platform uses an experimental feature without opting in we will fail.

text/0000-lifecycle-experimental-api-mode.md
ekcasey: looks good!

text/0000-lifecycle-prelease-version-and-experimental-section.md
sclevine: @hone

hone: @ekcasey how does that work with the buildpack API side of things? Is it all features if we have multiple or is it an array of features?

hone: @ekcasey I've updated this based on our convo last week to have a single env var set by the platform to control turning this on/off. By default we'll warn.

ekcasey (approved):

sclevine:
FCP closing on 11/11

jromero (approved):

PR #116, opened 2020-09-15
Add Extension Spec for Builder 

dfreilich:
[Readable](https://github.com/dfreilich/rfcs/blob/extension-spec-builder/text/0000-builder-spec.md)

### DIFF
--- a/text/0000-builder-spec.md
+++ b/text/0000-builder-spec.md
@@ -132,8 +132,8 @@ Additionally, a builder requires:
 * The image config's Labels field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
 * The image config's Labels field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [buildpack layers](#layers)
 
-If the builder contains an optional [lifecycle descriptor file][lifecycle-descriptor-rfc], it also requires:
-* The image config's Labels field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
+In accordance with the [lifecycle descriptor rfc][lifecycle-descriptor-rfc], it also requires:
+* The image config's Labels field has the label io.buildpacks.lifecycle.version, set to the lifecycle version (greater than `0.9.0`)
 * The image config's Labels field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
 
 #### Order
@@ -175,7 +175,7 @@ The `io.buildpacks.builder.metadata` data should look like:
   ],
   "createdBy": {
     "name": "<creator of builder>", 
-    "version": "<builder version>"
+    "version": "<version of tool used to create builder>"
   }
 }
 ```
--- a/text/0000-builder-spec.md
+++ b/text/0000-builder-spec.md
@@ -25,22 +25,22 @@ A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecif
 
 # Motivation
 [motivation]: #motivation
-Builders are a vital part of the CNB ecosystem as it currently stands; they are used to distribute the lifecycle and buildpacks to every platform, including:
+Builders are a vital part of the CNB ecosystem as it currently stands; they are used to distribute the lifecycle and buildpacks to many platforms, including:
 * pack
 * Tekton
 * pack-orb
 * kpack
 etc.
 
-However, at this point, builders are underspecified, and a rigorous technical specification would fill a needed hole in the current documentation.
+However, at this point, builders are unspecified, and a rigorous technical specification would guarantee a stable API for platforms to develop against.
 
 # What it is
 [what-it-is]: #what-it-is
 A Builder consists of four items:
 * Buildpacks &rarr; A list of buildpack references, and an ordering of buildpacks
 * Lifecycle &rarr; A version of a lifecycle implementation
-* Stack’s build image &rarr; An environment used at build-time (the metadata contains a reference to the run image required, and potentially a list of run-image mirrors)
-* Metadata, describing the builder (description, who it was created by, etc)
+* Stack’s build image &rarr; An environment used at build-time
+* Metadata, describing the builder (description, who it was created by, etc) and the stack run images (potentially a list of run-image mirrors as well)
 
 Users of `pack` interact with builders in the following ways:
 * `pack build` &rarr; requires a builder to construct an image
@@ -116,6 +116,7 @@ A builder is composed of at least the following directories/files:
 - /cnb/order.toml
 - /cnb/stack.toml
 - /layers
+- /platform
 - /workspace
 ```
 
@@ -129,13 +130,14 @@ A builder's environment is the build-time environment of the stack, and as such,
 
 Additionally, a builder requires:
 * The image config's WorkingDir should be set
-* The image config's Label field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
-* The image config's Label field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
-* The image config's Label field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [layers](#layers)
+* The image config's Labels field has the label io.buildpacks.builder.api, set to a string (defaulting to `0.1`)
+* The image config's Labels field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
+* The image config's Labels field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
+* The image config's Labels field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [buildpack layers](#layers)
 
 If the builder contains an optional [lifecycle descriptor file][lifecycle-descriptor-rfc], it also requires:
-* The image config's Label field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
-* The image config's Label field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
+* The image config's Labels field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
+* The image config's Labels field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
 
 #### Order
 The `io.buildpacks.buildpack.order` data should look like:
@@ -159,7 +161,6 @@ The `io.buildpacks.builder.metadata` data should look like:
 ```json
 {
   "description": "<Some description>",
-  "api": "<some builder API version | default 0.1 >",
   "stack": {
     "runImage": {
       "image": "<some/run-image>",
@@ -175,17 +176,6 @@ The `io.buildpacks.builder.metadata` data should look like:
 	  "homepage": "http://geocities.com/top-bp"
 	}
   ],
-  "lifecycle": {
-    "version": "<version of lifecycle in builder>",
-    "apis":  {
-        "buildpack": {
-          "deprecated": ["0.1"],
-          "supported": ["1.2", "1.3"]},
-        "platform": {
-          "deprecated": [],
-          "supported": ["2.3", "2.4"]}
-    }
-  },
   "createdBy": {
     "name": "<creator of builder>", 
     "version": "<builder version>"
@@ -257,7 +247,7 @@ We should add a cross-reference to the extension builder specification, noting t
 
 # Drawbacks
 [drawbacks]: #drawbacks
-N/A
+Development Speed &rarr; By specifying the builder spec, it will require more ceremony to make changes, like one to builder metadata
 
 # Alternatives
 [alternatives]: #alternatives
@@ -276,7 +266,7 @@ Some similar PRs that led to the development of specifications are:
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
-This RFC should lead to changes in the distribution spec.
+This RFC should lead to a new extension spec, and small changes to the distribution and platform spec.
 
 [//]: <> (Links)
 [buildpacks-spec]: https://github.com/buildpacks/spec/blob/main/buildpack.md
--- a/text/0000-builder-spec.md
+++ b/text/0000-builder-spec.md
@@ -21,7 +21,7 @@ A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecif
 * **[Platform][platform-docs]** &rarr; A **platform** uses a lifecycle, buildpacks (packaged in a builder), and application source code to produce an OCI image.
 * **[Stack][stack-docs]** &rarr; A stack provides the buildpack lifecycle with build-time and run-time environments in the form of images.
 * **[Lifecycle][lifecycle-docs]** &rarr; The lifecycle orchestrates buildpack execution, then assembles the resulting artifacts into a final app image.
-* **Extension Spec** &rarr; An extension spec defines soemthing that is an optional add-on for Cloud Native Buildpacks systems, but not required to be considered `spec` compliant.
+* **Extension Spec** &rarr; An extension spec defines something that is an optional add-on for Cloud Native Buildpacks systems, but not required to be considered `spec` compliant.
 
 # Motivation
 [motivation]: #motivation
@@ -121,12 +121,12 @@ A builder is composed of at least the following directories/files:
 A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs].
 
 Additionally, a builder requires:
-* The image config's WorkingDir should be set
+* The image config's WorkingDir should be set to a directory in the builder
 * The image config's Env field has the environment variable `CNB_APP_DIR` set to the application directory of the build environment
-* The image config's Env field has the environment variable `CNB_LAYERS_DIR` set
-* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set
-* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set
-* The image config's Env field has the environment variable `SERVICE_BINDING_ROOT` set if any services are bound
+* The image config's Env field has the environment variable `CNB_LAYERS_DIR` set to a directory in the builder
+* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set to a directory in the builder
+* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set to a directory in the builder
+* The image config's Env field has the environment variable `SERVICE_BINDING_ROOT` set to a directory in the builder, if any services are bound
 * The image config's Labels field has the label io.buildpacks.builder.api, set to a string (defaulting to `0.1`)
 * The image config's Labels field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
 * The image config's Labels field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
--- a/text/0000-builder-spec.md
+++ b/text/0000-builder-spec.md
@@ -10,7 +10,9 @@
 
 # Summary
 [summary]: #summary
-A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecified][spec-issue-builder]. This RFC proposes that we should add a rigorous description of builders in the [distribution specification][distribution-spec], and clarify the usage/existence of builders in the [platform specification][platform-spec]
+A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecified][spec-issue-builder]. This RFC proposes that we should add a rigorous versioned specification of builders as an extension to the core specification, and clarify the existence of builders in the:
+* [distribution specification][distribution-spec]
+* [platform specification][platform-spec]
 
 # Definitions
 [definitions]: #definitions
@@ -19,6 +21,7 @@ A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecif
 * **[Platform][platform-docs]** &rarr; A **platform** uses a lifecycle, buildpacks (packaged in a builder), and application source code to produce an OCI image.
 * **[Stack][stack-docs]** &rarr; A stack provides the buildpack lifecycle with build-time and run-time environments in the form of images.
 * **[Lifecycle][lifecycle-docs]** &rarr; The lifecycle orchestrates buildpack execution, then assembles the resulting artifacts into a final app image.
+* **Extension Spec** &rarr; An extension spec defines soemthing that is an optional add-on for Cloud Native Buildpacks systems, but not required to be considered `spec` compliant.
 
 # Motivation
 [motivation]: #motivation
@@ -49,7 +52,7 @@ Users of `pack` interact with builders in the following ways:
 * `list-trusted-builders` &rarr; List Trusted Builders
 * `create-builder` &rarr; Create builder image
 
-Users of `pack-orb`, `tekton`, and `kpack` interact with Builders through supplying a builder to use for building applications.
+Users of `pack-orb` and `tekton` interact with Builders through supplying a builder to use for building applications. Users of `kpack` can create and use builders.
 
 A sample inspected builder (using `pack inspect-builder`), looks like:
 ```bash
@@ -116,7 +119,7 @@ A builder is composed of at least the following directories/files:
 - /workspace
 ```
 
-In addition, builders can build in environment variables by defining env-files in
+In addition, builders MAY build in environment variables by defining env-files in
 ```
 - /platform/env/<env-files>
 ```
@@ -132,20 +135,7 @@ Additionally, a builder requires:
 
 If the builder contains an optional [lifecycle descriptor file][lifecycle-descriptor-rfc], it also requires:
 * The image config's Label field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
-* The image config's Label field has the label io.buildpacks.lifecycle.apis, set to:
-```json
-{
-  "buildpack": {
-    "deprecated": ["<list of versions>"],
-    "supported": ["<list of versions>"]
-  },
-  "platform": {
-    "deprecated": ["<list of versions>"],
-    "supported": ["<list of versions>"]
-  }
-}
-```
-
+* The image config's Label field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
 
 #### Order
 The `io.buildpacks.buildpack.order` data should look like:
@@ -169,6 +159,7 @@ The `io.buildpacks.builder.metadata` data should look like:
 ```json
 {
   "description": "<Some description>",
+  "api": "<some builder API version | default 0.1 >",
   "stack": {
     "runImage": {
       "image": "<some/run-image>",
@@ -239,15 +230,30 @@ The `io.buildpacks.buildpack.layers` data should look like:
 }
 ```
 
+### Lifecycle APIs
+The `io.buildpacks.lifecycle.apis` data should look like:
+```json
+{
+  "buildpack": {
+    "deprecated": ["<list of versions>"],
+    "supported": ["<list of versions>"]
+  },
+  "platform": {
+    "deprecated": ["<list of versions>"],
+    "supported": ["<list of versions>"]
+  }
+}
+```
+
 ## Spec Additions
-### [Buildpack specification][buildpacks-spec]
-Builders don't seem immediately relevant to the buildpack specification; as such, I wouldn't add mention of builders here. 
+### Extension Spec
+Builders should be specified as an extension spec, and explicitly versioned, using a `<major>.<minor>` pattern, and following the [established spec api release pattern][spec-api-branches]. The version should be persisted in the `io.buildpacks.builder.metadata`, as a top-level `api` field. Versioning will start with 0.1.
 
 ### [Platform specification][platform-spec]
-We can add a note, clarifying that in platforms that use builders, the Stack's build image is the core of the builder. 
+We should add a cross-reference to the extension builder specification, noting that a builder is a fully-functioning Stack build image.
 
 ### [Distribution specification][distribution-spec]
-We should have a rigorous definition of the builder here, with a description of all the files and metadata necessary and allowed. 
+We should add a cross-reference to the extension builder specification, noting that a builder is also a valid artifact, and is specified in the extension spec.
 
 # Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs], including:
+* The image config's User field is set to a non-root user with a writable home directory.
+* The image config's Env field has the environment variable CNB_STACK_ID set to the stack ID.
+* The image config's Env field has the environment variable CNB_USER_ID set to the user †UID/‡SID of the user specified in the User field.
+* The image config's Env field has the environment variable CNB_GROUP_ID set to the primary group †GID/‡SID of the user specified in the User field.
+* The image config's Label field has the label io.buildpacks.stack.id set to the stack ID. (string)
+* The image config's Label field has the label io.buildpacks.stack.mixins set to a JSON array containing mixin names for each mixin applied to the image.
+
+Additionally, a builder requires:
+* The image config's WorkingDir should be set
```
dfreilich, line 129: Should this be a requirement?

ekcasey: In the future the working dir shouldn't matter https://github.com/buildpacks/spec/pull/128.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+### Env Vars/Labels
+A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs], including:
+* The image config's User field is set to a non-root user with a writable home directory.
+* The image config's Env field has the environment variable CNB_STACK_ID set to the stack ID.
+* The image config's Env field has the environment variable CNB_USER_ID set to the user †UID/‡SID of the user specified in the User field.
+* The image config's Env field has the environment variable CNB_GROUP_ID set to the primary group †GID/‡SID of the user specified in the User field.
+* The image config's Label field has the label io.buildpacks.stack.id set to the stack ID. (string)
+* The image config's Label field has the label io.buildpacks.stack.mixins set to a JSON array containing mixin names for each mixin applied to the image.
+
+Additionally, a builder requires:
```
ekcasey, line 128: We should include `io.buildpacks.lifecycle.version` and `io.buildpacks.lifecycle.apis` labels from https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md#lifecycle-labels.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+### [Distribution specification][distribution-spec]
+We should have a rigorous definition of the builder here, with a description of all the files and metadata necessary and allowed. 
+
+# Drawbacks
+[drawbacks]: #drawbacks
+N/A
+
+# Alternatives
+[alternatives]: #alternatives
+Specifying the builder is something that, to my mind, should definitely done; the question is, how. We could have an entirely separate `extension spec`, defining builders and how they should be used.
```
ekcasey, line 242: I think a separate extension spec is appropriate, given that a platform could choose not to use builders and instead work with stacks, lifecycles, and buildpacks directly.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+### Env Vars/Labels
+A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs], including:
+* The image config's User field is set to a non-root user with a writable home directory.
+* The image config's Env field has the environment variable CNB_STACK_ID set to the stack ID.
+* The image config's Env field has the environment variable CNB_USER_ID set to the user †UID/‡SID of the user specified in the User field.
+* The image config's Env field has the environment variable CNB_GROUP_ID set to the primary group †GID/‡SID of the user specified in the User field.
+* The image config's Label field has the label io.buildpacks.stack.id set to the stack ID. (string)
+* The image config's Label field has the label io.buildpacks.stack.mixins set to a JSON array containing mixin names for each mixin applied to the image.
```
ekcasey, line 126: Instead of enumerating the requirements of a `build-image` maybe it is best to simply leave the link to the spec, so that this section doesn't fall out of date.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+## Structure of Builders
+### Files/Directories
+A builder is composed of at least the following directories/files:
+```
+- /cnb/buildpacks/...<buildpack ID>/<buildpack version>/
+- /cnb/lifecycle/<lifecycle binaries>
+- /cnb/order.toml
+- /cnb/stack.toml
+- /layers
+- /platform/env/<env-files>
```
ekcasey, line 115: I don't think we should expect and certainly not require environment variable files to be baked into the builder image. These are supplied by the user for a specific build and are typically mounted in (although `pack` does bake these into an ephemeral builder image).

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+* Tekton
+* pack-orb
+* kpack
+etc.
+
+However, at this point, builders are underspecified, and a rigorous technical specification would fill a needed hole in the current documentation.
+
+# What it is
+[what-it-is]: #what-it-is
+A Builder consists of four items:
```
ekcasey, line 35: I think the stack information including a reference to the `run-image` and potentially `run-image-mirrors` should be called out separately here.

ekcasey (commented):
In my mind the big missing piece here is a API number or similar concept for builder images. An API number would allow us to change label schemas and file paths in an explicit and predictable way.

text/0000-builder-spec.md
dfreilich: Super interesting! That's only for platform api 0.5, though, so I assume it should be kept in at the moment?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+* **[Platform][platform-docs]** &rarr; A **platform** uses a lifecycle, buildpacks (packaged in a builder), and application source code to produce an OCI image.
+* **[Stack][stack-docs]** &rarr; A stack provides the buildpack lifecycle with build-time and run-time environments in the form of images.
+
+# Motivation
+[motivation]: #motivation
+Builders are a vital part of the CNB ecosystem as it currently stands; they are used to distribute the lifecycle and buildpacks to every platform, including:
+* pack
+* Tekton
+* pack-orb
+* kpack
```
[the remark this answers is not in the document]
dfreilich: Interesting! I hadn't realized that change.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecified][spec-issue-builder]. This RFC proposes that we should add a rigorous description of builders in the [distribution specification][distribution-spec], and clarify the usage/existence of builders in the [platform specification][platform-spec]
+
+# Definitions
+[definitions]: #definitions
+* **[Builder][builder-docs]** &rarr; A **builder** is an image that bundles all the bits and information on how to build your apps, such as buildpacks and build-time image, as well as executes the buildpacks against your app source code.
```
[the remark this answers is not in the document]
dfreilich: 👍 

I grabbed that description from the docs, so I'll take that as feedback for docs.

text/0000-builder-spec.md
dfreilich: 41b9a4ae471e3490de686f66e81fa8bbd999a262

text/0000-builder-spec.md
dfreilich: 41b9a4ae471e3490de686f66e81fa8bbd999a262

text/0000-builder-spec.md
dfreilich: done in 41b9a4ae471e3490de686f66e81fa8bbd999a262

text/0000-builder-spec.md
```diff
@@ -0,0 +1,270 @@
...
+Specifying the builder is something that, to my mind, should definitely done; the question is, how. We could have an entirely separate `extension spec`, defining builders and how they should be used. 
+ > Given that one of the key uses of builders is in distributing buildpacks, I thought it should at least be discussed in some depth in the [distribution spec][distribution-spec], and given that the buildpackage makeup is described there, I thought it appropriate to just describe it fully in the distribution spec. 
+
+# Prior Art
+[prior-art]: #prior-art
+Some similar PRs that led to the development of specifications are:
+* [Distribution Specification][distrib-spec-rfc]
+* [Service Binding Specification][service-binding-rfc]
+
+# Unresolved Questions
```
jkutner, line 267: Should we also define the schema of the `builder.toml`?

dfreilich: It could make sense to do so - but on the other hand, `builder.toml` is used in pack, but should it be standardized for other platforms as well? It feels like more of an implementation detail of how pack creates builders.

dfreilich: Seemed like consensus in Working Group was not to define it here.

dfreilich:
Feedback from WG meeting (09/16):
- It should be in extension spec, because platforms don't need to use builders at all. Add a link in platform spec noting that a builder is a fully functioning build-environment of a stack. kpack, despite having different style builders, still tries to support the same spec of builders. 
- Builder API should certainly be versioned. There was a suggestion for using OCI [media-type labels](https://github.com/opencontainers/image-spec/blob/master/media-types.md#oci-image-media-types) to define the version on the builder image, with a cross-cutting concern that we wouldn't want to lose support for any registries just because we use an OCI compliant feature that isn't necessarily supported on every registry. 

I'll rework it using that feedback.

text/0000-builder-spec.md
dfreilich: Incorporated

nebhale (approved):

jkutner (approved):

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* **[Buildpack][buildpacks-docs]** &rarr; A **buildpack** is a unit of work that inspects your app source code and formulates a plan to build and run your application.
+* **[Platform][platform-docs]** &rarr; A **platform** uses a lifecycle, buildpacks (packaged in a builder), and application source code to produce an OCI image.
+* **[Stack][stack-docs]** &rarr; A stack provides the buildpack lifecycle with build-time and run-time environments in the form of images.
+* **[Lifecycle][lifecycle-docs]** &rarr; The lifecycle orchestrates buildpack execution, then assembles the resulting artifacts into a final app image.
+* **Extension Spec** &rarr; An extension spec defines soemthing that is an optional add-on for Cloud Native Buildpacks systems, but not required to be considered `spec` compliant.
+
+# Motivation
+[motivation]: #motivation
+Builders are a vital part of the CNB ecosystem as it currently stands; they are used to distribute the lifecycle and buildpacks to every platform, including:
```
ekcasey, line 28:
```suggestion
Builders are a vital part of the CNB ecosystem as it currently stands; they are used to distribute the lifecycle and buildpacks to many platforms, including:
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* pack
+* Tekton
+* pack-orb
+* kpack
+etc.
+
+However, at this point, builders are underspecified, and a rigorous technical specification would fill a needed hole in the current documentation.
```
ekcasey, line 35:
```suggestion
However, at this point, builders are unspecified, and a rigorous technical specification would guarantee a stable API for platforms to develop against.
```
This is a bit of a nitpick. We could easily _document_ builders as `pack` defines them but specifying them goes a step further.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+A Builder consists of four items:
+* Buildpacks &rarr; A list of buildpack references, and an ordering of buildpacks
+* Lifecycle &rarr; A version of a lifecycle implementation
+* Stack’s build image &rarr; An environment used at build-time (the metadata contains a reference to the run image required, and potentially a list of run-image mirrors)
```
ekcasey, line 42: The run-image and mirror references don't actually exist on the stack `build-image`. They are added in a label and a `stack.toml` file only when a builder is created.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+```
+- /platform/env/<env-files>
+```
+
+### Env Vars/Labels
+A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs].
+
+Additionally, a builder requires:
+* The image config's WorkingDir should be set
+* The image config's Label field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
```
ekcasey, line 132:
```suggestion
* The image config's Labels field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* The image config's Label field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
```
ekcasey, line 133:
```suggestion
* The image config's Labels field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* The image config's Label field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [layers](#layers)
```
ekcasey, line 134:
```suggestion
* The image config's Labels field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [buildpack layers](#layers)
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+
+If the builder contains an optional [lifecycle descriptor file][lifecycle-descriptor-rfc], it also requires:
+* The image config's Label field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
```
ekcasey, line 137:
```suggestion
* The image config's Labels field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* The image config's Label field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
```
ekcasey, line 138:
```suggestion
* The image config's Labels field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
```

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+    "version": "<version of lifecycle in builder>",
+    "apis":  {
+        "buildpack": {
+          "deprecated": ["0.1"],
+          "supported": ["1.2", "1.3"]},
+        "platform": {
+          "deprecated": [],
+          "supported": ["2.3", "2.4"]}
+    }
+  },
```
ekcasey, line 188: Given this information exists in the `io.buildpacks.lifecycle.*` labels I think we can pull it out of the builder metadata. The only reason the lifecycle version and the old form of the APIs:
```
  "lifecycle": {
    "version": "0.9.1",
    "api": {
      "buildpack": "0.2",
      "platform": "0.3"
      }
   },
   ...
```
are in the current label is for backwards compatibility. I don't think we should specifying the backwards compatible data, even if we continue to include in `pack create-builder` for some period of time.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+      ]
+    }
+  },
+  "buildpacks": [
+    {
+      "id": "<buildpack id>",
+	  "version": "<buildpack version>",
+	  "homepage": "http://geocities.com/top-bp"
+	}
+  ],
```
ekcasey, line 177: Do we need this information in this label if it is already in `io.buildpacks.buildpack.layers`?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+	}
+]
+```
+
+#### Metadata
+The `io.buildpacks.builder.metadata` data should look like:
+```json
+{
+  "description": "<Some description>",
+  "api": "<some builder API version | default 0.1 >",
```
ekcasey, line 162: Given this is new in this RFC maybe we should move this into a separate `io.buildpacks.builder.api` label, now that we are preferring more specific labels?

hone:
Is a file like `builder.toml` `pack` specific or something included here?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+
+### [Platform specification][platform-spec]
+We should add a cross-reference to the extension builder specification, noting that a builder is a fully-functioning Stack build image.
+
+### [Distribution specification][distribution-spec]
+We should add a cross-reference to the extension builder specification, noting that a builder is also a valid artifact, and is specified in the extension spec.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+N/A
```
ekcasey, line 260: For the benefits of the formal specification we trade off speed. For example, it will require more ceremony to make changes to builder metadata.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+* [Distribution Specification][distrib-spec-rfc]
+* [Service Binding Specification][service-binding-rfc]
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+* This RFC doesn't assume the requirement of a PATH variable, or any other open RFCs  
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+This RFC should lead to changes in the distribution spec.
```
ekcasey, line 279:
```suggestion
This RFC should lead to a new extension spec.
```

dfreilich:
@hone It came up here: https://github.com/buildpacks/rfcs/pull/116#discussion_r489655243. Consensus in WG was that it was `pack` specific.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+## Structure of Builders
+### Files/Directories
+A builder is composed of at least the following directories/files:
+```
+- /cnb/buildpacks/...<buildpack ID>/<buildpack version>/
+- /cnb/lifecycle/<lifecycle binaries>
+- /cnb/order.toml
+- /cnb/stack.toml
+- /layers
+- /workspace
```
ekcasey, line 119: After some thought I think we should add `/platform` in here to reduce possible permissions issue reading volumes mounted volumes under platform.

We may also want to build in `/platform/bindings` and specify that `$SERVICE_BINDING_ROOT` should be set to `/platform/bindings` in the builder image. This is a convenience for any platform that wants to support bindings.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+In addition, builders MAY build in environment variables by defining env-files in
+```
+- /platform/env/<env-files>
+```
```
ekcasey, line 125: I don't think we should be encouraging folks to distribute `/platform/env/<env-files>` on a builder image. Although pack creates an ephemeral builder with these files, it would be natural for a platform like `tekton` to mount a volume over `/platform/env` to provide env vars to a buildpack. Any env files baked into a distributable builder would be present in some platforms and not in others, leading to weird behavior.

text/0000-builder-spec.md
ekcasey, line 119: I am wondering...
Do we actually want to specify these literal file paths or if we want to specify things like, "there is an application directory and `CNB_APP_DIR` is set to the app dir in the builder image environment."

The latter would give us more flexibility to for example change `/layers` to `/build/layers` or something like that in the future. Not hard coding the default paths in the building spec but instead using the env vars keeps our options open.

text/0000-builder-spec.md
```diff
@@ -0,0 +1,295 @@
...
+- Author(s): [@dfreilich](https://github.com/dfreilich)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+A Builder is a critical tool in the CNB ecosystem, and yet, is largely [unspecified][spec-issue-builder]. This RFC proposes that we should add a rigorous versioned specification of builders as an extension to the core specification, and clarify the existence of builders in the:
+* [distribution specification][distribution-spec]
```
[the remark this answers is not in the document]
dfreilich: I think that since we use builders in a lot of contexts, to distribute things, it would be weird for us not to at least acknowledge their existence in the distribution spec, even if we don't require them. Similar to the addition in the platform spec that builders are fully functioning stack build image, say that builders are similar artifacts to buildpackages, and MAY be used by some platforms.

text/0000-builder-spec.md
dfreilich: I had meant to say that it was on the metadata of a builder, and not a stack, but I'll make that more clear.

text/0000-builder-spec.md
dfreilich: I thought the bindings spec was deprecated in favor of the K8s spec?

dfreilich: If the k8s spec just requires that $SERVICE_BINDING_ROOT be set, why would we restrict it to `platform/bindings`?

text/0000-builder-spec.md
dfreilich: But if some builders are doing it, is it better to leave it as an undocumented easter-egg feature?

text/0000-builder-spec.md
dfreilich: Done in 211b74c7c7443b353cfa111cc15c92abb60369a9

text/0000-builder-spec.md
dfreilich: Done in 211b74c7c7443b353cfa111cc15c92abb60369a9

text/0000-builder-spec.md
dfreilich: Done in 211b74c7c7443b353cfa111cc15c92abb60369a9

text/0000-builder-spec.md
dfreilich: I won't spam you with too many more emails, but did all of these suggestions in 211b74c7c7443b353cfa111cc15c92abb60369a9

text/0000-builder-spec.md
dfreilich: I think that having the literal file paths could be helpful to concretize what a builder looks like, but I'm also happy to change it, if consensus is against that.

text/0000-builder-spec.md
natalieparellano: Do we have any examples of non-ephemeral builders that are distributing env vars this way?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,285 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+## Structure of Builders
+### Files/Directories
+A builder is composed of at least the following directories/files:
+```
+- /cnb/buildpacks/...<buildpack ID>/<buildpack version>/
+- /cnb/lifecycle/<lifecycle binaries>
+- /cnb/order.toml
```
natalieparellano, line 116: Does this duplicate information in the `io.buildpacks.buildpack.order` label?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,285 @@
...
+- /cnb/stack.toml
```
natalieparellano, line 117: Does it make sense to call out what's in these files? I believe this information is also captured in `io.buildpacks.builder.metadata`. It's weird that it's in both places.

text/0000-builder-spec.md
ekcasey: TBH I don't consider packs ephemeral builder to be builders formally and therefore I wouldn't say that "some builders are doing it".

text/0000-builder-spec.md
ekcasey: The information needs exist in both a label and a file so it can be read by a platform (from the label) and by the lifecycle (from the file).

text/0000-builder-spec.md
dfreilich: To clarify, I have added `/platform` in, and I'm happy to build in `/platform/bindings` as well, I just don't know enough about the current state of service bindings, and how it's used in the CNB ecosystem. Do current platforms set `$SERVICE+BINDING_ROOT`?

dfreilich: Should we do this for all file paths, or just for `CNB_APP_DIR`?

text/0000-builder-spec.md
dfreilich: Those are already defined in the [Platform documentation](https://github.com/buildpacks/spec/blob/main/platform.md), so I don't think there's a major need

text/0000-builder-spec.md
dfreilich: Looking at the [k8s spec](https://k8s-service-bindings.github.io/spec/#application-projection), I see that it demands that `$SERVICE_BINDING_ROOT` be set. I think, though, that given this comment (https://github.com/buildpacks/rfcs/pull/116#discussion_r493777489), we probably shouldn't require it be set to a specific directory. What do you think?

dfreilich:
@buildpacks/core-team There were 2 open questions from @ekcasey I wanted to get feedback for:
1. Should we specify a dedicated directory for service bindings, like `/platform/bindings`, and require builders to set `$SERVICE_BINDING_ROOT` to `/platform/bindings`?
1. Do we actually want to specify any literal file paths or should we use `CNB_X` env vars to be set to an appropriate directory?

nebhale:
@dfreilich 

> Should we specify a dedicated directory for service bindings, like /platform/bindings, and require builders to set $SERVICE_BINDING_ROOT to /platform/bindings?

I’m ambivalent about whether we should confine bindings to `/platform/bindings`.  We were really strict in the early days of the spec, but I think we’ve been slouching towards a bit more of a lassie faire attitude towards things like this, recently.  Given that at runtime, the developer/platform is free to mount the bindings anywhere in the filesystem, I’m not sure it’s worthwhile for us to be more restrictive any more.  Regardless of what we choose (mandatory location, flexible location) we should require that `$SERVICE_BINDING_ROOT` **MUST** be set if services are bound.

dfreilich:
> Regardless of what we choose (mandatory location, flexible location) we should require that $SERVICE_BINDING_ROOT MUST be set if services are bound.

@nebhale Added that in in 61420f4. I felt that if there's no strict requirement from our end that `SERVICE_BINDING_ROOT`, it doesn't make sense to restrict it to `/platforms/binding`

nebhale (approved):

text/0000-builder-spec.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Env Vars/Labels
+A builder's environment is the build-time environment of the stack, and as such, requires all of the [build image specifications][build-image-specs].
+
+Additionally, a builder requires:
+* The image config's WorkingDir should be set
+* The image config's Env field has the environment variable `CNB_APP_DIR` set to the application directory of the build environment
+* The image config's Env field has the environment variable `CNB_LAYERS_DIR` set
+* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set
+* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set
```
ekcasey, line 128: We should specify that the directories these env vars point to should exist in the builder image.

ekcasey (approved):

hone (approved):

sclevine (approved):

text/0000-builder-spec.md
```diff
@@ -0,0 +1,285 @@
...
+* [platform specification][platform-spec]
+
+# Definitions
+[definitions]: #definitions
+* **[Builder][builder-docs]** &rarr; A **builder** is an image that bundles all the bits and information on how to build your apps, such as buildpacks, an implementation of the lifecycle, and build-time environment that platforms may use when executing the lifecycle.
+* **[Buildpack][buildpacks-docs]** &rarr; A **buildpack** is a unit of work that inspects your app source code and formulates a plan to build and run your application.
+* **[Platform][platform-docs]** &rarr; A **platform** uses a lifecycle, buildpacks (packaged in a builder), and application source code to produce an OCI image.
+* **[Stack][stack-docs]** &rarr; A stack provides the buildpack lifecycle with build-time and run-time environments in the form of images.
+* **[Lifecycle][lifecycle-docs]** &rarr; The lifecycle orchestrates buildpack execution, then assembles the resulting artifacts into a final app image.
+* **Extension Spec** &rarr; An extension spec defines soemthing that is an optional add-on for Cloud Native Buildpacks systems, but not required to be considered `spec` compliant.
```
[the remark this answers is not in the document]
dfreilich: Fixed in 94207149e7534c608c8b926652ba12cdccff66b8

text/0000-builder-spec.md
dfreilich: Fixed in 94207149e7534c608c8b926652ba12cdccff66b8

natalieparellano (approved):

text/0000-builder-spec.md
```diff
@@ -0,0 +1,282 @@
...
+  "buildpacks": [
+    {
+      "id": "<buildpack id>",
+	  "version": "<buildpack version>",
+	  "homepage": "http://geocities.com/top-bp"
+	}
+  ],
+  "createdBy": {
+    "name": "<creator of builder>", 
+    "version": "<builder version>"
```
jromero, line 178: Isn't this really more like "version of tool used to create builder"?

text/0000-builder-spec.md
```diff
@@ -0,0 +1,282 @@
...
+* The image config's Env field has the environment variable `CNB_PLATFORM_DIR` set to a directory in the builder
+* The image config's Env field has the environment variable `SERVICE_BINDING_ROOT` set to a directory in the builder, if any services are bound
+* The image config's Labels field has the label io.buildpacks.builder.api, set to a string (defaulting to `0.1`)
+* The image config's Labels field has the label io.buildpacks.builder.metadata, set to a JSON object representing [Builder Metadata](#metadata)
+* The image config's Labels field has the label io.buildpacks.buildpack.order, set to a JSON object representing an [Order](#order)
+* The image config's Labels field has the label io.buildpacks.buildpack.layers, set to a JSON object representing the [buildpack layers](#layers)
+
+If the builder contains an optional [lifecycle descriptor file][lifecycle-descriptor-rfc], it also requires:
+* The image config's Labels field has the label io.buildpacks.lifecycle.version, set to the lifecycle version
+* The image config's Labels field has the label io.buildpacks.lifecycle.apis, set to a JSON object representing the [lifecycle apis](#lifecycle-apis)
```
jromero, line 137: I'm trying to comprehend how a platform would be able to use a lifecycle without know it's API information given that this wording makes all that information optional. Was that discussed somewhere?

dfreilich: Good catch. I added these requirements in, after @ekcasey reviewed it [here](https://github.com/buildpacks/rfcs/pull/116#discussion_r489611119). It sounds from the [lifecycle-descriptor rfc](https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md#backwards-compatibility) that the the metadata for older versions was contained in the `io.buildpacks.builder.metadata` label:
> Currently lifecycle API and verison information is contained in the io.buildpacks.builder.metadata label. 

But, I removed that from the `io.buildpacks.builder.metadata` label after @ekcasey noted that we shouldn't have the old style in the new requirements. I'll update that to reflect that we assume those labels will be present.

sclevine:
FCP closing on 11/4

PR #120, opened 2020-10-23
RFC: add stacks.base_image label

dumez-k:


text/0000-base-image-label.md
```diff
@@ -0,0 +1,73 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.
+
+- Define the target persona: stack user, stack author
+
+The label can be **optionally** added by stack maintainers to stack image:
+ 
+`io.buildpacks.stack.base_image:` SHA of stack base image
```
ekcasey, line 31: which `SHA`? Is this the [image ID](https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md#imageid) or the manifest digest?

text/0000-base-image-label.md
```diff
@@ -0,0 +1,73 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This is the technical portion of the RFC, where you explain the design in sufficient detail.
+
+The `base_image` label will be added as a `Label` in the stack's Dockerfile.
+Example:
+
+`LABEL io.buildpacks.stack.base_image=e3c1c2659c2468ea8ac8058c501293882b0a55622840eee53525f90b59d31a33`
```
ekcasey, line 42: Without the registry and repository, this doesn't seem like enough information for a user to actually find the image.

text/0000-base-image-label.md
```diff
@@ -0,0 +1,73 @@
...
+# Summary
+[summary]: #summary
+
+Building off of this earlier [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0050-stack-metadata.md) outlining stack metadata,
+we would like to add a label to expose the stack base image SHA in the metadata of all stack-cnb images.
+
+# Motivation
+[motivation]: #motivation
+
+- Allows stack users and authors to easily see which base-image that each stack cnb image is based off of.
```
ekcasey, line 20: I am assuming that by base-image we mean something like `ubuntu:bionic`?

nebhale (approved):

text/0000-base-image-label.md
dumez-k: image ID as defined [here](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) but I think this is the same concept as what you're referring to

text/0000-base-image-label.md
dumez-k: yeah this is definitely a good point, we should change this to the fully qualified path

text/0000-base-image-label.md
dumez-k: No actually this means the base image of the image that we provide AKA  `paketobuildpacks/build:full`

text/0000-base-image-label.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `base_image` label will be added as a `Label` in the stack's Dockerfile.
+Example:
+
+`LABEL io.buildpacks.stack.base_image=paketobuildpacks/build@sha256:e3c1c2659c2468ea8ac8058c501293882b0a55622840eee53525f90b59d31a33`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One downside is it adds slightly more output to `inspect-image` commands which some users may find extraneous.
```
ekcasey, line 47: How do you expect this to be reported by `pack inspect-image`? Currently this output include a base image, which is the build-image (the build-image is the base of the app image)

e.g.
```
Base Image:
  Reference: 523e5e0ad0892f136eebebcaaa6fd60cd744d9b449ade675c3d2a4cad9c4de3e
  Top Layer: sha256:83c184a138eb51d80858fed5f484dbf0cf62aff327ba4c9364f5c81eb5bd0303
```

text/0000-base-image-label.md
```diff
@@ -0,0 +1,73 @@
...
+[motivation]: #motivation
+
+- Allows stack users and authors to easily see which base-image that each stack cnb image is based off of.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.
+
+- Define the target persona: stack user, stack author
```
ekcasey, line 27: I can see why this is useful to a stack author, but when/why would a stack consumer care about this information?

ekcasey (commented):
I have some misgivings about this because stack base image isn't an established domain concept. For example, I [wasn't entirely sure what we were referring to](https://github.com/buildpacks/rfcs/pull/120/files#r513633096).

I wonder if this applies only to stack authors who are building stack images in a specific way (creating an intermediate base image from which stack images are built). In which case, maybe this doesn't need to be a specified label? There is nothing currently stopping stack authors from adding any arbitrary label (e.g. `io.paketo.base_image`) to the image.

text/0000-base-image-label.md
dumez-k: Really good point that the "stack base image" isn't an established domain concept. Completely agree and seeing as we can as you pointed out add arbitrary labels anyway, going to go ahead and close this RFC. 

Thanks @ekcasey !

PR #123, opened 2020-11-06
 RFC: Create stackify repo

martyspiewak:
[Readable](https://github.com/buildpacks/rfcs/blob/b22decf895459a6c229f2dbb783f83a95d33a7a4/text/0000-create-stackify-repo.md)

### DIFF
--- a/text/0000-create-stackify-repo.md
+++ b/text/0000-create-stackify-repo.md
@@ -47,21 +47,21 @@ Stackify will be run once for each image (build and run):
 ```
 stackify 
   --base-image string
-    	Source image ref
+      Source image ref
   --destination string
-    	Destination for stack image
+      Destination for stack image
   --stack-id string
       Stack ID to add to image metadata
   --certs-dir string
-    	Path to directory with CA Certs to add to image
+      Path to directory with CA Certs to add to image
   --mixin-labels string
-    	Comma-separated list of mixins to add to labels
+      Comma-separated list of mixins to add to labels
   --packages string
-    	Comma-separated list of packages to add to image
+      Comma-separated list of packages to add to image
   --user-id string
-    	CNB User ID
+      CNB User ID
   --group-id string
-    	CNB Group ID
+      CNB Group ID
   --build
       Add build image metadata (must specify either --build or --run)
   --run
--- a/text/0000-create-stackify-repo.md
+++ b/text/0000-create-stackify-repo.md
@@ -29,19 +29,103 @@ Stackify will be a tool that will be useful to users who would like to customize
 
 Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
 
+Stackify can be integrated into Pack in the form of a `pack create-stack` command.
+
 # How it Works
 [how-it-works]: #how-it-works
+For stacks defined by the buildpacks project, Stackify will add mixins labels for each package installed on the image. If Stackify is used to give add a stack ID other than one defined by the buildpacks project, mixin labels will not be added.
+
+If Stackify is used with an already valid stack image, the `CNB_*` labels and users will not be changed unless new user and group ID's are passed, in which case the labels and user will be overwritten.
+
+If run on the image being "stackify-ed", Stackify can be used to add packages and ca-certificates.
+
 Stackify will use Kaniko to create additional image layers with the additional packages and ca-certificates that users choose to add to their base image. It will also use GGCR to manipulate the image metadata.
+
+
+### Proposed Interface
+Stackify will be run once for each image (build and run):
+```
+stackify 
+  --base-image string
+    	Source image ref
+  --stack-image string
+    	Destination for stack image
+  --certs-dir string
+    	Path to directory with CA Certs to add to image
+  --group-id string
+    	CNB Group ID
+  --image-type string
+    	Type of image to create (build or run)
+  --mixins string
+    	Comma-separated list of mixins to add to labels
+  --packages string
+    	Comma-separated list of packages to add to image
+  --stack-id string
+      Stack ID to add to image metadata
+  --user-id string
+    	CNB User ID
+```
+
+This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
+
+`pack create-stack my-stack -c /path/to/config.toml`
+
+Config file:
+```
+[build]
+image = "paketobuildpacks/build:base"
+outputTag = "some-registry/some-custom-stack-build-image"
+
+[run]
+image = "paketobuildpacks/run:base"
+outputTag = "some-registry/some-custom-stack-run-image"
+
+[caCerts]
+
+[caCerts.build]
+path = "/path/to/dir/with/build-certs"
+
+[caCerts.run]
+path = "/path/to/dir/with/run-certs"
+
+[[packages]]
+name = "cowsay"
+
+[[packages]]
+name = "fortune"
+phase = "build"
+
+[[packages]]
+name = "rolldice"
+phase = "run"
+
+[[mixins]]
+name = "set=build-utils"
+phase = "build"
+
+[[mixins]]
+name = "set=run-utils"
+phase = "run"
+
+[[mixins]]
+name = "set=shared-utils"
+
+[user]
+userID = 1000
+groupID = 1000
+```
 
 # Drawbacks
 [drawbacks]: #drawbacks
+* Stackify must be run on the image being "stackify-ed" in order to add ca-certificates and packages.
+  * Pack have to run stackify within the image that is being modified.
 * If we do not do enough validation, users may be able to successfully use stackify with incompatible or invalid base images.
 * Some of the CNB metadata can be difficult to generate, particularly the mixins
 
 # Alternatives
 [alternatives]: #alternatives
 * Without this tool users can still use Dockerfiles to extend base images and add the CNB metadata.
-    * However, this requires them to be aware of all of the relevant metadata needed on the image and maintain a Dockerfile and pipelines to continuously update their images. 
+    * However, this requires them to be aware of all of the relevant metadata needed on the image and maintain a Dockerfile to continuously update their images. 
 
 # Prior Art
 [prior-art]: #prior-art
@@ -51,7 +135,6 @@ N/A
 [unresolved-questions]: #unresolved-questions
 * How much validation should be done on the images, if any?
 * It will be challenging to generate a list of common mixins between the build and run images. Is it ok that we will just have `build:<mixin>` and `run:<mixin>` labels and no shared mixins?
-* Can this later be integrated into pack, perhaps in the form of a new command (e.g. `pack create-stack`)?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-create-stackify-repo.md
+++ b/text/0000-create-stackify-repo.md
@@ -29,11 +29,11 @@ Stackify will be a tool that will be useful to users who would like to customize
 
 Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
 
-Stackify can be integrated into Pack in the form of a `pack create-stack` command.
+Stackify can be integrated into Pack in the form of a `pack stack create` command.
 
 # How it Works
 [how-it-works]: #how-it-works
-For stacks defined by the buildpacks project, Stackify will add mixins labels for each package installed on the image. If Stackify is used to give add a stack ID other than one defined by the buildpacks project, mixin labels will not be added.
+For stacks defined by the buildpacks project, Stackify will add mixin labels for each package installed on the image. If Stackify is used to add a stack ID other than one defined by the buildpacks project, mixin labels will not be added.
 
 If Stackify is used with an already valid stack image, the `CNB_*` labels and users will not be changed unless new user and group ID's are passed, in which case the labels and user will be overwritten.
 
@@ -48,66 +48,84 @@ Stackify will be run once for each image (build and run):
 stackify 
   --base-image string
     	Source image ref
-  --stack-image string
+  --destination string
     	Destination for stack image
+  --stack-id string
+      Stack ID to add to image metadata
   --certs-dir string
     	Path to directory with CA Certs to add to image
-  --group-id string
-    	CNB Group ID
-  --image-type string
-    	Type of image to create (build or run)
-  --mixins string
+  --mixin-labels string
     	Comma-separated list of mixins to add to labels
   --packages string
     	Comma-separated list of packages to add to image
-  --stack-id string
-      Stack ID to add to image metadata
   --user-id string
     	CNB User ID
+  --group-id string
+    	CNB Group ID
+  --build
+      Add build image metadata (must specify either --build or --run)
+  --run
+      Add run image metadata (must specify either --build or --run)
 ```
 
-This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
+This will be called under the hood by the `pack stack create` command, which will take a TOML config file. Pack will have to copy the stackify image onto each base image and then run stackify with the specified inputs for each image.
 
-`pack create-stack my-stack -c /path/to/config.toml`
+This command can be invoked by passing in `--run-image`, `--build-image`, or both, depending on which images you would like to have created.
+
+`pack stack create my-stack --run-image <run-tag> --build-image <build-tag> -c /path/to/config.toml`
 
 Config file:
 ```
+[stack]
+id = "(required)"
+maintainer = "(optional)"
+homepage = "(optional)"
+description = "(optional)"
+
+[stack.distro]
+name = "(optional)"
+version = "(optional)"
+
+[stack.metadata]
+# arbitrary keys and values
+
 [build]
-image = "paketobuildpacks/build:base"
-outputTag = "some-registry/some-custom-stack-build-image"
+base-image = "paketobuildpacks/build:base"
 
-[run]
-image = "paketobuildpacks/run:base"
-outputTag = "some-registry/some-custom-stack-run-image"
+[[build.ca-certs]]
+uri = "/path/to/dir/with/build-certs"
 
-[caCerts]
+[[build.ca-certs]]
+uri = "file:///path/to/ca-cert.pem"
 
-[caCerts.build]
-path = "/path/to/dir/with/build-certs"
+[[build.ca-certs]]
+uri = "https://example.com/ca-cert.pem"
 
-[caCerts.run]
-path = "/path/to/dir/with/run-certs"
-
-[[packages]]
+[[build.packages]]
 name = "cowsay"
 
-[[packages]]
+[[build.packages]]
 name = "fortune"
-phase = "build"
 
-[[packages]]
+[run]
+base-image = "paketobuildpacks/run:base"
+
+[[run.ca-certs]]
+path = "/path/to/dir/with/run-certs"
+
+[[run.packages]]
+name = "cowsay"
+
+[[run.packages]]
 name = "rolldice"
-phase = "run"
 
-[[mixins]]
-name = "set=build-utils"
-phase = "build"
+[[mixin-labels]]
+name = "build:set=build-utils"
 
-[[mixins]]
-name = "set=run-utils"
-phase = "run"
+[[mixin-labels]]
+name = "run:set=run-utils"
 
-[[mixins]]
+[[mixin-labels]]
 name = "set=shared-utils"
 
 [user]
@@ -118,7 +136,7 @@ groupID = 1000
 # Drawbacks
 [drawbacks]: #drawbacks
 * Stackify must be run on the image being "stackify-ed" in order to add ca-certificates and packages.
-  * Pack have to run stackify within the image that is being modified.
+  * Pack has to run stackify within the image that is being modified.
 * If we do not do enough validation, users may be able to successfully use stackify with incompatible or invalid base images.
 * Some of the CNB metadata can be difficult to generate, particularly the mixins
 
--- a/text/0000-create-stackify-repo.md
+++ b/text/0000-create-stackify-repo.md
@@ -1,0 +1,58 @@
+# Meta
+[meta]: #meta
+- Name: Create stackify repo
+- Start Date: 2020-11-06
+- Authors: @martyspiewak @mdelillo @dumez-k
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+Create a new repo called “stackify” which can be used to turn OCI images into CNB compliant stack images.
+
+# Definitions
+[definitions]: #definitions
+N/A
+
+# Motivation
+[motivation]: #motivation
+Currently, for users to extend the published stack images, or turn their own images into stack images, they have to create Dockerfiles which take the base images, apply their changes, and then set the CNB metadata. It can be difficult to generate and set all of the relevant metadata, especially the mixins.
+
+Stackify will be a tool that they can use to perform all of these manipulations to their base image without Dockerfiles.
+
+
+# What it is
+[what-it-is]: #what-it-is
+Stackify will be a tool that will be useful to users who would like to customize the published stack images or turn their own base images into CNB compliant stack images.
+
+Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
+
+# How it Works
+[how-it-works]: #how-it-works
+Stackify will use Kaniko to create additional image layers with the additional packages and ca-certificates that users choose to add to their base image. It will also use GGCR to manipulate the image metadata.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+* If we do not do enough validation, users may be able to successfully use stackify with incompatible or invalid base images.
+* Some of the CNB metadata can be difficult to generate, particularly the mixins
+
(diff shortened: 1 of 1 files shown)


text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+Stackify will be a tool that will be useful to users who would like to customize the published stack images or turn their own base images into CNB compliant stack images.
+
+Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
+
+# How it Works
+[how-it-works]: #how-it-works
+Stackify will use Kaniko to create additional image layers with the additional packages and ca-certificates that users choose to add to their base image. It will also use GGCR to manipulate the image metadata.
```
ekcasey, line 34: Is this tool intended to be specific to the `io.buildpacks.stacks.bionic` stack. This is the only stack defined by project and therefor the only stack where we can translate image diff -> mixin name.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Motivation
+[motivation]: #motivation
+Currently, for users to extend the published stack images, or turn their own images into stack images, they have to create Dockerfiles which take the base images, apply their changes, and then set the CNB metadata. It can be difficult to generate and set all of the relevant metadata, especially the mixins.
+
+Stackify will be a tool that they can use to perform all of these manipulations to their base image without Dockerfiles.
+
+
+# What it is
+[what-it-is]: #what-it-is
```
ekcasey, line 27: It would be nice to have a proposed user interface in this RFC.

jkutner: Specifically, I'd like to see the interface for the "customize" part. Do we want to define a new TOML file that can include `mixins`, labels, etc?

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+* How much validation should be done on the images, if any?
+* It will be challenging to generate a list of common mixins between the build and run images. Is it ok that we will just have `build:<mixin>` and `run:<mixin>` labels and no shared mixins?
+* Can this later be integrated into pack, perhaps in the form of a new command (e.g. `pack create-stack`)?
```
jkutner, line 54: I'd like to see `pack create-stack` as part of this proposal.

nebhale (approved):

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+* If we do not do enough validation, users may be able to successfully use stackify with incompatible or invalid base images.
+* Some of the CNB metadata can be difficult to generate, particularly the mixins
+
+# Alternatives
+[alternatives]: #alternatives
+* Without this tool users can still use Dockerfiles to extend base images and add the CNB metadata.
+    * However, this requires them to be aware of all of the relevant metadata needed on the image and maintain a Dockerfile and pipelines to continuously update their images.
```
jabrown85, line 44: I think we said the "pipelines to continuously update their images" will remain regardless of stackify.

nebhale (approved):

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+Stackify will use Kaniko to create additional image layers with the additional packages and ca-certificates that users choose to add to their base image. It will also use GGCR to manipulate the image metadata.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+* If we do not do enough validation, users may be able to successfully use stackify with incompatible or invalid base images.
+* Some of the CNB metadata can be difficult to generate, particularly the mixins
+
```
jabrown85, line 40: I think we should elaborate on what would happen if you passed an existing builder to this new program. If the image already has `CNB_*` labels and users, what would the expected output be? Should you be able to `stackify heroku/buildpacks:18 org/mynewstack`?

text/0000-create-stackify-repo.md
martyspiewak: Addressed in latest update. Thanks!

text/0000-create-stackify-repo.md
martyspiewak: Addressed in latest update. Thanks!

text/0000-create-stackify-repo.md
martyspiewak: Removed- thanks!

text/0000-create-stackify-repo.md
martyspiewak: Addressed in latest update. Thanks!

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+```
+[build]
+image = "paketobuildpacks/build:base"
+outputTag = "some-registry/some-custom-stack-build-image"
+
+[run]
+image = "paketobuildpacks/run:base"
+outputTag = "some-registry/some-custom-stack-run-image"
+
+[caCerts]
```
jkutner, line 83: Using camel-case feels a little odd here to me. But there isn't much prior art in in our other TOML files. I'm ok if we decide to standardize on camel-case, but we should probably be explicit about the decision.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+    	Source image ref
+  --stack-image string
+    	Destination for stack image
+  --certs-dir string
+    	Path to directory with CA Certs to add to image
+  --group-id string
+    	CNB Group ID
+  --image-type string
+    	Type of image to create (build or run)
+  --mixins string
```
jkutner, line 59: But this will _not_ actually try to install any of these mixins right? I wonder if we should call the flag `--mixin-labels` or something.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+path = "/path/to/dir/with/build-certs"
+
+[caCerts.run]
+path = "/path/to/dir/with/run-certs"
+
+[[packages]]
+name = "cowsay"
+
+[[packages]]
+name = "fortune"
```
jabrown85, line 95: I'd prefer `build:fortune` just to match what we do elsewhere.

jabrown85 (approved):

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+image = "paketobuildpacks/build:base"
+outputTag = "some-registry/some-custom-stack-build-image"
+
+[run]
+image = "paketobuildpacks/run:base"
+outputTag = "some-registry/some-custom-stack-run-image"
+
+[caCerts]
+
+[caCerts.build]
```
jabrown85, line 85:
```suggestion
[build.ca-certs]
```

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+path = "/path/to/dir/with/build-certs"
+
+[caCerts.run]
+path = "/path/to/dir/with/run-certs"
+
+[[packages]]
```
jabrown85, line 91: What do we think about matching `project.toml` type of syntax?
```suggestion
[[build.packages]]

[[run.packages]]

[[packages]]
```

jabrown85 (approved):

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+`pack create-stack my-stack -c /path/to/config.toml`
+
+Config file:
+```
+[build]
+image = "paketobuildpacks/build:base"
+outputTag = "some-registry/some-custom-stack-build-image"
+
+[run]
+image = "paketobuildpacks/run:base"
```
ekcasey, line 80: I know this is just an example and not the point but could the base image be something as simple as `ubuntu:bionic`? I think that would be ideal. Or, are you imagining that there is still a step that need to come before this where a Dockerfile is used to generate a "base image"?

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+outputTag = "some-registry/some-custom-stack-run-image"
+
+[caCerts]
+
+[caCerts.build]
+path = "/path/to/dir/with/build-certs"
```
ekcasey, line 86:
```suggestion
[[build.ca-certs]]
uri = "path/to/dir/with/build-certs"

[[build.ca-certs]]
uri = "file:///path/to/ca-cert.pem"

[[build.ca-certs]]
uri = "https://example.com/ca-cert.pem"
```
I don't see any reason not to allow `https` uris in additional to file paths. Also accepting just a single file in addition to directory containing multiple certsmight make things more flexible for end users.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+    	Comma-separated list of mixins to add to labels
+  --packages string
+    	Comma-separated list of packages to add to image
+  --stack-id string
+      Stack ID to add to image metadata
+  --user-id string
+    	CNB User ID
+```
+
+This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
```
ekcasey, line 69: Can we be more explicit about how this works. I am assuming `pack` would copy `stackify` into the image and then run it?

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+    	Destination for stack image
+  --certs-dir string
+    	Path to directory with CA Certs to add to image
+  --group-id string
+    	CNB Group ID
+  --image-type string
+    	Type of image to create (build or run)
+  --mixins string
+    	Comma-separated list of mixins to add to labels
+  --packages string
```
ekcasey, line 61: What exactly is supported here?

I am assuming `stackify` won't, on day one, know about all of the package managers in every linux distro (or the location where CA certs should be installed on the filesystem). Will we only support specified stack IDs? Not care about stack IDs but only support ubuntu?

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+
+This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
+
+`pack create-stack my-stack -c /path/to/config.toml`
+
+Config file:
+```
+[build]
+image = "paketobuildpacks/build:base"
+outputTag = "some-registry/some-custom-stack-build-image"
```
ekcasey, line 77: For consistency with `create-builder` and `package-buildpack` I would prefer that users must pass the destination tags on the command line.

For example a user could run `pack create-stack` with `--run-image <tag>` `--build-image <tag>` or both. This has an added benefit in that users may not always want to build both images. If for example, I am rebuilding a build image because of a vulnerability in a package that exists only on the run image, I may not want to wait for an identical copy of the run image to build.

text/0000-create-stackify-repo.md
hone: @ekcasey do you envision supporting custom packages or that's out of scope, i.e. roll your own.

ekcasey: @hone My instinct is that would be out of scope for this RFC. But my comment above wasn't trying to suggest a different set of supported stacks/packages, but rather, trying to understand more concretely what is being proposed.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+Stackify will be a tool that they can use to perform all of these manipulations to their base image without Dockerfiles.
+
+
+# What it is
+[what-it-is]: #what-it-is
+Stackify will be a tool that will be useful to users who would like to customize the published stack images or turn their own base images into CNB compliant stack images.
+
+Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
+
+Stackify can be integrated into Pack in the form of a `pack create-stack` command.
```
dfreilich, line 32: Given the new `stack` subcommand, and https://github.com/buildpacks/rfcs/pull/93, can we change this to `pack stack create`?

text/0000-create-stackify-repo.md
dfreilich: @ekcasey I agree with the second point (creating a functionality for users to recreate either the build or run image, but not always both), but I'm not sure I feel the first point. Since users are already going to be passing in a config file, I'm not sure I get the utility of having them also pass information in through flags.

text/0000-create-stackify-repo.md
dfreilich: I agree, camel-case feels odd in configuration files. It's also the case with `outputTag`.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+
+[user]
+userID = 1000
+groupID = 1000
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+* Stackify must be run on the image being "stackify-ed" in order to add ca-certificates and packages.
+  * Pack have to run stackify within the image that is being modified.
```
dfreilich, line 121:
```suggestion
  * Pack has to run stackify within the image that is being modified.
```

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+### Proposed Interface
+Stackify will be run once for each image (build and run):
+```
+stackify 
+  --base-image string
+    	Source image ref
+  --stack-image string
+    	Destination for stack image
```
natalieparellano, line 52: I struggled with this for a moment because to me there is no such thing as a "stack" image, but rather a build + run image grouping. I noticed you also have `--image-type` flag - would it be possible to replace stack image with `--run-image` or `--build-image`, and the tool could infer the type based on which flag was used? Just a thought.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+name = "set=shared-utils"
+
+[user]
+userID = 1000
+groupID = 1000
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+* Stackify must be run on the image being "stackify-ed" in order to add ca-certificates and packages.
```
dfreilich, line 120: To this point: are there any requirements on the image, in order for stackify to be run within it? Can it be run on a totally scratch image? A windows container?

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+# Motivation
+[motivation]: #motivation
+Currently, for users to extend the published stack images, or turn their own images into stack images, they have to create Dockerfiles which take the base images, apply their changes, and then set the CNB metadata. It can be difficult to generate and set all of the relevant metadata, especially the mixins.
+
+Stackify will be a tool that they can use to perform all of these manipulations to their base image without Dockerfiles.
+
+
+# What it is
+[what-it-is]: #what-it-is
+Stackify will be a tool that will be useful to users who would like to customize the published stack images or turn their own base images into CNB compliant stack images.
```
dfreilich, line 28: This sounds like stackify is something new that is being created, rather than an existing tool/repo being donated to the project. Are we trying to keep that tone throughout the RFC, or is there any objection to explicitly linking to the current project?

dfreilich (commented):
In general, I'm super happy to see this happen, I think it definitely adds a lot of value to the project. Thanks for driving this!

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+
+Users can use stackify to add packages and ca-certificates to a base image as well as add all required CNB metadata to the image.
+
+Stackify can be integrated into Pack in the form of a `pack create-stack` command.
+
+# How it Works
+[how-it-works]: #how-it-works
+For stacks defined by the buildpacks project, Stackify will add mixins labels for each package installed on the image. If Stackify is used to give add a stack ID other than one defined by the buildpacks project, mixin labels will not be added.
```
hone, line 36:
```suggestion
For stacks defined by the buildpacks project, Stackify will add mixins labels for each package installed on the image. If Stackify is used to add a stack ID other than one defined by the buildpacks project, mixin labels will not be added.
```

hone:
I'm :+1: on the idea, but I'd like to see the outstanding comments addressed.

text/0000-create-stackify-repo.md
martyspiewak: I think the plan on day one would be to support adding packages to any Bionic based image (by checking `/etc/os-release`) and then expanding support as the Buildpacks project creates new stack IDs.

text/0000-create-stackify-repo.md
martyspiewak: I agree that we should allow building just one of the stack images. For consistency with `create-builder` and `package-buildpack` we'd be ok passing those tags as flags.

text/0000-create-stackify-repo.md
martyspiewak: Perhaps we can change `--stack-image` to `--destination` or `--image-destination` and we can remove the `--image-type` flag and pass either `--build` or `--run`?

text/0000-create-stackify-repo.md
martyspiewak: That makes sense.

text/0000-create-stackify-repo.md
martyspiewak: On day one, it would only support bionic images but we would hope to expand that later on.

text/0000-create-stackify-repo.md
martyspiewak: We have a first implementation of this that we've worked on but not to all of the specifications in this RFC. It is also currently in a private repo so we couldn't link to it.

text/0000-create-stackify-repo.md
martyspiewak: This makes sense. We'll change that.

text/0000-create-stackify-repo.md
martyspiewak: I think changing to `--mixin-labels` makes sense, but it would still make sense to keep that separate from `packages`. This would allow users who are using a custom stack ID to make use of both of these features.

text/0000-create-stackify-repo.md
martyspiewak: With the `[build.packages]` change this doesn't seem like it would be necessary anymore.

text/0000-create-stackify-repo.md
martyspiewak: This makes sense- we'll change to that.

text/0000-create-stackify-repo.md
martyspiewak: This makes sense- we'll change to that.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+    	CNB User ID
+```
+
+This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
+
+`pack create-stack my-stack -c /path/to/config.toml`
+
+Config file:
+```
+[build]
```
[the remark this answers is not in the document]
martyspiewak: This seems like it would be worthwhile to add. It seems like the `released` date should be something stackify can set with the time stackify was called.

I'm not sure what we should do about that second question- seems like something the project as a whole needs to resolve.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+name = "fortune"
+phase = "build"
+
+[[packages]]
+name = "rolldice"
+phase = "run"
+
+[[mixins]]
+name = "set=build-utils"
+phase = "build"
```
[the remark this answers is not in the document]
martyspiewak: I think given the suggest to make a top level `[build]` and `[run]` section, we can have a third top level section `[mixin-labels]` and there we can have users pass them in with the `build:` and `run:` prefixes as you suggested.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+This will be called under the hood by the `pack create-stack` command, which will take a TOML config file:
+
+`pack create-stack my-stack -c /path/to/config.toml`
+
+Config file:
+```
+[build]
+image = "paketobuildpacks/build:base"
```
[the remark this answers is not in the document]
martyspiewak: This makes sense- we'll change this.

text/0000-create-stackify-repo.md
martyspiewak: We've spiked on this and in our current implementation you cannot start with `ubuntu:bionic` because it requires `ca-certificates` to be present to pull down the image manifest. That said, though, we would like to support starting from `ubuntu:bionic`.

text/0000-create-stackify-repo.md
martyspiewak: This seems valuable to add.

text/0000-create-stackify-repo.md
martyspiewak: Yes that is how it would work. We'll be more explicit about that.

text/0000-create-stackify-repo.md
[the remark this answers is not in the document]
ekcasey: Maybe I am overthinking the label name issue. Let not worry about that.

Label name aside, maybe it would make the most sense if we could set these values either in a top level `[stack]` table (apply to both build and run images) or set any of these fields besides `id` (which must be the same for both images) in the `[build]` or `[run]` tables, to allow different values to be set on build and run images.

text/0000-create-stackify-repo.md
ekcasey: When and why does `stackify` pull down the image manifest. When this become a `pack create-stack` command, is this something `pack` could do on the host rather than something `stackify` must do inside the container?

text/0000-create-stackify-repo.md
ekcasey: @dfreilich If we wanted to allow users to specify the destination tag in the file, I think we should do that for `pack builder create` and `pack buildpack package` as well. I think that discussion should be separate from this RFC (in a `pack` issue?) and we should make the change everywhere simultaneously to keep things as consistent and predictable as possible for users.

I.E I think this RFC only needs to include usage that is consistent with the current `pack` conventions. If we change the `pack` conventions then either:
1. This RFC has already been implemented and we change the usage at the same time we change that of the other commands
2. This RFC has not yet been implemented, but it is okay that the usage in the RFC is slightly different than what gets implemented due to `pack` changes in the interim.

text/0000-create-stackify-repo.md
martyspiewak: It pulls down the manifest in order to manipulate the image metadata. I think this could probably be done by `pack`, yes.

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,159 @@
...
+  --certs-dir string
+      Path to directory with CA Certs to add to image
+  --mixin-labels string
+      Comma-separated list of mixins to add to labels
+  --packages string
+      Comma-separated list of packages to add to image
+  --user-id string
+      CNB User ID
+  --group-id string
+      CNB Group ID
```
samj1912, line 64: I am curious about how stackify would handle the cases when the specified user-id/group-id do not exist on the base image. Would it create them or would it throw an error?

martyspiewak: It would create them.

dfreilich (approved):

text/0000-create-stackify-repo.md
samj1912: Gotcha! I hit some errors in the past when I was trying to use extremely minimal base images as runtime images as they didn't have useradd or even /bin/sh. I'm hoping that this would be able to handle such situations by manually creating the required files.

ekcasey (approved):

text/0000-create-stackify-repo.md
```diff
@@ -0,0 +1,159 @@
...
+    * However, this requires them to be aware of all of the relevant metadata needed on the image and maintain a Dockerfile to continuously update their images. 
+
+# Prior Art
+[prior-art]: #prior-art
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+* How much validation should be done on the images, if any?
+* It will be challenging to generate a list of common mixins between the build and run images. Is it ok that we will just have `build:<mixin>` and `run:<mixin>` labels and no shared mixins?
```
jromero, line 155: If I'm not mistaken, this an implementation detail/challenge and not a limitation of the current proposed schema (`mixin-labels`). I'm not sure what the challenge is but it does seem to be a concern that may be addressed separately as long as the expectation (aka contract is adhered to).

jromero (approved):

text/0000-create-stackify-repo.md
martyspiewak: For stacks that are defined by the buildpacks project (i.e. `io.buildpacks.stacks.bionic`), stackify can generate most of the mixins automatically since for the bionic stack, mixins (other than set mixins) correspond to packages on the image. Therefore, users would only specifcy `mixin-labels` on a bionic stack if they want to add set mixins to the image metadata.

The challenge is that since stackify is run separately on each image, it can't itself know which mixins are on both images and which are on only one image. Therefore, it will not be able to generate mixin labels like `apt` on both images, but will instead have to put `build:apt` on the build image and `run:apt` on the run image.

Perhaps this is a concern that can be addressed separately, just wanted to clarify the issue.

hone (approved):

PR #124, opened 2020-11-13
merge-rfc issues usage

ekcasey:
These improvements prevent maintainers from accidentally forgetting to add issues when merging an RFC. Also, auto-generate links from copy-pastable issues references to make this script easier to use.

* Adds -n flag to explicitly indicate that no issues should be linked
* Require either -i or -n flag
* Accept issues in the format <org>/<repo>#<number> and create links

Signed-off-by: Emily Casey <ecasey@vmware.com>


nebhale (approved):
Awesome-sauce.  Sorry for being so dumb that you needed to do this!

PR #125, opened 2020-12-01
RFC for buildpack registry search api

elbandito:


jkutner (approved):

sclevine (approved):

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  ```
+  ```json
+  {
+      "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
+      "license": "MIT",
+      "ns":"projectriff",
+      "name":"command-function",
+      "latest_version": "1.4.1",
+      "versions": "/buildpacks/1/versions",
+      "weekly_downloads": 8993,
```
ekcasey, line 99: Where will we get `weekly_downloads` from? Dockerhub specifically might expose this but I don't think this information is available in the standard [registry API](https://github.com/opencontainers/distribution-spec). Even in Dockerhub I believe `pull_count` is only [available for a repository](https://hub.docker.com/v2/repositories/buildpacksio/lifecycle/), not a specific image.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "issues": 44,
```
ekcasey, line 100: What type of `issues`? Github issues? How would we get this?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+- **GET /buildpacks/:id**
+
+  Retrieves metadata for a specific buildpack.  This will contain more detailed data.
+  ```
+  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks/1
+  ```
+  ```json
+  {
+      "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
+      "license": "MIT",
```
ekcasey, line 94: `description` and `license` might be different on different buildpack images (i.e. different versions of the same ID).

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f"
+  }
+  ```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In the initial implementation, we can take advantage of the existing [registry](https://github.com/buildpacks/registry-index) repository to pull buildpack data from.  Pack already performs a clone/pull on this repo for its registry cache.  To keep ensure up-to-date buildpack data, we'd have a polling loop that performs a `git pull`.  This polling would take place in a separate background process.  
+
+Each time the local repository has been updated via `git pull`, buildapack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
```
ekcasey, line 110: It might be wise to be more vague here to leave room for iteration on the implementation without this RFC becoming out of date. We really only need to say that each time the repository is updated with `git pull` a searchable representation of the data is generated.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "addr":"gcr.io/projectriff/streaming-http-adapter@sha256:b202d9ec203e882ee7e3c599d9e867617f909c8b6123e4ce942af47db6e58c45"
+    }
+  ]
+  ```
+
+- **GET /buildpacks/:id**
+
+  Retrieves metadata for a specific buildpack.  This will contain more detailed data.
+  ```
+  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks/1
```
ekcasey, line 89:
```suggestion
  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks/projectriff/command-function
```
Or are we associating numerical IDs with each buildpack?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  ```
+  ```json
+  {
+      "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
```
ekcasey, line 93: For fields like `description` that are not currently in the `registry-index` I think we should probably either
a) add them to the index
b) not return them from the search API
in order to avoid problem like dockerhub rate limiting.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  
+    We already have buildpack metadata stored in the [registry](https://github.com/buildpacks/registry-index).  
+    Customers can simply explore this repository for buildpack information.  Perhaps someone else in the community will
+    come up with their own solution that can be donated to CNB.
+
+  - Having a separate, independent service responsible for:
+    1. pulling from the registry-index repository
+    2. augmenting buildpack data
+    3. persisting it to a hosted PostgresDB  
+
```
ekcasey, line 142: Another option: we could implement search purely on the client side (with `pack` for example), instead of providing a service. This would have the benefit of searching across all user-configured registries not just https://github.com/buildpacks/registry-index.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+- **GET /buildpacks**
+  
+  Retrives all the buildpacks in the registry
+
+- **GET /buildpacks?query=text**
+
+  Retrievs all the buildpacks that statisfy the search query.  For example:
+  ```
+  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks?query=projectriff
```
ekcasey, line 53: Is query intended to query fields like `description`? or just the fields that are currently populated in the index?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "name":"command-function",
+      "version":"1.4.1",
+      "yanked":false,
+      "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f"
+    },
+    {
+      "ns":"projectriff",
+      "name":"java-function",
+      "version":"1.4.1",
+      "yanked":false,"addr":"gcr.io/projectriff/java-function@sha256:5eabea8f7b2c09074ec196fe0c321006fb5ad8f282cc918520286d8a0007196f"},
```
ekcasey, line 68:
```suggestion
      "yanked":false,
      "addr":"gcr.io/projectriff/java-function@sha256:5eabea8f7b2c09074ec196fe0c321006fb5ad8f282cc918520286d8a0007196f"}
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+    {
+      "ns":"projectriff",
+      "name":"node-function",
+      "version":"1.4.1",
+      "yanked":false,
+      "addr":"gcr.io/projectriff/node-function@sha256:194298b826c15bb079c59aed99968d7678a6e1f7a882c9d7f61811e0990717ba"},
```
ekcasey, line 74:
```suggestion
      "addr":"gcr.io/projectriff/node-function@sha256:194298b826c15bb079c59aed99968d7678a6e1f7a882c9d7f61811e0990717ba"}
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks/1
+  ```
+  ```json
+  {
+      "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
+      "license": "MIT",
+      "ns":"projectriff",
+      "name":"command-function",
+      "latest_version": "1.4.1",
+      "versions": "/buildpacks/1/versions",
```
jromero, line 98: This "versions" endpoint isn't detailed in this RFC. I'm curious as to how that would look. I have a few concerns about it's contents in relation to this endpoint dependent on what it looks like based on some of the expected use cases from the frontend I was designing.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  ```
+  $ curl https://glacial-plateau-95350.herokuapp.com/buildpacks/1
+  ```
+  ```json
+  {
+      "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
+      "license": "MIT",
+      "ns":"projectriff",
+      "name":"command-function",
+      "latest_version": "1.4.1",
```
jromero, line 97: It seems like all of the other content in this object are values of _the_ latest buildpack. Also, is there a need for a "versions" endpoint?

Maybe a structure such as... ?

```json
{
  "latest": {
    "description": "The Command Function Buildpack is a Cloud Native Buildpack V3 that provides riff Command Function Invoker to functions",
    "license": "MIT",
    "ns": "projectriff",
    "name": "command-function",
    "version": "1.4.1",
    "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f"
  },
  "versions": [ // in decending semver order
    "1.4.1",
    "1.3.1"
  ]
}
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "ns":"projectriff",
+      "name":"streaming-http-adapter",
+      "version":"1.4.0",
+      "yanked":false,
+      "addr":"gcr.io/projectriff/streaming-http-adapter@sha256:b202d9ec203e882ee7e3c599d9e867617f909c8b6123e4ce942af47db6e58c45"
+    }
+  ]
+  ```
+
+- **GET /buildpacks/:id**
```
jromero, line 85:
```suggestion
- **GET /buildpacks/:ns/:name**
```

jromero, line 85: I would also expect a `GET /buildpacks/:ns/:name/:version` so a client can get details of a specific version.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "weekly_downloads": 8993,
+      "issues": 44,
+      "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f"
+  }
+  ```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In the initial implementation, we can take advantage of the existing [registry](https://github.com/buildpacks/registry-index) repository to pull buildpack data from.  Pack already performs a clone/pull on this repo for its registry cache.  To keep ensure up-to-date buildpack data, we'd have a polling loop that performs a `git pull`.  This polling would take place in a separate background process.
```
jromero, line 108: I don't see how pack's mention here ties to this RFC/section.

text/0000-buildpack-registry-search-api.md
jromero, line 110:
```suggestion
Each time the local repository has been updated via `git pull`, buildpack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
```

jromero, line 110: What fields will be searched upon? I assume `ns` and `name`. Are there others? It may be worth making it explicit here.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+Each time the local repository has been updated via `git pull`, buildapack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
+
+In addition, a server process will expose endpoints and handle incoming request for the GET endpoints mentioned earlier.  For search requests, the text will be extracted from the `query` query parameter and used to search against the normalized buildpacks index.   Results will be added to the server response, as a list of JSON objects.
+
+*Note:  The Distribution Team maintainers can manage this service, and could even set-up a pager (but only best effort, and during working hours).
```
jromero, line 114: We should probably open it up to verified project contributors as well so there is less stress or demand on just a few and also have a bit more knowledge transfer.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  - Customers will be able to quickly find relevant buildpacks for their use cases.
+
+# What it is
+[what-it-is]: #what-it-is
+
+An external API service that exposes endpoints for retrieving buildpack metadata.  Initially, I envision this service providing a few GET endpoints to support plain text query searches, and individual buildpack data retrieval. 
+
+### Supported Endpoints
+
+- **GET /buildpacks**
```
hone, line 45: We aren't going to be big enough where this will be on the db, but I've noticed that npm did [deprecate their `./-/all` endpoint](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#get-v1search).

If we are going to do this, do we need to look into pagination as well?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  
+  Retrives all the buildpacks in the registry
```
hone, line 47:
```suggestion
  Retrieves all the buildpacks in the registry
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+- **GET /buildpacks?query=text**
```
hone, line 49: I like the REST semantics here. If we get rid of the index should we try to do something like [`/search`](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#get-v1search)?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+      "ns":"projectriff",
+      "name":"node-function",
+      "version":"1.4.1",
+      "yanked":false,
+      "addr":"gcr.io/projectriff/node-function@sha256:194298b826c15bb079c59aed99968d7678a6e1f7a882c9d7f61811e0990717ba"},
+    {
+      "ns":"projectriff",
+      "name":"streaming-http-adapter",
+      "version":"1.4.0",
+      "yanked":false,
```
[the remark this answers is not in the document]
elbandito: @ekcasey initially, yes.  Maybe after some initial customer feedback, we find that customers only care about valid buildpacks, at which point it might make sense to filter them out.   I also wonder if there's a use case for searching explicitly for yanked buildpacks.

text/0000-buildpack-registry-search-api.md
elbandito: I should probably have called out that I embellished this meta-data to illustrate what it _may_ look like.  You're correct, it will depend on what these external services can provide e.g. DockerHub, ECR, GCR, etc.   But I think it would be valuable to try and some level of metrics, once we know what's really possible in future iterations.

text/0000-buildpack-registry-search-api.md
elbandito: Again, some example data that could be useful, but I envision this being Github issues.  I think it would be nice to somehow indicate if there are any outstanding issues associated with a given Buildpack, and even provide a link the Github issues tab for connivence.

text/0000-buildpack-registry-search-api.md
elbandito: Good to know.  So then I would expect a user would see a list of all the versions for a specific buildpack, and some of those version may have a different `description` and `license` field.   I assume this could be helpful as it would shed some light on the evolution of the buildpack.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+  Customers will struggle to find relevant buildpacks to aid them in their development.  
+ 
+ 
+# Prior Art
+[prior-art]: #prior-art
+
+  1. [Ruby Gems](https://rubygems.org/)
+  2. [npm](https://www.npmjs.com/)
+  3. [Rust](https://crates.io/)
```
hone, line 158: Sadly I couldn't find any public APIs that are documented outside of code for crates.

text/0000-buildpack-registry-search-api.md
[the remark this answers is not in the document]
jkutner: Prior art: rubygems includes yanked versions:
https://rubygems.org/gems/nokogiri/versions

text/0000-buildpack-registry-search-api.md
hone, line 89: we should probably use a valid buildpack id here

text/0000-buildpack-registry-search-api.md
jkutner: we can get things like `description` from Dockerhub by pulling the manifest only right? would we still hit rate limits?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  4. [Dockerhub](https://hub.docker.com/)
+
+# Unresolved Questions
```
hone, line 161: Is this API part of the extension spec and if so should we look into versioning? We probably should do versions at least through headers but it's pretty common to do it via URL as well. Stripe has written a good [post on API versioning](https://stripe.com/blog/api-versioning) using dates instead of version numbers.

On top of that should we be looking into standards like OpenAPI or JSONAPI? There are some good tools like [swagger](https://swagger.io/) that can help here.

text/0000-buildpack-registry-search-api.md
hone, line 114: we should add a note that it's explicitly out of scope.

text/0000-buildpack-registry-search-api.md
jkutner: actually, we do this when the buildpack is published, to v[erify that it's a valid buildpack](https://github.com/buildpacks/github-actions#verify-buildpackage-action)

text/0000-buildpack-registry-search-api.md
hone, line 68:
```suggestion
      "yanked":false,"addr":"gcr.io/projectriff/java-function@sha256:5eabea8f7b2c09074ec196fe0c321006fb5ad8f282cc918520286d8a0007196f"
    },
```

text/0000-buildpack-registry-search-api.md
natalieparellano: If we're collecting extra data for each buildpack, another field that would be useful is `homepage`

text/0000-buildpack-registry-search-api.md
hone, line 74:
```suggestion
      "addr":"gcr.io/projectriff/node-function@sha256:194298b826c15bb079c59aed99968d7678a6e1f7a882c9d7f61811e0990717ba"
    },
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,172 @@
...
+  Customers will struggle to find relevant buildpacks to aid them in their development.  
+ 
+ 
+# Prior Art
+[prior-art]: #prior-art
+
+  1. [Ruby Gems](https://rubygems.org/)
+  2. [npm](https://www.npmjs.com/)
+  3. [Rust](https://crates.io/)
+  4. [Dockerhub](https://hub.docker.com/)
```
dfreilich, line 159: An additional prior art you may want to add is Tekton's new [Hub](https://hub-preview.tekton.dev/) for seeing Tekton tasks

text/0000-buildpack-registry-search-api.md
elbandito: Currently, the query field is compared against all the metadata fields that represent a buildpack in the registry.  We can of course tune this however we please.

text/0000-buildpack-registry-search-api.md
elbandito: It doesn't really.  I think it probably got sucked into my conscience thought because we're already doing I was thinking about how Pack also does a direct clone of the registry repo.  I can remove it.

text/0000-buildpack-registry-search-api.md
elbandito: We should support pagination for large responses from search queries, but I agree that the `GET /buildpacks` probably doesn't make sense.

text/0000-buildpack-registry-search-api.md
elbandito: Nice, thanks!

text/0000-buildpack-registry-search-api.md
elbandito: At this stage, I don't really have a strong sense for what fields would be deemed useful to a customer.  I'm just speculating and drawing inspiration from some of the other repos out there, listed in the prior art section below.

text/0000-buildpack-registry-search-api.md
elbandito: We could implement this entirely on the client side for both `pack` and a website.  But I don't like the tight coupling it would create for the web-based client.   However, I think we should definitely do this for `pack`.   Along these lines, it sounds like from the last WG, we might also want to consider a registry spec, so that `pack` would work against other private registries.  For the web case, I see it being very specific for CNB.   However, with loser coupling, customers could more easily re-use our API service and re-skin it with their own search dashboard (website).

text/0000-buildpack-registry-search-api.md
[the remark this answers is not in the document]
elbandito: @jromero good suggestion on the endpoint.  You're correct about having pagination.

text/0000-buildpack-registry-search-api.md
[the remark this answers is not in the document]
elbandito: Still unknown how we might calculate this info.  For now, to get rid of any confusion, I'm going to remove these.

text/0000-buildpack-registry-search-api.md
elbandito: Going to remove it for now.

text/0000-buildpack-registry-search-api.md
elbandito: Sounds good to me!  I just didn't want to volunteer others to participate, but if others are willing, the more the merrier.

text/0000-buildpack-registry-search-api.md
elbandito: agreed

text/0000-buildpack-registry-search-api.md
elbandito: That's how I'm leaning as well.  Something like `buildpacks/:ns/:name`

text/0000-buildpack-registry-search-api.md
elbandito: Given very small number of endpoints on this server, I'm not terribly worried about selecting a standard such as OpenAPI or JsonAPI, but I'll take a look at these.   I need to think about the versioning.

text/0000-buildpack-registry-search-api.md
elbandito: I like these suggestions.  I'm going to make some adjustments to these endpoints, shortly.

text/0000-buildpack-registry-search-api.md
elbandito: Great idea!

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,208 @@
...
+      "name":"command-function",
+      "latestVersion":"1.4.1",
+      "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f",
+      "versions": {
+        "1.4.1": {
+          "link": "https://registry.buildpacks.io/buildpacks/projectriff/command-function/1.4.1"
+        },
+        "1.3.9": {
+          "link": "https://registry.buildpacks.io/buildpacks/projectriff/command-function/1.3.9"
+        }
```
jromero, line 62: It would be great if the data model was more alike to that of `/buildpacks/:ns/:name/:version` so that there is only one data structure to represent a "buildpack" when consuming the API. 

For example:

```suggestion
     "latest": {
          "ns":"projectriff",
          "name":"command-function",
          "version":"1.4.1",
          "license": "MIT",
          "description": "...",
          "addr":"gcr.io/projectriff/command-function@sha256:99f9054abb73635a9b251b61d3627a8ff86508c767f9d691c426d45e8758596f"
      },
      "versions": {
        "1.4.1": {
          "link": "https://registry.buildpacks.io/buildpacks/projectriff/command-function/1.4.1"
        },
        "1.3.9": {
          "link": "https://registry.buildpacks.io/buildpacks/projectriff/command-function/1.3.9"
        }
```

text/0000-buildpack-registry-search-api.md
jkutner: I think we might want/need to encode the the `/` as `%2F`. We did this in the v2 buildpack registry so that we could use other types of IDs (like a number), and still have the `:id` be a single element in the path.

text/0000-buildpack-registry-search-api.md
elbandito: Good suggestion, @jromero

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,232 @@
...
+  ```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In the initial implementation, we can take advantage of the existing [registry](https://github.com/buildpacks/registry-index) repository to pull buildpack data from.  To keep ensure up-to-date buildpack data, we'd have a polling loop that performs a `git pull`.  This polling would take place in a separate background process.  
+
+Each time the local repository has been updated via `git pull`, buildpack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
+
+Search fields will include `ns`, and `name`.
```
jromero, line 163: Now that we've agreed on description being stored should we add description? Back to my case of someone searching for "java" and the buildpack myorg/jvm supports java as stated in the description.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,232 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+An external API service that exposes endpoints for retrieving buildpack metadata.  Initially, I envision this service providing a few GET endpoints to support plain text query searches, and individual buildpack data retrieval.  The API endpoints will be versioned, and follow OpenAPI and/or Json Schema/API standards.
+
+### Supported Endpoints
+
+- **GET /search?matches=text**
+
+  Retrievs all the buildpacks that satisfy the search query.  If the response contains more than 20 buildpacks, pagination will be used.  For example:
```
jromero, line 45: 1. I can't think of a single API that returns paging data only after X amount of results if that is what this statement is saying. I think paging data should always be return.
2. Can we get some details on what that would look like and the query parameters available to control paging?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,232 @@
...
+
+Search fields will include `ns`, and `name`. 
+
+In addition, a server process will expose endpoints and handle incoming request for the GET endpoints mentioned earlier.  For search requests, the text will be extracted from the `query` query parameter and used to search against the normalized buildpacks index.   Results will be added to the server response, as a list of JSON objects.
+
+*Note:  Initially, the Distribution Team maintainers can manage this service, and could even set-up a pager (but only best effort, and during working hours).  In the future, we maybe able to extend this to other verified project contributors.
+
+### Future Work
+
+- Query fields will be broken-up into separate, finer grained fields e.g. `ns`, `name`, and `yarnked`.
```
jromero, line 171: Typo

```suggestion
- Query fields will be broken-up into separate, finer grained fields e.g. `ns`, `name`, and `yanked`.
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,232 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+- more vulnerable due to Github outages
+- less optimal than using actual DB technology
+- constraint by what's in the index registry repository.  We might want to augment the metadata, with new fields that don't makes sense to have in the index repository.  I think this maybe an issue for the `/buildpacks/1` endpoint that *could* have download data, etc. things that don't belong in the actual repo.
+- without a DB we can't easily do things like ORDER_BY, or use ASC|DESC
```
jromero, line 180: Is this concern still valid? My understanding is that there would be a format backing data store.

text/0000-buildpack-registry-search-api.md
jromero: Re: versioning

My suggestion: just slap a v1 prefix to the base url. ie. `.../v1/buildpacks/....`

This also typically makes organization of the implementation easier.

jromero (changes requested):

text/0000-buildpack-registry-search-api.md
jkutner: No, @elbandito have agreed to maintain a database for the service

text/0000-buildpack-registry-search-api.md
jkutner: I think we should follow something like Github:

* A `page` param
* A `Link` header with format `<url>; rel="<type>"`

https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination

text/0000-buildpack-registry-search-api.md
jkutner: https://github.com/buildpacks/rfcs/pull/127

I don't think we should block this RFC on changes to `buildpack.toml` though.

text/0000-buildpack-registry-search-api.md
elbandito: yep, that's the route I went down

text/0000-buildpack-registry-search-api.md
elbandito: No longer valid - I'll update.  As @jkutner said, we're moving forward with having a DB

text/0000-buildpack-registry-search-api.md
elbandito: In the past, I've also used `limit` and `offset` params.  For example:
```
GET /items?limit=20&offset=20
```

But, I agree that the Github approach seems reasonable.  

Does this RFC need to commit to a specific pagination implementation?

text/0000-buildpack-registry-search-api.md
elbandito: @jromero I added a subsection that calls this out.

text/0000-buildpack-registry-search-api.md
jromero: Re: pagination commitment

In my opinion it would be preferable because the spec that comes out of this would need to if we hope to make this available as an interface for other registries.

jkutner: I don't think we're going to spec this. Given that we have the `/v1/` version schema, it can definitely change in the future. The goal here should be to capture where we want to start. I'm fine with `limit` and `offset` too.

jromero: > I don't think we're going to spec this. Given that we have the `/v1/` version schema, it can definitely change in the future. The goal here should be to capture where we want to start. I'm fine with `limit` and `offset` too.

Maybe I'm confused but I was under the impression that there would be a spec for the API similar to how we have a spec for the [registry storage](https://github.com/buildpacks/spec/blob/main/extensions/buildpack-registry.md#backing-storage). The purpose for the spec would be so that this API interface can be reimplemented by other non-official registries. Not spec'ing how pagination works would simply mean that the spec/interface is incomplete.

To be clear, I'm not proposing that this be clearly defined in the RFC (given I have no strong opinion on either choice) or that we have a spec before the implementation starts. But if our goal is what I understand it to be it would need to be spec'd.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+Each time the local repository has been updated via `git pull`, buildpack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
+
+Search fields will include `ns`, and `name`. 
+
+In addition, a server process will expose endpoints and handle incoming request for the GET endpoints mentioned earlier.  For search requests, the text will be extracted from the `query` query parameter and used to search against the normalized buildpacks index.   Results will be added to the server response, as a list of JSON objects.
+
+*Note:  Initially, the Distribution Team maintainers can manage this service, and could even set-up a pager (but only best effort, and during working hours).  In the future, we maybe able to extend this to other verified project contributors.
+
+### Future Work
```
jromero, line 173: We should add the idea of "tying" an API endpoint to a storage location here or under "What related issues do you consider out of scope for this RFC".

The use case in mind is:

> If the user uses a buildpack registry that lives in _some unofficial repo_, how does a tool (such as `pack`) know where to reach the API to search for buildpacks?

text/0000-buildpack-registry-search-api.md
jkutner: Not at this time. We're not seeking to support unofficial registries right now, and it's unclear if we ever would. The goal right now is to support the frontend for the official registry only (this is common for interfaces like npmjs.org and rubygems.org for example)

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Versioned Endpoints
+
+All the endpoints defined below will be version via a `vN` prefix to the base URL. e.g. `https://registry.buildpacks.io/v1/<endpoint>`.
+
+### Supported Endpoints
+
+- **GET /search?matches=text**
+
+  Retrieves all the buildpacks that satisfy the search query.  If the response contains more than significant number of buildpacks, pagination will be used.  The current approach to pagination is to take inspiration from how [Github](https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination).
```
ekcasey, line 49:
```suggestion
  Retrieves all the buildpacks that satisfy the search query.  If the response contains more than significant number of buildpacks, pagination will be used.  The current approach to pagination is to take inspiration from [Github](https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination).  
```

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+In the initial implementation, we can take advantage of the existing [registry](https://github.com/buildpacks/registry-index) repository to pull buildpack data from.  To keep ensure up-to-date buildpack data, we'd have a polling loop that performs a `git pull`.  This polling would take place in a separate background process.  
+
+Each time the local repository has been updated via `git pull`, buildpack data will be processed/normalized into a single JSON object, where plain text searches can be used against it.   Fields in this JSON object will be re-indexed as searchable fields, which will be compared against the search text during the retrieval algorithm.
+
+Search fields will include `ns`, and `name`. 
+
+In addition, a server process will expose endpoints and handle incoming request for the GET endpoints mentioned earlier.  For search requests, the text will be extracted from the `query` query parameter and used to search against the normalized buildpacks index.   Results will be added to the server response, as a list of JSON objects.
```
ekcasey, line 169:
```suggestion
In addition, a server process will expose endpoints and handle incoming request for the GET endpoints mentioned earlier.  For search requests, the text will be extracted from the `matches` query parameter and used to search against the normalized buildpacks index.   Results will be added to the server response, as a list of JSON objects.
```
We renamed the query param to `matches` in the example but the text still says `query` in a couple places. I assume we want to change the text?

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+*Note:  Initially, the Distribution Team maintainers can manage this service, and could even set-up a pager (but only best effort, and during working hours).  In the future, we maybe able to extend this to other verified project contributors.
+
+### Future Work
+
+- Query fields will be broken-up into separate, finer grained fields e.g. `ns`, `name`, and `yanked`.
```
ekcasey, line 175: I wonder if instead of breaking up query into fine-grained fields we want an exact match for params like `ns`. E.g. I can pass `matches=function&ns=heroku` to find all Heroku function buildpacks.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+- more vulnerable due to Github outages
+- less optimal than using actual DB technology
+- constraint by what's in the index registry repository.  We might want to augment the metadata, with new fields that don't makes sense to have in the index repository.  I think this maybe an issue for the `/buildpacks/1` endpoint that *could* have download data, etc. things that don't belong in the actual repo.
```
ekcasey, line 183: Given we have non-indexed fields like `description` above is this drawback out of date, as we aren't tying ourselves strictly to the fields in the index?

ekcasey (approved):

text/0000-buildpack-registry-search-api.md
jkutner: I don't think we should be worried about `pack` trying to search unofficial registries at this time. There's no guarantee other registries will support the search API, and most unofficial registries (in the near future) will just be Git repos).

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+All the endpoints defined below will be version via a `vN` prefix to the base URL. e.g. `https://registry.buildpacks.io/v1/<endpoint>`.
+
+### Supported Endpoints
+
+- **GET /search?matches=text**
+
+  Retrieves all the buildpacks that satisfy the search query.  If the response contains more than significant number of buildpacks, pagination will be used.  The current approach to pagination is to take inspiration from how [Github](https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination).  
+  ```
+  $ curl https://registry.buildpacks.io/v1/search?matches=projectriff
```
hone, line 51: do we want to namespace this with "api", i.e. <https://registry.buildpacks.io/v1/search?matches=projectriff>? This way <https://registry.buildpacks.io> could be where the frontend lives.

text/0000-buildpack-registry-search-api.md
jromero: I have UX concerns about this longterm but can see that we omit this for the time being until it's a little more proven and tested. ...With caution that we don't add constraints/assumptions that make this a lot harder to incorporate in the future.

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,235 @@
...
+  5. [Tekton Hub](https://hub-preview.tekton.dev/)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+
+- What parts of the design do you expect to be resolved through implementation of the feature?
+  
+  I expect to have an API with a GET endpoint with a `query` parameter that returns a list of matching buildpacks from the registry index repo.
```
hone, line 229: Do we need to change this to `matches`?

text/0000-buildpack-registry-search-api.md
elbandito: @hone like: `api.registry.buildpacks.io` or `api.buildpacks.io`?

text/0000-buildpack-registry-search-api.md
elbandito: yeah?

text/0000-buildpack-registry-search-api.md
jromero: FWIW: From my systems management days, I like subdomain based routing plus it looks nice. +1 on `https://api.buildpacks.io`.

jkutner: This isn't really intended to be an api for all of buildpack.io though. I think it should be:

* `https://registry.buildpacks.io` - frontend
* `https://registry.buildpacks.io/api` - API

hone: Yeah, I didn't want to presume this would be an API for all of buildpacks or what that means. At that point it'd be <https://api.buildpacks.io/registry>?

nebhale (approved):

text/0000-buildpack-registry-search-api.md
```diff
@@ -0,0 +1,242 @@
...
+
+An external API service that exposes endpoints for retrieving buildpack metadata against the official CNB buildpack registry.  It's NOT the goal of this API to support private buildpack registries.  I envision this service providing a few GET endpoints to support plain text query searches, and individual buildpack data retrieval.  The API endpoints will be versioned, and follow OpenAPI and/or Json Schema/API standards.
+
+### Versioned Endpoints
+
+All the endpoints defined below will be version via a `vN` prefix to the base URL. e.g. `https://registry.buildpacks.io/api/v1/<endpoint>`.
+
+### Supported MIME types
+
+`Accept: application/vnd.buildpacks+json`
```
hone, line 47: awesome to see this. I imagine in the future we'll probably do something similar to what [GitHub does](https://docs.github.com/en/free-pro-team@latest/rest/overview/media-types).

hone (approved):
I'm fine with this moving forward as is right now. I share similar concerns that @jromero in the longer term, but we can address that as we evolve.

jkutner:
Final Comment Period with merge disposition, closing on 24 December, 2020.

hone:
@jkutner more realistically I imagine we'll close this Jan 6, 2021 when we're back from the break.

jromero (approved):

PR #127, opened 2020-12-12
Added New Buildpack Descriptor Keys RFC

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/new-buildpack-toml-keys/text/0000-new-buildpack-toml-keys.md)

### DIFF
--- a/text/0000-new-buildpack-toml-keys.md
+++ b/text/0000-new-buildpack-toml-keys.md
@@ -38,7 +38,7 @@ See the spec changes below for full details.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- It introduces some non-machine-readable fields to an otherwise machine readable descriptor.
+- It introduces some free form fields to an otherwise heavily structured descriptor.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-new-buildpack-toml-keys.md
+++ b/text/0000-new-buildpack-toml-keys.md
@@ -57,7 +57,11 @@ This information could be stored in some other way, completely independent of ea
 # Prior Art
 [prior-art]: #prior-art
 
-Discuss prior art, both the good and bad.
+For prior art, it's very common for to see these fields in the file:
+* [gemspec](https://guides.rubygems.org/specification-reference/) (Ruby)
+* [`package.json`](https://docs.npmjs.com/cli/v6/configuring-npm/package-json) (JavaScript)
+* [`Cargo.toml`](https://doc.rust-lang.org/cargo/reference/manifest.html) (Rust)
+* [`pom.xml`](https://maven.apache.org/pom.html) (Java)
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-new-buildpack-toml-keys.md
+++ b/text/0000-new-buildpack-toml-keys.md
@@ -1,0 +1,88 @@
+# Meta
+[meta]: #meta
+- Name: New Buildpack Descriptor Keys
+- Start Date: 2020-12-12
+- Author(s): [@jkutner](@jkutner)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal for new keys in the `buildpack.toml` file.
+
+# Definitions
+[definitions]: #definitions
+
+* _buildpack descriptor_ - The `buildpack.toml` file that describes a buildpack.
+
+# Motivation
+[motivation]: #motivation
+
+New optional fields in the `buildpack.toml` would allow us to introduce more information into the Buildpack Registry search API and the web interface.
+
+# What it is
+[what-it-is]: #what-it-is
+
+* Add `description` as a free text field
+* Add `licenses` table
+
+# How it Works
+[how-it-works]: #how-it-works
+
+See the spec changes below for full details.
+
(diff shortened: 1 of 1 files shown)


text/0000-new-buildpack-toml-keys.md
```diff
@@ -0,0 +1,88 @@
...
+
+We would not be able to include this information in the Buildpack Registry.
+
+## Put the new keys in `package.toml`
+
+If we want to separate this information from the `buildpack.toml`, we could introduce it to `package.toml`, and only use it when creating a buildpackpage. This would be compatible with the Buildpack Registry because it only support buildpackages.
+
+## Put the new information in the Registry only
+
+This information could be stored in some other way, completely independent of each version (i.e. a name for the buildpack regardless of version). This could be stored in the registry index, or in some annex to it.
```
hone, line 55: I still think we could do this just in the registry, but you probably want at least license to be per version b/c OSS projects can change licenses.

text/0000-new-buildpack-toml-keys.md
```diff
@@ -0,0 +1,88 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
```
hone, line 60:
```suggestion
For prior art, it's very common for to see these fields in the file:
* [gemspec](https://guides.rubygems.org/specification-reference/) (Ruby)
* [`package.json`](https://docs.npmjs.com/cli/v6/configuring-npm/package-json) (JavaScript)
* [`Cargo.toml`](https://doc.rust-lang.org/cargo/reference/manifest.html) (Rust)
* [`pom.xml`](https://maven.apache.org/pom.html) (Java)
```

hone (approved):

text/0000-new-buildpack-toml-keys.md
```diff
@@ -0,0 +1,88 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+See the spec changes below for full details.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It introduces some none machine readable fields to an otherwise machine readable descriptor.
```
ekcasey, line 40:
```suggestion
- It introduces some non-machine-readable fields to an otherwise machine readable descriptor.
```
typo, but also, is this true? I find it confusing.

The fields can be parsed and printed automatically? It's no less machine-readable than the existing `metadata` table.

text/0000-new-buildpack-toml-keys.md
```diff
@@ -0,0 +1,88 @@
...
+# Definitions
+[definitions]: #definitions
+
+* _buildpack descriptor_ - The `buildpack.toml` file that describes a buildpack.
+
+# Motivation
+[motivation]: #motivation
+
+New optional fields in the `buildpack.toml` would allow us to introduce more information into the Buildpack Registry search API and the web interface.
+
```
ekcasey, line 25: `description` and licenses would also be valuable output from `pack inspect-buildpack`.

text/0000-new-buildpack-toml-keys.md
ekcasey: Including this information in `buildpack.toml` and the derived labels makes a lot of sense to me, even in a world where the registry does not exist. It is analogous to the description field in the builder metadata and the `licenses` table in `project.toml`.

Come to think of it... `project.toml` could probably benefit from a description too...

ekcasey (approved):

jromero (approved):

text/0000-new-buildpack-toml-keys.md
hone: Standardization is a big pro for me.

hone:
I'd love to piggyback on this RFC but by no means a blocking suggestion. Can we add a `keywords` field too? npm uses this where package authors can use these for categorizing their own packages.

text/0000-new-buildpack-toml-keys.md
jkutner: yea, i couldn't think of the right way to describe these fields. What I mean is that they are structured in a way that a machine could do something with the value other than store it. I think every other value (except `name`) can be parsed/processed in a way that a machine can change it's behavior based on the values

ekcasey: not `metadata` necessarily?

jabrown85 (approved):

elbandito (approved):

text/0000-new-buildpack-toml-keys.md
jkutner: addressed in 3ffe0a7

jkutner:
@hone `keywords` added in de408cf

text/0000-new-buildpack-toml-keys.md
```diff
@@ -0,0 +1,94 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+* Add `description` as a free text field
+* Add `licenses` table
+* Add `keywords` table
```
nebhale, line 31: Should keywords be a table? I would have guessed a list.

jkutner: agree. I have a list in the spec details below. will update this

jkutner: fixed in 42e2ad9

nebhale (approved):

nebhale:
Final Comment Period with merge disposition, closing on 20 January, 2021.

dfreilich (approved):

PR #128, opened 2020-12-17
Add Lifecycle Configuration File

jabrown85:
[Readable](https://github.com/buildpacks/rfcs/pull/128/files?short_path=85f520c#diff-85f520c3374e64b29919aa4c99421761a200127051a30c8a92ed2767d62b2315)

### DIFF
--- a/text/0000-add-lifecycle-config.md
+++ b/text/0000-add-lifecycle-config.md
@@ -35,7 +35,7 @@ In order to allow platforms to have more dynamic configurations, a new configura
 
 Target persona: platform operator, platform implementor
 
-A new `lifecycle.toml` will be read by lifecycle. By default, this file will be searched at `/cnb/lifecycle.toml`. `CNB_CONFIG_PATH` would be available to override this location.
+A new `lifecycle.toml` will be read by lifecycle. By default, this path searched will be `/cnb/config/lifecycle.toml`. `CNB_CONFIG_PATH` would be available to override this location.
 
 An example `lifecycle.toml` might be produced by the platform with various inputs (source code, platform specific configuration, etc.):
 ```
@@ -54,7 +54,7 @@ The keys of the `[config]` table are expected to match the env var names, minus 
 # How it Works
 [how-it-works]: #how-it-works
 
-Lifecycle will read in `CNB_CONFIG_PATH` or the default of `/cnb/lifecycle.toml` and prioritize the values found over lifecycle's defaults and before present env vars.
+Lifecycle will read in `CNB_CONFIG_PATH` or the default of `/cnb/config/lifecycle.toml` and prioritize the values found over lifecycle's defaults and before present env vars.
 
 If there is no configuration found, lifecycle will resolve configuration as it does today (env vars, defaults).
 
@@ -70,6 +70,7 @@ Another configuration file.
 
 - What other designs have been considered?
   - Lifecycle could load env vars from a known file location. e.g. `source lifecycle.env` or similar prior to resolving lifecycle configuration.
+  - Lifecycle could only allow for non-path based inputs like `log_level` in `lifecycle.toml`, but add an additional path to to a known platform location. Example would be `/cnb/config/order.toml` for `CNB_ORDER_PATH`. This would allow platforms to use a volume at `/cnb/config` and write files that are fed into lifecycle in addition to a `lifecycle.toml` for non-path based configuration.
 - Why is this proposal the best?
   - A well defined configuration file feels less convoluted and less stack specific
 - What is the impact of not doing this?
--- a/text/0000-add-lifecycle-config.md
+++ b/text/0000-add-lifecycle-config.md
@@ -11,52 +11,66 @@
 # Summary
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


text/0000-add-lifecycle-config.md
```diff
@@ -0,0 +1,95 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+Target persona: platform operator, platform implementor
+
+A new `lifecycle.toml` will be read by lifecycle. By default, this path searched will be `/cnb/config/lifecycle.toml`. `CNB_CONFIG_PATH` would be available to override this location.
+
+An example `lifecycle.toml` might be produced by the platform with various inputs (source code, platform specific configuration, etc.):
+```
+[config]
```
jkutner, line 42: I wonder if we should try to keep these keys matching the `lifecycle` flags exactly. So:

* `app` instead of `app_dir`
* `order` instead of `order_path`

OTOH, these more closely match the env vars

jkutner (approved):

text/0000-add-lifecycle-config.md
jabrown85: I went ahead and changed the names to be kebab-case, but I kept them matching the env variants for now. I was thinking it might make sense to be able to set some things that maybe aren't available as flags in the future. Like `CNB_REGISTRY_AUTH` isn't a flag, but you still might want to do the work to get that auth setup in the same pod. Similarly, there may be things like lifecycle feature flags or version selection that could be useful to have here that is likely to be available as an env var but may not have a flag name.

text/0000-add-lifecycle-config.md
```diff
@@ -0,0 +1,117 @@
...
+
+Why should we *not* do this?
+
+Another configuration file.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Lifecycle could introduce a `prepare` phase that processes a project's `project.toml` to generate the input files needed by detector, much like `pack` does today.
```
hone, line 86: does the implementation of `prepare` change anything for you here?

hone (approved):

text/0000-add-lifecycle-config.md
jabrown85: If prepare outputs an `order.toml` and puts it in a location that `detector` picks up, I would be OK putting a pause on this RFC. That is my primary need right now.

text/0000-add-lifecycle-config.md
```diff
@@ -0,0 +1,117 @@
...
+  - Lifecycle could load env vars from a known file location. e.g. `source lifecycle.env` or similar prior to resolving lifecycle configuration.
+  - Lifecycle could only allow for non-path based inputs like `log-level` in `config.toml`, but add an additional path to to a known platform location. Example would be `/cnb/config/order.toml` for `CNB_ORDER_PATH`. This would allow platforms to use a volume at `/cnb/config` and write files that are fed into lifecycle in addition to a `config.toml` for non-path based configuration.
```
yaelharel, line 88:
```suggestion
  - Lifecycle could only allow for non-path based inputs like `log-level` in `config.toml`, but add an additional path to a known platform location. Example would be `/cnb/config/order.toml` for `CNB_ORDER_PATH`. This would allow platforms to use a volume at `/cnb/config` and write files that are fed into lifecycle in addition to a `config.toml` for non-path based configuration.
```

text/0000-add-lifecycle-config.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+An example config resolution would be, in prioritized order, CLI flag, os env `CNB_ORDER_PATH`, `[config.toml][config]order_path`, then the `lifecycle` default of `/cnb/order.toml`.
+
+The keys of the `[config]` table are expected to match the env var names, minus the `CNB_` prefix and kebab-case.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Lifecycle will read in `CNB_CONFIG_PATH` or the default of `/cnb/config.toml` and prioritize the values found over lifecycle's defaults and before present env vars.
```
yaelharel, line 71: What is the prioritization of the cli flag vs a configuration file?
I guess it is:
1. cli flags
2. configuration file
3. env vars
4. default
right? Maybe it's worth adding it.

text/0000-add-lifecycle-config.md
```diff
@@ -0,0 +1,117 @@
...
+
+An example `config.toml` might be produced by the platform with various inputs (target source code, platform specific configuration, etc.):
+```
+[config]
+order-path = "/artifacts/generated-order.toml"
+no-color = true
+log-level = "debug"
+```
+
+An example config resolution would be, in prioritized order, CLI flag, os env `CNB_ORDER_PATH`, `[config.toml][config]order_path`, then the `lifecycle` default of `/cnb/order.toml`.
```
yaelharel, line 64: What about env vars that are also defined in the config file?
For example:
`log-level = "debug"` in the config file and `EnvLogLevel` as a variable that the user manually set.

jabrown85:
I'm going to close this for now. [This spec change](https://github.com/buildpacks/spec/pull/182) will get me most of the way there by being able to optionally provide an `order.toml` that does not belong to the builder. Thanks for the review comments, I'll make sure to bring them in if I resurrect this.

PR #132, opened 2021-01-05
Move layer types to new table in layer.toml

natalieparellano:
[Readable](https://github.com/natalieparellano/rfcs/blob/layer-table/text/0000-layer-table.md)

### DIFF
--- a/text/0000-layer-table.md
+++ b/text/0000-layer-table.md
@@ -67,7 +67,7 @@ This change should be implemented with Buildpack API 0.6 so that it goes hand-in
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Since we are moving the `launch`, `build`, and `cache` keys, could they be re-named for further clarity?
+- **Out of scope:** Since we are moving the `launch`, `build`, and `cache` keys, could they be re-named for further clarity?
 - Is `types` the right name for this table?
 - Would any changes be necessary to [`io.buildpacks.lifecycle.metadata`](https://github.com/buildpacks/spec/blob/main/platform.md#iobuildpackslifecyclemetadata-json)? This label includes `launch`, `build`, and `cache` keys for each referenced layer. If so, should the lifecycle standardize this label for buildpacks implementing different Buildpack APIs?
 
--- a/text/0000-layer-table.md
+++ b/text/0000-layer-table.md
@@ -19,7 +19,7 @@ Introduces a new table in `<layers>/<layer>.toml` to enable Bash buildpacks to e
 # Motivation
 [motivation]: #motivation
 
-- Why should we do this? [RFC 0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md) described a change (to be implemented in Buildpack API 0.6) whereby buildpacks must opt in to layer re-use - i.e., each `<layers>/<layer>.toml` will be restored with no `launch`, `build`, or `cache` flags set and buildpacks must modify the restored TOML to set these flags.
+[RFC 0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md) described a change (to be implemented in Buildpack API 0.6) whereby buildpacks must opt in to layer re-use - i.e., each `<layers>/<layer>.toml` will be restored with no `launch`, `build`, or `cache` flags set and buildpacks must modify the restored TOML to set these flags.
 
 The RFC assumed that Bash buildpacks would be able to easily opt in to layer re-use by appending lines to the TOML, e.g.:
 
--- a/text/0000-layer-table.md
+++ b/text/0000-layer-table.md
@@ -1,0 +1,86 @@
+# Meta
+[meta]: #meta
+- Name: Move layer types to new table in layer.toml
+- Start Date: 2021-01-05
+- Author(s): natalieparellano
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Introduces a new table in `<layers>/<layer>.toml` to enable Bash buildpacks to easily opt in to layer re-use.
+
+# Definitions
+[definitions]: #definitions
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this? [RFC 0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md) described a change (to be implemented in Buildpack API 0.6) whereby buildpacks must opt in to layer re-use - i.e., each `<layers>/<layer>.toml` will be restored with no `launch`, `build`, or `cache` flags set and buildpacks must modify the restored TOML to set these flags.
+
+The RFC assumed that Bash buildpacks would be able to easily opt in to layer re-use by appending lines to the TOML, e.g.:
+
+`echo "launch = true" >> layer.toml`
+
+However, because `launch`, `build`, and `cache` are top level keys, the above command would add a `launch` key to the `[metadata]` table instead of the top level.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC proposes moving the `launch`, `build`, and `cache` keys into a new `types` table in `<layers>/<layer>.toml`:
+
+```toml
+[types]
+launch = true
+build = false
+cache = false
+```
+
+A Bash buildpack could then write something like the following:
+
+```bash
+echo "[types]" >> layer.toml
+echo "launch = true" >> layer.toml
+```
+
+This change should be implemented with Buildpack API 0.6 so that it goes hand-in-hand with the opt in layer re-use feature.
+
+- If a buildpack implementing Buildpack API less than 0.6 tries to write `<layers>/<layer>.toml` with a `[types]` table, the lifecycle should fail.
+- If a buildpack implementing Buildpack API 0.6 or above tries to write `<layers>/<layer>.toml` with a top level `launch`, `build`, or `cache` key, the lifecycle should fail.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Why should we *not* do this? Having `launch`, `build`, and `cache` as top level keys potentially makes more sense in describing a layer.
+
+# Alternatives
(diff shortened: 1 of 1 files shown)


text/0000-layer-table.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+echo "[types]" >> layer.toml
+echo "launch = true" >> layer.toml
+```
+
+This change should be implemented with Buildpack API 0.6 so that it goes hand-in-hand with the opt in layer re-use feature.
+
+- If a buildpack implementing Buildpack API less than 0.6 tries to write `<layers>/<layer>.toml` with a `[types]` table, the lifecycle should fail.
+- If a buildpack implementing Buildpack API 0.6 or above tries to write `<layers>/<layer>.toml` with a top level `launch`, `build`, or `cache` key, the lifecycle should fail.
```
natalieparellano, line 52: Actually I'm not sure if this is necessary because the lifecycle won't ever present the file as it was written to a later build's buildpack... maybe a warning would be sufficient.

ekcasey (approved):

text/0000-layer-table.md
ekcasey: Failure seems better to me, so buildpack authors get early explicit feedback that their top-level keys won't work with the new API.

hone:
@danielleadams @Malax would love to get your input here since you maintain some Heroku/Salesforce buildpacks already using `layer.toml`.

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Why should we *not* do this? Having `launch`, `build`, and `cache` as top level keys potentially makes more sense in describing a layer.
```
hone, line 59: This is also a breaking change.

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Bash buildpack authors could prepend to the TOML file.
+- What is the impact of not doing this? Potentially aggravating extra labor for Bash buildpack authors.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Since we are moving the `launch`, `build`, and `cache` keys, could they be re-named for further clarity?
```
hone, line 70: I think for simplicity and your motivation this should be out of scope of this RFC, but it'd be good to couple another RFC with the change for `0.6` or whenever this rolls out if we wanted to do it.

I believe we were considering the following name changes from [packfile](https://github.com/sclevine/packfile/blob/master/docs/schema.md):
* `build` => `expose`
* `launch` => `export`
* `cache` => `store`
```

File: text/0000-layer-table.md [hunk 4]
```diff
@@ -0,0 +1,89 @@
...
+- Is `types` the right name for this table?
```
hone, line 71: I think `types` is a reasonable name. The biggest pushback in my head is when I think of types though I tend to think of a singular type, i.e. I can't be `build`, `launch`, and `cache`. `properties`?

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Why should we *not* do this? Having `launch`, `build`, and `cache` as top level keys potentially makes more sense in describing a layer.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Bash buildpack authors could prepend to the TOML file.
```
hone, line 64: This is pretty gross, but maybe a bash stdlib would help solve this problem?

```bash
echo -e "launch=true\n$(cat layer.toml)" > layer.toml
```

text/0000-layer-table.md
hone: How would it work if you have a group of buildpacks where there are some that are `< 0.6` and `0.5`?

nebhale (approved):

jkutner (approved):

text/0000-layer-table.md
Malax: To be fair, RFC 0052 is already a breaking change and this RFC tries to make the transition to `0.6` for bash buildpacks easier.

text/0000-layer-table.md
Malax: Given that they're already called layer types in the spec (https://github.com/buildpacks/spec/blob/main/buildpack.md#layer-types), `types` sounds good to me.

Malax (approved):

text/0000-layer-table.md
ekcasey: > How would it work if you have a group of buildpacks where there are some that are < 0.6 and 0.5?

We can validate the layers from each buildpack using that buildpack's API. A group containing multiple APIs doesn't make this any more complicated.

danielleadams (approved):

text/0000-layer-table.md
natalieparellano: I want to say that the lifecycle would be responsible doing the conversion here. When restoring `<layers>/<layer>.toml` it would write the file in the format that each buildpack expects to find it - i.e., for older buildpacks, preserving the top-level keys, and for newer buildpacks, stripping them out. And the lifecycle when reading `<layers>/<layer>.toml` would parse it with the schema that is valid for each buildpack.

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+# Motivation
+[motivation]: #motivation
+
+[RFC 0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md) described a change (to be implemented in Buildpack API 0.6) whereby buildpacks must opt in to layer re-use - i.e., each `<layers>/<layer>.toml` will be restored with no `launch`, `build`, or `cache` flags set and buildpacks must modify the restored TOML to set these flags.
+
+The RFC assumed that Bash buildpacks would be able to easily opt in to layer re-use by appending lines to the TOML, e.g.:
+
+`echo "launch = true" >> layer.toml`
+
+However, because `launch`, `build`, and `cache` are top level keys, the above command would add a `launch` key to the `[metadata]` table instead of the top level.
```
dfreilich, line 28: I don't fully get this – can you explain this a bit more deeply? If buildpacks are restricted to writing non-top level keys, then how would they be able to write a top-level types array?

natalieparellano: After the buildpack makes its edits the file would look like:
```
[metadata]
...stuff...

[types]
launch = true
build = false
cache = false
```

Without `[types]`, `launch = true` etc. would be keys under `[metadata]`.

dfreilich: Ahh. Hence prepending would work, because it's assured to be a top level key.

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+
+- Bash buildpack authors could prepend to the TOML file.
+- What is the impact of not doing this? Potentially aggravating extra labor for Bash buildpack authors.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- **Out of scope:** Since we are moving the `launch`, `build`, and `cache` keys, could they be re-named for further clarity?
+- Is `types` the right name for this table?
+- Would any changes be necessary to [`io.buildpacks.lifecycle.metadata`](https://github.com/buildpacks/spec/blob/main/platform.md#iobuildpackslifecyclemetadata-json)? This label includes `launch`, `build`, and `cache` keys for each referenced layer. If so, should the lifecycle standardize this label for buildpacks implementing different Buildpack APIs?
```
hone, line 72: Trying to circle back from the last discussion, did we mark this as out of scope or did we want to try to align label changes? @ekcasey

ekcasey: I lean towards yes, let's standardize the label. But there is one caveat Once you build an image with this platform API it could have implications for a rebuild with an older lifecycle that doesn't understand the new format. If lifecycle downgrade is a sticking point for folks I would rather keep the label the same and move forward for now.

text/0000-layer-table.md
```diff
@@ -0,0 +1,89 @@
...
+```bash
+cat >> launch.toml <<EOL
+[types]
+launch = true
+EOL
+```
+
+This change should be implemented with Buildpack API 0.6 so that it goes hand-in-hand with the opt in layer re-use feature.
+
+- If a buildpack implementing Buildpack API less than 0.6 tries to write `<layers>/<layer>.toml` with a `[types]` table, the lifecycle should fail.
```
yaelharel, line 53: A buildpack that implements Buildpack API less than 0.6 won't need to write `launch`, `build` and `cache` because [RFC0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md) doesn't affect it, right?
If yes, will `lifecycle` fail if the buildpack tries to write these values at the top level or will it be just part of `metadata` and there is no problem with it?

natalieparellano: That's right, `<layers>/<layer>.toml` will be restored with the top level keys intact for older buildpack apis. If a buildpack were to append those keys to the file they would just be part of metadata.

yaelharel: I'm not sure but maybe it's worth adding it to the RFC.

text/0000-layer-table.md
natalieparellano: I'm not sure I follow the need to change the label (this suggestion was more compelling when we were talking about changing the key names). The spec does require that analyzed.toml and `io.buildpacks.lifecycle.metadata` have the same schema, but the translation from <layer>.toml to analyzed.toml is more hand wavy; as it is, today <layer>.toml has `metadata` and the label has `data` to represent the same thing, so further translating <layer>.toml feels okay to me.

natalieparellano: Hmm... I may have been mistaken about what is `data` and what is `metadata` ...checking.

text/0000-layer-table.md
natalieparellano: Added in 2f2d27347272f6bcd538a71dd93c673c0ff2c69c

text/0000-layer-table.md
ekcasey: I buy that argument. Let's leave it alone for now and we can update the label and `analyzed.toml` in the future if we wish.

PR #133, opened 2021-01-07
Windows build images use SIDs for User/Group ownership

micahyoung:
[Readable](https://github.com/micahyoung/buildpacks-rfcs/blob/windows-security-identifiers/text/0000-windows-security-identifiers.md)


jabrown85:
This might be out of scope for this, but how does kubernetes translate things like this in for windows based pods?
```
securityContext:
    fsGroup: <group-id>
```

I'm trying to imagine writing a single platform that can target windows and linux builders with a single implementation. Right now, a platform may read the `CNB_GROUP_ID`  from the target builder and pop that into the `securityContext` for the pod so that volume ownership is as expected.

text/0000-windows-security-identifiers.md
```diff
@@ -0,0 +1,257 @@
...
+
+As any other end-user, I use the stack images as usual.
+
+> If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+
+All migration needs should be covered by an API bump and for implementations, changing the values they use for `-uid` and `-gid` to SID strings. Platforms that accept older API versions will need to continue to convert `CNB_USER_ID` into equivalent Security descriptors. See detailed migration steps below ([High-level migration steps](#high-level-migration-steps))
+
+> If applicable, describe the differences between teaching this to existing users and new users.
+
+SID usage will be very similar to Linux, though almost always using the SID for the standard Docker Windows user `ContainerUser`. Any Linux-specific documentation or existing Windows samples for Windows run images would need to change to reflect the new variables for Windows.
```
ekcasey, line 87: It it worth having `create-builder` automatically set `CNB_USER_SID` and `CNB_GROUP_SID` if it find that the build image is using one of the well known users? This might reduce burden on end users.

text/0000-windows-security-identifiers.md
```diff
@@ -0,0 +1,257 @@
...
+> The image config's Env field has the environment variable CNB_USER_ID set to the user †UID/‡SID of the user specified in the User field.
+
+Instead, Windows stack authors should specify *alternative* env vars for Windows `CNB_OWNER_SID`/`CNB_GROUP_SID` (instead of `CNB_USER_ID`/`CNB_GROUP_ID`) where the values are Windows Security Descriptor strings (ex: `CNB_OWNER_SID=S-1-5-93-2-1`). Elsewhere, in the platform spec (example: [creator](https://github.com/buildpacks/spec/blob/313078611d8a7925cb69a241df58e7d749d7f364/platform.md#creator)), lifecycle binaries are called with `-uid <uid>` and `-gid <gid>` flags. Instead, these flags shoud be defined as `-uid <†UID/‡SID>` or `-gid <†UID/‡SID>` to accept SIDs. Finally, all platforms/implementations should use this SID values to generate Windows image layer entries and container files with Security Descriptors instead of leaving them blank or converting from POSIX UID/GIDs.
```
yaelharel, line 16:
```suggestion
Instead, Windows stack authors should specify *alternative* env vars for Windows `CNB_OWNER_SID`/`CNB_GROUP_SID` (instead of `CNB_USER_ID`/`CNB_GROUP_ID`) where the values are Windows Security Descriptor strings (ex: `CNB_OWNER_SID=S-1-5-93-2-1`). Elsewhere, in the platform spec (example: [creator](https://github.com/buildpacks/spec/blob/313078611d8a7925cb69a241df58e7d749d7f364/platform.md#creator)), lifecycle binaries are called with `-uid <uid>` and `-gid <gid>` flags. Instead, these flags should be defined as `-uid <†UID/‡SID>` or `-gid <†UID/‡SID>` to accept SIDs. Finally, all platforms/implementations should use this SID values to generate Windows image layer entries and container files with Security Descriptors instead of leaving them blank or converting from POSIX UID/GIDs.
```

text/0000-windows-security-identifiers.md
```diff
@@ -0,0 +1,257 @@
...
+
+# Definitions
+[definitions]: #definitions
+* *Security Identifier*: A unique value of variable length that is used to identify a Windows *Security Principal*. Expressed as a string. Maximum 256 bytes.
+  * [Microsoft SID definition](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/78eb9013-1c3a-4970-ad1f-2b1dad588a25) 
+  * [SID components](https://docs.microsoft.com/en-us/windows/win32/secauthz/sid-components)
+  * [SID formal spec](https://docs.microsoft.com/en-us/windows/win32/secauthz/security-identifiers)
+  * [Well-known Windows SIDs](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/81d92bba-d22b-4a8c-908a-554ab29148ab)
+  * [Container-specific SIDs in moby/moby](https://github.com/moby/moby/blob/751d5f879a4f625bd32b08481bbb45e5d2db5b6c/pkg/system/syscall_windows.go)
+* *Security Principal*: Any entity that can be authenticated by the Windows operating system, such as a user account, a computer account, or a thread or process that runs in the security context of a user or computer account, or the security groups for these accounts
```
yaelharel, line 32: I would put this definition before the definition of "Security Identifier" because "Security Principal" is part of the "Security Identifier" definition.

text/0000-windows-security-identifiers.md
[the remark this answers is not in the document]
micahyoung: As discussed in working group, this suggestion makes a lot of sense, for consistency and simplicity of implementation. It should be easier to validate type-specific flags than inferring type. I'll update the RFC.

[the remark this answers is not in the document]
micahyoung: Definitely very relevant and while the [current lifecycle implementation](https://github.com/buildpacks/lifecycle/blob/77735c391b3f7b8376616a04cad8013c46b16e59/priv/user_windows.go) is unimplemented, it should be feasible to do so using Windows APIs like [AdjustTokenPrivileges](https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-adjusttokenprivileges?redirectedfrom=MSDN) or [LogonUserW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonuserw) (which have wrappers in [x/sys/windows](https://pkg.go.dev/golang.org/x/sys/windows#AdjustTokenPrivileges))

text/0000-windows-security-identifiers.md
micahyoung: Great suggestion, particularly since Docker's [hard-coded](https://github.com/moby/moby/blob/751d5f879a4f625bd32b08481bbb45e5d2db5b6c/pkg/system/syscall_windows.go#L50-L51) `ContainerUser`/`ContainerAdministrator` SIDs that will most commonly used, aren't well-known Windows SIDs. Having most stacks follow this pattern could also give us a mitigation if the SIDs ever changed for some.

I'll update the RFC with this suggestion

micahyoung:
> This might be out of scope for this, but how does kubernetes translate things like this in for windows based pods?
> 
> ```
> securityContext:
>     fsGroup: <group-id>
> ```
> 
> I'm trying to imagine writing a single platform that can target windows and linux builders with a single implementation. Right now, a platform may read the `CNB_GROUP_ID` from the target builder and pop that into the `securityContext` for the pod so that volume ownership is as expected.

@jabrown85 I think the Windows equivalent `securityContext` ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-runasusername/)) would be: 
```
  securityContext:
    windowsOptions:
      runAsUserName: "ContainerUser"
```

... so you'd likely end up grabbing the `CNB_USER_SID` instead to do the same for a Windows pod. 

Does that line up with the use-case? Would the Group vs User mismatch be problematic?

text/0000-windows-security-identifiers.md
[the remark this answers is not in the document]
micahyoung: Changed all refs to `-usid`

jabrown85:
> Does that line up with the use-case? Would the Group vs User mismatch be problematic?

That will probably be 👍

text/0000-windows-security-identifiers.md
```diff
@@ -0,0 +1,253 @@
...
+    nil,                                // system ACL
+    nil,                                // discretionary ACL
+  )
+// ... 
+```
+
+
+The `StringToSid` function converts the string SID to a struct and `SetNamedSecurityInfo` changes the Owner field of the file's on-disk Security Descriptor. The `nil` fields - Group, DACL, SACL - are not set here but can be set in this API. 
+
+Unfortunately these functions can't be used in any cross-platform way, as both `windows.StringToSid` and `windows.SetNamedSecurityInfo` wrap underlying Windows-only API syscalls.
```
natalieparellano, line 126: Maybe it's worth calling out here why this is problematic - e.g., pack on Linux/Darwin building Windows builders / buildpackages I believe?

text/0000-windows-security-identifiers.md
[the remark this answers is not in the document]
micahyoung: After digging in a bit more, it seems this impersonation behavior _should_ work for Windows. There's not a lot of well documented examples out there but I stumbled on this [Powershell Proof-of-concept](https://github.com/HarmJ0y/Misc-PowerShell/blob/master/Get-System.ps1) demonstrating how  an Administrator user can impersonate the `SYSTEM` user - technically to gain, rather than lose privileges - but translating this to Golang should be feasible, though maybe not trivial. I confirmed it works on a `servercore` (b/c only servercore has Powershell) and I assume it's likely to work on nanoserver for binaries using the same Windows syscalls.

As for the specifics of establishing daemon socket connections and `chown`ing files before dropping privileges, I'd have to continue to investigate, but it feels close enough that I assume we could find a workaround, if not use the same implementation.

But all the inputs as far as user/group SIDs are sufficient to do impersonation.

text/0000-windows-security-identifiers.md
micahyoung: I was going to add some text for that option but was wondering if it starts to contradict valid build image requirements of the spec:

Under [**Build Image**](https://github.com/buildpacks/spec/blob/313078611d8a7925cb69a241df58e7d749d7f364/platform.md#:~:text=CNB_USER_ID%20set%20to%20the%20user)
> The platform MUST ensure that:
>
> The image config's Env field has the environment variable CNB_USER_ID set to the user †UID/‡SID of the user specified in the User field.

(We'll be tweaking the var name here of course)

But I'm now feeling like, while this helps with `pack` builders, this may cause some delayed confusion in a case where a Stack Author makes a "build image" w/o the env vars, that works with `pack` but, potentially much later, a platform like `kpack` fails to validates the original build image while trying to make a cluster builder or something.

Perhaps, instead we could add a more helpful error message during `pack builder create` - when the env vars are missing, suggest the value to use, based on a lookup table for `ContainerUser`, `ContainerAdministrator` (and perhaps even a `/etc/passwd` check for Linux images, if we wanted to go that far).

text/0000-windows-security-identifiers.md
micahyoung: Great call. Now updated

ekcasey (approved):

jabrown85 (approved):

jkutner (approved):

text/0000-windows-security-identifiers.md
```diff
@@ -0,0 +1,250 @@
...
+
+Up until now, CNB platforms and implementations either ignore or implicitly convert POSIX UIDs/GIDs to Windows SIDs to create appropriately permissioned files in image layers. This RFC proposes changing the implementations/platforms (via the spec) so all existing references to a POSIX *UID/GID* will become a Windows *Owner SID/Group SID* for Windows images, environment variables, and file/directory entries in layers and volumes.
+
+There are various technical details involved with this design change, detailed below.  
+
+# Definitions
+[definitions]: #definitions
+* *Security Principal*: Any entity that can be authenticated by the Windows operating system, such as a user account, a computer account, or a thread or process that runs in the security context of a user or computer account, or the security groups for these accounts
```
yaelharel, line 26: There are many "or" in this sentence. Maybe it's better to have a list instead.

sclevine (approved):

natalieparellano (approved):

yaelharel (approved):

text/0000-windows-security-identifiers.md
hone: Do we need to close the loop on this?

ekcasey:
Final Comment Period with merge disposition, closing on 17 February, 2021

PR #138, opened 2021-02-11
Adding builder-key-in-project-descriptor RFC proposal.

wburningham:
[readable](https://github.com/wburningham/rfcs/blob/main/text/0000-builder-key-in-project-descriptor.md)

This is my first time submitting a feature/idea via an RFC pull request here so let me know if I have missed anything or done something incorrectly.

### DIFF
--- a/text/0000-builder-key-in-project-descriptor.md
+++ b/text/0000-builder-key-in-project-descriptor.md
@@ -68,7 +68,7 @@ or
 id = "<string>"
 ...
 
-[[build]]
+[build]
 builder = "<string>" # optional
 ...
 ```
--- a/text/0000-builder-key-in-project-descriptor.md
+++ b/text/0000-builder-key-in-project-descriptor.md
@@ -64,7 +64,7 @@ or
 Since a builder can be selected multiple ways, the following defines the order of precedence when selecting a builder:
 
 1. `--builder` flag (highest precedence)
-2. `project.toml` → `uild.builder`
+2. `project.toml` → `build.builder`
 3. configured builder (set via pack config default-builder)
 
 ## project.toml (TOML)
--- a/text/0000-builder-key-in-project-descriptor.md
+++ b/text/0000-builder-key-in-project-descriptor.md
@@ -61,6 +61,12 @@ or
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
+Since a builder can be selected multiple ways, the following defines the order of precedence when selecting a builder:
+
+1. `--builder` flag (highest precedence)
+2. `project.toml` → `uild.builder`
+3. configured builder (set via pack config default-builder)
+
 ## project.toml (TOML)
 
 ```toml
--- a/text/0000-builder-key-in-project-descriptor.md
+++ b/text/0000-builder-key-in-project-descriptor.md
@@ -61,7 +61,7 @@ or
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
-## proejct.toml (TOML)
+## project.toml (TOML)
 
 ```toml
 [project]
--- a/text/0000-builder-key-in-project-descriptor.md
+++ b/text/0000-builder-key-in-project-descriptor.md
@@ -1,0 +1,74 @@
+# Meta
+[meta]: #meta
+- Name: Builder key in project descriptor
+- Start Date: 2021-02-11
+- Author(s): @wburningham
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This is a proposal for a new `builder` key in the `project.toml` file.
+
+# Definitions
+[definitions]: #definitions
+
+* _project descriptor_ - The `project.toml` file that describes a project.
+
+# Motivation
+[motivation]: #motivation
+
+New optional `builder` field in the `project.toml` would allow configuration via file rather than CLI flags. This is useful when teams/companies have scaffolding/templetazing tools to help developers initialize a new project (ex: an internal tool/system can generate a `project.toml` with a default custom builder that matches the project). It is expected that a project/repo can store/serialize a `builder` in a parsable format (ie the `project.toml` rather than a `--builder` flag in a bash script). Being able to search repos to identify consumers of a specifier builder is helpful for auditing and communciating changes (hard to do when the `--builder` flag is specified in an unknown script/location).
+
+# What it is
+[what-it-is]: #what-it-is
+
+* Add `builder` as a text field
+
+# How it Works
+[how-it-works]: #how-it-works
+
+See the spec changes below for full details.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It surfaces the question around which/how-many CLI flags of `pack` should be made available in the project descriptor file (should all CLI flags be configurable in a project descriptor?).
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+
+This would require teams 
+1. to add some other propriety parsable file to store a builder and pre-parse it before invoking `pack`
+or
+2. fork `pack` to support parsing a builder from the project descriptor
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [Waypoint](https://www.waypointproject.io/plugins/pack) has a `builder` key likely because it is not configurable in the project descriptor. 
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+-Which/How-many CLI flags of `pack` should be made available in the project descriptor file?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+## proejct.toml (TOML)
+
+```toml
+[project]
+id = "<string>"
+...
+
+[[build]]
+builder = "<string>" # optional
+...
+```


text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+* _project descriptor_ - The `project.toml` file that describes a project.
+
+# Motivation
+[motivation]: #motivation
+
+New optional `builder` field in the `project.toml` would allow configuration via file rather than CLI flags. This is useful when teams/companies have scaffolding/templetazing tools to help developers initialize a new project (ex: an internal tool/system can generate a `project.toml` with a default custom builder that matches the project). It is expected that a project/repo can store/serialize a `builder` in a parsable format (ie the `project.toml` rather than a `--builder` flag in a bash script). Being able to search repos to identify consumers of a specifier builder is helpful for auditing and communciating changes (hard to do when the `--builder` flag is specified in an unknown script/location).
```
samj1912, line 24: I quite like this proposal! Curious about a possible edge-case - How would this key interact with a platform if someone explicitly specifies a different builder.

For example `pack build image-name --builder another-builder -d project.toml`. Would it error/refuse to build, or would it be overridden? I guess I am also curious what the existing expected behaviour for the same with `buildpacks` key vs providing an explicit flag via the command line that contradicts with it. and https://github.com/buildpacks/rfcs/blob/main/text/0019-project-descriptor.md.

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [Waypoint](https://www.waypointproject.io/plugins/pack) has a `builder` key likely because it is not configurable in the project descriptor. 
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+-Which/How-many CLI flags of `pack` should be made available in the project descriptor file?
```
ekcasey, line 59: Hmmm... `run-image` and `default-process-type` seem like no-brainers to me.

Other candidates:
* `--cache-image`, but there is some complexity here as `pack` currently only supports this when `--publish` is true
* `--tag`(s). This one gets a little weird because we make a slightly unintuitive distinction between the `<image>` (also a tag) and addition `--tag`s on the command line.

I think we should go with the easy wins for now :).

text/0000-builder-key-in-project-descriptor.md
dfreilich: I would assume that an explicit CLI flag value should override a project descriptor. That's what we do with run-mirrors, for instance

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+This would require teams 
+1. to add some other propriety parsable file to store a builder and pre-parse it before invoking `pack`
+or
+2. fork `pack` to support parsing a builder from the project descriptor
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [Waypoint](https://www.waypointproject.io/plugins/pack) has a `builder` key likely because it is not configurable in the project descriptor.
```
dfreilich, line 54: This is super interesting to note.

text/0000-builder-key-in-project-descriptor.md
dfreilich: I would hold off on cache-image for now, while we think over how we want to handle caching... https://github.com/buildpacks/pack/issues/1077

dfreilich (approved):
Overall, I think this is great! Good easy win.

dfreilich:
@wburningham Can you add a signoff to your commits, to fix the DCO?

text/0000-builder-key-in-project-descriptor.md
samj1912: That's good to know!


Probably off topic for this PR but - is this documented somewhere in the spec/rfc? If not, might be a good idea to enforce consistent behavior with descriptor overrides.

dfreilich: Definitely agree that it should be added in the RFC here, at least. :D

text/0000-builder-key-in-project-descriptor.md
jabrown85: @ekcasey why would you think `run-image` if the `builder` would have the run image specified?

+1 to `default-process-type` as something to support here

jabrown85 (approved):

text/0000-builder-key-in-project-descriptor.md
ekcasey: @jabrown85 an explicitly passed run image would take precedence over the default encoded in the builder. I imagine most users would be happy with the default, but it's a straightforward thing to configure and I don't see any reason not to make that option available in `project.toml` in addition to the existing `--run-image` flag?

nebhale (approved):

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+## proejct.toml (TOML)
+
+```toml
+[project]
+id = "<string>"
+...
+
+[[build]]
+builder = "<string>" # optional
```
jromero, line 72: This is likely a typo but `build` should be a table as opposed to an array.

```suggestion
[build]
builder = "<string>" # optional
```

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+-Which/How-many CLI flags of `pack` should be made available in the project descriptor file?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+## proejct.toml (TOML)
```
jromero, line 64:
```suggestion
## project.toml (TOML)
```

jromero (approved):
Aside from my typo comments, this looks really good. 👍

ekcasey (approved):

wburningham:
What is the next step now that there are sufficient approvals?

ekcasey:
> What is the next step now that there are sufficient approvals?

We still need either @jkutner or @hone to approve.

Also, I was happy to approve without waiting for the addition `run-image` or `default-process-type` because this is a net positive change, but did you want to add these keys (as discussed above)?

wburningham:
>Also, I was happy to approve without waiting for the addition `run-image` or `default-process-type` because this is a net positive change, but did you want to add these keys (as discussed above)?

I don't need to wait on `run-image` or `default-process-type`.

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+See the spec changes below for full details.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It surfaces the question around which/how-many CLI flags of `pack` should be made available in the project descriptor file (should all CLI flags be configurable in a project descriptor?).
```
hone, line 39: Definitely agree here, but for simplicity I'd mark all that stuff out of scope.

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+
+This would require teams 
+1. to add some other propriety parsable file to store a builder and pre-parse it before invoking `pack`
```
hone, line 47: @wburningham did you look at just jamming builder into `[metadata]` at all for your use case? Not that that should deter you from upstreaming this change! I'm more interested in how extensible people find `project.toml`.

hone (approved):
I'm :+1: this change. One thing to note is this is adding an extension spec feature into another extension spec.

sclevine (approved):

ekcasey:
Final Comment Period with merge disposition, closing on 3 March, 2021.

text/0000-builder-key-in-project-descriptor.md
jromero: @wburningham could we add the proposed order of precedence to this RFC?

My thinking would be that this is the desired order of precedence:

1. `--builder` 👈 highest precedence
2. `project.toml` → `build.builder` 
3. configured builder (set via `pack config default-builder`)  👈 lowest precedence

jromero (changes requested):
We should explicitly define the order of precedence. See https://github.com/buildpacks/rfcs/pull/138#discussion_r588681780

text/0000-builder-key-in-project-descriptor.md
wburningham: I've documented the order of precedence in my latest commit.

samj1912:
The only other comment I have about something like this is that this is the first time we are adding something to the project descriptor that is not achievable through a buildpack or a lifecycle process that runs inside the builder, as it will have to choose the builder before hand.

This means that the platform will have to run this step, possibly outside of the lifecycle. The possible issue I see with this is that this means that each platform will have to implement a project descriptor parser and it won't be possible to have something like a prepare step which is shipped as a part of the lifecycle and is platform agnostic, unless we provide a way to switch builders after the build process has started.

jromero (approved):

sclevine:
@jromero @ekcasey do we have any issues created for this RFC?

Acceptance is currently blocked on issue creation.

text/0000-builder-key-in-project-descriptor.md
```diff
@@ -0,0 +1,80 @@
...
+
+-Which/How-many CLI flags of `pack` should be made available in the project descriptor file?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+Since a builder can be selected multiple ways, the following defines the order of precedence when selecting a builder:
+
+1. `--builder` flag (highest precedence)
+2. `project.toml` → `uild.builder`
```
samj1912, line 67:
```suggestion
2. `project.toml` → `build.builder`
```

dfreilich:
@sclevine It looks like this can be merged – there aren't really any relevant implementation issues to be created, and there's already a PR open (https://github.com/buildpacks/pack/pull/1116) to implement this.

samj1912 (approved):

PR #139, opened 2021-02-24
pack-package-refactoring

dwillist:
Signed-off-by: dwillist <dthornton@vmware.com>


text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+│   ├── pack/ → current pack cli
+│   └── fetcher/ → cli to download buildpacks
+├── pkg/
+│   ├── archive/ → lib for converting multiple sources into OCI layers
+│   ├── client/ → current pack client library
+│   ├── descriptor/ → lib for processing project descriptor
+│   ├── inspect/ → lib for inspecting various CNB resources (app image, buildpack, builder, stack, ...)
+│   ├── internal/ → all non-exported go code
+│   ├── stack/ → lib for mixin validation & stack labels
+│   ├── stack/internal/ → keep internal details close to exported APIs.
```
ekcasey, line 44: Given that whatever would live here could only be used by the `stack` package, should we just include it in the `stack` package but not export it?

Also, this package would be named `internal` which is a bit strange. If we really need an internal package (the logic is used by multiple exported packages) maybe we want `pkg/internal/X` where X describes the whatever utility it is exposing. I think it would be easier to maintain a useable and well organized public interface for `stacks` etc. if any stack specific functions had to go in the external package and internal packages were reserved for coherent but non-domain-level concepts (e.g. `blob`)

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+- ease the process of contribution by having well defined connectable components.
+- Follow a general process of if data is externally visible on a pack produced artifact. We should provide structs for decoding.
+
+To accomplish the above we propose the following package structure:
+
+```
+.
+├── cmd/
+│   ├── pack/ → current pack cli
+│   └── fetcher/ → cli to download buildpacks
```
ekcasey, line 36: This seems like a brand new feature rather than a refactor. Maybe this should be out of scope.

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+│   ├── descriptor/ → lib for processing project descriptor
+│   ├── inspect/ → lib for inspecting various CNB resources (app image, buildpack, builder, stack, ...)
+│   ├── internal/ → all non-exported go code
+│   ├── stack/ → lib for mixin validation & stack labels
+│   ├── stack/internal/ → keep internal details close to exported APIs.
+│   ├── builder/ → lib for builder labels & builder manipulation
+│   ├── builder/internal/ → ...
+│   ├── buildpackage/ → lib for buildpackage label & buildpack extraction from buildpackages
+│   ├── buildpackage/internal/ → ...
+│   └── resolve/ → lib for resolving buildpacks via Buildpacks URIs
```
ekcasey, line 49: Given a lifecycle `prepare` phase may eventually want to use some of this logic (resolving buildpack uris, pulling buildpacks) and we don't want the lifecycle to import pack, should we think about creating a new repo (`distribution`?) for these packages?

We could do this later but it might make sense to do know, while we are refactoring to avoid problems disentangling later (if for example internal packages are used by packages that should be moved and also by packages that should remain in this repo).

text/0000-pack-package-refactor.md
jromero: It is my understanding that this structures is based on some [future thinking hypotheticals](https://github.com/buildpacks/community/discussions/51#discussion-1608992). I'd propose keeping it in the RFC but making note that it's "future thinking".

```suggestion
│   └── fetcher/ → cli to download buildpacks (future thinking)
```

text/0000-pack-package-refactor.md
jromero: Part of me is hesitant to move stuff into a new repository at this moment because we don't yet have the right seams and am afraid that the effort would prevent us from moving forward to even a slightly better structure.

In my mind, this RFC can be seen as the first step into potentially moving stuff out. I would think of it as a "promotion" strategy for any logic that exists in pack that we may want to redistribute.

ie:
`/pack/archive` -> `imgutil/archive`
`/pack/pkg/resolve` -> `<TBD>/resolve`

ekcasey: The makes sense. We can promote as a first step and think about extraction later if it is warranted.

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+# Summary
+[summary]: #summary
+
+This RFC proposes a restructuring of the `pack` codebase into reusable modular packages.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+    - Maintaining a modular and re-usable codebase enables contributors to be more effective, and encourages new adopters.
```
jromero, line 20:
```suggestion
- Why should we do this?
    - Maintaining a modular and re-usable codebase enables contributors to be more effective, and encourages new adopters. 
    - Allow for better consumption of pack as a library by having more isolated and focused packages.
    - Identify and create logical seams that can be used to then more easily redistribute packages.
```

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+│   ├── buildpackage/internal/ → ...
+│   └── resolve/ → lib for resolving buildpacks via Buildpacks URIs
+└── go.mod
+```
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved through implementation of the feature?
+Most of the refactoring needed to achieve this package structure. This might be a PR where we should agree on a possible package structure, start the implementation and make updates to this RFC as needed.
```
jromero, line 57: Not sure I understand the question and thereby the answer. Could you elaborate?

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
+# Meta
+[meta]: #meta
+- Name: (fill in the feature name: My Feature)
```
jromero, line 3:
```suggestion
- Name: Refactor Pack go packages
```

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,58 @@
...
+
+To accomplish the above we propose the following package structure:
+
+```
+.
+├── cmd/
+│   ├── pack/ → current pack cli
+│   └── fetcher/ → cli to download buildpacks
+├── pkg/
+│   ├── archive/ → lib for converting multiple sources into OCI layers
```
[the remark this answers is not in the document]
jromero: This has [already started](https://github.com/buildpacks/imgutil/pull/73) but I wouldn't say it's a blocker for this RFC.

text/0000-pack-package-refactor.md
ekcasey: Personally I would rather cut it from this RFC. I have misgivings about breaking `pack` into smaller binaries and believe worthy of its own RFC.

The `(future thinking)` caveat makes me feel slightly better, but also I don't see a reason to include this at all. The question of where we would put future CLIs doesn't seem like a controversy that need resolution (they always go in `cmd` by convention).

text/0000-pack-package-refactor.md
dwillist: The question is from the generic `0000-template.md`. The general thinking was this entire RFC seems like we are proposing a top down view of how we would like things to look. When actually performing the refactor we might decide that adjustments should be made. I believe it would be prudent to keep this RFC open and adjust it as we do the refactoring work.

jromero:
Based on our conversionation during the sub-team sync, I believe the next steps for this RFC would be to:

1. Focus goal on the intention as opposed to the end result.
2. Provide an end result for what the expected outcome may look like as soon as the RFC is approved.
3. Provide an end result for the _potential_ long-term vision of what the structure may be.

text/0000-pack-package-refactor.md
dwillist: I think re-framing the goals of this rfc as focusing on a process to pull functionality out means we can defer this conversation until we need to distribute additional CLIs

dwillist:
Hey maintainers any feedback would be appreciated @jromero @dfreilich

jkutner:
@dwillist can you add `.idea/` to `.gitignore`. It's ok to have it in this PR IMO

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,103 @@
+# Meta
+
```
jkutner, line 2: Are there any alternative proposals for the structure?

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,103 @@
...
+4) Extract code from external packages `/pkg/<pkg-name>/...` into a separate repository (this should require an RFC)
+
+The goal of this RFC is to:
+- Reach consensus on the overall code promotion process for the `pack` repo
+- Reach consensus on a "probable" target structure for `pack`
+
+The goal of this RFC is NOT:
+- to precisely define what packages we want & what their contents/functionality should be.
+
+
```
jkutner, line 38: everything below here feels like "How it works". Can you add that header?

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,103 @@
...
+This is a long term guess about which functionality we would want to pull out into packages as well as additional 
+commands that could be used by other platforms (such as Tekton)
+
+```
+.
+├── cmd/
+│   ├── pack/ → current pack cli
+│   ├── env-vars/ → cli for handling platform env vars.
+│   └── fetcher/ → cli to download buildpacks.
+├── pkg/
```
jkutner, line 91: Are there any examples of other projects that have a similar structure to this?

jkutner: internal to `buildpacks` or external

text/0000-pack-package-refactor.md
dwillist: Added 👍

jromero (approved):
🚀 I'm all for this refactor. As I work on Tekton even more, there are certain parts of `pack` that I'd like to encapsulate and externalize to be reused on the Tekton pipeline. Simply being able to gain access to them would greatly reduce what otherwise may be duplicate efforts.

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+[summary]: #summary
+
+This RFC outlines a promotion strategy and process for refactoring `pack` into consumable components as needed.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+    - Establish some shared vision about the codebase we want to manage.
+    - General refactoring benifits:
```
dfreilich, line 21:
```suggestion
    - General refactoring benefits: 
```

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+    - General refactoring benifits: 
+        - Maintaining a modular and re-usable codebase enables contributors to be more effective, and encourages new adopters.
+        - Ease consumption of pack as a library by having isolated and focused packages
+        - Create logical seams to redistribute packages.
+
+# What it is
+[what-it-is]: #what-it-is
+The general flow of this process is
+1) Unorganized features exist & are added to the `internal` package, these are not consumable by external programs
+2) Extract code from the `interal` package to `/pkg/<pkg-name>/...` to enable external consumption.
```
dfreilich, line 30:
```suggestion
1) Extract code from the `interal` package to `/pkg/<pkg-name>/...` to enable external consumption.
```
etc all the way down, to ease in case of reordering

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+│    └── tests
+├── pkg
+│    └── archive
+├── project
+├── registry
+├── resources
+└── tools
+    └── pedantic_imports
+```
+
```
dfreilich, line 66:
```suggestion
Further, the structure itself is very intimidating to a new-comer - what is important? Where does the definition/implementation live, and why is it split into multiple components without clear organization?
```

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+.
+├── builder
+├── buildpackage
+├── cmd
+│    └── pack
+├── config
+├── internal/...
+├── logging
+├── out
+│    └── tests
```
dfreilich, line 57:
```suggestion
```

These don't exist by default, they are just created through the testing structure.

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+```
+.
+├── cmd/
+│   └── pack/ → current pack cli
+├── pkg/
+│   ├── archive/ → lib for converting multiple sources into OCI layers
+│   ├── client/ → current pack client library
+│   ├── descriptor/ → lib for processing project descriptor
+│   ├── inspect/ → lib for inspecting various CNB resources (app image, buildpack, builder, stack, ...)
+│   ├── internal/ → all non-exported go code
```
dfreilich, line 80: Should `internal` be inside the `pkg` dir, or a top-level package? I'm not sure what best/typical practices here are

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+│   ├── stack/ → lib for mixin validation & stack labels
+│   ├── builder/ → lib for builder labels & builder manipulation
+│   ├── buildpackage/ → lib for buildpackage label & buildpack extraction from buildpackages
+│   └── resolve/ → lib for resolving buildpacks via Buildpacks URIs
+└── go.mod
+
+archive → pulled to a separate repo.
```
dfreilich, line 106: Assumedly part of `imgutil`, per the discussion here: https://github.com/buildpacks/imgutil/pull/73

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+# Alternatives
+[alternatives]: #alternatives
+- We proceed with ad-hoc refactorings
+- A more or less ambitions short term target.
```
dfreilich, line 112:
```suggestion
- A more or less ambitious short term target.
```

text/0000-pack-package-refactor.md
```diff
@@ -0,0 +1,118 @@
...
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+    - Establish some shared vision about the codebase we want to manage.
+    - General refactoring benifits: 
+        - Maintaining a modular and re-usable codebase enables contributors to be more effective, and encourages new adopters.
+        - Ease consumption of pack as a library by having isolated and focused packages
+        - Create logical seams to redistribute packages.
+
```
dfreilich, line 25: Add a reference to the discussion we had here?  https://github.com/buildpacks/community/discussions/51

dfreilich (approved):
overall, definitely happy to see this!

text/0000-pack-package-refactor.md
dwillist: [This](https://github.com/golang-standards/project-layout) seems to say top level is expected.

text/0000-pack-package-refactor.md
dwillist: Thanks for this added 👍

jromero:
TODO: @dwillist, remove binaries from "future" possibility.

text/0000-pack-package-refactor.md
jromero: See Prior Art section

hone (approved):

PR #140, opened 2021-03-02
Project Descriptor Domains RFC

hone:


text/0000-project-descriptor-domains.md
```diff
@@ -0,0 +1,104 @@
...
+[what-it-is]: #what-it-is
+
+The Project Descriptor will move to using reverse domains for top level keys as namespaces. This does two things:
+
+1. Allows any top level customization.
+1. Removes conflict of top level keys.
+
+The current schema as part of this proposal would look something like this:
+
+```TOML
```
jabrown85, line 33: It might be useful to have an example of a `project.toml` that does NOT use buildpacks at all. Maybe a hypothetical `Dockerfile` example might look like. Or even a `[com.megasaasplatform]` that allows has a few example keys. Using nextjs/fly.io or similar as a basis maybe?

hone: @jabrown85 I added a fly.io example.

hone:
Feedback from @jromero during the platform subteam sync:
* add `api` field under `[project]` or the top level
* clarify that `api` field is optional for reverse domain namespaces

jabrown85 (approved):

hone:
From WG today: need to solve the top level domain issue. By using `[project]`, it squats on the ".project" TLD. `_` can be used in bare TOML keys that aren't allowed in domain names.

jkutner:
I propose special casing the `project` TLD to resolve the issue. 

1. We forbid the `project` table from being customized (thus, you cannot add keys to `[project]` or `[project.mydomain]`
2. As an escape hatch, we allow `["project.mydomain"]` and interpret it as `project.mydomain` table.

Optionally, we forbid all TLD tables. This means that an org owning `.google` cannot create `[google] key = "value"`, but it prevents confusion when `[google]` is alongside `[project]`

hone:
@jkutner as discussed in the WG, forbidding just TLDs doesn't really solve the problem b/c what if I have `[project.foo]`. We're just squatting on the entire TLD.

text/0000-project-descriptor-domains.md
```diff
@@ -0,0 +1,155 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+This suprsedes the [Project Descriptor Flexibility RFC](https://github.com/buildpacks/rfcs/pull/95).
+
+## Special Case Project Namespace
+During the WG, there was a suggestion to just special case the "project" TLD. This would have the following implications:
+
+1. Forbid the project table from being customized (thus, you cannot add keys to '[project]` or `[project.mydomain]`)
+1. As an escape hatch, the project descriptor still allows the key quoted. For example, `["project.mydomain"]` and interpret it as the "project.mydomain" table.
```
Malax, line 140:
```suggestion
2. As an escape hatch, the project descriptor still allows the key quoted. For example, `["project.mydomain"]` and interpret it as the "project.mydomain" table.
```

text/0000-project-descriptor-domains.md
```diff
@@ -0,0 +1,155 @@
...
+
+- This is a backwards breaking change and the schemas aren't compatible at all.
+- TOML doesn't use whitespacing to determine nesting, so the full key is used which can be verbose.
+
+# Alternatives
+[alternatives]: #alternatives
+
+This suprsedes the [Project Descriptor Flexibility RFC](https://github.com/buildpacks/rfcs/pull/95).
+
+## Special Case Project Namespace
```
Malax, line 136: Alternatively, we could define that the root level of this TOML does not use reverse domain keys and move them under a generic `config` (or similar) key. This would allow us to use descriptive keys for things that are generic (such as `[project]`) and still allow folks using the `.project` TLD to use this feature without workarounds.

This would also allow us to keep the `build` key if we'd like to keep it.

Example:
```
[project]
id = ""
name = ""
# ...

[config.com.example]
awesome = "sauce"
```

I'm not sure how much I personally like this, but it seems to be a viable alternative.

text/0000-project-descriptor-domains.md
hone: This is just how markdown does numbered lists so you don't need to manually increment them.

jkutner (approved):

sclevine (approved):

ekcasey (approved):

jromero (approved):

nebhale (approved):

jromero:
@buildpacks/platform-maintainers - please create issues for this. (update parsing of `project.toml`)

sclevine:
@buildpacks/implementation-maintainers @buildpacks/distribution-maintainers needs issues

jromero:
@nebhale I see you added the `issues-created/core` but not finding a spec issue. If I'm not mistaken this should have one right?

sclevine:
@jromero to create issue and merge

PR #142, opened 2021-03-08
Add RFC for displaying contributer graph

aemengo:
[Readable](https://github.com/aemengo/rfcs/blob/add-contributer-graph/text/0000-add-contributer-graph.md)

[Mockup](https://user-images.githubusercontent.com/4236888/110389457-8dc2b500-8032-11eb-9dc1-3101dd35aff6.gif)

### DIFF
--- a/text/0000-add-contributer-graph.md
+++ b/text/0000-add-contributer-graph.md
@@ -56,7 +56,7 @@ The GitHub API is queried and a graph is rendered from the information. Links be
 [drawbacks]: #drawbacks
 
 - Not all contributors may want their organization affliations displayed
-- The buildpacks community might not want this information publically depicted
+- The buildpacks community might not want this information publicly depicted
 
 # Prior Art
 [prior-art]: #prior-art
--- a/text/0000-add-contributer-graph.md
+++ b/text/0000-add-contributer-graph.md
@@ -1,0 +1,71 @@
+# Meta
+[meta]: #meta
+- Name: Add Contributer Graph
+- Start Date: 2020-03-08
+- Author(s): aemengo
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes adding a customized contributer graph to the buildpacks [website](https://buildpacks.io/community/). Please see [mockup](#prior-art) for concept.
+
+# Definitions
+[definitions]: #definitions
+
+* Contribution - Opening a issue, Creating a PR, or Creating an RFC
+* Organization - The corporation represented by those making a contribution
+* OSS community - People engaging in discussion, usage and contribution to an OSS project
+
+# Motivation
+[motivation]: #motivation
+
+Tracking the health of an OSS community can be a very intensive activity. It can consist of keeping tabs on GitHub issues across multiple key repositories, or being an active participant in a multiple OSS slack channels, or keeping up with blog posts, etc. Although these collectively yield the most accurate insight into the health of an OSS community, not everyone can devote the time and attention on a consistent basis.
+
+For these others that are still invested in the health of the community, quantitative metrics are a useful recourse. These "others" could be prospective engineers deciding whether to immerse themselves within the community, or hobbyists looking to champion buildpacks within their organizations.
+
+The expected outcome is that more productive discussions about community engagement can occur when centered around the metrics attempting to depict it.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Please note: _This RFC does not intend to provide something fully realized. Instead, it aims to introduce the concept and have it be shaped by discussion, feasibility, and utility._
+
+At a high level, it is a graph that lives on the buildpacks [website](https://buildpacks.io/community/). The graph will show the full lifecycle of an OSS contribution, which could include the following:
+
+- RFC -> Issue/PR -> Shipped
+- Issue/PR -> Shipped
+
+A contribution is indicated as a data point and linked to other contributions via lines. The graph is seperated into the most critical repositiories of the buildpacks project. Contributions are color-coded by the participants' organization. Contributions are hyperlinked to GitHub issues. The graph aims to show 3 metrics within a given time interval:
+
+- Total amount of contributions
+- Percentage of contributions by organization
+- Average turnaround time from RFC -> Issue or Issue -> Shipped. (_This metric is currently illustrated poorly in the mockup_)
+
+There is also a CSV print-out option for those that prefer.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The GitHub API is queried and a graph is rendered from the information. Links between contributions can be extracted from labels on issues. If it proves to be too much logic for the frontend, it can be moved to a backend service.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Not all contributors may want their organization affliations displayed
+- The buildpacks community might not want this information publically depicted
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [Mockup](https://user-images.githubusercontent.com/4236888/110389457-8dc2b500-8032-11eb-9dc1-3101dd35aff6.gif)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What metrics to show
+- What the design looks like
+- How to implement


text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+
+At a high level, it is a graph that lives on the buildpacks [website](https://buildpacks.io/community/). The graph will show the full lifecycle of an OSS contribution, which could include the following:
+
+- RFC -> Issue/PR -> Shipped
+- Issue/PR -> Shipped
+
+A contribution is indicated as a data point and linked to other contributions via lines. The graph is seperated into the most critical repositiories of the buildpacks project. Contributions are color-coded by the participants' organization. Contributions are hyperlinked to GitHub issues. The graph aims to show 3 metrics within a given time interval:
+
+- Total amount of contributions
+- Percentage of contributions by organization
```
jkutner, line 45: I wonder how important this is at a top-level. I also wonder if it will be confusing or undefined when we get outside contributions. Maybe we can instead annotated each contribution (i.e. make this info a peer of the contributor name)

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Not all contributors may want their organization affliations displayed
+- The buildpacks community might not want this information publically depicted
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [Mockup](https://user-images.githubusercontent.com/4236888/110389457-8dc2b500-8032-11eb-9dc1-3101dd35aff6.gif)
```
jkutner, line 64: In the mockup, it's split out by project (pack, lifecycle, spec, etc). I wonder if we should lump them all together. This helps with things like github-actions that get infrequent updates. The spec might need to remain it's own thing.

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+The expected outcome is that more productive discussions about community engagement can occur when centered around the metrics attempting to depict it.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Please note: _This RFC does not intend to provide something fully realized. Instead, it aims to introduce the concept and have it be shaped by discussion, feasibility, and utility._
+
+At a high level, it is a graph that lives on the buildpacks [website](https://buildpacks.io/community/). The graph will show the full lifecycle of an OSS contribution, which could include the following:
+
+- RFC -> Issue/PR -> Shipped
```
samj1912, line 39: I think it might be a better workflow to have issue open > issue in progress (either as PR(s) or as an answer on the issue for support questions) > issue closed. A single issue can also span multiple or 0 PRs so we may essentially end up with a more complex graph here. 



We could have releases marking merged PRs that have been shipped.

samj1912: That was these issues can be tracked through different repositories by related links in github and you could annotate different components with different colors. That way the same lifecycle is applicable to an issue that starts as an RFC or as a support ticket.

text/0000-add-contributer-graph.md
samj1912: I agree, although it's important to know that some organizations are funding/backing up a project, I think it sort of deters individual contributions if you see that all the changes are dominated by a set of organizations and it might make certain individuals feel that their contributions may not be accepted if they are not associated with these organizations? ( I may be overthinking here)

aemengo: Thanks so much for these words. They go a long way in helping me define how to make this useful.

@samj1912 I especially like what you said about how it might get interpreted. When I have time, my next draft will be more contributor focused.

text/0000-add-contributer-graph.md
aemengo: It's partly intended to be educational. By splitting it out, it supposed to direct _newcomers_ towards the most prominent repos and how active they tend to be. Perhaps there's a way to do both what you and I said well.

text/0000-add-contributer-graph.md
aemengo: Very interesting. And you don't feel like information is lost?

You don't feel like we are omitting parts of the full journey for a given RFC?

samj1912: I was imagining something like above to trace a lifetime of a single connected component in the contribution graph. You could tag each note with the person who open the issue/PR for example. 


People could drill down into the above from a higher level view like 

This above summary/high level view could also lead to pie charts that can be pivoted in different ways like %age contributions by different organizations or users. 

LMK what you think :D

samj1912: You can also add more repositories on the left in their supposed order of dependency. For eg, we could add docs or samples to the bottom.

This way if a particular issue/PR has a lifecycle that starts and ends with docs, we can just hide other components.

aemengo: Sam, thanks so much. A picture is worth a thousand words.

These are great idea. Tbh, I don't know how I feel about pie charts w/ different pivots. Feels too "businessy". 

But I like what you're saying about higher level and then an in-depth view. I just want something appealing at that higher-level, that makes people want to click further. I'm thinking something more contributer-focused but not sure yet.

text/0000-add-contributer-graph.md
ekcasey: I am curious how we would display workstreams that cut across projects. For example imagine a new pack feature that requires a platform API change. The change would take the following journey to become available to pack users (omitting issues for now):

RFC -> Spec PR -> Lifecycle Issue -> LIfecycle PR -> Pack PR (support new API) -> Pack PR (expose new option to users) -> Docs PR

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes adding a customized contributer graph to the buildpacks [website](https://buildpacks.io/community/). Please see [mockup](#prior-art) for concept.
+
+# Definitions
+[definitions]: #definitions
```
dfreilich, line 17: I think it could be helpful to define `health` of an `OSS community` as well, and clarify why it is important to be conscious of, both for the maintainers, as well as for organizations/individuals who are thinking of adopting it.

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+At a high level, it is a graph that lives on the buildpacks [website](https://buildpacks.io/community/). The graph will show the full lifecycle of an OSS contribution, which could include the following:
+
+- RFC -> Issue/PR -> Shipped
+- Issue/PR -> Shipped
+
+A contribution is indicated as a data point and linked to other contributions via lines. The graph is seperated into the most critical repositiories of the buildpacks project. Contributions are color-coded by the participants' organization. Contributions are hyperlinked to GitHub issues. The graph aims to show 3 metrics within a given time interval:
+
+- Total amount of contributions
+- Percentage of contributions by organization
+- Average turnaround time from RFC -> Issue or Issue -> Shipped. (_This metric is currently illustrated poorly in the mockup_)
```
dfreilich, line 46:
```suggestion
- Average turnaround time from RFC -> Issue or Issues -> Shipped. (_This metric is currently illustrated poorly in the mockup_)
```
`+100000`

Ideally with an eye for reducing this turnaround time

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+
+There is also a CSV print-out option for those that prefer.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The GitHub API is queried and a graph is rendered from the information. Links between contributions can be extracted from labels on issues. If it proves to be too much logic for the frontend, it can be moved to a backend service.
```
dfreilich, line 53: Would associations with organization be done by the DCO, or is it something that we keep track of? If it's by DCO, not every person uses their work email for their contributions

text/0000-add-contributer-graph.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Not all contributors may want their organization affliations displayed
+- The buildpacks community might not want this information publically depicted
```
dfreilich, line 59:
```suggestion
- The buildpacks community might not want this information publicly depicted
```

text/0000-add-contributer-graph.md
dfreilich, line 64:
```suggestion
![Mockup](https://user-images.githubusercontent.com/4236888/110389457-8dc2b500-8032-11eb-9dc1-3101dd35aff6.gif)
```
will this get the gif to display?

samj1912: I believe the alternative render I posted below should be able to handle the these situations. It essentially tries to find connected components in the contribution graph and starts from a root node. Edges are decided based on time of creation so there shouldn't be any issues with finding a deterministic graph for any kind change spanning across tracked repositories.

aemengo: Yes it will, but it can be obnoxious

jromero:
Hi @aemengo, just checking in. It seems like it's been a few days since the last activity on this RFC. There seems to be some indication of updates coming soon. Is that correct?

aemengo:
@jromero Unfortunately, this RFC isn't quite top of mind for me right now. And I don't have updates planned soon.

PR #145, opened 2021-03-11
RFC: Combine and organize metadata file locations across lifecycle

sambhav:
[Readable](https://github.com/samj1912/rfcs/blob/layer-metadata/text/0000-layer-metadata-files.md)

It's my first time creating an rfc so not sure if I did everything right here.

### DIFF
--- a/text/0000-layer-metadata-files.md
+++ b/text/0000-layer-metadata-files.md
@@ -27,43 +27,52 @@ The proposal is as follows -
 The current structure of the `<layers>` folder is -
 
 ```
-<layers>
-├── launch.toml
-├── build.toml
-├── store.toml
-├── layer-x.toml
-├── layer-x
-│   └── ...
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
+├── group.toml
+├── plan.toml
+├── <escaped_buildpack_id> (buildpack arg `$1` )
+│   ├── build.toml
+│   ├── launch.toml
+│   ├── store.toml
+│   ├── <layer>.toml
+│   └── <layer>
+└── config
+    └── metadata.toml
 ```
+
 
 The proposal is to change this structure to - 
 
 
 ```
-<buildpack-workspace>
-├── launch.toml
-├── build.toml
-├── store.toml
+`/output` (configurable with `CNB_OUTPUT_DIR` or -output-dir) 
+├── group.toml
+├── plan.toml
 ├── layers
-│   ├── layer-x.toml
-│   ├── layer-x
-│   └── ...
-└── ...
+│   └── <escaped_buildpack_id> (`CNB_BP_LAYERS_DIR` )
+│       ├── <layer>.toml
+│       └── <layer>
+└── config
+    ├── metadata.toml
+    └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR`  )
+        ├── build.toml
+        ├── launch.toml
+        └── store.toml
 ```
 
-The top level `<buildpack-workspace>` would house any files and folder that are applicable at a `buildpack` level. Any layer specific files and directory structure would be moved to the `layers` sub-directory inside the `<buildpack-workspace>`.
+The top level `<output-dir>` would be the mounted volume that contains the output from buildpacks. The `<layers>` sub-directory would contains the layer contributions by each buildpack under an `<escaped-buildpack-id>` directory. This directory would be passed to the buildpack via the environment variable `CNB_BP_LAYERS_DIR`. Similarly any config files would reside in the `<config>`, under a sub-directory for each buildpack. This directory would be passed to the buildpack via the environment variable `CNB_BP_CONFIG_DIR`.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-The lifecycle phases that reference the `<layers>` directory would now take in `<buildpack-workspace>` directory instead. The behavior with respect to the various metadata files and the lifecycle would remain the same, however the `analysis`, `restore`, `build` and `export` phases would have to be updated to look for the various layer directories inside the `layers` sub-folder in the `buildpack-workspace`.
+The lifecycle phases that reference the `<layers>` directory would now take in `<output-dir>` directory instead. The behavior with respect to the various metadata files and the lifecycle would remain the same, however the `analysis`, `restore`, `build` and `export` phases would have to be updated to look for the various layer directories inside the `layers` and `config` sub-folder in the `<output-dir>`.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 Why should we *not* do this?
 
-This will break backwards compatibility for numerous buildpacks that don't use a set of bindings like `libcnb`. This would also result in a one-time loss of layer caches for app images.
+This will break backwards compatibility for numerous buildpacks that don't use a set of bindings like `libcnb`. This would also result in a one-time loss of layer caches for app images. It would also result in longer path names. We would also have to re-work the export logic to account for the export of the various files which are partially present in the `config` sub-directory and the `layers` sub-directory.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -74,17 +83,86 @@ This will break backwards compatibility for numerous buildpacks that don't use a
 
 We could add a prefix to each of the layer metadata files like `layer.`. This would allow us to re-use the cached layers but may lead to awkwardness around naming the layer metadata files - for eg. if a layer is called `layer.build` the metadata file would end up being `layer.layer.build.toml`
 
+```
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
+├── group.toml
+├── plan.toml
+├── <escaped_buildpack_id> (buildpack arg `$1` )
+│   ├── build.toml
+│   ├── launch.toml
+│   ├── store.toml
+│   ├── layer.<layer>.toml
+│   └── <layer>
+└── config
+    └── metadata.toml
+```
+
 ## Alternative 2
 
 We could add a prefix for each of the special metadata files and reserve it for CNB usage. For eg. `launch.toml` could be renamed to `cnb.launch.toml`. This way, we will only be prohibiting layers that have a prefix `cnb.` or a layer named `cnb`. The downsides to this alternative are the fact that the metadata files would not be usable across builds and we would still be prohibiting a class of layer names that may be valid.
 
-- Why is this proposal the best?
+```
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
+├── group.toml
+├── plan.toml
+├── <escaped_buildpack_id> (buildpack arg `$1` )
+│   ├── cnb.build.toml
+│   ├── cnb.launch.toml
+│   ├── cnb.store.toml
+│   ├── <layer>.toml
+│   └── <layer>
+└── config
+    └── metadata.toml
+```
 
-Although this is a drastic change, this also open up possibilities in the future to introduce other top level concepts inside the `<buildpacks-workspace>` apart from just `layers` while still allowing layer names to be free-form and without any restrictions.
 
-- What is the impact of not doing this?
+## Alternative 3
+
+The top level `<buildpack-workspace>` would house any files and folder that are applicable at a `buildpack` level. Any layer specific files and directory structure would be moved to the `layers` sub-directory inside the `<buildpack-workspace>`.
+
+```
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
+├── group.toml
+├── plan.toml
+├── <escaped_buildpack_id> (buildpack arg `$1` )
+│   ├── build.toml
+│   ├── launch.toml
+│   ├── store.toml
+│   └── layers
+│       ├── <layer>.toml
+│       └── <layer>
+└── config
+    └── metadata.toml
+```
+
+
+## Alternative 4
+
+```
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers-dir) 
+├── group.toml
+├── plan.toml
+├── <escaped_buildpack_id> (`CNB_BP_LAYERS_DIR`)
+│   ├── <layer>.toml
+│   └── <layer>
+└── config
+    ├── metadata.toml
+    └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR`  )
+        ├── build.toml
+        ├── launch.toml
+        └── store.toml
+```
+
+This structure doesn't break layer re-use and has the same length for path names but the top level directory is called `layers` when it houses more than just layers. And also the `config` directory may clash with the `escaped_buildpack_id`.
+
+## Why is this proposal the best?
+
+Although this is a drastic change, this also open up possibilities in the future to introduce other top level concepts inside the `CNB_BP_CONFIG_DIR` apart from the current ones while still allowing layer names to be free-form and without any restrictions.
+
+## What is the impact of not doing this?
 
 We keep having this ambiguity and possible issues in the future when we want to add more of such special files.
+
 
 # Prior Art
 [prior-art]: #prior-art
--- a/text/0000-layer-metadata-files.md
+++ b/text/0000-layer-metadata-files.md
@@ -34,10 +34,15 @@ The current folder structure is -
 │   └── <creator/builder/etc>
 ├── order.toml
 └── stack.toml
+/platform (configurable with `CNB_PLATFORM_DIR`)
+├── env 
+└   └── <env>
 /layers (configurable with `CNB_LAYERS_DIR` or -layers)
+├── analyzed.toml
 ├── group.toml
 ├── plan.toml
-├── order.toml (written by a platform, optionally)
+├── order.toml (written by a platform, optionally - prior to `lifecycle` execution)
+├── project-metadata.toml (written by a platform, optionally - prior to `lifecycle` execution, configurable with `CNB_PROJECT_METADATA_PATH`)
 ├── report.toml
 ├── <escaped_buildpack_id> (buildpack arg `$1`)
 │   ├── build.toml
@@ -61,17 +66,14 @@ This proposal is to change this folder structure to -
 │   └── <creator/builder/etc>
 ├── order.toml
 └── stack.toml
+/platform (configurable with `CNB_PLATFORM_DIR`)
+├── env 
+└   └── <env>
 /workspace (configurable with `CNB_WORKSPACE_DIR`)
 ├── app (configurable with `CNB_APP_DIR ` or -app)
 │    └── <app source code to transform>
-├── platform-config (configurable with `CNB_PLATFORM_CONFIG_DIR`)
-│   └── order.toml (written by a platform, optionally - prior to `lifecycle` execution `platform/prepare`)
-├── lifecycle-config (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
-│   └── group.toml
-│   └── plan.toml
-│   └── report.toml
 ├── layers (configurable with `CNB_LAYERS_DIR` or -layers)
-│   └── new-buildpack (`CNB_BP_LAYERS_DIR`)
+│   └── new-buildpack (`CNB_BUILDPACK_LAYERS_DIR`)
 │       ├── <layer>.toml
 │       └── <layer>
 │   └── old-buildpack (buildpack arg `$1`)
@@ -80,12 +82,21 @@ This proposal is to change this folder structure to -
 │       ├── store.toml
 │       ├── <layer>.toml
 │       └── <layer>
-├── layers-config (configurable with `CNB_LAYERS_CONFIG_DIR`)
-│    ├── metadata.toml
-│    └── new-buildpack (`CNB_BP_CONFIG_DIR`)
+├── config
+│   └── layers (configurable with `CNB_LAYERS_CONFIG_DIR`)
+│       ├── metadata.toml
+│       └── new-buildpack (`CNB_BUILDPACK_CONFIG_DIR`)
 │        ├── build.toml
 │        ├── launch.toml
 │        └── store.toml
+│   └── lifecycle (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
+│       ├── analyzed.toml
+│       ├── group.toml
+│       ├── plan.toml
+│       └── report.toml
+│   └── platform (configurable with `CNB_PLATFORM_CONFIG_DIR`)
+│       ├── order.toml (written by a platform, optionally - prior to `lifecycle` execution)
+└       └── project-metadata.toml (written by a platform, optionally - prior to `lifecycle` execution, configurable with `CNB_PROJECT_METADATA_PATH`)
 ```
 
 # How it Works
--- a/text/0000-layer-metadata-files.md
+++ b/text/0000-layer-metadata-files.md
@@ -1,0 +1,67 @@
+# Meta
+[meta]: #meta
+- Name: Disambiguate layer metadata files from Application metadata files
+- Start Date: 2021-03-11
+- Author(s): [samj1912](https://github.com/samj1912)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This is a proposal to change the location of layer specific TOML metadata files from application specific TOML metadata files like `launch.toml` or `build.toml` in order to disambiguate special metadata files from layer metadata files that can have similar names.
+
+# Motivation
+[motivation]: #motivation
+
+Currently all the metadata files live under the `<layers>` folder. This means that application metadata files like `launch.toml`, `build.toml` or `store.toml` share the same directory as layer metadata files. Since the layers could possibly be named `launch`, `build` or `store` this leads to ambiguity, conflicts and/or restrictions on layer names and their associated metadata files that should not exist. This also prevents us from possibly adding new metadata files in the future without breaking backwards compatibility with buildpacks who may have been creating layers with the same name. This proposal seeks to disambiguate special metadata files from free-form layer metadata files.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal is as follows - 
+
+All layer metadata files should be prefixed with the string `layer.`. For example, the current implementation expects layer specific metadata file for a layer `<layers>/example` to be located at `<layers>/example.toml`, we now propose that the associated layer metadata file be located at `<layers>/layer.example.toml`.
+
+This RFC would also prohibit buildpack authors from putting any files in the `<layers>` directory that is not currently a recognized metadata file or begins with the `layer.` prefix. This is to allow the buildpack API to accommodate more of special metadata files like `build.toml` without breaking any backwards compatibility in an unambiguous fashion.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would look for layer metadata files at `<layers>/layer.<layer>.toml` instead of `<layers>/<layer>.toml`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+This will break backwards compatibility for numerous buildpacks that don't use a set of bindings like `libcnb`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+
+We could have separate folders for layers and their metadata files and the application metadata files.
+
+- Why is this proposal the best?
+
+It seemed easier to add a prefix instead of creating a sub-folder in order to provide proper namespacing for layer metadata.
+
+- What is the impact of not doing this?
+
+We keep having this ambiguity and possible issues in the future when we want to add more of such special files.
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [RFC 0053](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
+
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+<!-- TODO -->
+
--- a/text/0000-layer-metadata-files.md
+++ b/text/0000-layer-metadata-files.md
@@ -56,7 +56,7 @@ The top level `<buildpack-workspace>` would house any files and folder that are 
 # How it Works
 [how-it-works]: #how-it-works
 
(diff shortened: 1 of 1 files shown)


ekcasey (approved):

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,67 @@
...
+[motivation]: #motivation
+
+Currently all the metadata files live under the `<layers>` folder. This means that application metadata files like `launch.toml`, `build.toml` or `store.toml` share the same directory as layer metadata files. Since the layers could possibly be named `launch`, `build` or `store` this leads to ambiguity, conflicts and/or restrictions on layer names and their associated metadata files that should not exist. This also prevents us from possibly adding new metadata files in the future without breaking backwards compatibility with buildpacks who may have been creating layers with the same name. This proposal seeks to disambiguate special metadata files from free-form layer metadata files.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal is as follows - 
+
+All layer metadata files should be prefixed with the string `layer.`. For example, the current implementation expects layer specific metadata file for a layer `<layers>/example` to be located at `<layers>/example.toml`, we now propose that the associated layer metadata file be located at `<layers>/layer.example.toml`.
```
jkutner, line 26: Could we make this optional to allow for arbitrary names like `jre.toml`, but then `layer.build.toml` would still work for a `build/` layer?

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,67 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+
+We could have separate folders for layers and their metadata files and the application metadata files.
+
+- Why is this proposal the best?
+
+It seemed easier to add a prefix instead of creating a sub-folder in order to provide proper namespacing for layer metadata.
```
yaelharel, line 51: It might be only me but although it seemed easier to add a prefix, I think that it'll be clearer to create a sub-folder.
I think about 2 examples where adding adding a prefix could be a little weird:
* A long layer name with dots in the middle.
For example: `foo.bar.test` will become `layer.foo.bar.test`.
* A layer name that already contains the word "layer".
For example:`layer.test` will become `layer.layer.test`, and `my.test.layer` will become `layer.my.test.layer`.

I don't have a strong opinion on this, just a slight preference.

samj1912:
I updated the rfc with the sub-directory proposal instead of the prefix one as discussed in the last WG meeting.

text/0000-layer-metadata-files.md
jromero: I agree with this sentiment. 

I would like to throw another potential alternative. Instead of making the layer metadata file _more_ special, we make the other files (`build.toml`, `launch.toml` `store.toml`) special by prefixing those.

For example:

Prefixing non-layer metadata files with `cnb.` and adding restrictions to the spec such as "Layers may not be prefixed with `cnb.`".

This would yield a file structure such as...

```
.
└── <layer>/
    ├── cnb.build.toml
    ├── cnb.launch.toml
    ├── cnb.store.toml
    ├── <layer>.toml
    └── */
        └── <other-content>
```

We do something similar with [environment variables](https://github.com/buildpacks/spec/blob/main/buildpack.md#provided-by-the-platform):

> The environment variable prefix CNB_ is reserved. It MUST NOT be used for environment variables that are not defined in this specification or approved extensions.

jromero:
> I updated the rfc with the sub-directory proposal instead of the prefix one as discussed in the last WG meeting.

This makes it sound like the sub-directory structure would be the primary idea but I'm still seeing prefixes as the primary proposal. Am I missing something?

samj1912:
> > I updated the rfc with the sub-directory proposal instead of the prefix one as discussed in the last WG meeting.
> 
> This makes it sound like the sub-directory structure would be the primary idea but I'm still seeing prefixes as the primary proposal. Am I missing something?

Urgh - I think I forgot to push.

text/0000-layer-metadata-files.md
samj1912: @jkutner also proposed a similar idea during the WG meeting. This might be a good compromise between our current system and the primary proposal! Will add it as an alternative tomorrow :)

jromero:
Really like the directory restructure. The term `buildpack-workspace` is a bit of a mouthful so I may bikeshed on it a bit but it totally makes sense IMO.

jromero (approved):

text/0000-layer-metadata-files.md
samj1912: Added this alternative as well.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,99 @@
...
+│   └── ...
+└── ...
+```
+
+The top level `<buildpack-workspace>` would house any files and folder that are applicable at a `buildpack` level. Any layer specific files and directory structure would be moved to the `layers` sub-directory inside the `<buildpack-workspace>`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would stages the reference `<layers>` directory would now take in `<buildpack-workspace>` directory instead. The behavior with respect to the various metadata files and the lifecycle would remain the same, however the `analysis`, `restore`, `build` and `export` phases would have to be updated to look for the various layer directories inside the `layers` sub-folder in the `buildpack-workspace`.
```
natalieparellano, line 59:
```suggestion
The lifecycle phases that reference the `<layers>` directory would now take in `<buildpack-workspace>` directory instead. The behavior with respect to the various metadata files and the lifecycle would remain the same, however the `analysis`, `restore`, `build` and `export` phases would have to be updated to look for the various layer directories inside the `layers` sub-folder in the `buildpack-workspace`.
```

jabrown85 (approved):
I like the structure change.

natalieparellano:
> Really like the directory restructure. The term `buildpack-workspace` is a bit of a mouthful so I may bikeshed on it a bit but it totally makes sense IMO.

`buildspace`?

text/0000-layer-metadata-files.md
```diff
@@ -18,37 +18,63 @@ This is a proposal to change the location of layer specific TOML metadata files
...
+
+```
+<buildpack-workspace>
+├── launch.toml
+├── build.toml
+├── store.toml
+├── layers
+│   ├── layer-x.toml
+│   ├── layer-x
+│   └── ...
```
yaelharel, line 50: I think you need to add a few spaces in this line as appears in line 36.

yaelharel (approved):

ekcasey:
Capturing a discussion from WG, Stephen proposed:

1. Moving "top-level" files to a different "config" directory so that the layers dir provided to the buildpack is exclusively for layers and their metadata.
2. Providing the location of both the layers and config directories as environment variables, rather than positional arguments

Elaborating on this idea and digging the details we would need to work out:

current directory structure:
```
`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
 ├── group.toml
 ├── plan.toml
 └── <escaped_buildpack_id> (buildpack arg `$1` )
    ├── build.toml
    ├── launch.toml
    ├── store.toml
    ├── <layer>.toml
    └── <layer>
  └── config
    └── metadata.toml
```

Assuming we want to reusing the existing layers volume for the new files instead of introducing a new required volume, I think these are the possible future directory structures?
Option 1:
```
`/layers` (configurable with `CNB_LAYERS_DIR` or -layers) 
 ├── group.toml
 ├── plan.toml
 └── <escaped_buildpack_id> (`CNB_BP_LAYERS_DIR` )
    ├── <layer>.toml
    └── <layer>
  └── config
     └── metadata.toml
     └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR` )
          ├── build.toml
          ├── launch.toml
          └── store.toml
```
pros:
* doesn't break layer reuse
* same-length for layer path names
cons:
* calling the top-level `layers` even though it is layers + config

Option 2:
```
`/build` (configurable with `CNB_BUILD_DIR` or -layers) 
 ├── group.toml
 ├── plan.toml
 └── l ( or `layers?`)
      └── <escaped_buildpack_id> (`CNB_BP_LAYERS_DIR` )
          ├── <layer>.toml
          └── <layer>
 └── c ( or `config`?)
      ├── metadata.toml
      └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR`  )
          ├── build.toml
          ├── launch.toml
          └── store.toml
```

pros:
* doesn't incorrectly call the entire top level directory `layers` (it's really too bad we used `workspace` for the app dir).
cons:
* breaks layer reuse
* `layers/layers`? `layers/l`? (both are bad, one is shorter)
* longer path names
* right now we export `layers/config`, if we implement this before eliminating `metadata.toml` we would need to partially export `/layers/config`.


In the above examples names of specific directories or env vars could change. E.g:
1. Do we want to reuse `CNB_LAYERS_DIR` to mean something different in the buildpack context instead of a creating a new `CNB_BP_LAYERS_DIR`?
1. `build` is just a placeholder. `workspace` actually seems the most logical to me but we already used it for the app dir.

But I think the overall structures represent the two obvious options.

jkutner:
I'm +1 on @ekcasey's "option 1". I'm also fine with `/layers/config`. I don't think it's exceptionally weird.

jabrown85:
I like @ekcasey's Option 2 - with a name change from `/build` to (`/output`, `/result`, `/out`, `/cnb/out`) or something along those lines. 

I am also ok with Option 1.

samj1912:
I also prefer option 2 with the same suggestions that @jabrown85 gave.

jkutner:
fwiw, I'm fine with option 2 also

nebhale (approved):

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+│   └── <layer>
+└── config
+    ├── metadata.toml
+    └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR`  )
+        ├── build.toml
+        ├── launch.toml
+        └── store.toml
+```
+
+This structure doesn't break layer re-use and has the same length for path names but the top level directory is called `layers` when it houses more than just layers. And also the `config` directory may clash with the `escaped_buildpack_id`.
```
ekcasey, line 156: > And also the `config` directory may clash with the `escaped_buildpack_id`.

This is true of alternatives 1-3 as well I think.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Why is this proposal the best?
+
+Although this is a drastic change, this also open up possibilities in the future to introduce other top level concepts inside the `CNB_BP_CONFIG_DIR` apart from the current ones while still allowing layer names to be free-form and without any restrictions.
```
ekcasey, line 160: I find this very convincing. I like this alternative because it give us more flexibility to make non-breaking additions to the spec in the future 🙂

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+└── config
+    ├── metadata.toml
+    └── <escaped_buildpack_id> (`CNB_BP_CONFIG_DIR`  )
+        ├── build.toml
+        ├── launch.toml
+        └── store.toml
+```
+
+The top level `<output-dir>` would be the mounted volume that contains the output from buildpacks. The `<layers>` sub-directory would contains the layer contributions by each buildpack under an `<escaped-buildpack-id>` directory. This directory would be passed to the buildpack via the environment variable `CNB_BP_LAYERS_DIR`. Similarly any config files would reside in the `<config>`, under a sub-directory for each buildpack. This directory would be passed to the buildpack via the environment variable `CNB_BP_CONFIG_DIR`.
+
```
ekcasey, line 64: We probably need to elaborate on compatibility a bit here. For example, if my build includes two buildpacks, one that implements the new API that includes this change and another older buildpack. If I am using the new platofrm API the directory structure end up looking like like the following?
```
`/output` (configurable with `CNB_OUTPUT_DIR` or -output-dir) 
├── group.toml
├── plan.toml
├── layers
│   └── new-buildpack (`CNB_BP_LAYERS_DIR` )
│       ├── <layer>.toml
│       └── <layer>
│   └── old-buildpack (buildpack arg `$1` )
│       ├── build.toml
│       ├── launch.toml
│       ├── store.toml
│       ├── <layer>.toml
│       └── <layer>
└── config
    ├── metadata.toml
    └── new-buildpack (`CNB_BP_CONFIG_DIR`  )
        ├── build.toml
        ├── launch.toml
        └── store.toml
```

If I am using the old platform API... this gets weirder. We have specified the locations of files like `metadata.toml`, `group.toml`, and the `<layer>` directories and `<layer>.toml` files in the platform API. We haven't specified the location of `build.toml` `launch.toml` etc. at the platform level so we could basically use alternative 4. e.g. 
```
`/layers` (configurable with `CNB_LAYERS_DIR` or -layers-dir) 
├── group.toml
├── plan.toml
├── new-buildpack (`CNB_BP_LAYERS_DIR`)
│   ├── <layer>.toml
│   └── <layer>
├── old-buildpack (buildpack arg `$1` )
│   ├── build.toml
│   ├── launch.toml
│   ├── store.toml
│   ├── <layer>.toml
│   └── <layer>
└── config
    ├── metadata.toml
    └── new-buildpack (`CNB_BP_CONFIG_DIR`  )
        ├── build.toml
        ├── launch.toml
        └── store.toml
```

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+[how-it-works]: #how-it-works
+
+The lifecycle phases that reference the `<layers>` directory would now take in `<output-dir>` directory instead. The behavior with respect to the various metadata files and the lifecycle would remain the same, however the `analysis`, `restore`, `build` and `export` phases would have to be updated to look for the various layer directories inside the `layers` and `config` sub-folder in the `<output-dir>`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+This will break backwards compatibility for numerous buildpacks that don't use a set of bindings like `libcnb`. This would also result in a one-time loss of layer caches for app images. It would also result in longer path names. We would also have to re-work the export logic to account for the export of the various files which are partially present in the `config` sub-directory and the `layers` sub-directory.
```
ekcasey, line 75: > This would also result in a one-time loss of layer caches for app images.

For cached layers this might not be true, we could simply restore the layer contents to the new path. This would provide a one-time loss of launch layers however. We should maybe call that out explicitly. E.g. the restorer will not restore layer metata without layer contents if the bp layers dir path has changed, either due to a platform API change (a buildpack API change might be fine, see comment above regarding compatibility strategies). We must also record the platform API that was used on the app image in order to enable this.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+│   ├── store.toml
+│   ├── layer.<layer>.toml
+│   └── <layer>
+└── config
+    └── metadata.toml
+```
+
+## Alternative 2
+
+We could add a prefix for each of the special metadata files and reserve it for CNB usage. For eg. `launch.toml` could be renamed to `cnb.launch.toml`. This way, we will only be prohibiting layers that have a prefix `cnb.` or a layer named `cnb`. The downsides to this alternative are the fact that the metadata files would not be usable across builds and we would still be prohibiting a class of layer names that may be valid.
```
natalieparellano, line 102: > the metadata files would not be usable across builds

Do you mean that if a buildpack upgrades from the old api to the new api the metadata file can't be re-used? I think the lifecycle could handle the re-naming if that's required.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+    └── metadata.toml
+```
+
+
+## Alternative 3
+
+The top level `<buildpack-workspace>` would house any files and folder that are applicable at a `buildpack` level. Any layer specific files and directory structure would be moved to the `layers` sub-directory inside the `<buildpack-workspace>`.
+
+```
+`/layers` (configurable with `CNB_LAYERS_DIR` or -layers)
```
natalieparellano, line 124: Do you mean `<buildpack-workspace>`? Would there be a new environment variable to configure it?

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+└── config
+    └── metadata.toml
+```
+
+
+The proposal is to change this structure to - 
+
+
+```
+`/output` (configurable with `CNB_OUTPUT_DIR` or -output-dir)
```
natalieparellano, line 48: Maybe it's worth calling out explicitly that this will require a platform api change.

text/0000-layer-metadata-files.md
natalieparellano: Though not impossible this seems like a lot of edge cases to handle. I could see the debugging experience in these cases being rather interesting :)

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,176 @@
...
+│   ├── store.toml
+│   └── layers
+│       ├── <layer>.toml
+│       └── <layer>
+└── config
+    └── metadata.toml
+```
+
+
+## Alternative 4
```
natalieparellano, line 139: I like this one as the simplest to understand and implement :)

text/0000-layer-metadata-files.md
yaelharel: I totally agree! I will add that I think it's worth making a one drastic change over a small change that we'll need to change again soon.

natalieparellano: Do we have ideas of things we might eventually want to add under `/output` besides `/layers` and `/config`?

jabrown85:
Here is my take after our recent discussions.
```
/cnb
├── buildpacks
│   └── <buildpacks>
├── lifecycle
│   └── <creator/builder/etc>
├── app (configurable with `CNB_APP_DIR ` or -app)
│   └── <app source code to transform>
├── build-workspace (configurable with `CNB_WORKSPACE_DIR`)
│   ├── platform-config (configurable with `CNB_PLATFORM_CONFIG_DIR`)
│   │   └── order.toml (written by a platform, optionally - prior to `lifecycle` execution `platform/prepare`)
│   ├── lifecycle-config (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
│   │   └── group.toml
│   │   └── plan.toml
│   ├── layers (configurable with `CNB_LAYERS_DIR`)
│   │   └── new-buildpack (`CNB_BP_LAYERS_DIR` )
│   │       ├── <layer>.toml
│   │       └── <layer>
│   │   └── old-buildpack (buildpack arg `$1` )
│   │       ├── build.toml
│   │       ├── launch.toml
│   │       ├── store.toml
│   │       ├── <layer>.toml
│   │       └── <layer>
│   └── layers-config (configurable with `CNB_LAYERS_CONFIG_DIR`)
│       ├── metadata.toml
│       └── new-buildpack (`CNB_BP_CONFIG_DIR`  )
│           ├── build.toml
│           ├── launch.toml
│           └── store.toml
```
Platforms can mount a volume at `/build-workspace` or mount different volumes at each sub-directory if they have a need to do so.

jkutner (approved):

jabrown85:
```
/builder
├── buildpacks
│   └── <buildpacks>
├── lifecycle
│   └── <creator/builder/etc>
├── order.toml
└── stack.toml
/cnb (configurable with `CNB_WORKSPACE_DIR`)
├── app (configurable with `CNB_APP_DIR ` or -app)
│    └── <app source code to transform>
├── platform-config (configurable with `CNB_PLATFORM_CONFIG_DIR`)
│   └── order.toml (written by a platform, optionally - prior to `lifecycle` execution `platform/prepare`)
├── lifecycle-config (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
│   └── group.toml
│   └── plan.toml
├── layers (configurable with `CNB_LAYERS_DIR`)
│   └── new-buildpack (`CNB_BP_LAYERS_DIR` )
│       ├── <layer>.toml
│       └── <layer>
│   └── old-buildpack (buildpack arg `$1` )
│       ├── build.toml
│       ├── launch.toml
│       ├── store.toml
│       ├── <layer>.toml
│       └── <layer>
├── layers-config (configurable with `CNB_LAYERS_CONFIG_DIR`)
│    ├── metadata.toml
│    └── new-buildpack (`CNB_BP_CONFIG_DIR`  )
│        ├── build.toml
│        ├── launch.toml
│        └── store.toml
```

How does this look?

nebhale (approved):

jabrown85:
@samj1912 can you fetch upstream on your fork? 

I am not sure how to properly take over this PR 😬 . I checked it out via `gh pr checkout 145` and pushed...but I think it only lives in this repo somehow. I noticed the fork has a "fetch upstream" button but I don't have write permissions to do it.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,184 @@
...
+├── app (configurable with `CNB_APP_DIR ` or -app)
+│    └── <app source code to transform>
+├── platform-config (configurable with `CNB_PLATFORM_CONFIG_DIR`)
+│   └── order.toml (written by a platform, optionally - prior to `lifecycle` execution `platform/prepare`)
+├── lifecycle-config (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
+│   └── group.toml
+│   └── plan.toml
+│   └── report.toml
+├── layers (configurable with `CNB_LAYERS_DIR` or -layers)
+│   └── new-buildpack (`CNB_BP_LAYERS_DIR`)
```
sclevine, line 74: Since `CNB_BP_LAYERS_DIR` is used more often than `CNB_LAYERS_DIR`, could we pick a shorter name for the former?

Maybe something like:
- `CNB_ALL_LAYERS_DIR` for `/workspace/layers`
- `CNB_ALL_CONFIG_DIR` for `/workspace/layers-config`
- `CNB_LAYERS_DIR` for `/workspace/layers/<buildpack-id>`
- `CNB_CONFIG_DIR` for `/workspace/layers-config/<buildpack-id>`?

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,184 @@
...
+├── buildpacks
+│   └── <buildpacks>
+├── lifecycle
+│   └── <creator/builder/etc>
+├── order.toml
+└── stack.toml
+/workspace (configurable with `CNB_WORKSPACE_DIR`)
+├── app (configurable with `CNB_APP_DIR ` or -app)
+│    └── <app source code to transform>
+├── platform-config (configurable with `CNB_PLATFORM_CONFIG_DIR`)
```
sclevine, line 67: Would it make sense to nest all the `-config` files in a new directory?
E.g.,
```
/workspace/config/platform
/workspace/config/lifecycle
/workspace/config/layers
```
This feels more organized to me, and it would also allow platforms to put all the metadata files on a dedicated volume.

sclevine (changes requested):
Bikeshed

text/0000-layer-metadata-files.md
jabrown85: I'm fine with `/workspace/config/*` as the default location for the config directories.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,184 @@
...
+│   └── order.toml (written by a platform, optionally - prior to `lifecycle` execution `platform/prepare`)
+├── lifecycle-config (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
+│   └── group.toml
+│   └── plan.toml
+│   └── report.toml
```
natalieparellano, line 72: This isn't really a "config" file - is there another place for it?

text/0000-layer-metadata-files.md
jabrown85: I have a minor concern with using `CNB_LAYERS_DIR` for `CNB_BP_LAYERS_DIR` since the former is already a config var for `/layers`. We would be moving `CNB_LAYERS_DIR` from a platform env var to a buildpack env var, which feels a bit weird.

text/0000-layer-metadata-files.md
jabrown85: Yeah, this whole directory is more like `lifecycle-output`. Now that I think about it...`analyzed.toml` is missing from the example and needs a new home. I assumed it would go here, but the name `config` is maybe not the best given the contents.

text/0000-layer-metadata-files.md
sclevine: Ah, good point. Maybe `CNB_ALL_LAYER_DIR` and `CNB_LAYER_DIR`?

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,184 @@
...
+```
+/cnb (provided by builder image)
+├── buildpacks
+│   └── <buildpacks>
+├── lifecycle
+│   └── <creator/builder/etc>
+├── order.toml
+└── stack.toml
+/layers (configurable with `CNB_LAYERS_DIR` or -layers)
+├── group.toml
```
yaelharel, line 38: I don't think that it's important to add it to the RFC, but just in case someone thinks differently, I just want to remind that before platform API 0.5, the default paths of the following files were in the working directory:
```
- group.toml
- plan.toml
- analyzed.toml
- project-metadata.toml
- report.toml
```
So if the working directory was set to `layers` and then the `layers` directory was set to a different directory using `-layers`, these files still exist under `layers`.

For more information, please see https://github.com/buildpacks/lifecycle/issues/393.

text/0000-layer-metadata-files.md
yaelharel: I think that `project-metadata.toml` is also missing.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,184 @@
...
+├── plan.toml
+├── order.toml (written by a platform, optionally)
+├── report.toml
+├── <escaped_buildpack_id> (buildpack arg `$1`)
```
natalieparellano, line 42: For new buildpacks running on the old platform, do we set `CNB_BP_CONFIG_DIR` equal to `CNB_BP_LAYERS_DIR`?

text/0000-layer-metadata-files.md
natalieparellano: I still like `CNB_BP_LAYERS_DIR` for its obviousness.

text/0000-layer-metadata-files.md
ekcasey: @jabrown85 Can we add this caveat to the updated draft? If the layers dir changes (either because the default changed between APIs or due to user-provided configuration) we cannot reuse launch layers from the previous image and therefore should not restore the metadata. This is already an issue if the change is user configured but it is an edge case that will suddenly become a very common occurrence (terrible bug that affect everyone) if we change the default layers dir between API versions.

We will probably need an additional RFC to add metadata to `analyzed.toml` that allows us to make that determination... :/

text/0000-layer-metadata-files.md
ekcasey: Right now the buildpack's layers dir is provided to it as a positional arg

> /bin/build <layers[EIC]> <platform[AR]> <plan[ER]>, Working Dir: <app[AI]>

Are we proposing replacing all positional args in the buildpack API with environment variables as part of that RFC? J I think that is a good idea but it's a big enough change that I think we should call this out explicitly in a separate section and provide the names for the env vars that will replace the plan and platform arguments as well.

ekcasey: If we go with `CNB_BP_LAYERS_DIR` should we switch from `CNB_BUILDPACK_DIR` to `CNB_BP_DIR` for consistency?

nebhale (approved):

text/0000-layer-metadata-files.md
jabrown85: I decided to change them to `_BUILDPACK_` to match `CNB_BUILDPACK_DIR`. They are long, but explicit and match our current standards that buildpacks are already built against.

text/0000-layer-metadata-files.md
jabrown85: Updated default location

text/0000-layer-metadata-files.md
jabrown85: Added

text/0000-layer-metadata-files.md
jabrown85: I think we would need to.

text/0000-layer-metadata-files.md
jabrown85: @ekcasey I would prefer to not tackle the positional argument changes in this RFC.

ekcasey: If we aren't switching from position arguments to environment variables then I am unsure what `CNB_BUILDPACK_LAYERS_DIR` is referring to above. The layers dir is currently a positional argument and nobody would be setting or reading this environment variable.

ekcasey: Also, if we are introducing at least one environment variable `CNB_BUILDPACK_CONFIG_DIR` it does seem to me like an appropriate time to convert the rest (just for the sake of internal consistency).

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,211 @@
...
+│        ├── launch.toml
+│        └── store.toml
+│   └── lifecycle (configurable with `CNB_LIFECYCLE_CONFIG_DIR`)
+│       ├── analyzed.toml
+│       ├── group.toml
+│       ├── plan.toml
+│       └── report.toml
+│   └── platform (configurable with `CNB_PLATFORM_CONFIG_DIR`)
+│       ├── order.toml (written by a platform, optionally - prior to `lifecycle` execution)
+└       └── project-metadata.toml (written by a platform, optionally - prior to `lifecycle` execution, configurable with `CNB_PROJECT_METADATA_PATH`)
```
natalieparellano, line 99: Would this make sense to go in the platform directory? In talking through https://github.com/buildpacks/pack/issues/1221 it seems like it could be helpful for the platform to mount a directory where it could provide configuration and receive output.

text/0000-layer-metadata-files.md
```diff
@@ -0,0 +1,211 @@
...
+│       └── <layer>
+│   └── old-buildpack (buildpack arg `$1`)
+│       ├── build.toml
+│       ├── launch.toml
+│       ├── store.toml
+│       ├── <layer>.toml
+│       └── <layer>
+├── config
+│   └── layers (configurable with `CNB_LAYERS_CONFIG_DIR`)
+│       ├── metadata.toml
```
natalieparellano, line 87: Should this be a child of `config`?

hone (approved):

nebhale (approved):

hone:
Putting on the backburner for now due to other active RFCs.

XavierMP14 (approved):

PR #146, opened 2021-03-15
Recommend different build/runtime user ids

samj1912:
This is based off the discussion in the last CNB office hours.

https://docs.google.com/document/d/1CwixRknj3gAXbi-fTI1-chwtMfy6G3GhdH17aEUjhK0/edit

(March 11 2021)

[Readable](https://github.com/samj1912/rfcs/blob/run-uid/text/0000-run-uid.md)

### DIFF
--- a/text/0000-run-uid.md
+++ b/text/0000-run-uid.md
@@ -29,7 +29,7 @@ Run-time user: The user which is referenced in the `User` field of the run image
 
 - Why should we do this?
 
-Currently the [Platform Spec](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) says that the run-time image's `User` along with its `CNB_USER_ID` environment variable must be set to the same user and uid as the build image.
+Currently the [Platform Spec](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) says that the run-time image's `User` along with its `CNB_USER_ID` environment variable must be set to the same user and UID†/SID‡ (OwnerSid) as the build image.
 
 This prevents use cases where you may want the run-time user to not have the permissions to modify the built image during run-time and in general restrict the set of permissions on it. 
 
@@ -39,12 +39,12 @@ This limitation seems to be enforced only by the spec and from my limited testin
 
 - What is the expected outcome?
 
-Modification of the spec to allow a different run-time user than build-time. We can still enforce them being a part of the same unix group to avoid too much divergence from run-time and build-time images which may cause issues, while at the same time providing the flexibility to accommodate above use-cases.
+Modification of the spec to allow a different run-time user than build-time. We can still enforce them being a part of the same primary group †GID/‡SID (GroupSid) to avoid too much divergence from run-time and build-time images which may cause issues, while at the same time providing the flexibility to accommodate above use-cases.
 
 # What it is
 [what-it-is]: #what-it-is
 
-A change to the spec to lift the restriction on the run-time user id being the same as the build-time user id.
+A change to the spec to lift the restriction on the run-time user being the same as the build-time user.
 
 # How it Works
 [how-it-works]: #how-it-works
@@ -57,3 +57,10 @@ Since pack and in turn the lifecycle already supports this, I don't imagine we w
 Does this RFC entail any proposed changes to the core specifications or extensions? 
 
 Yes, see above.
+
+
+# Notes
+
+When a word or bullet point is prefixed with a †, it SHALL be assumed to apply only to Linux stacks.
+
+When a word or bullet point is prefixed with a ‡, it SHALL be assumed to apply only to Windows stacks.
--- a/text/0000-run-uid.md
+++ b/text/0000-run-uid.md
@@ -1,6 +1,6 @@
 # Meta
 [meta]: #meta
-- Name: Allowing different run-time and build-time uids
+- Name: Recommending different run-time and build-time users
 - Start Date: 2021-03-15
 - Author(s): [@samj1912](https://github.com/samj1912)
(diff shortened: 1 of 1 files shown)


josegonzalez:
Is there any chance we can extend this to GID as well for similar reasons?

text/0000-run-uid.md
```diff
@@ -0,0 +1,56 @@
...
+
+This proposal proposes a change to the spec to allow for a different run-time user than the build-time user for various scenarios where you may want the run-time user to have different (ideally fewer) permissions than the build-time user.
+
+# Definitions
+[definitions]: #definitions
+
+Build image: An OCI image that provides the base of the build environment.
+Run image: An OCI image that provides the base from which app images are built.
+Build-time user: The user which is referenced in the `User` field of the build image.
+Run-time user: The user which is referenced in the `User` field of the run image.
```
ekcasey, line 22:
```suggestion
Build image: An OCI image that provides the base of the build environment.

Run image: An OCI image that provides the base from which app images are built.

Build-time user: The user which is referenced in the `User` field of the build image.

Run-time user: The user which is referenced in the `User` field of the run image.
```
I assume we want these to render on different lines.

text/0000-run-uid.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+
+Currently the [Platform Spec](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) says that the run-time image's `User` alongwith its `CNB_USER_ID` environment variable must be set to the same user and uid as the build image.
```
ekcasey, line 29:
```suggestion
Currently the [Platform Spec](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) says that the run-time image's `User` along with its `CNB_USER_ID` environment variable must be set to the same user and uid as the build image.
```

ekcasey (approved):

samj1912:
I went through the points raised at https://youtu.be/Kd_GMqMHkOo and I think they are all valid.

The major point that I wanted to focus on here is the fact that through something like dockerfiles you can easily switch between users during the build process and buildpacks API doesn't have a way to replicate something like this. 

An alternative proposal (which would probably not be as simple as the current rfc given that this rfc already works for pack and kpack and the only thing that really forbids it is the spec) that I can think of could be that -

The buildpack can advertise the user it wants to run as (except root?) and the lifecycle ensures that the buildpack is run as that user and the files it creates will be exported with the given user. The main caveats I see here are as follows - 

There would be 2 special users, the build user and the runtime user.

The initial set of files in the app dir will be owned by the build user. Same goes for the app directory.

If a buildpack wishes to create layers/files/directories that are writable by the runtime user it should be run as the runtime user.

This could be a part of the buildpack.toml as the user field.

```
[buldpack.user]
user_type = "custom"
uid = random_uid
```

or 

```
[buildpack.user]
user_type = "runtime"
```

or 
```
[buildpack.user]
user_type = "buildtime"
```

samj1912:
I also think that the build-time and run-time users are a property of a stack and it is inherently part of the BP contract to account for these users when advertising the stacks it is compatible with.

We should be able to make this change without affecting existing buildpacks that assume compatibility with currently existing stacks and introduce a new set of stacks (if needed) in the future with different runtime users and the buildpack authors can then opt-in to support this new stack.

jkutner (approved):

jromero (approved):

text/0000-run-uid.md
```diff
@@ -0,0 +1,59 @@
...
+
+This prevents use cases where you may want the run-time user to not have the permissions to modify the built image during run-time and in general restrict the set of permissions on it. 
+
+The above restriction seems to impose limitations on such use cases which are valid and are supposed to be in line with the Cloud Native Buildpack policy of secure container environments (for eg. we prevent building/running as root user).
+
+This limitation seems to be enforced only by the spec and from my limited testing, it looks like platforms like [pack](https://github.com/buildpacks/pack) and [kpack](https://github.com/pivotal/kpack) don't actually enforce this and stacks with different run-time and build-time users seem to work as expected.
+
+- What is the expected outcome?
+
+Modification of the spec to allow a different run-time user than build-time. We can still enforce them being a part of the same unix group to avoid too much divergence from run-time and build-time images which may cause issues, while at the same time providing the flexibility to accommodate above use-cases.
```
natalieparellano, line 42: I wonder what constraint this translates to for Windows. Looking at https://github.com/buildpacks/rfcs/blob/main/text/0076-windows-security-identifiers.md I think it would mean requiring that the group SIDs are the same.

samj1912: Added some clarification on windows - does that look better?

jabrown85 (approved):

nebhale (approved):

samj1912:
I have updated the proposal to **recommend** a different user instead of simply allowing it.

sclevine:
FCP closing 4/21/21

sclevine:
@hone reminder, FCP is past date

hone:
@jromero @dfreilich are there issues that need to be made for platform/pack from this?

jromero:
> @jromero @dfreilich are there issues that need to be made for platform/pack from this?

No issues on the platform side (outside of spec change).

jromero:
@buildpacks/implementation-maintainers / @buildpacks/core-team need issues to be created. 🙏 Thanks!

natalieparellano:
No issues needed on implementation repos, added "issues-created" label.

PR #147, opened 2021-03-22
Fix grammar and markdown in Asset Cache RFC

edmorley:
- `s/weather/whether/`
- fix mismatched backticks


Malax:
However, it would be nice if it would actually depend on the weather. 🌧️

edmorley:
Updated to also fix a markdown issue on the same line.

hone (approved):

samj1912 (approved):

edmorley:
I spotted one more mismatched backtick, updated again to fix that too.

PR #148, opened 2021-03-23
[BLOCKED] Allow setting default command arguments that can be overridden by the user

samj1912:

[Readable](https://github.com/samj1912/rfcs/blob/command/text/0000-command-overrides.md)

Notes - 

I would prefer the alternative that we were talking about that would bring us closer to the k8s interpretation of these terms (command = list of strings containing the command to execute and a list of arguments to pass to it, args = list of arguments that are default but can be overridden by the user), but if we are concerned about b/w compat - this is a good compromise. Also, couldn't think of a better key than `additional_default_args` that is descriptive and concise. Suggestions welcome.

### DIFF
--- a/text/0000-command-overrides.md
+++ b/text/0000-command-overrides.md
@@ -33,16 +33,25 @@ The launcher should be able to determine if arguments have been passed by the us
 # What it is
 [what-it-is]: #what-it-is
 
-The proposal essentially involves adding an additional `additional_default_args` key in the `Launch Metadata` (launch.toml) that is a list of strings. This list of strings would be a set of additional arguments that would be appended to the existing command invocation if the user does not provide any arguments themselves.
+The proposal involves adding an changing the `command` key to be a string array instead of a single string. This array would be the command and its associated arguments that would be set by default and any additional arguments passed by the user would be appended to it. The `args` array would refer to a set of default arguments that will be appened to the `command` array if no arguments are passed by the user during runtime.
+
 
 ```toml
 [[processes]]
 type = "<process type>"
-command = "<command>"
-args = ["<arguments>"]
-additional_default_args = ["<additional>", "<default>", "<arguments>"]
+command = ["command", "along", "with", "additional", "arguments"]
+args = ["default", "arguments", "that", "can", "be", "overridden"]
 direct = false
 ```
+
+This would bring the definition of `command` and `args` to be consistent with the `Kubernetes` definition of the same. This table summarizes the field names used by Docker/OCI, Kubernetes, the current API and the proposed API.
+
+
+| OCI/Docker   | Kubernetes | Proposed  | Current            |
+| ------------ | ---------- | --------- | ------------------ |
+| `ENTRYPOINT` | `command`  | `command` | `command` + `args` |
+| `CMD`        | `args`     | `args`    | N/A                |
+
 
 # How it Works
 [how-it-works]: #how-it-works
@@ -53,9 +62,8 @@ Let's see how this proposal will work through an example. Let's say we have a pr
 ```toml
 [[processes]]
 type = "hi"
-command = "echo"
-args = ["Hello"]
-additional_default_args = ["world"]
+command = ["echo", "Hello"]
+args = ["World"]
 direct = <direct>
 ```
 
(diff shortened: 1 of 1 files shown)


text/0000-command-overrides.md
```diff
@@ -0,0 +1,132 @@
...
+[what-it-is]: #what-it-is
+
+The proposal essentially involves adding an additional `additional_default_args` key in the `Launch Metadata` (launch.toml) that is a list of strings. This list of strings would be a set of additional arguments that would be appended to the existing command invocation if the user does not provide any arguments themselves.
+
+```toml
+[[processes]]
+type = "<process type>"
+command = "<command>"
+args = ["<arguments>"]
+additional_default_args = ["<additional>", "<default>", "<arguments>"]
```
jkutner, line 43: is there a situation where you'd want to provide _both_ `args` and `additional_default_args`?

jkutner: oh i think i see. `args` _cannot_ be overridden but `additional_default_args` can be?

text/0000-command-overrides.md
```diff
@@ -0,0 +1,132 @@
...
+
+- What is the expected outcome?
+
+The launcher should be able to determine if arguments have been passed by the user or not and if they have, it should override a default list of arguments by the once specified by the user.
+
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal essentially involves adding an additional `additional_default_args` key in the `Launch Metadata` (launch.toml) that is a list of strings. This list of strings would be a set of additional arguments that would be appended to the existing command invocation if the user does not provide any arguments themselves.
```
jkutner, line 36: Maybe just `default-args` as the name of the key?

text/0000-command-overrides.md
samj1912: Yes!

text/0000-command-overrides.md
```diff
@@ -0,0 +1,132 @@
...
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+N/A
+
+# Alternatives
+[alternatives]: #alternatives
+
+The other alternative to support something like this is to change `command` from a string to an array of strings specifying the command and the additional arguments it will always be run with and `args` to become a set of default arguments that can be over-ridden by the user.
```
jabrown85, line 107: I think I would prefer to change `command` to match existing OCI platforms/standards.

I get the backwards compatibility concern - but is that a large concern if this is only for Buildpack API > `0.X`?  Is the concern that buildpack authors won't notice the breaking change?

This is what this alternative is proposing right? Command becoming an array and args being 100% override-able.
```toml
[[processes]]
type = "hi"
command = ["echo", "Hello"]
args = ["world"]
direct = <direct>
```

samj1912: Yup, this would have been ideal, but the only reason I didn't go with this is because it was backwards incompatible.

samj1912:
I have switched the main/alternative proposal given the discussion in the last WG meeting.

text/0000-command-overrides.md
```diff
@@ -0,0 +1,149 @@
...
+
+- What is the expected outcome?
+
+The launcher should be able to determine if arguments have been passed by the user or not and if they have, it should override a default list of arguments by the once specified by the user.
+
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal involves adding an changing the `command` key to be a string array instead of a single string. This array would be the command and its associated arguments that would be set by default and any additional arguments passed by the user would be appended to it. The `args` array would refer to a set of default arguments that will be appened to the `command` array if no arguments are passed by the user during runtime.
```
jabrown85, line 36:
```suggestion
The proposal involves changing the `command` key to be a string array instead of a single string. This array would be the command and its associated arguments that would be set by default and any additional arguments passed by the user would be appended to it. The `args` array would refer to a set of default arguments that will be appended to the `command` array if no arguments are passed by the user during runtime.
```

jabrown85 (approved):
I like taking this step towards a more normalized `command` and `arg` model

hone:
This proposal looks great. Since it's a breaking change, I think it's worth talking about and considering the impact since everyone is doing it this way now. If someone passes a String vs. an array, will lifecycle catch and point to this breaking change?

jromero (approved):

text/0000-command-overrides.md
```diff
@@ -0,0 +1,149 @@
...
+Discuss prior art, both the good and bad.
+
+- [RFC #45](0045-launcher-arguments.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+
+Should we go with the alternative proposal or should we make a backwards incompatible change for a possible improvement in user experience.
```
jromero, line 139: Given that we have a buildpack API which would affect the behaviour along with the fact that we can guide buildpack authors via [migration guides](https://buildpacks.io/docs/reference/spec/migration/) to the new functionality, I don't think this would be as impactful as perceived.

natalieparellano: Would this also require a bump to the platform API? I'm looking at this line under [launcher](https://github.com/buildpacks/spec/blob/main/platform.md#launcher):
```
MUST append any user-provided <args> to process arguments
```

text/0000-command-overrides.md
```diff
@@ -0,0 +1,151 @@
...
+
+- What is the expected outcome?
+
+The launcher should be able to determine if arguments have been passed by the user or not and if they have, it should override a default list of arguments by the once specified by the user.
+
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal involves adding an changing the `command` key to be a string array instead of a single string. This array would be the command and its associated arguments that would be set by default and any additional arguments passed by the user would be appended to it. The `args` array would refer to a set of default arguments that will be appened to the `command` array if no arguments are passed by the user during runtime.
```
yaelharel, line 36:
```suggestion
The proposal involves adding and changing the `command` key to be a string array instead of a single string. This array would be the command and its associated arguments that would be set by default and any additional arguments passed by the user would be appended to it. The `args` array would refer to a set of default arguments that will be appened to the `command` array if no arguments are passed by the user during runtime.
```

text/0000-command-overrides.md
```diff
@@ -0,0 +1,151 @@
...
+
+Let's see how this proposal will work through an example. Let's say we have a process as defined below - 
+
+
+```toml
+[[processes]]
+type = "hi"
+command = ["echo", "Hello"]
+args = ["World"]
+direct = <direct>
```
yaelharel, line 67: I'm not sure why but `<direct>` is marked in red.

sclevine:
Per working group discussion, potentially blocked on larger changes to the command/args model.

samj1912:
Superceded by https://github.com/buildpacks/rfcs/pull/168. Please follow that instead.

PR #149, opened 2021-03-25
Pack cache options

jromero:
Allow users to configure the various cache options.

[rendered](https://github.com/buildpacks/rfcs/blob/pack-cache/text/0000-pack-cache-options.md)

### DIFF
--- a/text/0000-pack-cache-options.md
+++ b/text/0000-pack-cache-options.md
@@ -27,7 +27,7 @@ Allow users to configure the various cache options.
 * **lifecycle** - Component that executes in a container running all the necessary buildpacks and producing the final app image.
 
 #### Cache Types
-* **build cache** - Cache container `build` layers. That is layers that were created by buildpacks and marked as `build=true` and `cache=true`.
+* **build cache** - Cache containing layers that should be restored on subsequent builds. That is layers that were created by buildpacks and marked as `cache=true`.
 * **launch cache** - Cache containing `launch` layers. That is layers that were created by buildpacks and marked as `launch=true` (regardless of `cache` option).
 
 #### Lifecycle (Platform API) flags
--- a/text/0000-pack-cache-options.md
+++ b/text/0000-pack-cache-options.md
@@ -1,0 +1,161 @@
+# Meta
+[meta]: #meta
+- Name: Pack cache options
+- Start Date: (fill in today's date: YYYY-MM-DD)
+- Author(s): [@jromero](https://github.com/jromero), [@dwillist](https://github.com/dwillist)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+<!-- One paragraph explanation of the feature. -->
+
+Allow users to configure the various cache options.
+
+# Definitions
+[definitions]: #definitions
+
+<!-- Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with. -->
+
+#### General
+
+* **bind mounts** - Directories mounted from the docker host system onto the running container.
+* **named volumes** - Volumes that are referenced by name within docker and not directly associated with the host file system.
+* **lifecycle** - Component that executes in a container running all the necessary buildpacks and producing the final app image.
+
+#### Cache Types
+* **build cache** - Cache container `build` layers. That is layers that were created by buildpacks and marked as `build=true` and `cache=true`.
+* **launch cache** - Cache containing `launch` layers. That is layers that were created by buildpacks and marked as `launch=true` (regardless of `cache` option).
+
+#### Lifecycle (Platform API) flags
+* **`launch-cache` flag** - A flag accepted by the `exporter` phase (and `creator` binary) of the lifecycle to specify a directory where `launch` layers may be stored.
+* **`cache-dir` flag** -  A flag accepted by the `exporter` phase (and `creator` binary) of the lifecycle to specify a directory where `cache` layers may be stored.
+* **`cache-image` flag** -  A flag accepted by the `exporter` phase (and `creator` binary) of the lifecycle to specify an image on a registry where `cache` layers may be stored.
+
+# Motivation
+[motivation]: #motivation
+
+The platform spec currently provides two formats for caching build layers. These formats are a directory or an image.
+
+Additionally, there is an option to specify a _`launch-cache`_ directory to cache launch layers.
+
+> NOTE: Currently _`launch-cache` flag_ is only used by the current implementation of the lifecycle when _`-daemon` flag_ is provided.
+
+Pack, as a consumer of the lifecycle using the platform spec, uses these options in the following ways:
+
+* By default, `pack build` would create two named volumes
+
+<!-- - Why should we do this?
+- What use cases does it support?
+- What is the expected outcome? -->
+
+##### Use Cases
+
+* As a Pack user, I would like to build using a docker daemon while storing _`build` cache_ as an image (aka _`cache-image`_) in a registry. ([buildpacks/pack#1024](https://github.com/buildpacks/pack/issues/1024#issuecomment-762901676))
+* As a Pack user using `pack` on BitBucket Pipelines, I would like to be able to store the _`build` cache_ and _`launch` cache_ on the file system. ([bitbucket discussion](https://community.atlassian.com/t5/Bitbucket-discussions/Using-CNCF-Buildpacks-on-Bitbucket-Pipelines/td-p/1516690))
+* As a Pack user, I would like to be able to specify the name of the volumes used by pack for caching. ([buildpacks/pack#665](https://github.com/buildpacks/pack/issues/665))
+
+# What it is
+[what-it-is]: #what-it-is
+
+<!-- This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users. -->
+
+A new variadic flag that enables configuring different cache types geared towards app developers or operators that are using pack in very specific (and sometimes restrictive) ways that require them to configure caching in a more advanced fashion.
+
+The options available via this new flag would have sensible defaults based on the current behaiviour with the exception of the `--cache-image` which would be deprecated and later removed to reduce confusion. Overall UX would remain the same while allowing for greater flexibility.
+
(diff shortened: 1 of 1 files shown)


jromero:
poke @buildpacks/platform-maintainers

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,222 @@
...
+<!-- Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with. -->
+
+#### General
+
+* **bind mounts** - Directories mounted from the docker host system onto the running container.
+* **named volumes** - Volumes that are referenced by name within docker and not directly associated with the host file system.
+* **lifecycle** - Component that executes in a container running all the necessary buildpacks and producing the final app image.
+
+#### Cache Types
+* **build cache** - Cache container `build` layers. That is layers that were created by buildpacks and marked as `build=true` and `cache=true`.
```
ekcasey, line 30:
```suggestion
* **build cache** - Cache containing layers that should be restored on subsequent builds. That is layers that were created by buildpacks and marked as `cache=true`.
```
Layers don't need to be `build=true` to be cached.

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,222 @@
...
+
+An additional benefit of this format is that it can later be expanded on if more options for a specific format were deamed necessary.
+
+```
+pack build --cache 'type=<type>;format=<format>;<additional-options>' --cache ...
+```
+
+### `cache` flag options
+* _**type**_ - The type of cache to configure [`build`, `launch`] (default: `build`).
+* _**format**_ - The format in which to store that cache [`dir`, `image`, `volume`] (default: `volume`).
```
ekcasey, line 84: I am hesitant to expose the launch cache concept to users. It is a hack necessary to produce adequate performance when exporting to a docker daemon. Instead f exposing this concept to users I would rather drop it entirely if we stop supporting the daemon as an export target (as we have previously discussed).

Making it directly configurable only opens up potential confusion imho, for example if we used this cache during a registry build it would actually decrease performance. Also, neither layer contents or layer metadata are restored from this cache, making it confusing to buildpack authors concerned with the layer reuse lifecycle.

ekcasey: Do we have existing use cases for configuring the launch cache? Could we solve this problem another way?

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,222 @@
...
+pack build --cache 'type=<type>;format=<format>;<additional-options>' --cache ...
+```
+
+### `cache` flag options
+* _**type**_ - The type of cache to configure [`build`, `launch`] (default: `build`).
+* _**format**_ - The format in which to store that cache [`dir`, `image`, `volume`] (default: `volume`). 
+
+###### Format specific options
+
+* `dir`
```
ekcasey, line 88: Should we call this type `bind` instead? And use `source` instead of `path`. If I am familiar with the `--volume` flag syntax, using something slightly different here feels like it would be hard to remember.

If we ever wanted to make the in-container path configurable we could add a `target` option. `path` might be ambiguous if we are setting host and container paths, or a host image and a container path.

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,222 @@
...
+Why should we *not* do this? -->
+
+#### Pros
+
+- The fomat is extensible to allow for additional cache formats or additional options.
+
+#### Cons
+
+- There is added complexity in the syntax since it's freeform.
+    - Given that the target use cases are more advanced this should be too impactful.
```
ekcasey, line 144: Not saying it isn't worth it, but I don't think we should minimize the impact of creating complex custom usage. Ever since we introduce the buildpack URIs I find myself checking the docs frequently because I have trouble remembering the syntax, and I would consider myself an "advanced" user.

text/0000-pack-cache-options.md
dwillist: Think [the linked ticket](https://community.atlassian.com/t5/Bitbucket-discussions/Using-CNCF-Buildpacks-on-Bitbucket-Pipelines/td-p/1516690) would require the launch cache to be configurable. We could potentially work around this by making the following changes in pack:
- Explicitly create a launch cache volume before running any of the lifecycle phases.
  - If this errors out because of permissions just ignore it and continue on without a launch cache
  - If it works yay mount it into the lifecycle containers!
- Force people to use image caches in these environments (this is already an ask [here](https://github.com/buildpacks/pack/issues/1024))

jromero: As @dwillist mentions, we _need_ to expose users to the launch cache in order to be able to configure it for very specific use cases. Another example just came up in pack-orb (https://github.com/buildpacks/pack-orb/issues/25#issuecomment-824086257).

> The reason for the current requirement appears to be on trying to cache the docker volumes created. Once we have a way to configure cache volumes via pack, we should be able to remove that dependency.

I'm on the boat for removing it and IMO that is one of the benefits of this proposal. In the future, the `type` option would likely not exist.

text/0000-pack-cache-options.md
jromero: That's interesting and honestly confusing but I'll take your word for it. :)

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,222 @@
...
+- If additional options are necessary, a new flag must be introduced.
+
+# Prior Art
+[prior-art]: #prior-art
+<!-- 
+Discuss prior art, both the good and bad. -->
+
+* [Docker CLI `--mount` flag](https://docs.docker.com/storage/bind-mounts/):
+    `--mount type=bind,source="$(pwd)"/target,target=/app`
+
```
jromero, line 206:
```suggestion
* [Docker CLI `--mount` flag](https://docs.docker.com/storage/bind-mounts/):
    `--mount type=bind,source="$(pwd)"/target,target=/app`
* [BuildKit `--output` flag](https://github.com/moby/buildkit#output):
    `--output type=image,name=docker.io/username/image,push=true`
```

text/0000-pack-cache-options.md
jromero: I'm okay with this. I opted for `dir` and `path` to make it the most understandable for certain users. My thought process involved the idea that we shouldn't expose the internals (where the cache dir would be mounted) but that may be wrong long-term.

Additionally, I see the value of aligning it with pre-existing concepts.

nebhale (approved):

text/0000-pack-cache-options.md
jromero: Don't disagree but would you rather the alternative be used or is there another proposed option?

jkutner (approved):

jromero:
@dfreilich Could you please review this RFC? Your input would be greatly appreciated.

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,224 @@
...
+    * _**name**_ - Name of the volume to use to save as cache. (default: `pack-<type>-<app-uuid>`)
+
+##### Examples
+
+```bash
+# specify named volume build cache
+pack build --cache 'type=build;format=volume;name=my-build-cache'
+
+# specify bind mount cache (relative to current working directory)
+pack build --cache 'type=build;format=bind;source=./my-build-cache'
```
dfreilich, line 102:
```suggestion
pack build --cache 'type=build;format=bind;name=./my-build-cache'
```
Where possible, I'd like to keep the same option names for all options (if that's sensible), to minimize the loss factor when switching cache types.

text/0000-pack-cache-options.md
```diff
@@ -0,0 +1,224 @@
...
+
+### `cache` flag options
+* _**type**_ - The type of cache to configure [`build`, `launch`] (default: `build`).
+* _**format**_ - The format in which to store that cache [`bind`, `image`, `volume`] (default: `volume`). 
+
+###### Format specific options
+
+* `bind`
+    * _**source**_ - Path to the local directory to use as cache.
+* `image`
```
dfreilich, line 90: And, to clarify (per https://github.com/buildpacks/pack/issues/1024#issuecomment-762901676), this image could be a local image, or would it still need to be a registry image?

dfreilich (approved):
Overall, I think this strikes a good balance of functionality and expansiveness, while also paring down our flags a bit (which is a good thing, to my mind). 

My only concern is that we be extraordinarily verbose when describing the options, even in the `pack build` description, because I personally find `docker mount...` options to be confusing, and I don't want this to end up confusing users and never getting used as a result.

text/0000-pack-cache-options.md
jromero: The linked comment does bring up the _future_ potential for an additional flag similar to [buildkit](https://github.com/moby/buildkit#output):

```
push=<boolean>
```

ie:

```
type=build;format=image;name=io.docker.io/myorg/my-cache:build;push=false
```

text/0000-pack-cache-options.md
jromero: I can't seem to find @ekcasey's comment that inspired the change from `format=dir;path=./my-build-cache` to match Docker's [`mount`](https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag) format.

Personally, I think it's more confusing to use `name` because it doesn't clearly depict what "name" means in this context whereas `source` is pretty clear, especially if you are familiar with `mount`.

jromero:
RE: @dfreilich 

> My only concern is that we be extraordinarily verbose when describing the options, even in the `pack build` description, because I personally find `docker mount...` options to be confusing, and I don't want this to end up confusing users and never getting used as a result.

This is also a [concern of mine](https://github.com/buildpacks/rfcs/blob/pack-cache/text/0000-pack-cache-options.md#cons) but with the lack of good [alternatives](https://github.com/buildpacks/rfcs/blob/pack-cache/text/0000-pack-cache-options.md#alternatives) and the [motivation](https://github.com/buildpacks/rfcs/blob/pack-cache/text/0000-pack-cache-options.md#motivation) for this feature as an "advanced" feature it makes me feel more comfortable with the change.

I'm completely open to any _better_ proposed alternative.

text/0000-pack-cache-options.md
ekcasey: Could we do a hybrid of this proposal and the listed alternative and remove `type` from this syntax. e.g. provide `--build-cache` and `--launch-cache` (or `--layers-cache` and `--daemon-cache`?) flags? I think this would allow `pack` to provide more meaningful help text for each type of cache and would allow us to more explicitly remove concepts like the launch cache, if we ever wanted to do that.

dwillist (approved):

text/0000-pack-cache-options.md
jromero: @dwillist / @dfreilich what do y'all think about this proposal? I'm okay with `--build-cache` and `--launch-cache` and agree with the sentiment.

hone:
@dfreilich @dwillist are there any issues needed for this?

PR #151, opened 2021-04-01
Add BOM to layer content metadata

samj1912:


text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+Keep it as it is and provide a similar functionality in language bindings like `libcnb`. Currently the layer content metadata files are already being used to store and regenerate the `BOM` in buildpacks from the `paketo` project.
+
+# Prior Art
+[prior-art]: #prior-art
+
+<!-- TODO -->
```
samj1912, line 80: @ekcasey can you point to the examples you were talking about in paketo where you store the BOM in layer metadata and then restore it in the next build?

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,106 @@
...
+
+Often times `BOM` describes the artifacts present in the layer. Having this present in the `launch.toml` or `build.toml` which are ephemeral between builds means that the `BOM` has to be generated for each build, even if the layer itself has not changed and is cached. 
+
+This leads to surprising behaviors for users at times when they may be generating a `BOM` as part of the layer generation process and may find that their `BOM` is missing if the `BOM` generation process is not called when the layer is assumed to be cached.
+
+Having the `BOM` be a part of the layer content metadata also means that we can pin-point which layer contributed a BOM entry without having to implement extra logic in downstream buildpacks.
+
+- What is the expected outcome?
+
+The lifecycle is able to generate the final `BOM` by combining the `BOM` entries from top level metadata files like `launch.toml` and `build.toml` and also the ones from `<layer-name>.toml` which handling layer caching correctly i.e. if the layer is cached and re-used in the new build - the `BOM` associated with that layer will also be re-used.
```
natalieparellano, line 31: Would it make more sense to remove `bom` entirely from `launch.toml` and `build.toml`? What would be the advantage in having it defined in more than one place?

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,106 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+The layer content metadata files would be updated to include - 
+
+```
+[[bom]]
+name = "<dependency name>"
+
+[bom.metadata]
```
natalieparellano, line 42: Should we have `bom.build` and `bom.launch`?

natalieparellano: Ohh I see - https://github.com/buildpacks/rfcs/pull/151/files#diff-d31996aeb587aab806ae658bfb0199878b47d25545090827fa6b81723ca2d033R59 handles build vs. launch dependencies.

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,106 @@
...
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image.
+If the layer has `types.build` set to `true` then all the `bom` entries will be contributed to the build environment.
+If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file. If `types.cache` is set to `false` all the `bom` entries from the previous build must be cleaned before/during the export.
```
natalieparellano, line 61: I'm having a little trouble understanding this case. What would happen if `launch=true` `cache=false`?

text/0000-bom-in-layer-metadata.md
samj1912: I guess one reason to have the buildpack bom is so that bom that is not associated with a layer still has a place to stay. For eg app related bom?

text/0000-bom-in-layer-metadata.md
samj1912: Yeah, I imagined if the bom describes a layer then the bom type should be derived from layer type. Makes it easier to reason about.

text/0000-bom-in-layer-metadata.md
samj1912: Added some clarification - hope that helps. LMK if it is still ambigious/lacks clarity.

text/0000-bom-in-layer-metadata.md
natalieparellano: I see. I am just curious, what would be an example of a BOM entry that isn't tied to a layer?

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+  
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. `BOM` entries that are not related to a specific layer should go in the buildpack level `BOM` entries in `launch.toml` or `build.toml`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be contributed to the build environment.
```
natalieparellano, line 60:
```suggestion
- If the layer has `types.build` set to `true` then all the `bom` entries will be written to `report.toml`.
```

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+  - metadata MAY contain additional data describing the dependency.
+  
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. `BOM` entries that are not related to a specific layer should go in the buildpack level `BOM` entries in `launch.toml` or `build.toml`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image.
```
natalieparellano, line 59:
```suggestion
- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image in `io.buildpacks.build.metadata` and `<layers>/config/metadata.toml`.
```

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+- If the layer has `types.build` set to `true` then all the `bom` entries will be contributed to the build environment.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
```
natalieparellano, line 62:
```suggestion
```

natalieparellano: (duplicate)

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. `BOM` entries that are not related to a specific layer should go in the buildpack level `BOM` entries in `launch.toml` or `build.toml`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be contributed to the build environment.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
```
natalieparellano, line 61: This part is still a bit confusing to me. Maybe it makes sense to describe the state of the world at two points: when the `<layer>.toml` is restored, and after the build is completed.

I would think when the `<layer>.toml` is restored, it should contain the BOM entries that were there before. At the end of the build, if the layer is not re-used, the BOM entries should be cleared.

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
+- If `types.cache` is set to `false`, but `types.launch` is set to `true` before a rebuild, the layer content metdata file should be restored without the `BOM` section. If the buildpack sets `types.launch` to `true` after the re-build, and leaves an empty layer directory AND an empty BOM section, the BOM should be restored from the previous app image along with the layer. If the buildpack deletes the `<layer-name>.toml` file or fails to set `types.launch` to true, the `BOM` related to that layer should be cleared. If the buildpack leaves a non-empty layer directory or a non-empty bom section, the value of `bom` after the re-build should be used.
```
natalieparellano, line 63: What is the advantage of restoring `<layer>.toml` without a BOM? I would think the information contained there could be useful to the buildpack in deciding whether to re-use the layer.

natalieparellano: My instinct is that the lifecycle should just trust that whatever BOM is present in the metadata file at the end of the build is the one that the buildpack intended to contribute. But maybe I'm missing something...

text/0000-bom-in-layer-metadata.md
natalieparellano: Perhaps it's worth noting, for a layer with `types.cache = true` but `build` and `launch` `false`, there's (currently) no place for that BOM to go except the cache - it doesn't end up in the app or `report.toml`.

text/0000-bom-in-layer-metadata.md
samj1912: Something like a C++ or go buildpack that builds the binary in the app workspace. Any build related BOM for the compiled binary won't be in a layer but would come from the app workspace. Would be weird to create a layer just to add a BOM.

text/0000-bom-in-layer-metadata.md
samj1912: > What is the advantage of restoring `<layer>.toml` without a BOM? I would think the information contained there could be useful to the buildpack in deciding whether to re-use the layer.

I am a bit cautious about this as that would mean we would have both `metadata` and `bom` to store data to make decisions related to `cache` but maybe I am overthinking here.

samj1912: > My instinct is that the lifecycle should just trust that whatever BOM is present in the metadata file at the end of the build is the one that the buildpack intended to contribute. But maybe I'm missing something...

Hmm - the only reason I added this specific behavior is because I wanted to tie the layer's lifecycle to the BOM.

text/0000-bom-in-layer-metadata.md
natalieparellano: Discussed in https://github.com/buildpacks/rfcs/pull/151#discussion_r608809595

text/0000-bom-in-layer-metadata.md
natalieparellano: Could we still remove the `bom` from `build.toml`?

text/0000-bom-in-layer-metadata.md
natalieparellano: I just think it would be a lot easier to understand :)

sclevine (approved):

text/0000-bom-in-layer-metadata.md
samj1912: Updated it, does that look better?

text/0000-bom-in-layer-metadata.md
samj1912: Added this case as well.

text/0000-bom-in-layer-metadata.md
natalieparellano: This looks great!

natalieparellano (approved):

yaelharel (approved):

text/0000-bom-in-layer-metadata.md
ekcasey: The paketo Java buildpacks have some related prior art but not the exact thing I think you are asking for. The paketo Java buildpacks regenerate the BOM everytime even if the layer is reused.

prior art:

1. `libpak` also has logic to autogenerate BOM entries for certain types of layers, and it uses the layer types to determine whether it needs a launch or build BOM entry - https://github.com/paketo-buildpacks/libpak/blob/a7abf79a2a53ada27c56d48db765454405ab8259/layer.go#L139-L148
1. We include the layer name in the BOM metadata.

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+
+[metadata]
+# buildpack-specific data
+
+[[bom]]
+name = "<string>"
+
+[bom.metadata]
+# Additional metadata
+```
```
ekcasey, line 108: I tried to make this a suggestion but couldn't figure out how to put a code block in a suggestion...


A new `layer` key should be added to the `bom` entries in `io.buildpacks.build.metadata` - 
```javascript
{
  "processes": [
    {
      "type": "<process-type>",
      "command": "<command>",
      "args": [
        "<args>"
      ],
      "direct": false
    }
  ],
  "buildpacks": [
    {
      "id": "<buildpack ID>",
      "version": "<buildpack version>",
      "homepage": "<buildpack homepage>"
    }
  ],
  "bom": [
    {
      "name": "<bom-entry-name>",
      "layer": "<layer-name>",
      "metadata": {
        // arbitrary buildpack provided metadata
      },
      "buildpack": {
        "id": "<buildpack ID>",
        "version": "<buildpack version>"
      }
    },
  ],
 "launcher": {
    "version": "<launcher-version>",
    "source": {
      "git": {
        "repository": "<launcher-source-repository>",
        "commit": "<launcher-source-commit>"
      }
    }
  }
}
```


Paketo buildpack add the layer name to the BOM entries so consumers of the BOM know which layer contains what. Do we want to spec this data now that it is available? For entries that are not associated with a layer we could:
1.  omit the `layer` key
2. assume it is in the `app` layer. The drawback here is that a buildpack could theoretically be adding BOM entries describing the stack image itself? But I think we would rather provide a mechanism that allows the stack to BYOBOM entries instead of encouraging this pattern.

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,108 @@
...
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image in `io.buildpacks.build.metadata` and `<layers>/config/metadata.toml`.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be written to `report.toml`.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
+- If `types.cache` is set to `false`, but `types.launch` is set to `true` before a rebuild, the layer content metdata file should be restored with the `BOM` section. If the buildpack sets `types.launch` to `true` after the re-build, the final value of `bom` after the re-build should be used.
+- If neither `types.launch` or `types.build` are set to `true` the `bom` entries will not be exported anywhere. However they will be preserved in the layer metadata file if `types.cache` is set to `true`.
```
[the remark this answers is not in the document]
samj1912: I ended up going with always add to build bom if someone added a bom entry.

text/0000-bom-in-layer-metadata.md
samj1912: Went with option 1

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,162 @@
...
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image in `io.buildpacks.build.metadata` and `<layers>/config/metadata.toml`.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be written to `report.toml`.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
+- If `types.cache` is set to `false`, but `types.launch` is set to `true` before a rebuild, the layer content metdata file should be restored with the `BOM` section. If the buildpack sets `types.launch` to `true` after the re-build, the final value of `bom` after the re-build should be used.
+- If neither `types.launch` or `types.build` are set to `true` but the layer is present during the export phase the `bom` entries should STILL be written to `report.toml` as the contents of this layer may have been used to generate other artifacts in the same buildpack and someone went through the effort of adding these entries to the `bom`.
```
natalieparellano, line 63: I feel that this slightly contradicts the spirit of https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md in that 0052 made it so that buildpacks by "doing nothing" don't get layer re-use. Here, if a buildpack does nothing (doesn't set `types` but also doesn't delete the layer.toml file) the layer is still "there" during export and would be included in the BOM even if it really didn't contribute to the build at all. 

This is sort of a tough spot because we can't really know if a buildpack did use a layer in the **current** build that was restored from a previous build.

text/0000-bom-in-layer-metadata.md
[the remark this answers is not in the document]
natalieparellano: Added a comment here that is (I think) a continuation of this conversation: https://github.com/buildpacks/rfcs/pull/151#discussion_r617727774

nebhale (approved):

text/0000-bom-in-layer-metadata.md
natalieparellano: Summarizing from WG: adding to the build BOM is the right thing to do because even if a layer wasn't actually used, its presence in the build container could still be noteworthy. I think that satisfies the concern :)

jkutner (approved):

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,162 @@
...
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. `BOM` entries that describe contribution to the application directory should go in the buildpack-level `BOM` entries in `launch.toml` or `build.toml`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image in `io.buildpacks.build.metadata` and `<layers>/config/metadata.toml`.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be written to `report.toml`.
+- If the layer has `types.cache` set to `true` and if the layer is reused in the next build, the respective `bom` entries should be re-populated from the cached layer content metadata file.
```
yaelharel, line 61: Two questions:
* In each of these bullets, the two flags that are not indicated, won't influence to what file the `bom` entries will be written to, right?
If this is correct, maybe it's worth to write it down.
* To what file/s the `bom` entries will be written to if the three flags are set to true?

text/0000-bom-in-layer-metadata.md
```diff
@@ -0,0 +1,162 @@
...
+  
+The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. `BOM` entries that describe contribution to the application directory should go in the buildpack-level `BOM` entries in `launch.toml` or `build.toml`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for the generation of the appropriate metadata files and labels given partial `BOM` entries in the `launch.toml`, `build.toml` and `<layer-name>.toml`. The generation of these artifacts given `BOM` entries in `<layer-name>.toml` would be controlled as follows - 
+
+- If the layer has `types.launch` set to `true` then all the `bom` entries will be contributed to the app image in `io.buildpacks.build.metadata` and `<layers>/config/metadata.toml`.
+- If the layer has `types.build` set to `true` then all the `bom` entries will be written to `report.toml`.
```
ekcasey, line 60:
```suggestion
- If the layer directory exists than all the `bom` entries will be written to `report.toml`.
```
This is what I thought you meant by https://github.com/buildpacks/rfcs/pull/151#discussion_r615155684 and the direction I think we should take things in based on one of the conversation in WG. If the layer directory exists, the layer was present during this build, and therefore the entries go into the build bom, (even if it is an ignored layer).

samj1912: This is covered in L63

text/0000-bom-in-layer-metadata.md
samj1912: Added this point at the end.

sclevine:
FCP ending 5/12

hone (approved):

hone:
@samj1912 do you have the case where other buildpacks will read the `<layer.toml>` for the BOM? For instance, a npm buildpack may use the nodejs-engine buildpack's layer to determine the node version installed without having to shell out.

samj1912:
> @samj1912 do you have the case where other buildpacks will read the `<layer.toml>` for the BOM? For instance, a npm buildpack may use the nodejs-engine buildpack's layer to determine the node version installed without having to shell out.

I don't think I imagined such a use case as a part of this RFC. I personally also see this being a tough pattern to use between buildpacks. Because the npm buildpack will need to know both the location of layer.toml and the format of the BOM created by the nodejs-engine buildpack.

 I guess the nodejs engine buildpack could set and env var to point to the layer.toml but in that case it should also be able to set env vars specifying the node version which might be an easier way to retrieve this information.

PR #152, opened 2021-04-05
Add RFC for pack interaction mode

aemengo:
[Readable](https://github.com/aemengo/rfcs/blob/add-pack-interact/text/0000-add-pack-interact.md)

[Mockup](https://storage.googleapis.com/artifacts.cf-garden-windows-dev.appspot.com/pack-interact-edit.mp4)


samj1912:
The mockup looks absolutely amazing! This would be a massive improvement in debugging and introducing buildpacks to new users. 

We have also had a request for a somewhat related feature called hijack - https://github.com/buildpacks/pack/issues/62 that seems to be along similar lines and I suppose the changes needed to display individual buildpack build stages from lifecycle back to the platform would be along the same lines to support a feature like hijack. Is it worth adding it to prior art and the unresolved issues section?

S/N - IMO I think it might be worth splitting the CVE part with pack interact that deals with building process, but that's upto you :)

dfreilich (commented):
Wildly cool concept. I'm having trouble thinking how the CVE detection would work, though - I assume it would be with a specific image? Or is it based on the builder, if it detects the builder is out of date? That would probably require more metadata from builder improvements, letting us know what CVEs it fixed since the previous one, etc...

aemengo:
@samj1912 @dfreilich I thought about it and like your comments.

Although I want it in there, I'm just don't sure how CVE would work yet. We can split it up as a follow-up RFC. Along w/ hijack too.

aemengo:
I gather that it's usually the maintainers that create the issues... If this RFC is accepted:

I would like responsibility for creating the relevant issues. And
I would like responsibility for bringing contributors to work on this.

jromero:
@buildpacks/platform-maintainers  please review (myself included)

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The **casual application developer** has trouble seeing the value proposition of buildpacks. They also encounter a high learning curve when using our tooling. We've seen evidence of this through user interviews and other channels. By providing a more interactive experience, the casual application developer will be more enticed to remain within the buildpacks ecosystem. Pack will act as the gateway tool to more advanced parts of the ecosystem.
+
+# What it is
+[what-it-is]: #what-it-is
+
+When a casual application developer runs `pack interact` on the command-line in the working directory of their application, they are provided with an interactive panel where they can perform activities related to converting code into runnable images. The following **2 user workflows** will be introduced in the interactive panel:
```
jromero, line 30: One of the major things missing from this RFC is how this feature incorporate all the possible options allowed via `pack build`.

For example, how do I provide additional buildpacks, build-time env vars, etc. The full list of options below.

I wonder if an alternative would be to do something like `pack build -i` / `pack build --interact`. I understand that it might defeat some purpose in "interaction" but I think it would greatly simplify the maintenance of the feature.

```
Flags:
  -B, --builder string              Builder image (default "cnbs/sample-builder:bionic")
  -b, --buildpack strings           Buildpack to use. One of:
                                      a buildpack by id and version in the form of '<buildpack>@<version>',
                                      path to a buildpack directory (not supported on Windows),
                                      path/URL to a buildpack .tar or .tgz file, or
                                      a packaged buildpack image name in the form of '<hostname>/<repo>[:<tag>]'
                                    Repeat for each buildpack in order, or supply once by comma-separated list
  -r, --buildpack-registry string   Buildpack Registry by name
      --cache-image string          Cache build layers in remote registry. Requires --publish
      --clear-cache                 Clear image's associated cache before building
  -D, --default-process string      Set the default process type. (default "web")
  -d, --descriptor string           Path to the project descriptor file
      --docker-host string          Address to docker daemon that will be exposed to the build container.
                                    If not set (or set to empty string) the standard socket location will be used.
                                    Special value 'inherit' may be used in which case DOCKER_HOST environment variable will be used.
                                    This option may set DOCKER_HOST environment variable for the build container if needed.

  -e, --env stringArray             Build-time environment variable, in the form 'VAR=VALUE' or 'VAR'.
                                    When using latter value-less form, value will be taken from current
                                      environment at the time this command is executed.
                                    This flag may be specified multiple times and will override
                                      individual values defined by --env-file.
                                    Repeat for each env in order, or supply once by comma-separated list
                                    NOTE: These are NOT available at image runtime.
      --env-file stringArray        Build-time environment variables file
                                    One variable per line, of the form 'VAR=VALUE' or 'VAR'
                                    When using latter value-less form, value will be taken from current
                                      environment at the time this command is executed
                                    NOTE: These are NOT available at image runtime."
  -h, --help                        Help for 'build'
      --lifecycle-image string      Custom lifecycle image to use for analysis, restore, and export when builder is untrusted.
      --network string              Connect detect and build containers to network
  -p, --path string                 Path to app dir or zip-formatted file (defaults to current working directory)
      --publish                     Publish to registry
      --pull-policy string          Pull policy to use. Accepted values are always, never, and if-not-present. (default "always")
      --run-image string            Run image (defaults to default stack's run image)
  -t, --tag strings                 Additional tags to push the output image to.
                                    Repeat for each tag in order, or supply once by comma-separated list
      --trust-builder               Trust the provided builder
                                    All lifecycle phases will be run in a single container (if supported by the lifecycle).
      --volume stringArray          Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.
                                    - 'host path': Name of the volume or absolute directory path to mount.
                                    - 'target path': The path where the file or directory is available in the container.
                                    - 'options' (default "ro"): An optional comma separated list of mount options.
                                        - "ro", volume contents are read-only.
                                        - "rw", volume contents are readable and writeable.
                                        - "volume-opt=<key>=<value>", can be specified more than once, takes a key-value pair consisting of the option name and its value.
                                    Repeat for each volume in order, or supply once by comma-separated list

Global Flags:
      --no-color     Disable color output
  -q, --quiet        Show less output
      --timestamps   Enable timestamps in output
  -v, --verbose      Show more output
```

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. **Break between detection and compilation.** Buildpacks should be known as a tool that writes (90%) of your Dockerfiles for you. You should be allowed to **see and intervene** with the "build plan" autogenerated for you. And when you **(b) build**, you see the **implementation buildpacks** (as steps) being traversed. It is important that a little description is provided for each implementation buildpack to provide insight into what is happening.
+1. **Updates.** One thing Dockerfiles can't do is update hardcoded dependencies. We should be able to show when buildpacks, lifecycle, stack, etc, can be upgraded so that the user gets this sense of value.
```
jromero, line 33: Is "updates" here synonymous with running `pack rebase` or something completely different?

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+- buildpacks/pack: `pack interact` prompts you to choose from suggested builders [epic/pack-interact]
+- buildpacks/pack: `pack interact` prompts you to choose a build tag [good first issue][epic/pack-interact]
+- buildpacks/pack: `pack interact` conducts detect phase and shows interactive panel [epic/pack-interact]
+- buildpacks/pack: `pack interact` shows **image:**, **builder:** info on interactive panel [good first issue][epic/pack-interact]
+- buildpacks/pack: `pack interact` shows **plan:** info on interactive panel [epic/pack-interact]
+- buildpacks/pack: `pack interact` shows when updates are available [epic/pack-interact]
+- buildpacks/pack: `pack interact` allows you to **(u) update** [epic/pack-interact]
+- buildpacks/pack: `pack interact` allows you to **(b) build** [epic/pack-interact]
+- buildpacks/pack: `pack interact` allows you to **(e) edit** build plan before build [status/needs-discussion][epic/pack-interact]
+- buildpacks/pack: `pack interact` allows and prompts you to **(r) rebase** [epic/pack-interact]
```
jromero, line 81: Can we get some clarification on "update" vs "rebase"?

jromero (changes requested):
Really like the idea but would like a little more detail. Added some clarifying questions.

text/0000-add-pack-interact.md
aemengo: > how do I provide additional buildpacks, build-time env vars, etc

I would expect a user specify more details like that through the *(e) edit)* option. As of right now, I don't have a clear idea what it would look like - and I don't want to rush a solution for it. If we think it's out of scope for this RFC, then I'm fine with it.

Regardless, I don't want a user inundated with all the options at the "interactive panel".

> how about `pack build --interact`?

I don't have strong feelings here.

text/0000-add-pack-interact.md
aemengo: **Rebase** meaning when you update the `run:` image of your application
**Update** meaning when you update anything else (lifecycle, buildpack, etc)

I want to draw the distinction for the user because you would do 1 or the other for different reasons. You would _rebase_ for security reasons. And you _update_ primarily to opt-in to potentially new functionality. Furthermore, by drawing the distinction, it bumps the "number special things that buildpacks can do".

text/0000-add-pack-interact.md
aemengo: **Rebase** meaning when you update the `run:` image of your application
**Update** meaning when you update anything else (lifecycle, buildpack, etc)

I want to draw the distinction for the user because you would do 1 or the other for different reasons. You would _rebase_ for security reasons. And you _update_ primarily to opt-in to potentially new functionality. Furthermore, by drawing the distinction, it bumps the "number special things that buildpacks can do".

dwillist:
From working group & platform subteam sync discussion it seems like there is alignment around adding a feature like this to pack. The exact details of an implementation should be left up to subsequent RFC & issues as needed.

text/0000-add-pack-interact.md
hone: @jromero do we need interact to support everything `build` vs `interact` at least initially?

I do wonder for env stuff if we push them to use `project.toml` instead?

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+When a casual application developer runs `pack interact` on the command-line in the working directory of their application, they are provided with an interactive panel where they can perform activities related to converting code into runnable images. The following **2 user workflows** will be introduced in the interactive panel:
+
+1. **Break between detection and compilation.** Buildpacks should be known as a tool that writes (90%) of your Dockerfiles for you. You should be allowed to **see and intervene** with the "build plan" autogenerated for you. And when you **(b) build**, you see the **implementation buildpacks** (as steps) being traversed. It is important that a little description is provided for each implementation buildpack to provide insight into what is happening.
```
hone, line 32: @aemengo Per your definition above, you only mean the list of buildpacks and not the Build Plan from the spec generated by the buildpacks? This seems feature is really cool, but has some large implications of how pack deals with it before handing it off to lifecycle.

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+1. **Updates.** One thing Dockerfiles can't do is update hardcoded dependencies. We should be able to show when buildpacks, lifecycle, stack, etc, can be upgraded so that the user gets this sense of value.
+
+---
+
+This RFC acts as a gateway to subsequent functionality. Follow-up RFCs could include:
+
+1. **CVE detection.** A user should be able to view CVEs that a **(r) rebase** will remove.
+1. **Build Hijack.** A user should be able to open a shell for debugging, into a failed build container.
+1. **Edit Build Plan.** A user should be able to modify a build plan, before building.
```
hone, line 41: Is this the same as the "**Break between detection and compilation.** or something different?

text/0000-add-pack-interact.md
hone: If this is two separate things for clarity in the RFC should this be two different numbered items?

text/0000-add-pack-interact.md
```diff
@@ -0,0 +1,83 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+One can make extensive use of [rivo/tview](github.com/rivo/tview) to provide a UI, and handle user events, on the command-line. In Golang.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- None
```
hone, line 55: One major drawback of this is the maintenance cost. This is not a small feature and provides a whole net new path that will need to tested and maintained.

text/0000-add-pack-interact.md
jromero: But isn't an update the same as a rebuild (aka build). I'm still confused as to why there would be a distinction. From pack's perspective we only have two operations. `build` and `rebase`. An update as you describe it is simply a `build` that points to a pre-existing image name as the output. Are you proposing the introduction of a new operation?

text/0000-add-pack-interact.md
aemengo: Added

text/0000-add-pack-interact.md
aemengo: I'm thinking that it's something different.

* **Break between detection and compilation** just allows you to see and continue with the build plan.
* **Edit Build Plan** is something enabled with **(e) edit**. You perform it during the _break_ above. Since I don't know what it looks like yet, I think it should be reserved for a separate RFC.

text/0000-add-pack-interact.md
aemengo: Agreed. The motivation is ROADMAP: [Configurability](https://github.com/buildpacks/community/blob/main/ROADMAP.md#configurability).

From a user perspective, it's a very natural way to achieve "Better escape hatches, flexibility, and ways of configuring buildpack builds."

If you see a way to decrease the implications that I'm introducing, I'd be very grateful to hear.

text/0000-add-pack-interact.md
aemengo: I'd like your help in being more clear in the RFC. **Rebase** is not specified in that section because it is functionality that's already present. It is by the same logic that **Build** is not specified in the that section of the RFC.

text/0000-add-pack-interact.md
aemengo: > An update as you describe it is simply a build that points to a pre-existing image name as the output.

Yes

> Are you proposing the introduction of a new operation?

Yes. But I admit that I have not made the parameters very clear here. Much like a language package manager can lock dependencies at a specific version, I think it's a good pattern to follow. And these specifc versions can be bumped w/ **(u) update**.

Here's an example:
```
paketo-buildpacks/go-dist:0.3.0
```

changes to 

```
paketo-buildpacks/go-dist:0.3.1
```

aemengo:
Feedback at sync meeting, on 05/12/21, was that this RFC brings more questions than it answers.

As agreed upon, we'll put this RFC into draft mode. And I will make more tightly scoped RFC in its place.

PR #154, opened 2021-04-07
Additions to RFC template

natalieparellano:
[Readable](https://github.com/natalieparellano/rfcs/blob/additions-to-template/text/0000-rfc-release-notes.md)

### DIFF
--- a/text/0000-rfc-release-notes.md
+++ b/text/0000-rfc-release-notes.md
@@ -1,0 +1,82 @@
+# Meta
+[meta]: #meta
+- Name: Additions to RFC template
+- Start Date: 4/7/2021
+- Author(s): natalieparellano
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The existing process for creating text for release notes is largely manual. This RFC proposes adding new fields to the RFC template that could be parsed by (future) automation and added to issues and ultimately release notes via more automation.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this? Automate a manual process that is currently being done by contributors and maintainers.
+- What use cases does it support? Release notes for: spec, lifecycle, pack, etc.
+- What is the expected outcome? More consistency in release notes text and less time spent on this process.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Define the target persona: project contributor.
+
+Example: For opt-in layer caching, we had the following release notes:
+
+* spec: `Buildpacks must explicitly opt-in to layer re-use by appending to <layers>/<layer>.toml (#132, #209, RFC 0052).`
+* lifecycle: `When using buildpack API 0.6 or greater, buildpacks must explicitly opt-in to layer re-use by appending to <layers>/<layer>.toml (#537 by @yaelharel)`
+* migration guide: 
+```text
+Buildpacks must now explicitly opt-in to layer re-use by setting the launch, build, and cache keys in <layers>/<layer>.toml. If buildpacks do not modify <layers>/<layer>.toml, the layer will behave like a temporary directory, available only to the authoring buildpack, existing for the duration of a single build - even if the buildpack in the previous build set any of these keys to true.
+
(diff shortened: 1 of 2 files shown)


text/0000-rfc-release-notes.md
```diff
@@ -0,0 +1,82 @@
...
+cache = true
+EOL
+```
+
+It would be nice if this text was provided by the RFC author in the RFC template, so that when issues are generated from the RFC, the issues would already contain this text. Then when release notes are generated from those issues, the release notes would already contain this text. We could even automate the creation of the migration guides.
+
+Example: In the RFC template:
+```text
+# Release Notes (OPTIONAL)
+Provide a one-sentence summary of the feature that could be used to announce this feature to buildpacks users.
```
ekcasey, line 51: Lot's of RFCs touch multiple components which might need slightly different release notes. For example if an RFC add support for platform provided labels to the lifecycle and `pack` allows those users to set those labels with a flag you might want:
"Allows platform to provide additional image labels"
vs.
"Allows users to set labels during `pack build` with the `--label` flag
vs.
"Supports labels in project.toml`

Most release note autogeneration uses the issue or PR title generate release notes. So maybe https://github.com/buildpacks/rfcs/blob/rfc-issue-generation/text/0000-rfc-issue-generation.md already gives us a way to automate this in a per-component way?

text/0000-rfc-release-notes.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Migration Guide (OPTIONAL)
+Provide a one- to two-paragraph explanation of the changes required to update buildpacks and/or platforms to use this feature. Example: <link to an existing migration guide>.
+```
+
+Note that the text could still be edited in the event that the RFC author didn't foresee everything that would need to be included, as does happen. But it would give us a place to start from.
```
hone, line 57: Are you saying the text should be edited in the existing RFC after close or just as a basis for the final form?

text/0000-rfc-release-notes.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Higher burden of effort on RFC authors (though this could be made optional).
```
hone, line 62: If we want to make it optional, what recourse do we have? Are we still stuck with the mostly manual option? Would only SOME of the RFCs be automated still move the needle enough for you?

text/0000-rfc-release-notes.md
hone: @ekcasey are you suggesting to do this using comments after the fact?

hone:
My biggest concern (though not a strong one) is the risk of this feeling daunting for non regular contributors. I do think if we do this it should be optional. The RFC template is becoming quite large. For everyone, I've heard feedback from multiple people it's already a lot of work to do a RFC, do you imagine adding these sections would add a lot more time?

text/0000-rfc-release-notes.md
natalieparellano: I don't think we should edit the RFC.

text/0000-rfc-release-notes.md
natalieparellano: Maybe not :/ In the end, probably there is always going to be someone crafting text prior to a release. 

I was imagining that this line

> If applicable, provide sample error messages, deprecation warnings, or migration guidance.

WRT migration guidance could become parseable. But to your point, the benefits gained might not be enough to justify the effort in setting this up.

text/0000-rfc-release-notes.md
natalieparellano: Maybe this is something that could be solved by editing generated issue titles after the fact.

If https://github.com/buildpacks/rfcs/blob/rfc-issue-generation/text/0000-rfc-issue-generation.md uses PR title to generate the issue title, then issues would have names like:

* [RFC 0073] Offline Buildpackages RFC
* [RFC 0075] Move analyze phase before detect

which aren't that great for generating release notes, but the maintainers of the component repos could then edit the issue titles so that they are more useful.

I was originally thinking about asking RFC authors to choose more descriptive PR titles, but maybe it's too much. It might be overly cumbersome when writing an RFC.

natalieparellano:
The comments here have got me thinking that there are better ways to tackle this problem. I'm going to close this one for now.

PR #155, opened 2021-04-16
Add a build-write flag for layers

samj1912:
Signed-off-by: Sambhav Kothari <sambhavs.email@gmail.com>


text/0000-build-write-flag.md
```diff
@@ -0,0 +1,116 @@
...
+1. Buildpack #1 provides a python binary that can be used by subsequent buildpacks but is also used in the app image
+2. Buildpack #2 uses the python binary to invoke pip to provide further dependencies.
+
+Unbeknownst to the buildpack author, python has a side-effect on execution where if it's installation directory is writable, it creates new cache objects or updates existing files. As a result, execution of buildpack #2 causes the content of the layers provided by buildpack #1 to change, causing the final layer imageID to change and hampers layer reproducibility and new layers have to be pushed out to the registry.
+
+This could easily be fixed if buildpack #1 only exposed its build layers as read-only to subsequent buildpacks or if it discarded the modifications made by subsequent buildpacks essentially making it "read-only" in the context of the build process.
+
+### Example use case for read and write build layers
+
+There are various use cases when you may want a common workspace that multiple buildpacks can collaborate on during the build process and this workspace needs to be cached, hence it can be a temporary directory.
```
jkutner, line 48: One example might be an android-sdk buildpack that sets up `ANDROID_SDK_ROOT` (pointing to a layer), which a gradle-buildpack may later write into when gradle runs.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,116 @@
...
+# Summary
+[summary]: #summary
+
+This RFC proposes the introduction of a  `build-write` flag in addition to the `build` flag. The idea behind this flag is to expose build layers that are both read-write (if `build-write` is set to `true`) or build layers that are read-only (if `build-write` is set to `false`) for subsequent buildpacks.
+
+# Definitions
+[definitions]: #definitions
+
+Build layers: Layers from a specific buildpack that are available to subsequent buildpacks during the build process.
+layerize: The process of converting a layer directory to a tarball which will be used to create the final application image during the export phase.
```
jkutner, line 20: i don't think we need to go so far as to define what "layerize" does; only that it freezes the layer. It could just copy the directory off somewhere, or use a tempfs, etc

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,116 @@
...
+Layer content meteadata TOML - s
+
+```toml
+[types]
+  launch = false
+  build = false
+  # New flag
+  # If build is set to false, this flag will have no effect.
+  # The default is false.
+  build-write = false
```
jkutner, line 111: alternative:

```
mode = "(r|w)"
```

jkutner (approved):

text/0000-build-write-flag.md
hone: I like SOME definition here, just so I know what the RFC means when the word is used.

text/0000-build-write-flag.md
hone: Would love to see more concrete examples laid out like the Python one you did for "read-only" and Joe's Android example. Adding more flags to `layer.toml` increases the complexity, so for me it needs to be justified.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,116 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle will have to layerize the layers marked `build-write` as `false` during the `build` phase of the lifecycle instead of the `export` phase.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Possibly mixing the concerns of the `build` and the `export` phases.
+
```
hone, line 72: There are already 3 different layer "types", adding a 4th one increases the mental model for buildpack authors.

text/0000-build-write-flag.md
hone: I wonder if we're running into the limits of booleans here. By default it would be "read".

hone:
I think this a good idea, but it does add both complexity and encourages coupling. Another buildpack needs to know how to fetch or find the directory. This can be done via an env var like the android SDK example or just accidental like the Python one. We have the working dir + slices which is the current place that shared writing can happen. I think being able to show why we don't want to pollute this directory could go a long way. There is an inherit idea today that a layer created by a buildpack is "owned" by it and this changes that.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,121 @@
...
+### Example use case for read and write build layers
+
+There are various use cases when you may want a common workspace that multiple buildpacks can collaborate on during the build process and this workspace needs to be cached, hence it can be a temporary directory.
+
+In such a case a buildpack could provide a layer that is marked as `build-writable` i.e. subsequent buildpacks can modify its contents and the final state of the layer at the end of the build process is what is exported out.
+
+One example might be an Android SDK buildpack that sets up ANDROID_SDK_ROOT (pointing to a layer), which a Gradle buildpack may later write into when gradle runs.
+
+Another example can be a CCache buildpack that sets up `CCACHE_DIR` (pointing to a layer), which a CMake buildpack may use for its build cache.
```
ekcasey, line 54: I worry about how this interacts with https://github.com/buildpacks/rfcs/pull/151. If for example a CCache buildpack sets up a `CCACHE_DIR` and the CMake buildpack installs things into it, those things should ideally end up in the build BOM. Should the downstream buildpack then edit the layer content metadata of the ccache layer. Just having to figure out the path to the layer content metadata seems to break some important abstractions.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,121 @@
...
+- Possibly mixing the concerns of the `build` and the `export` phases.
+- Added complexity of more flags for users.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+
+We could change the the layer so that it is owned by root or a different user between the build steps of different buildpacks so that subsequent buildpacks cannot modify the layers created by other buildpacks. This would require elevated privileges during the build phase of the lifecycle.
+
```
ekcasey, line 84: @samj1912 not sure if you brought up an idea similar to this in WG but what if, as an alternative we introduce a new type of buildpack (name sure what to name it 😬) that runs twice during the build phase. It could run wherever it is placed in the order and generate layers, and run again after all the downstream buildpacks have executed, at which point it could write the layer content metadata. The example CCache buildpack mentioned above would be responsible for:
1. Scaffolding the `CCACHE_DIR` at the beginning of the build
2. Understanding the structure of the cache and recording its contents in the BOM at the end of the build

All normal buildpacks would create read-only layers.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,121 @@
...
+# Summary
+[summary]: #summary
+
+This RFC proposes the introduction of a  `build-write` flag in addition to the `build` flag. The idea behind this flag is to expose build layers that are both read-write (if `build-write` is set to `true`) or build layers that are read-only (if `build-write` is set to `false`) for subsequent buildpacks.
+
+# Definitions
+[definitions]: #definitions
+
+Build layers: Layers from a specific buildpack that are available to subsequent buildpacks during the build process.
+layerize: The process of converting a layer directory to a tarball which will be used to create the final application image during the export phase.
```
yaelharel, line 20:
```suggestion
Build layers: Layers from a specific buildpack that are available to subsequent buildpacks during the build process.

layerize: The process of converting a layer directory to a tarball which will be used to create the final application image during the export phase.
```

You can either add an empty line between the definitions or just add them in bullets. Just so they won't appear in one long line.

text/0000-build-write-flag.md
```diff
@@ -0,0 +1,121 @@
...
+
+The lifecycle will have to layerize the layers marked `build-write` as `false` during the `build` phase of the lifecycle instead of the `export` phase.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Possibly mixing the concerns of the `build` and the `export` phases.
+- Added complexity of more flags for users.
+
+# Alternatives
```
yaelharel, line 78: I saw that you wrote under the "spec changes" section that if `build` is set to false, `build-write` will have no effect, but I think it can be confusing.
What do you think about the alternative of changing the `build` flag (or introduce one flag that will be used instead) to have 3 values:
* none (like `build=false`)
* read (like `build=true` and `build-write=false`)
* write (like `build=true` and `build-write=true`).

text/0000-build-write-flag.md
yaelharel: @hone, I agree. Please see my suggestion [here](https://github.com/buildpacks/rfcs/pull/155/files#r631302970).

samj1912:
Per the last RFC deep dive meeting, I have converted this RFC to only cover the "layerization" part. I will create a sibling RFC for "shared-layers" which should be implemented as an atomic change along with this RFC.

text/0000-read-only-build-layers.md
```diff
@@ -0,0 +1,98 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Updates to the lifecycle so that `build` layers created by buildpacks should be layerized immediately after the buildpack that provided the layer is finished with its build process. This way any modification made by the future buildpacks would be ignore while still making this layer available to future buildpacks.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle will have to layerize the layers marked `build` as `true` during the `build` phase of the lifecycle instead of the `export` phase.
```
hone, line 60: Is there a concern that if we "layerize" at this point that a subsequent buildpack may make changes that aren't part of launch since they're "thrown away"?

nebhale (approved):

text/0000-read-only-build-layers.md
```diff
@@ -0,0 +1,98 @@
...
+[unresolved-questions]: #unresolved-questions
+
+The impact this would have on the performance of the export step. The exact details of which parts of the lifecycle build and export phases that need to be changed.
+
+How to handle the cases when users want a common build workspace that is cached. This will be covered in a subsequent RFC and that RFC and this RFC should be implemented as an "atomic" change to avoid existing use cases from being broken without a proper migration path.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+None.
```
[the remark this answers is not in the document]
ekcasey: Another thing to consider is the calculation of the layer diffIDs. We currently do this in parallel while archiving avoiding the need to read the layer data from disk a second time. To avoid degrading the performance of the build we will need to move this calculation to the build phase and record the results for the exporter to use.

text/0000-read-only-build-layers.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+TODO
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+The impact this would have on the performance of the export step. The exact details of which parts of the lifecycle build and export phases that need to be changed.
```
ekcasey, line 91: Given the process of archiving layers in currently sequential, if we simply moving layer archiving from the export to the build phase the performance of a the build operation as a whole will not change. However, the duration of the build phase will increase while the duration of the export phase will decrease. Interleaving the slow of archiving and calculating diff IDs between the execution of individuals buildpacks may impact the perception users have of the performance of the system. Also it will impact the time of an iteration loop on a failed build

text/0000-read-only-build-layers.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+- Why is this proposal the best?
+
+This proposal allows you to achieve the above use cases without changing the expected UID/GID of the layers and without elevated permissions during the build phase.
+
+- What is the impact of not doing this?
+
+We could have unknown layer modifications by buildpacks which is currently a spec violation.
+
```
ekcasey, line 82: Another possible alternative that would avoid changes to the platform spec: Could we check modification time stamps and fail the build if a layer is modified after the buildpack that created the layer has finished executing?

text/0000-read-only-build-layers.md
ryanmoran: I worry about this. We (Paketo Buildpacks) sometimes get reports that a buildpack we maintain is slow to execute when the user is really reporting that parts of the restore or export phases are slow. I'd like to reduce those types of reports if at all possible and I'm concerned that if we are doing this "layerizing" in the build phase we may see an increase in reports that a buildpack is slow when it is really that the "layerizing" process is taking a while to perform its operation.

text/0000-read-only-build-layers.md
samj1912: This would also be a potentially time consuming operation as after each build it would have to recursively check each `layers` directory.

text/0000-read-only-build-layers.md
sclevine: We could layerize previous buildpack layers (non-`build=true`-only) in parallel with the rest of the build process. This would result in net faster builds.

sclevine (approved):

text/0000-read-only-build-layers.md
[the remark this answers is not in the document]
samj1912: Updated the implementation details slightly. The `builder` and `exporter` arguments would remain the same and the `builder` would just output layerized tars to `/layers/exported` which the `exporter` could read from. This would avoid a platform api change. We could re-evaulate this as a part of #145 and change it in the future to be explicitly passed but I would prefer that older platform apis with the newer lifecycle also get this behaviour.

text/0000-read-only-build-layers.md
```diff
@@ -59,6 +59,8 @@ Updates to the lifecycle so that `build` layers created by buildpacks should be
 
 The lifecycle will have to layerize the layers marked `build` as `true` during the `build` phase of the lifecycle instead of the `export` phase.
 
+The layerized output from the `build` phase of lifecycle could be stored at `/<layers>/@exported` (which should not clash with any buildpack IDs) where `<layers>`. The exporter will just read from this to construct the final image.
```
ekcasey, line 62: Can we add a example path to make sure I am understanding correctly, an exported build layer would end up at `<layers>/@exported/<buildpack ID>/<layer>.tar`?

text/0000-read-only-build-layers.md
```diff
@@ -59,6 +59,8 @@ Updates to the lifecycle so that `build` layers created by buildpacks should be
 
 The lifecycle will have to layerize the layers marked `build` as `true` during the `build` phase of the lifecycle instead of the `export` phase.
```
ekcasey, line 60: Should all the necessary layers (anything `cache=true` or `launch=true`) during build phase regardless of whether they are `build=true`? This might be good for two reasons
1. to prevent changes to launch layers from downstream buildpacks (although this is admittedly harder to do accidentally)
2. for consistency. It might be weird to layerize things in two different phases

ekcasey, line 60: We only need to layerize if `cache=true` or `launch=true`. I.E we don't actually need to layerize a `build=true`, `cache=false`, `launch=false` layer because we wouldn't do anything with the resulting layer.

jkutner (commented):
I'm removing my approval for now, because I thought we were working toward a solution that had an escape hatch.

text/0000-read-only-build-layers.md
```diff
@@ -0,0 +1,104 @@
...
+This could easily be fixed if buildpack #1 only exposed its build layers as read-only to subsequent buildpacks or if it discarded the modifications made by subsequent buildpacks essentially making it "read-only" in the context of the build process.
+
+## What is the expected outcome?
+
+The spec and lifecycle is modified to support the above use cases.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Updates to the lifecycle so that `build` layers created by buildpacks should be layerized immediately after the buildpack that provided the layer is finished with its build process. This way any modification made by the future buildpacks would be ignore while still making this layer available to future buildpacks.
```
jkutner, line 55: Can a buildpack indicate that a layer should not be layerized until all buildpacks have run?

If a subsequent buildpack modified a layer from an existing buildpack, can we save that as a new layer? This may be inefficient overall, but it makes it much easy to develop buildpacks in some cases.

jkutner:
I'm revisiting this RFC as I'm struggling to get `pip` to install packages outside of `PYTHONHOME`, while keeping `pip` as a distinct buildpack. Is it possible? Yes absolutely, but this puts a lot of burden on new buildpack authors who are used to working with these tools (`pip`, etc) the way they are meant to be used. By forcing layers to be read-only we are going against the grain.

hone:
> When deemed "ready", a team member will propose a "motion for final comment period (FCP)" along with a disposition of the outcome (merge, close, or postpone).

The proposed changes due to perf reasons are pretty drastic. This RFC seems to have stalled without a decision one way or another. I propose we "postpone" this and create an issue on lifecycle to try to at least warn users when this is happening. It may be expensive perf wise, but it maybe worth spiking and measure how much it would cost. Thoughts @ekcasey @natalieparellano ?

hone:
This is blocked currently on this spike: https://github.com/buildpacks/lifecycle/issues/703.

hone:
Since @ekcasey is on leave, I'd like to reassign this. Do you mind if I assign you @natalieparellano?

samj1912:
Coverting this to a draft while it is still blocked.

natalieparellano:
> The proposed changes due to perf reasons are pretty drastic. This RFC seems to have stalled without a decision one way or another. I propose we "postpone" this and create an issue on lifecycle to try to at least warn users when this is happening. It may be expensive perf wise, but it maybe worth spiking and measure how much it would cost. Thoughts @ekcasey @natalieparellano ?

@hone just to clarify, the spike is measuring the performance impact of notifying or failing if mod times change? Or something else?

mboldt:
I ran a couple experiments for buildpacks/lifecycle#703. Detecting which buildpacks modify which layers using a file system watcher was pretty efficient (added ~0.5 seconds to a ~12-second build of Spring Petclinic REST with Paketo buildpacks). More details and results in the [experiment repo](https://github.com/mboldt/lifecycle-703-read-only-layers).

So, it seems feasible for lifecycle to detect and report a buildpack changing another buildpack's layer.

On the broader topic, I've had some conversations with @yaelharel who is interested and has some good ideas to explore for read-only layers and change detection.

yaelharel:
While @mboldt ran some experiments with a hash function and a file system watcher (as described in [this issue](https://github.com/buildpacks/lifecycle/issues/703)), we talked about some ideas for this RFC:

- Hash - although the experiment that Mikey ran took too long, maybe it’s worth trying a different hash function.
- Git - if we would have Git on our file system, and everything is committed, we would be able to run git diff and see if anything was changed after each buildpacks run.
- Virtualization - use it to run the buildpacks in a way so they won’t have write access to the layers.
- Copy on write - use this mechanism for each buildpack run and then delete the changes so the actual layers won’t change.
- Search based on timestamps - record the timestamp before a buildpack starts running. At the end of the run, check the timestamp of the relevant directories, and if they changed, we can do a BFS/DFS search in order to find out what directories have changed. Probably it’s not a very good idea since buildpacks can modify the mtime information on files.

One tricky part is knowing ahead of time which buildpack should “own” a given layer. From a fresh build, it’s easy to see which buildpack creates a layer. But, with caching/restore, it’s a bit trickier as the layers already exist. We would need to track that somewhere but we think it’s possible to do so.

Another concern about some of the hard-stop methods is that, although layers should be read-only per spec, it is not enforced today, and so could break some buildpacks. If we don’t want to create a breaking change, we think that we can create some kind of a flag that will tell us whether we should only warn or error based on the version.

mboldt:
> One tricky part is knowing ahead of time which buildpack should “own” a given layer. From a fresh build, it’s easy to see which buildpack creates a layer. But, with caching/restore, it’s a bit trickier as the layers already exist. We would need to track that somewhere but we think it’s possible to do so.

This was my ignorance...the [platform API does include the buildpack ID in the layer path](https://github.com/buildpacks/spec/blob/main/platform.md#outputs-3), so this is not a problem as I first thought.

PR #143, opened 2021-03-08
RFC: Guidelines for Accepting Component-level Contributions

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/component-level-contrib/text/0000-component-level-contrib.md).

PR #159, opened 2021-05-12
Buildpack Authors Tooling Sub-Team

ekcasey:
[Readable](https://github.com/buildpacks/rfcs/blob/buildpacks-subteam/text/0000-buildpack-authors-tooling-subteam.md) 

cc @samj1912 @hone


samj1912:
I also had a draft at https://github.com/buildpacks/rfcs/tree/buildpack-authors-sub-team

samj1912:
cc: @dmikusa-pivotal

hone:
@dwillist here's the RFC I mentioned during Office Hours.

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+* [Rust Dev tools team](https://www.rust-lang.org/governance/teams/dev-tools)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Who are the maintainers?
```
hone, line 119: I talked with @jkutner, and I'll take this on as a maintainer from the core team with @samj1912 as the first contributor.

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+## Language Binding Responsibilties
+These are the responsibilities for maintaining any language binding:
+1. Implementing new APIs in a timely manner
+1. Release process, planning and updates
+1. Triaging issues and supporting users
+1. Documentation
+1. Buildpack Examples using these bindings
+1. Planning and executing improvements
+
+Today, this means taking ownership over the `libcnb` repo doing the above for it.
```
dwillist, line 72: I wonder if it might be beneficial to have a `staging` org for language bindings. This would give us the opportunity to see if they are mature enough to satisfy at the above requirements & also give them some additional visibility.

hone: @dwillist a good thought and maybe not specific to language bindings. I would think any code base coming in would make sense here and if it's something we create how do we experiment / test something without it requiring the same level of rigor we have for our existing software like `pack`?

dmikusa-pivotal:
Awesome! Super excited about this.

jromero (approved):

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Inline Buildpack Responsibilites
+The buildpack sub-team is responsible for sheparding and maintaining [Inline Buildpacks](https://github.com/buildpacks/rfcs/blob/main/text/0048-inline-buildpack.md).
```
dfreilich, line 75: What, concretely, does this mean at this point? Is this about the development of it? What about the pack and lifecycle aspects of inline buildpacks?

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+* [Rust Dev tools team](https://www.rust-lang.org/governance/teams/dev-tools)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Who are the maintainers?
+- Would this sub-team own the pack buildpack author tooling (like `pack buildpack new` and inline buildpacks)?
```
dfreilich, line 120: I think that partitioning aspects of `pack` is likely to cause things to slip between the crack. If anything, making `pack buildpack new` a very slim interface to some `libcnb` function that is called would be fantastic.

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+- What should the team name be?
+
```
dfreilich, line 122: Would this team also own the `samples` repo, at least with regards to the buildpacks? That is what comes to mind for me. I can see arguments both ways – perhaps the buildpacks sub-team only owns samples that use their lib bindings, or maybe they own all samples entirely (in which case, Windows is a complicated aspect to throw at a new team).

dfreilich (approved):
Glad to see more sub-teams joining! 



My main question at this point is about the `samples` repo - I'd love to see that explicitly called out and discussed.

text/0000-buildpack-author-subteam.md
hone: yeah, I think this team probably more than others may be doing cross team work whether in `pack` or `lifecycle` like you mentioned. I'm not entirely sure how we're going to thread that needle, but one way for `pack buildpack new` would be this sub-team would own the template repo and just PR stuff into `pack`, but ultimately platform would own the CLI side.

text/0000-buildpack-author-subteam.md
hone: @samj1912 and I talked about taking that in at least for the language binding case. I'd love to have input from @jromero, @jkutner, and the rest of the learning subteam that currently owns it today. We also talked about having an `exmaples/` directory in each language binding repo similar to [serenity-rs](https://github.com/serenity-rs/serenity/tree/current/examples).

samj1912: We also discussed this in the learning team sync last week, I think one ways @hone and I were thinking of making this work is that the language bindings could have small composable examples for buildpacks (like how to set env variables, how to create processes) as a reference on how to use these language bindings and is kept up to date with the latest versions of the BP api. Whereas samples could house some higher level e2e buildpack implementations which combine these concepts, for example our sample Java or bundler buildpacks. People could use the examples as a reference and samples as inspiration. We are hoping this could be a combined effort with the learning team.

text/0000-buildpack-author-subteam.md
samj1912: This was another aspect we were thinking about. I was hoping pack could simply house a templating library like https://github.com/tmrts/boilr or https://github.com/cookiecutter/cookiecutter that could just reference the latest version of the official templates.

That way the templating logic in pack could be generic and the BP author team could be responsible for the upkeep of the templates. This could also allow pack to allow using community maintained templates.

Just a thought - - nothing concrete but I imagine this could provide a nice separation of concerns :)

cc @jromero and @jkutner (I believe you created the original rfc around this)

text/0000-buildpack-author-subteam.md
samj1912: I think in this case we meant to concentrate more around gathering feedback around the inline buildpacks API and trying to make it more user friendly over time. (Or possibly supplementing it with a dsl like packfile).

This again would need to be done in collaboration with the implementation and platform teams.

text/0000-buildpack-author-subteam.md
samj1912: Should we also call this out in the rfc for accepting project level contributions once it's put in (I assume) the community repo? https://github.com/buildpacks/rfcs/pull/143

samj1912 (approved):

text/0000-buildpack-author-subteam.md
dwillist: @samj1912 That sounds like the perfect place IMO.

text/0000-buildpack-author-subteam.md
ekcasey: I think we should also include myself or @nebhale on the theory that someone who has experience maintaining the libcnb should be involved.

text/0000-buildpack-author-subteam.md
ekcasey: If the implementation of inline buildpacks in is pack (the current plan) I don't think the word maintaining is appropriate here. Components have maintainers but features shouldn't.

Maybe this responsibility belongs in "Collaboration with other teams"?

text/0000-buildpack-author-subteam.md
ekcasey: I agree with @dfreilich that responsibilities should be assigned at the component level. But as @samj1912 mentioned would could certainly extract buildpack author focused tool from pack (libraries or resources like templates) and give this new team responsibility for them.

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+
+If this team isn't made, `libcnb` will continue to live in the implementation team. That sub-team can then staff more focused contributors/maintainers for maintaing that library. Each other "Buildpack Author" focused project/tool will need to find an appropriate home. This will fragment and hurt the focus of having a single team built around it.
+
+## Distribution Team
+
+The distribution team already handles a part of the Buildpack Author experience which is publishing a buildpack and is currently a fairly small team. This would let the two sub-teams combine with expanded scope. This isn't a perfect match though, since the distribution team is well scoped at this point focusing on the registry, interacting with it as users, and operating it.
```
jkutner, line 109: If having lots of a small teams is a concern, I think this is a totally acceptable alternative. I don't have a strong opinion either way, but I'd probably start with this, and then split the team into two as needed.

jkutner: I suppose the question is whether or not the potential maintainer and contributors to "Buildpacks Author" team are comfortable also maintaining the Registry.

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
+# Meta
+[meta]: #meta
+- Name: Buildpacks Author Sub-Team
```
jkutner, line 3:
```suggestion
- Name: Buildpack Authors Sub-Team
```

text/0000-buildpack-author-subteam.md
```diff
@@ -0,0 +1,125 @@
...
+- Author(s): [@ekcasey](https://github.com/ekcasey), [@hone](https://github.com/hone), [@samj1912](https://github.com/samj1912)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the creation of a "Buildpacks Author" sub-team in the project governance with a charter of maintaining buildpack author tooling for creating buildpacks.
```
jkutner, line 14:
```suggestion
This RFC proposes the creation of a "Buildpack Authors" sub-team in the project governance with a charter of maintaining buildpack author tooling for creating buildpacks.
```

jkutner (changes requested):

text/0000-buildpack-author-subteam.md
samj1912: And vice-versa too right? If the distribution team is comfortable maintaining the repositories related to bp author tooling.

jkutner: Oh for sure, but it's just me and @elbandito, so I'm comfortable vouching for our comfort :)

text/0000-buildpack-author-subteam.md
hone: Sounds good to me. I do think it's more sheparding/championing issues/concerns which include some implementation peaces.

dwillist (approved):

hone:
@fg-j the "congress" a great idea, but I'm not sure making them maintainers would align b/c I wouldn't want to beholden them into working on libcnb or other tools if they don't actually want to/use it. Maybe the congress is just a thing the sub-team creates and holds as a place to have stakeholders with a diverse group during the sub-team sync. Ultimately, I'd love for other buildpack authors to see this sub-team as their champion within CNB.

If they have time and interest, there is always room on the team for people who want to work on solving buildpack authors problems though!

fg-j:
That makes sense @hone. I'll just say that from my perspective as a buildpack author who is not currently on a CNB subteam, this Buildpack Authors subteam will only be as effective as its outreach to buildpack authors 🙂 . I'm excited about this subteam helping facilitate flow of feedback and info between CNB and implementation projects!

hone:
@fg-j I agree 100%. I would ask, would you be interested in joining the sub-team syncs as a buildpack author?

text/0000-buildpack-author-subteam.md
samj1912, line 3:
```suggestion
- Name: Buildpacks Authors' Tooling Sub-Team
```

Adding this to capture the feedback from the last WG meeting. It was suggested we explicitly add "tooling" to the team name so that it is clear we are not maintaining buildpacks/authoring them but instead providing/maintaining the tooling used by BP authors.

fg-j:
@hone, definitely. And if not me, specifically, I think Paketo maintainers would probably make a point to have someone at those syncs if they started happening.

text/0000-buildpack-author-subteam.md
hone: I'd of course would love having you or @nebhale on the team, but didn't originally plan for it since I assumed you wouldn't have the time for it.

text/0000-buildpack-author-subteam.md
hone: I clarified this in d979f93. Let me know this makes sense.

text/0000-buildpack-author-subteam.md
hone: I tried to addressed this in d979f93.

hone:
@fg-j I added a section from our discussion in d979f93.

nebhale (approved):

hone (approved):

hone:
I'm proxying my approval for @ekcasey's yes vote she told me over slack DM.

ekcasey:
I can't formally submit an approving review as I originally opened the PR, but just in case there is any ambiguity I am +1 in favor of this proposal.

jkutner (approved):
I'm voting to approve, but I slightly feel we should start by merging this team with Distribution Team, and splitting when we grow to that point. I do think we should revisit the structure after a while to see if this was the right idea.

elbandito (approved):

hone:
@jkutner I wonder if doing the inverse is the right call. Let this sub-team get on its feet with a smaller focus and then expand scope if we still think that's the right idea.

nebhale:
Final Comment Period with merge disposition, closing on 9 June, 2021

PR #160, opened 2021-05-14
Add RFC for pack build visual mode

aemengo:
[Readable](https://github.com/aemengo/rfcs/blob/add-visual-pack-build/text/0000-add-visual-pack-build.md)

[Mockup](https://storage.googleapis.com/artifacts.cf-garden-windows-dev.appspot.com/pack-build-visual.mp4)

### DIFF
--- a/text/0000-add-visual-pack-build.md
+++ b/text/0000-add-visual-pack-build.md
@@ -1,0 +1,80 @@
+# Meta
+[meta]: #meta
+- Name: Add Visual Pack Build
+- Start Date: 2020-05-14
+- Author(s): aemengo
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: https://github.com/buildpacks/rfcs/pull/152
+
+# Summary
+[summary]: #summary
+
+Pack should provide a visual representation of the image build process. Please see [mockup](https://storage.googleapis.com/artifacts.cf-garden-windows-dev.appspot.com/pack-build-visual.mp4) for concept.
+
+# Definitions
+[definitions]: #definitions
+
+* [Pack](https://github.com/buildpacks/pack) - A tool within the buildpacks ecosystem for converting code into runnable images, in a local environment.
+
+# Motivation
+[motivation]: #motivation
+
+The **casual application developer** has trouble seeing the value proposition of buildpacks. They also encounter a high learning curve when using our tooling. We've seen evidence of this through user interviews and other channels.
+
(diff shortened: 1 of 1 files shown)


jromero (approved):

jromero:
@buildpacks/platform-maintainers can you please review?

hone (approved):

jabrown85 (approved):

dfreilich (approved):
Super interesting concept! It's definitely going to be a challenge to integrate it, and it may require us to restructure the codebase to do cleanly, but I can see the appeal, and I hope it'll help onboarding!

sclevine (approved):

text/0000-add-visual-pack-build.md
```diff
@@ -0,0 +1,80 @@
...
+[motivation]: #motivation
+
+The **casual application developer** has trouble seeing the value proposition of buildpacks. They also encounter a high learning curve when using our tooling. We've seen evidence of this through user interviews and other channels.
+
+For most of these developers, pack is their introduction to the buildpacks ecosystem. By providing a visual representation of the build process, they will be introduced to various buildpack concepts, and their relevance to the build process. 
+
+# What it is
+[what-it-is]: #what-it-is
+
+When a casual application developer runs `pack build [--interactive/-i]` on the command-line in the working directory of their application, they are presented with a terminal UI that depicts the build process. The command will depict the build process as four stages.
```
schneems, line 31: The current proposal is dynamic as it's streaming and changing, but it's not terribly "interactive". Is there something else we could re-name this to? If I was a new developer I would expect using an interactive flag might drop me off in some kind of a stateful "wizard" flow or allow me to change some settings and observe their effects in real time. As is, it's a different streaming output.

schneems (commented):
I like the output format idea and the extra information seems useful. I would ideally want all the extra pieces of information also under `--verbose` output. I use that flag for testing buildpacks  when debug mode is enabled and some of the extra info such as how the build order resolves isn't in the current output (https://github.com/heroku/cutlass/blob/main/lib/cutlass/pack_build.rb#L115)

text/0000-add-visual-pack-build.md
aemengo: Thanks for the feedback. I agree it's not "interactive", but the goal is that it eventually would be. However the scope of true interactivity proved to be very large, even for an RFC, so this is supposed to be a first pass.

See here for the original vision: https://github.com/buildpacks/rfcs/pull/152

schneems: Gotcha, is that the eventual goal and this is a stepping stone then? That makes sense as a path forward.

aemengo: > ...this is a stepping stone then?

Yep, absolutely!

PR #163, opened 2021-05-21
Add a proposal for a shared layers directory

samj1912:


text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC introduces the concept of shared build layers which multiple buildpacks can collaborate on/re-use during build time which are not meant to end up in the final application image and are cacheable. Typical use cases include shared caches for tools like CCache, pip cache etc.
+
+> NOTE: This RFC should be implemented as an atomic change alongside RFC https://github.com/buildpacks/rfcs/pull/155
```
hone, line 16:
```suggestion
> NOTE: This RFC should be implemented as an atomic change alongside [RFC#155](https://github.com/buildpacks/rfcs/pull/155)
```

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+All layers created in the `shared-layers` directory are build layers by default. The `cache` layer flag decides if the layer will be cached for the next run or not. This flag will follow the same semantics as [specified here.](https://github.com/buildpacks/spec/blob/main/buildpack.md#cached-layers). The `<shared-layer>.toml` for each layer will be restored with the same semantics as [specified here.](https://github.com/buildpacks/spec/blob/main/buildpack.md#cached-layers) with the `types` table being omitted during subsequent restores to avoid stale layers.
+
+A `shared-layer` is associated with a specific buildpack that created it and it is expected to be wiped if the associated buildpack is not involved during the build process. The `env` directory is co-located with the `shared-layer` so that the associated environment variables that may point to directories in the `shared-layer` follow the same lifecycle as the `shared-layer` itself.
```
hone, line 84: Is there ever a case that if that specific buildpack is gone that subsequent buildpacks would rely/still expect it to be there?

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+# buildpack-specific data
+```
+
+The `env` directory would be used to set build time only variables for subsequent buildpacks. It will follow an API similar to [Buildpack ENV API](https://github.com/buildpacks/spec/blob/main/buildpack.md#provided-by-the-buildpacks)
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+All layers created in the `shared-layers` directory are build layers by default. The `cache` layer flag decides if the layer will be cached for the next run or not. This flag will follow the same semantics as [specified here.](https://github.com/buildpacks/spec/blob/main/buildpack.md#cached-layers). The `<shared-layer>.toml` for each layer will be restored with the same semantics as [specified here.](https://github.com/buildpacks/spec/blob/main/buildpack.md#cached-layers) with the `types` table being omitted during subsequent restores to avoid stale layers.
```
hone, line 82: So a buildpack author would have to set:

```TOML
types.cache = true
```

hone:
I like this version of the proposal a lot better with a dedicated shared layer section.

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC introduces a new top level folder structure called `<shared-layers>` which shall live on the same level as the `<layers>` directory.
+
+The structure would look like - 
+
+```
+<shared-layers> (configurable with `CNB_SHARED_LAYERS_DIR`)
+└── <escaped-buildpack-id> `CNB_BP_SHARED_LAYERS_DIR`
```
hone, line 57: what's the `CNB_BP_SHARED_LAYERS_DIR` mean here?

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+    ├── <shared-layer>.toml
+    └── <shared-layer>
+        ├── <usable-directory>
+        └── env
+```
+
+Each buildpack will be passed an environment variable `CNB_BP_SHARED_LAYERS_DIR` with the location to the shared layers directory for a buildpack.
```
hone, line 64: Are we worried about collision of shared layered directories that we still want them to be "owned" by a buildpack?

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+[summary]: #summary
+
+This RFC introduces the concept of shared build layers which multiple buildpacks can collaborate on/re-use during build time which are not meant to end up in the final application image and are cacheable. Typical use cases include shared caches for tools like CCache, pip cache etc.
+
+> NOTE: This RFC should be implemented as an atomic change alongside RFC https://github.com/buildpacks/rfcs/pull/155
+
+# Definitions
+[definitions]: #definitions
+
+- shared layers: A build-time only layer which can be cached and re-used during builds. This layer is "created" by a specific buildpack but can be used by subsequent buildpacks during the build process. Unlike normal buildpack layers, these layers are not expected to follow the unix directory strcuture and the lifecycle SHOULD NOT modidy variables like `PATH`, `LD_LIBRARY_PATH` etc. with folders present in these layers.
```
jabrown85, line 21:
```suggestion
- shared layers: A build-time only layer which can be cached and re-used during builds. This layer is "created" by a specific buildpack but can be used by subsequent buildpacks during the build process. Unlike normal buildpack layers, these layers are not expected to follow the unix directory structure and the lifecycle SHOULD NOT modify variables like `PATH`, `LD_LIBRARY_PATH` etc. with folders present in these layers.
```

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+- Added maintanence complexity.
+- Additional complexity for buildpack authors to account for.
+
+# Alternatives
```
jabrown85, line 94: Should we list an alternative that is a simple build cache dumping ground (aka not owned by a specific buildpack)? Whatever is there is always restored, no automatic cleanup, etc.

text/0000-shared-layers.md
```diff
@@ -0,0 +1,117 @@
...
+
+- https://github.com/buildpacks/rfcs/pull/155
+- https://github.com/buildpacks/rfcs/pull/145
+
+The plan would be that this `<shared-layers>` directory would be relocated alongside the `<layers>` if the proposal at https://github.com/buildpacks/rfcs/pull/145 is accepted.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How would we accomodate for older platform APIs that may not have a mounted volume for `shared-layers`?
```
jabrown85, line 112: Does it matter if it is a volume or not? As long as the `shared-layers` folder exists and is writeable, things should be fine I think.

text/0000-shared-layers.md
ekcasey: @hone is that notation confusing given the RFC number doesn't match the PR number. E.g. if the referenced PR were merged today it would be RFC 0089.

text/0000-shared-layers.md
ekcasey: I think we may need to explicitly require users to set `build = true` for each of these layers. Otherwise it will be impossible to distinguish between a stale layer and a new `cache = false` shared layer.

text/0000-shared-layers.md
ekcasey: It will be impossible for the exporter to cache shared layers if they are not on a volume that is passed between phases. I think we have a couple options here:

1. Accept that shared layers won't be cached on older platforms and thus newer buildpacks that use this feature might be less performant on older platforms.
1. Have the builder create a shared layers dir under `layers`. It'll be ugly but given the requirement that a buildpack ID is in the form `<namespace>/<name>`, `shared` or `shared-layers` is not a valid buildpack ID and therefore there is no danger of a conflict.
1.  Have the builder copy these layers into the regular buildpack layers dir after the build is complete. This will cause some performance degradation on older platforms, but it probably isn't as dramatic as option 1.
1. Make this an optional feature of the buildpack API. e.g the platform MAY give you a shared layers dir, if not assumed `layers` is fine. This probably foists too much of the complexity onto buildpack authors.
1. Assume that by the time this is implemented the build phase will have taken over responsibility for archiving layers and putting them somewhere the exporter can read them https://github.com/buildpacks/rfcs/pull/155. But that change is a platform API change so it doesn't really solve the problem? Unless we ignore some things that are technically specified in the inter-phase API - if competing implementations of individuals phases is a non-goal, we potentially specified more than we should have...

jabrown85: The volume is only required for untrusted builder paths right? `creator` flows wouldn't have to use a volume. Subsequent builds would restore into the `shared-layers` dir from the cache volume/image right?

text/0000-shared-layers.md
hone: Yeah, it might be but we seem to use "RFC#XXX" for GitHub PR numbers and "RFC XXXX" for RFC numbers.

text/0000-shared-layers.md
hone: yeah, mostly just asking for clarification of the ask. This makes sense to me.

text/0000-shared-layers.md
hone: We should probably answer this in this RFC before it goes through.

samj1912: I went with 2. Although buildpacks published to the registry must be of the form `<namespace>/<name>` the spec doesn't prohibit buildpacks which don't have a namespace. It does however only allow alpha numeric + `./-` characters. So went with `@shared` instead. https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml

samj1912:
Given the recent discussions around different ways of caching things using the buildpacks API and the complexity surrounding them, we are converting this RFC to a draft until further needs/use cases arise.

This might go hand in hand with the asset packages pause.

PR #166, opened 2021-06-02
Add support for a structured BOM format

samj1912:
[Readable](https://github.com/samj1912/rfcs/blob/sbom/text/0000-sbom.md)

Note - This RFC only changes the BOM format for the existing bom tables. The RFCs should also be followed by RFCs that propose -

- BOMs for stacks 
- How to merge stack boms with BP boms during build and rebase

### DIFF
--- a/text/0000-sbom.md
+++ b/text/0000-sbom.md
@@ -64,6 +64,23 @@ The motivation for supporting multiple SBOM formats is because there is currentl
 Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
 It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json` for CycloneDX documents and `spdx.json` for SPDX documents. A buildpack can also output the `bom` in multiple formats i.e. both CycloneDX and SPDX. 
 
+Additionally the buildpack must also indicate the SBOM format(s) it exports using the in its `buildapck.toml`'s `sbom` key in the `buildpacks` table.
+
+The format of this key should look like - 
+
+```toml
+api = "0.x"
+
+[buildpack]
+id = "<buildpack ID>"
+name = "<buildpack name>"
+# This can be an array of supported SBOM formats by the buildpack.
+# Valid array values are "CycloneDX" and "SPDX" as of this RFC.
+sbom = ["CycloneDX"]
+```
+
+This information can be displayed on the registry and `pack inspect-buildpack`.
+
 The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
 
 The lifecycle would be responsible for taking all of the above `bom` files, and putting them in `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` for the layer `bom` files where `type` call be `build` or `launch` or `both`. Buildpack level BOM files will reside at `/layers/config/sbom/<buildpack-id>/build.bom.<ext>` and `/layers/config/sbom/<buildpack-id>/launch.bom.<ext>` for the `launch` and `build` `bom` respectively.
@@ -73,6 +90,7 @@ Additionally the lifecycle will be responsible for creating merged `bom` files, 
 Merging different formats (CycloneDX/SPDX/Legacy CNB format) is a non-goal and will not be supported. Since we will be making the individual `bom` files available as well, if users/platforms wish, they can choose to write conversion and merging logic but this will not be supported in the lifecycle.
 
 Since we will be preserving the original output files and if a buildpack chooses to output in both SPDX and CycloneDX, the lifecycle may eventually add support for SPDX merging in a future version if merging SPDX documents becomes better defined/efficient.
+
 
 # How it Works
 [how-it-works]: #how-it-works
--- a/text/0000-sbom.md
+++ b/text/0000-sbom.md
@@ -6,7 +6,7 @@
 - RFC Pull Request: (leave blank)
 - CNB Pull Request: (leave blank)
 - CNB Issue: (leave blank)
-- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+- Supersedes: [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md), [RFC 0053](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
 
 # Summary
 [summary]: #summary
@@ -19,10 +19,10 @@ This RFC proposes the following -
 # Definitions
 [definitions]: #definitions
 
-- SBOM/BOM: A software bill of materials (SBOM) is a list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
-- CycloneDX: CycloneDX is a lightweight software bill of materials (SBOM) standard designed for use in application security contexts and supply chain component analysis.
-- SPDX: Software Package Data Exchange (SPDX) is a file format used to document information on the software licenses under which a given piece of computer software is distributed.
-- SWID: SWID is a formal industry standard used by various commercial software publishers designed with software inventory and entitlements management in mind.
+- SBOM (Software Bill Of Materials) / BOM: A list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
+- CycloneDX: A lightweight SBOM standard designed for use in application security contexts and supply chain component analysis.
+- SPDX (Software Package Data Exchange): A file format used to document information on the software licenses under which a given piece of computer software is distributed.
+- SWID: A formal industry standard used by various commercial software publishers designed with software inventory and entitlements management in mind.
 
 # Motivation
 [motivation]: #motivation
--- a/text/0000-sbom.md
+++ b/text/0000-sbom.md
@@ -1,0 +1,118 @@
+# Meta
+[meta]: #meta
+- Name: Structured SBOMs
+- Start Date: 2021-06-02
+- Author(s): [@samj1912](https://github.com/samj1912)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following - 
+
+- Use of CycloneDX as the SBOM storage format.
+- Moving the SBOM from a label to a file in a separate layer.
+- Enhancements to the platform spec and pack to allow easy interaction with SBOMs including conversion between major SBOM formats.
+
+# Definitions
+[definitions]: #definitions
+
+- SBOM/BOM: A software bill of materials (SBOM) is a list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
+- CycloneDX: CycloneDX is a lightweight software bill of materials (SBOM) standard designed for use in application security contexts and supply chain component analysis.
+- SPDX: Software Package Data Exchange (SPDX) is a file format used to document information on the software licenses under which a given piece of computer software is distributed.
+- SWID: SWID is a formal industry standard used by various commercial software publishers designed with software inventory and entitlements management in mind.
+
+# Motivation
+[motivation]: #motivation
+
+## Why should we do this?
+
+Quoting [NTIA Whitepaper][NTIA-WP] - 
+> Modern software systems involve increasingly complex and dynamic supply chains. Unfortunately, the composition and functionality of these systems lacks transparency; this contributes substantially to cybersecurity risks, alongside the cost of development, procurement, and maintenance. This has broad implications in our interconnected world; risk and cost affect collective goods, like public safety and national security, in addition to the products and services upon which businesses rely.
+> We propose that increased supply chain transparency through SBOMs can reduce cybersecurity risks and overall costs by:
+> - Enhancing the identification of vulnerable systems and the root cause of incidents
+> - Reducing unplanned and unproductive work
+> - Supporting more informed market differentiation and component selection
+> - Reducing duplication of effort by standardizing formats across multiple sectors
+> - Identifying suspicious or counterfeit software components
+
+Currently since container images built by Cloud Native Buildpacks don't follow a generic file system layout, they may not be easily scannable by container scanning tools. We do however provide a BOM which should greatly help sidestep the entire manual scanning process and speed up things like CVE detection by directly providing the CVE scanner with a BOM.
+
+However, since we currently do not impose any standard for what the BOM should look like and since the metadata table in BOM is a freeform table, it is very hard to have consistent BOMs that can be used by CVE scanners.
+
+Having a structured and consistent SBOM that can be consumed by downstream systems could greatly enhance the security proposition around buildpacks.
+
+## What use cases does it support?
+
+- Consistent SBOM across buildpacks
+- Improved integration with other supply chain analysis and scanning tools
+
+## What is the expected outcome?
+
+- Use of CycloneDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
+- Moving the SBOM from a label to a file in a separate layer.
+- Enhancements to the platform spec and `pack` to allow easy interaction with SBOMs including conversion between major SBOM formats.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` may be `xml` or `json`. This is mostly to allow for different BOM generation tools to be used. This also allows us in the future to add support for `spdx`  tag documents or other file formats if needed.
+
+The above bom documents need to be created in CycloneDX format.
+
+The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
+
+The lifecycle would be responsible for taking all of the above `bom` files, merging it and storing it in `/layers/config/sbom.json`. This file would also be stored in a layer of its own and the lifecycle would be responsible for storing the layer digest in a label `io.buildpacks.bom.digest`.
+
+Platforms like `pack` would be responsible for fetching the `bom` from the above layer and displaying it to the user in a format of their choice via something like `pack inspect-image <image-name> --bom --format=spdx_2_2`.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This is the technical portion of the RFC, where you explain the design in sufficient detail.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+- Why is this proposal the best?
+- What is the impact of not doing this?
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+- What parts of the design do you expect to be resolved through implementation of the feature?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+Does this RFC entail any proposed changes to the core specifications or extensions? If so, please document changes here.
+Examples of a spec. change might be new lifecycle flags, new `buildpack.toml` fields, new fields in the buildpackage label, etc.
+This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
+
+# References
+
+The above RFC was inspired by [NTIA SBOM Formats and Standards Whitepaper - VERSION 202104XX][NTIA-WP] and borrows heavily from it for definitions and references.
+
+
+[NTIA-WP]: https://docs.google.com/document/d/1KEMRrjbMd6FV7-ZFCk-AVVi-QY7qRiYiylxFFUc5Y_8/edit#
+[CDX]: https://cyclonedx.org/
+[SPDX]: https://spdx.dev/
+[SWID]: https://nvd.nist.gov/products/swid
--- a/text/0000-sbom.md
+++ b/text/0000-sbom.md
@@ -13,7 +13,7 @@
 
 This RFC proposes the following - 
 
-- Use of CycloneDX as the SBOM storage format.
+- Use of CycloneDX and SPDX as the SBOM storage format.
 - Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to separate files.
 
 # Definitions
@@ -51,84 +51,35 @@ Having a structured and consistent SBOM that can be consumed by downstream syste
 
 ## What is the expected outcome?
 
-- Use of CycloneDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
+- Use of CycloneDX and SPDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
 - Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to separate files.
 
 # What it is
 [what-it-is]: #what-it-is
 
 Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
-It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json`
-
-The above bom documents need to be created in CycloneDX format.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json` for CycloneDX documents and `spdx.json` for SPDX documents.
 
 The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
 
-The lifecycle would be responsible for taking all of the above `bom` files, merging it while tagging it with the originating layer or buildpack the `bom` came from and storing it in `/layers/config/build.bom.cdx.json` and `/layers/config/launch.bom.cdx.json` respectively for `build` and `launch` boms.
+The lifecycle would be responsible for taking all of the above `bom` files, and putting them in `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` for the layer `bom` files where `type` call be `build` or `launch` or `both`. Buildpack level BOM files will reside at `/layers/config/sbom/<buildpack-id>/build.bom.<ext>` and `/layers/config/sbom/<buildpack-id>/launch.bom.<ext>` for the `launch` and `build` `bom` respectively.
+
+Additionally a stack author may choose to provide a `/cnb/image/merge-bom` binary in the base-image that takes in the above `/layers/config/sbom` folder as its input alongside the location of the final merged `build.bom.<ext>` and `launch.bom.<ext>`. The `merge-bom` binary can read as input the `/layers/config/sbom` to merge the build and launch `bom` as it sees fit and output them to a combined `build.bom.<ext>` and `launch.bom.<ext>`. The lifecycle will be responsible for invoking the `merge-bom` binary if it finds one, otherwise it will skip generating the merged bom.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-The lifecycle would be responsible for reading, merging and restoring the appropriate `bom` files. The lifecycle of these `bom` files would be tied to their respectice metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details. For merging the `bom` files, the lifecycle could replicate or use tooling from [CycloneDX-cli](https://github.com/CycloneDX/cyclonedx-cli) which has a merge operation. The only additional piece of information that the lifecycle would inject are `CycloneDX` [`metadata`](https://cyclonedx.org/use-cases/#properties--name-value-store) the following property keys -
+The lifecycle would be responsible for reading, merging and restoring the appropriate `bom` files. The lifecycle of these `bom` files would be tied to their respectice metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details. For merging the `bom` files, the lifecycle will simply call out the `merge-bom` binary with the appropriate inputs if one is provided, otherwise it will skip generating them.
 
-- `io.buildpacks.bom.buildpack.id` - Buildpack ID for the buildpack that created the BOM
-- `io.buildpacks.bom.layer.name` (Optional) - Set to the name of the layer if the `bom` was associated with a specific layer. 
+The output `bom` files could be uploaded to the registry following a format as proposed in https://github.com/sigstore/cosign/pull/439
 
-Thw lifecycle will put all the `build` related entries in the `/layers/config/build.bom.cdx.json` file and `launch` entries in `/layers/config/launch.bom.cdx.json`
+which looks like - 
+
+![](https://user-images.githubusercontent.com/1714486/125960784-2795510d-511f-40fb-91f2-f9f97eeb573b.png)
 
 # Backwards compatibility
 
-For `BOM` from older buildpacks the following mappings will be defined. Each `BOM` entry will be added as a `CycloneDX` component in the final components list.
-
-For eg. if we have the BOM table like below -
-
-```toml
-[[bom]]
-name = "example"
-[bom.metadata]
-version = "0.2.3"
-foo = "bar"
-foo_array = ["bar"]
-[[bom]]
-name = "example2"
-[bom.metadata]
-version = "0.0.1"
-foo = "bar"
-```
-
-It will be converted to - 
-
-```json
-{
-  "bomFormat": "CycloneDX",
-  "specVersion": "1.3",
-  "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79",
-  "version": 1,
-  "components": [
-    {
-      "type": "library",
-      "name": "example",
-      "version": "0.2.3",
-      "metadata": {
-          "foo": "bar",
-          "foo_array": ["bar"]
-      }
-    },
-    {
-      "type": "library",
-      "name": "example2",
-      "version": "0.0.1",
-      "metadata": {
-          "foo": "bar",
-      }
-    },
-  ]
-}
-```
-
-Since `CycloneDX v1.3` does not define a `metadata` key inside the `component` type, we can simply put the existing `metadata` table inside the `metadata` key in a `component` and use the `loose` JSON schema for `CycloneDX 1.3`.
-
-The `version` key inside the `metadata` table will be handled as a unique case and will be moved to a top level `version` key inside the `CycloneDX` component.
+TODO
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -138,17 +89,7 @@ This RFC adds a fair bit of complexity to the lifecycle and makes it responsible
 # Alternatives
 [alternatives]: #alternatives
 
-## Use a different SBOM format (SPDX)
-
-SPDX is the other leading SBOM format. It is a Linux Foundation project and is another front-runner in the SBOM eco-system. SPDX is heavily focused on compliance related metadata whereas CycloneDX is focused on security and supply chain analysis. SPDX also doesn't have as rich of an ecosystem when it comes to tooling for generating SBOMs as CycloneDX.
-
-For reference see - https://cyclonedx.org/tool-center/
-
-A more complete analysis of leading SPDX and CycloneDX tooling relevant to Buildpacks is documented at [sophiewigmore/bom](https://github.com/sophiewigmore/bom). This includes common SBOM field mapping, example SBOM outputs, scanning times.
-
-## Delegate SBOM merging to platform
-
-The main reason that the lifecycle needs to have a hard dependency on CycloneDX in this proposal is because we propose that the lifecycle merges the SBOM from various buildpacks and layers. If the lifecycle were to not do this, the content of the SBOM can be completely opaque to the lifecycle. This might provide greater flexibility and almost no dependency on the lifecycle side. Instead a platform like `pack` can be responsible for gathering the individual SBOM files and merging them together for the user. This can also be provided as a reusable library/CLI so that other platforms could do the same.
+Only use one SBOM format like `CycloneDX` or `SPDX` and bake it into the lifecycle.
 
 
 # Prior Art
@@ -160,9 +101,8 @@ The main reason that the lifecycle needs to have a hard dependency on CycloneDX 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Implementation details about how the lifecycle should be merging the SBOMs.
-- Moving the SBOM from the `io.buildpacks.build.metadata` label to a file in the output image to avoid large labels
 - Interactions with the stack SBOM and how to represent that and merge it with the Buildpack SBOM.
+- If the project will be responsible for providing some reference implementation of `merge-bom`. If yes, should it support `CycloneDX` as its first format?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-sbom.md
+++ b/text/0000-sbom.md
@@ -75,17 +75,15 @@ api = "0.x"
 id = "<buildpack ID>"
 name = "<buildpack name>"
 # This can be an array of supported SBOM formats by the buildpack.
-# Valid array values are "CycloneDX" and "SPDX" as of this RFC.
-sbom = ["CycloneDX"]
+# Valid array values are sbom media types based on https://www.iana.org/assignments/media-types/media-types.xhtml
+sbom = ["application/vnd.cyclonedx+json", "text/spdx+json"]
 ```
 
 This information can be displayed on the registry and `pack inspect-buildpack`.
 
-The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
+The `lifecycle` binaries would be responsible for taking all of the above `bom` files, and copying them in `/layers/config/sbom/<type>/<buildpack-id>/<layer-id>/bom.<ext>` for the layer `bom` files where `type` could be `build` or `launch`. Buildpack level BOM files will reside at `/layers/config/sbom/<type>/<buildpack-id>/bom.<ext>` where the type could be `launch` and `build` respectively.
 
-The lifecycle would be responsible for taking all of the above `bom` files, and putting them in `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` for the layer `bom` files where `type` call be `build` or `launch` or `both`. Buildpack level BOM files will reside at `/layers/config/sbom/<buildpack-id>/build.bom.<ext>` and `/layers/config/sbom/<buildpack-id>/launch.bom.<ext>` for the `launch` and `build` `bom` respectively.
-
-Additionally the lifecycle will be responsible for creating merged `bom` files, merging `bom` formats of the same type and outputting them at `/layers/config/sbom/build.bom.<ext>` or `/layers/config/sbom/launch.bom.<ext>` respectively. Initially this will only be supported for CycloneDX as it has a well defined and efficient way of merging multiple `bom` files. A reference implementation can be found at the [`cyclonedx-cli`](https://github.com/CycloneDX/cyclonedx-cli) project. In the future we may add support for merging SPDX `bom` files as well. 
+Additionally the `lifecycle` binaries will be responsible for creating merged `bom` files, merging `bom` formats of the same type and outputting them at `/layers/config/<type>/sbom/bom.<ext>`. Initially this will only be supported for CycloneDX as it has a well defined and efficient way of merging multiple `bom` files. A reference implementation can be found at the [`cyclonedx-cli`](https://github.com/CycloneDX/cyclonedx-cli) project. In the future we may add support for merging SPDX `bom` files as well. 
 
 Merging different formats (CycloneDX/SPDX/Legacy CNB format) is a non-goal and will not be supported. Since we will be making the individual `bom` files available as well, if users/platforms wish, they can choose to write conversion and merging logic but this will not be supported in the lifecycle.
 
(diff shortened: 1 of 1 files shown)


text/0000-sbom.md
```diff
@@ -0,0 +1,118 @@
...
+Having a structured and consistent SBOM that can be consumed by downstream systems could greatly enhance the security proposition around buildpacks.
+
+## What use cases does it support?
+
+- Consistent SBOM across buildpacks
+- Improved integration with other supply chain analysis and scanning tools
+
+## What is the expected outcome?
+
+- Use of CycloneDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
```
dmikusa-pivotal, line 55: If we use CycloneDX as the internal storage format, can it be converted to the other formats in a lossless manner? My concern would be a user needing format XYZ, but that format is missing info due to the conversion process.

Have you done any conversion demos/experiments? Like here's the BOM info for a container in CycloneDX format, I can convert this to SPDX and not lose data? Or possibly, can you convert CycloneDX -> SPDX -> CycloneDX and have the original document (although we may not care about that second conversion bac)? I think including a section on conversion details might be helpful in justifying this choice.

text/0000-sbom.md
```diff
@@ -0,0 +1,118 @@
...
+- Moving the SBOM from a label to a file in a separate layer.
+- Enhancements to the platform spec and `pack` to allow easy interaction with SBOMs including conversion between major SBOM formats.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` may be `xml` or `json`. This is mostly to allow for different BOM generation tools to be used. This also allows us in the future to add support for `spdx`  tag documents or other file formats if needed.
```
dmikusa-pivotal, line 63: >where `<ext>` may be `xml` or `json`

It seems that having support for multiple data representations would put extra burden on the lifecycle to understand all of those, convert, merge, and ultimately pull them into one representation in the CycloneDX format.

What about picking one standard format like "json", and pushing the formatting/conversion responsibilities to the integrator/buildpack (`libcnb` can probably help some with minimizing the work on an author too).

For example, if I want to use tool XYZ that only outputs in XML format. I as the buildpack author would be responsible for running the tool to get the SBOM and then converting it to JSON. The benefit of this, IMHO, would be that tool format/conversion issues would be on the user of the tool to fix, not on the lifecycle team.

text/0000-sbom.md
```diff
@@ -0,0 +1,118 @@
...
+
+The above bom documents need to be created in CycloneDX format.
+
+The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
+
+The lifecycle would be responsible for taking all of the above `bom` files, merging it and storing it in `/layers/config/sbom.json`. This file would also be stored in a layer of its own and the lifecycle would be responsible for storing the layer digest in a label `io.buildpacks.bom.digest`.
```
dmikusa-pivotal, line 69: Merging the files makes me worry a bit (maybe unnecessarily?). I start thinking about potential edge cases. Maybe it's unnecessary. What if each layer just had its own individual file. Then no merging is required. Instead of a single file like `sbom.json`, we could have `/layers/config/sbom/<layer>.json`.

I guess the downside of this would be that scanners would need to process multiple files, but that seems like something reasonable. It also seems like it would provide more detail, because the scanner could then know specifically which layer (and could trace back to the buildpack) that installed a particular thing. Maybe they could figure that out another way, but this would make it obvious.

text/0000-sbom.md
nishakm: One thing I would like to point out is that vulnerabilities are considered "dynamic" data. That is, if scanners find a vulnerability, it may not actually be something that manifests because reasons. My suggestion would be to create an internal key-value store containing a union of CycloneDX and SPDX metadata so the formats can be generated based on the use.

text/0000-sbom.md
nishakm: 100% this

text/0000-sbom.md
coderpatros: @dmikusa-pivotal whether or not you can losslessly convert from the CycloneDX format to others depends on what information you have. I wrote a converter for turning a CycloneDX BOM into an SPDX one and had to omit a lot of things simply because SPDX just doesn't support the same things as CycloneDX.

It's probably helpful to understand that the two formats originally came from two different needs. SPDX came from the need to address OSS license compliance. And CycloneDX came about to address security use cases because SPDX didn't.

The CycloneDX spec always supported custom extensions and namespaces. But we've added custom properties in v1.3. So you could potentially use that for storing additional information that either can't already be represented, or that is SPDX/other format specific.

text/0000-sbom.md
coderpatros: FYI CycloneDX and SPDX tooling already exists to convert between JSON and XML representations.

text/0000-sbom.md
dmikusa-pivotal: Thanks for sharing that. I wasn't aware of the two different target use cases. I can absolutely see users wanting information for security purposes but at the same time, I can see them wanting this information for OSS compliance reasons as well. Whatever we end up doing, probably needs to support both use cases. It sounds like the CycloneDX format, used internally could accommodate that, if we make use of custom extensions/namespaces/properties.

I think it would still be interesting to see some examples of how this might work out for an image. Kind of a PoC to prove it out.

text/0000-sbom.md
coderpatros: I don't know enough about this tooling. So I might be missing the mark here.

Layer level BOMs could be beneficial for automated analysis of each layer.

You can also create a nested BOM. With multiple layers as components under the top level container image.

And yet another option is to create separate BOMs for each layer. And then one final BOM that references all the individual layer BOMs.

They all have different pros and cons. But as someone using a particular container I probably don't care as much about the individual layers. I want to know what I'm using. And I most likely want it as a single BOM.

As someone building a container I probably want to know more per layer detail.

So a combination of the two could be beneficial depending on the use cases you are trying to serve.

dmikusa-pivotal: That's interesting and very flexible. I think my primary concern would be for the support of various tools that will consume this. I see it less as humans reading it and much more as automated tools procured by humans reading it.

Are all of these different ways to represent the layers in the  BOM required to be supported by the CycloneDX specification? Or are some optional under the spec? My worry would be picking a way to represent it, then having popular tool XYZ not support that part of the spec so be unable to read the BOM correctly. Thanks.

text/0000-sbom.md
coderpatros: Don't get me wrong. There is some overlap in the formats. My comment was more about where the formats have come from. And to explain that they both have strengths and weaknesses because of that.

CycloneDX already supports OSS license use cases that can be automated. But SPDX also supports some manual audit scenarios like https://spdx.github.io/spdx-spec/3-package-information/#316-comments-on-license and https://spdx.github.io/spdx-spec/4-file-information/#45-concluded-license

And from what I've heard SPDX is starting to work on support for more security use cases. Like last year they added support for package URL as an option for identifying a package.

text/0000-sbom.md
coderpatros: Yeah, most of the time it should be automated tools.

We don't put restrictions on implementations. It's an open standard that anyone can contribute to, adopt and use. I don't think it's reasonable for us to say you must do X, Y and Z if you are implementing CycloneDX.

But in terms of the simplest, most well supported, approach. I would say a flat list of all components in the final container image. This is the approach that Anchore Syft has taken too.

But as a builder, having the dependency graph information as well is very helpful. It helps you track down where exactly that vulnerable component was pulled in from. And if a tool doesn't support dependency graph information that doesn't matter, it's a separate element to the components and is ignored.

samj1912: This is what the lifecycle would be currently responsible for. After the RFC at https://github.com/buildpacks/rfcs/pull/151

the idea was for the lifecycle to automatically combine the BOM while injecting metadata indicating which layer it came from.  This was meant to solve issues like https://github.com/paketo-buildpacks/packit/issues/54 automatically.

IMHO it makes more sense for a container to have a single SBOM which can have layers as its components or can be a flat list. I would assume most container scanning tools currently produce/consume a single SBOM document and dividing it up into layers might make integrations harder.  Keeping it into a single document while still preserving metadata around which layer contributed to the bom entry might be the best of both worlds imho.

text/0000-sbom.md
samj1912: This makes sense to me. I guess JSON might be an easier target give that our current SBOM label is also json. I guess we can adopt a naming scheme as suggested in the NTIA document around 

`sbom.spdx.json` or `sbom.cdx.json` (I believe the cdx one was a suggesting but not a standard?)

samj1912:
Some example SBOMs created using Cyclonedx tools and how they look like after they are converted to spdx - 

Sample node npm project from paketo
https://gist.github.com/samj1912/37b8454ec882fe2dd20133774ff0e283

Sample go mod project from paketo
https://gist.github.com/samj1912/7a5f0e39ce0eff6d0dc3835581d6581a

samj1912:
I tried to use https://github.com/spdx/spdx-sbom-generator. 

This requires me to run `go mod vendor` before I can generate the sbom but the SPDX sbom it produces for go looks like - 
https://gist.github.com/63f36de13740d1b23702dab69af9b072

For `npm` it looks like - 
https://gist.github.com/a78b9b9e219c22b3af5f5bf8e7123640

samj1912:
I could use some help trying this out with other samples repositories and other tools (both on CDX and SPDX side so that we have a clear picture and can recommend an appropriate direct for CNB.)

@dmikusa-pivotal @ForestEckhardt @sophiewigmore if any of you from the paketo side would like to help collaborate on this RFC please lmk :) that would be a great help.

@coderpatros @nishakm your reviews are also extremely valuable. Thank you for taking a look at this RFC. I think with both of you we should have a good coverage across both CDX and SPDX :)

text/0000-sbom.md
sophiewigmore: +1 to @dmikusa-pivotal's point about seeing a demo of some sort of the format conversion process. 

From a buildpack author's perspective, I know the set of metadata I have the ability to pass along to the BoM, but from looking at both formats it's not abundantly clear what the best way to represent that metadata is to minimize data loss.  The benefit of having the data in the format now is that we have the full set of metadata, and other tools can take that raw metadata into the best structure for the requested format. I worry about the data loss that comes along with supporting multiple formats.

text/0000-sbom.md
nishakm: Tern uses an internal data model to represent all the keys and then uses plugins to generate each of the formats based on a internal-model-key to format-key template. Hence there is no data loss when generating each format. The loss happens when reading the formats for obvious reasons.

text/0000-sbom.md
nishakm: From my understanding of buildpacks, each layer represents the output of each buildpack. The reason why I advocate for layer level SBOMs is because it helps with SBOM reuse on the tool side. However, the final output of kpack can combine all the SBOMs into a single one.

nishakm:
> I tried to use https://github.com/spdx/spdx-sbom-generator.
> 
> This requires me to run `go mod vendor` before I can generate the sbom but the SPDX sbom it produces for go looks like 

Referencing https://github.com/spdx/spdx-sbom-generator/issues/1, `go mod vendor` doesn't give you the dependencies of a go project. I am not sure what "lifecycle" does to collect SBOM data, but if it can get the mount point of the container, Tern will be able to generate an SBOM for it.

samj1912:
> > I tried to use https://github.com/spdx/spdx-sbom-generator.
> > This requires me to run `go mod vendor` before I can generate the sbom but the SPDX sbom it produces for go looks like
> 
> Referencing [spdx/spdx-sbom-generator#1](https://github.com/spdx/spdx-sbom-generator/issues/1), `go mod vendor` doesn't give you the dependencies of a go project. I am not sure what "lifecycle" does to collect SBOM data, but if it can get the mount point of the container, Tern will be able to generate an SBOM for it.

I may be wrong but AFAICT tern can only scan container images once they are created, right? The tools we need around sbom generation for buildpacks need to work on a normal directory/filesystem during the container build itself. Is tern meant for such use cases? 

The other issue, although not a blocker/requirement is presence of go tooling for SBOM generation since a lot of the buildpacks are currently written in go and it would be ideal for them to use these SBOM generators as libraries instead of shelling out. (For buildpacks written in bash this is not a concern)

text/0000-sbom.md
samj1912: I have included some of the conversions below for reference. Apart from that, this RFC draft currently delegates the responsibility for conversion between SBOM formats to platforms so we could have the tooling that converts these fields losslessly either in `pack` or cli tools for other platforms like `kp cli`.

nishakm:
> I may be wrong but AFAICT tern can only scan container images once they are created, right? The tools we need around sbom generation for buildpacks need to work on a normal directory/filesystem during the container build itself. Is tern meant for such use cases?

This is actually new. So my apologies for not communicating that better. Tern can now generate an SBOM at container build time using `tern report --live /path/to/mounted/filesystem`.

> 
> The other issue, although not a blocker/requirement is presence of go tooling for SBOM generation since a lot of the buildpacks are currently written in go and it would be ideal for them to use these SBOM generators as libraries instead of shelling out. (For buildpacks written in bash this is not a concern)

This is something we'd love help with! Would you be able to join our upcoming community meeting? https://github.com/tern-tools/tern#community-meetings? We can discuss there.

cc: @rnjudge

text/0000-sbom.md
samj1912: Also to note  - CDX currently has a free for key value store (albeit the values have to be string only)
This can be added to any component in the SBOM - https://cyclonedx.org/use-cases/#properties--name-value-store
```json
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.3",
  "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79",
  "version": 1,
  "metadata": {
    "properties": [
      {
        "name": "Foo",
        "value": "Bar"
      }
    ]
  },
  "components": [
    {
      "type": "library",
      "name": "acme-library",
      "version": "1.0.0",
      "properties": [
        {
          "key": "Foo",
          "value": "456"
        }
      ]
    }
  ]
}
```

This can be used to store all this metadata (this would essentially be the same as storing it in the freeform metadata table in the current BOM spec). Whatever tooling was originally parsing out the free-form BOM table could be accommodated to look at these fields while at the same time we would get benefits from auto-generating the SBOM for components not installed directly through tarballs.

samj1912: From a quick scan at https://spdx.github.io/spdx-spec/  I couldn't find an equivalent in SPDX. Although I may not be looking in the right place.

rnjudge:
> I tried replicating my above experiments with tern but I wasn't getting any output. Maybe I am doing something wrong here as it seems to be expecting layers? Should I reach out to you at a separate slack channel or other communication medium?

Yes, Tern expects a mounted filesystem layer. We have a community meeting tomorrow (Tuesday, June 8th) at 3PM UTC/8AM PST where it might be easiest to discuss this live if the timing works for you. If not, there's a [slack channel](https://github.com/tern-tools/tern/blob/main/CONTRIBUTING.md#slack-channel) for ongoing discussions.

> We would also need the SBOM generation utility to be a standalone binary ideally which can work on various linux/windows operating systems. It looks like that might be tough with `tern` since it is a python tool and requires a python interpreter.

This is not the first time we have heard this feedback and are working on packing Tern as a debian package.

nishakm:
> We would also need the SBOM generation utility to be a standalone binary ideally which can work on various linux/windows operating systems. It looks like that might be tough with `tern` since it is a python tool and requires a python interpreter.

It's quite straightforward to build the Docker image for tern if you are concerned with portability. Even if we were to build a go library, it won't work natively on Mac or Windows because it uses linux syscalls.

samj1912:
Sorry for any confusion around our use case, buildpacks generate sbom during the build process, completely independent of the docker daemon or dockerfiles. This is independent of any post build scanning that a tool might do (which is where I think tern fits in). 

Most of the SBOM tools I tested above do some file system parsing and I don't think they need sys calls, and if they did, it won't work during the build process since the builds happen in an unprivileged environment. 

The sbom generation binary or library will be needed inside the build environment and in some cases the build environment can be pretty minimal which is why a standalone binary produced by go is an attractive option.

The build time user also doesn't have root privileges inside the build environment either so apt or dpkg installations don't work either. 

Either way, from what I can tell tern is meant for scanning containers and not files/directories. (See my above use cases around simply cloning the source repo and running the sbom generation tool)

text/0000-sbom.md
sophiewigmore: RE: defining the fields with someone from the Paketo team, @ForestEckhardt and I can do that with you for sure.

jabrown85:
This may be controversial, but I am not convinced this should be baked into buildpacks as a first-party thing. I am not saying the SBOM isn't useful or that this won't increase the security posture of _some_ buildpack built application images. 

I am hesitant to add a new format that lifecycle/pack has to know about and interact with. I would much rather give buildpacks the ability to add BOM/Licenses/SBOM/Cyclone/CodeCov/Whatever output they want into a `report.toml`-like file. That way paketo buildpacks could use cyclone while another builder uses SPDX.

I'm also not convinced that each buildpack is going to properly report everything it did. Or that another buildpack didn't modify the contents of a previous buildpack's layers. Will security tools end up having to scan the image anyway? 

I'm +1 for giving buildpacks and platforms the low level API tooling to allow them to produce these reports...but I am -1 on baking this in.

xtreme-conor-nosal:
Freely available scanners like Trivy and Grype look for package managers, lockfiles, and manifests inside the image (e.g. Gemfile, dpkg, apt, jar manifests). When a buildpack layer drops a binary into the filesystems, e.g. the java runtime environment, these tools can't identify it.

Enterprise binary analysis scanners like Blackduck can identify common binaries like the JRE, but don't support other languages like node.js. 

For both types of tools, there's a level of fuzzy matching to turn the filepaths, api groups, package names into a CPE's vendor, name, and target fields, which leads to both false negatives and false positives.

BoMs included in a buildpack may have errors, but that's also true of the existing tools. Combining a provided BoM with an extracted BoM could mitigate the shortcomings of both.

xtreme-conor-nosal:
To expand on the CVE scanning use case (focused on the content of the BoM rather than the schema or location):

CVE lookups based on a BoM will get better results with more CPE fields properly filled out, specifically vendor (which in some cases may be the API Group), language, and target_sw. 

As a concrete example, the java buildpack BoM includes:
```
{
      "name": "jre",
      "metadata": {
        "layer": "jre",
        "licenses": [
          {
            "type": "GPL-2.0 WITH Classpath-exception-2.0",
            "uri": "https://openjdk.java.net/legal/gplv2+ce.html"
          }
        ],
        "name": "BellSoft Liberica JRE",
        "sha256": "5bbb7b867ab797ace54aa98a76b7abcca6c5fa01338ee2907e97adb21150c414",
        "stacks": [
          "io.buildpacks.stacks.bionic",
          "org.cloudfoundry.stacks.cflinuxfs3"
        ],
        "uri": "https://github.com/bell-sw/Liberica/releases/download/11.0.11+9/bellsoft-jre11.0.11+9-linux-amd64.tar.gz",
        "version": "11.0.11"
      },
      "buildpacks": {
        "id": "paketo-buildpacks/bellsoft-liberica",
        "version": "8.0.0"
      }
    }
```
The vendor strings (BellSoft, bell-sw), package string (liberica), and publisher (paketo-buildpacks) would all be useful strings for a CVE search, ideally without extracting them from urls.

It would also be helpful if these fields match the upstream package names (e.g. bellsoft-jdk11.0.9+12-linux-amd64.deb installs the package bellsoft-java11). If Bellsoft reports bellsoft-java11 as impacted by a CVE, a scanner searching for "jre" based on the buildpack BoM, unless the buildpack maintainer also reports their own CPE string as impacted.

Additionally, scanners may need hints about the distro to determine what vulnerability feed to query (the ubuntu version used by io.buildpacks.stacks.bionic)

TLDR users of CVE scanners need to be able to retrieve (or assemble) accurate and complete CPEs and OS version from the BoM

coderpatros:
> CVE lookups based on a BoM will get better results with more CPE fields properly filled out, specifically vendor (which in some cases may be the API Group), language, and target_sw.

Depending on the component, vulnerability lookups are better served using package URLs https://github.com/package-url/purl-spec.

It's probably some time away, but the NVD will deprecate CPEs and likely replace them with SWID tags. So it may be worth considering the work being done on software identification here too https://github.com/usnistgov/swid-reg

ForestEckhardt:
> Depending on the component, vulnerability lookups are better served using package URLs https://github.com/package-url/purl-spec.

If you wouldn't mind elaborating on how the `purl` is able to be used to identify vulnerabilities, I understand `cpe` and `swid` but I am falling short on finding the integration point for `purl`. As far as I can understand the `purl` can help identify the package but when it comes to actual using the `purl` for look up I can't find anything.

samj1912:
> > Depending on the component, vulnerability lookups are better served using package URLs https://github.com/package-url/purl-spec.
> 
> If you wouldn't mind elaborating on how the `purl` is able to be used to identify vulnerabilities, I understand `cpe` and `swid` but I am falling short on finding the integration point for `purl`. As far as I can understand the `purl` can help identify the package but when it comes to actual using the `purl` for look up I can't find anything.

I believe here are some integrations \w purl - https://github.com/package-url/purl-spec#users-adopters-and-links

coderpatros:
> > Depending on the component, vulnerability lookups are better served using package URLs https://github.com/package-url/purl-spec.
> 
> If you wouldn't mind elaborating on how the `purl` is able to be used to identify vulnerabilities, I understand `cpe` and `swid` but I am falling short on finding the integration point for `purl`. As far as I can understand the `purl` can help identify the package but when it comes to actual using the `purl` for look up I can't find anything.

For software components, the two main sources of vulnerability information I know of are OSS Index and VulnDB.

OSS Index is provided by Sonatype. And is free to use.

VulnDB is provided by Risk Based Security. And is paid only.

But both sources go beyond the CVEs in the NVD.

Additionally, the centralised nature of CPEs can make them less than ideal to identify a lot of OSS components.

I'm not suggesting to leave them out if they can be accurately represented. Just that purl should be considered too.

text/0000-sbom.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Use of CycloneDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
+- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to a separate file.
+- Enhancements to the platform spec and `pack` to allow easy interaction with SBOMs including conversion between major SBOM formats.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json`
```
sophiewigmore, line 63: What is the reason for even creating the separate files for `layer`, `launch`, and `build` if we are just going to merge them together into one file down the line? 
Is there any intent here to have separate files so that users can also access the separate BOMs?

sclevine: One benefit of keeping them separate: the lifecycle could automatically insert metadata into each CycloneDX SBoM entry that maps the component to a layer digest and buildpack (and/or marks the component as build-only).

sclevine:
Re: support for multiple formats. Since buildpacks that output different formats would be incompatible with each other (i.e., spoil the SBoM), I think we should start with a single standardized format. The proposed design includes `.cdx.` as part of the file extension for every SBoM file. This gives us the ability to support additional formats in a controlled way (e.g., only when it's possible to convert between them, or such that one format is mandatory) in the future without making breaking changes. I support this approach, and suggest that we move forward with CycloneDX unless there is additional feedback.

text/0000-sbom.md
samj1912: Apart from what @sclevine said, each layer in a buildpack could have a different lifecycle (some could be launch only etc.). Keeping BOMs separate is an artifact of tying them to a similar lifecycle as their layers so that the BOMs and the contents they are describing don't drift apart. As for separate BOMs for diffirent buildpacks - It would get difficult for BP authors to write to a shared BOM file and random buildpacks could potentially modify BOMs for other buildpacks.

text/0000-sbom.md
```diff
@@ -0,0 +1,125 @@
...
+[prior-art]: #prior-art
+
+- [#RFC 53](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
+- [#RFC 87](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Implementation details about how the lifecycle should be merging the SBOMs.
+- Moving the SBOM from the `io.buildpacks.build.metadata` label to a file in the output image to avoid large labels
```
natalieparellano, line 109: Is there a way to ensure that the output file (with the BOM) and the image are always linked together? Would we put the SHA of the BOM file in a label on the image?

text/0000-sbom.md
ForestEckhardt: I think that your lifecycle description makes sense for the `build` and `launch` BOM files. I am still confused by the presence of the `layer` BOM file. What purpose does it serve and how would it be utilized?

text/0000-sbom.md
```diff
@@ -0,0 +1,125 @@
...
+[what-it-is]: #what-it-is
+
+Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json`
+
+The above bom documents need to be created in CycloneDX format.
+
+The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
+
+The lifecycle would be responsible for taking all of the above `bom` files, merging it while tagging it with the originating layer or buildpack the `bom` came from and storing it in `/layers/config/bom.cdx.json`.
```
ForestEckhardt, line 67: I am interested in the fact that all of the BOM files are merged into seemingly one BOM file. How do I distinguish things that were in the `build` BOM from those in the `launch` BOM? Is there a distinction at all in the final merged BOM? If there is no distinction why bother separating them above?

text/0000-sbom.md
samj1912: The current RFC just puts the combined BOM in a label as usual. I was hoping to have a separate RFC that moves it to a file which would do something similar (i.e. put the SHA of the BOM file in a label on the image)

text/0000-sbom.md
samj1912: There is a distinction. Each bom component entry has the following metadata properties added to it. https://github.com/samj1912/rfcs/blob/sbom/text/0000-sbom.md#how-it-works

samj1912: > io.buildpacks.bom.buildpack.id - Buildpack ID for the buildpack that created the BOM
> io.buildpacks.bom.types - Set to a comma seperated list of build or launch as applicable.
> io.buildpacks.bom.layer.name (Optional) - Set to the name of the layer if the bom was associated with a specific layer.

text/0000-sbom.md
ForestEckhardt: What goes in the `layer` BOM that does not go into `build` or `launch` BOM? I understand the separation in terms of the "lifecycles" but I am unclear where just the `layer` BOM comes in.

samj1912: Some details and BG are in this RFC - https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md

The idea was for the build and launch BOM to include changes/contribution to the app/workspace and the layer bom to have the layer specific bom.

samj1912: > The buildpack SHOULD NOT add bom to a layer content metadata file describing the contents that are not contributed by the same layer. BOM entries that describe contribution to the application directory should go in the buildpack-level BOM entries in launch.toml or build.toml.

text/0000-sbom.md
samj1912: Given the recent conversation in the OH - I guess we want these two to be separate files instead of just being able to distinguish between the two. Will make that change.

ForestEckhardt:
I have been doing to research into `SWID` and I am unclear where the burden of generating the `SWID` lies. Is it the original dependency maintainer? The consumer? A 3rd party host? Anything that could help clarify would be appreciated.

text/0000-sbom.md
```diff
@@ -0,0 +1,125 @@
+# Meta
+[meta]: #meta
+- Name: Structured SBOMs
+- Start Date: 2021-06-02
+- Author(s): [@samj1912](https://github.com/samj1912), [@sophiewigmore](https://github.com/sophiewigmore), [@ForestEckhardt](https://github.com/ForestEckhardt)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
```
ekcasey, line 9:
```suggestion
- Supersedes: [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md), [RFC 0053](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
```

text/0000-sbom.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following - 
+
+- Use of CycloneDX as the SBOM storage format.
+- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to a separate file.
```
ekcasey, line 17:
```suggestion
- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to separate files.
```

text/0000-sbom.md
```diff
@@ -0,0 +1,125 @@
...
+
+## What use cases does it support?
+
+- Consistent SBOM across buildpacks
+- Improved integration with other supply chain analysis and scanning tools
+
+## What is the expected outcome?
+
+- Use of CycloneDX as the SBOM storage format and updates the the specification and lifecycle to reflect the same.
+- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to a separate file.
```
ekcasey, line 55:
```suggestion
- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to separate files.
```

text/0000-sbom.md
ekcasey: How we will handle SBOM merging when a build contains both older and newer buildpacks that are producing freeform toml and cyclone dx BOMs respectively?

I can't think of many elegant ways to do this so we may need to do something dramatic? e.g. maybe one of
- support two different "merged" boms if we have two types of data
- have a hard cutoff in the platform API. If platform api is 0.x or above we produce a cycloneDX bom with whatever data we have in that format. If it is below we produce a CNB BOM with the same limitations. If we do this we may want to consider allowing newer buildpacks to produce data in both formats for a time to retain functionality on older platforms.

I kinda hate both of those options. Does anyone have a better idea? If not it may be a necessary evil.

ekcasey (commented):
In general I am very enthusiastic about this. However I think migration and compatibility is an issue that needs to be resolved and it might be particularly thorny for this RFC.

text/0000-sbom.md
sophiewigmore: @ekcasey That's a good point I hadn't considered. I don't HATE option 2 that you've proposed, but obviously I'd love to find a state in which we don't have to support both formats. 

I think the idea of supporting two different merged BOMs in bullet # 1 would be super confusing for someone to try to look at. It would also just be kind of hard to use scanning tools if the BOMs came in two formats but were both needed to make up the entire set of metadata.

sclevine: We could convert the existing free-form TOML to CycloneDX by placing the arbitrary TOML in the `properties` field of a `component`.

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+
+This RFC proposes the following - 
+
+- Use of CycloneDX as the SBOM storage format.
+- Moving the SBOM from `<layer>.toml`, `launch.toml` and `build.toml` to separate files.
+
+# Definitions
+[definitions]: #definitions
+
+- SBOM/BOM: A software bill of materials (SBOM) is a list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
```
yaelharel, line 22:
```suggestion
- SBOM (Software Bill Of Materials) / BOM: A list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
```

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+- CycloneDX: CycloneDX is a lightweight software bill of materials (SBOM) standard designed for use in application security contexts and supply chain component analysis.
```
yaelharel, line 23:
```suggestion
- CycloneDX: A lightweight SBOM standard designed for use in application security contexts and supply chain component analysis.
```

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+- SPDX: Software Package Data Exchange (SPDX) is a file format used to document information on the software licenses under which a given piece of computer software is distributed.
```
yaelharel, line 24:
```suggestion
- SPDX (Software Package Data Exchange): A file format used to document information on the software licenses under which a given piece of computer software is distributed.
```

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+- SWID: SWID is a formal industry standard used by various commercial software publishers designed with software inventory and entitlements management in mind.
```
yaelharel, line 25:
```suggestion
- SWID: A formal industry standard used by various commercial software publishers designed with software inventory and entitlements management in mind.
```

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for reading, merging and restoring the appropriate `bom` files. The lifecycle of these `bom` files would be tied to their respectice metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details. For merging the `bom` files, the lifecycle could replicate or use tooling from [CycloneDX-cli](https://github.com/CycloneDX/cyclonedx-cli) which has a merge operation. The only additional piece of information that the lifecycle would inject are `CycloneDX` [`metadata`](https://cyclonedx.org/use-cases/#properties--name-value-store) the following property keys -
+
+- `io.buildpacks.bom.buildpack.id` - Buildpack ID for the buildpack that created the BOM
+- `io.buildpacks.bom.layer.name` (Optional) - Set to the name of the layer if the `bom` was associated with a specific layer. 
+
+Thw lifecycle will put all the `build` related entries in the `/layers/config/build.bom.cdx.json` file and `launch` entries in `/layers/config/launch.bom.cdx.json`
```
yaelharel, line 77:
```suggestion
The lifecycle will put all the `build` related entries in the `/layers/config/build.bom.cdx.json` file and `launch` entries in `/layers/config/launch.bom.cdx.json`
```

text/0000-sbom.md
```diff
@@ -0,0 +1,180 @@
...
+foo = "bar"
+foo_array = ["bar"]
+[[bom]]
+name = "example2"
+[bom.metadata]
+version = "0.0.1"
+foo = "bar"
+```
+
+It will be converted to -
```
yaelharel, line 99: Maybe it's worth adding when it's going to be converted and who is responsible to do so.

ekcasey:
In working group on 07/22/21 we discussed:
1. Giving the lifecycle responsibility for merging the BOM in order to ensure more predictable results, across stacks and platforms.
2. Starting with support for CycloneDX but designing the process by which additional formats including SPDX could be added in the future.
3. Creating one merged BOM per format, explicitly making the merging of different formats (SPDX + CycloneDX + CNB legacy format) a non-goal. In order to create a valid BOM in multiple format buildpacks and stacks would need to generate partial BOMs in multiple formats.

hone:
I agree with @sclevine and @ekcasey that supporting multiple BOM formats adds a lot of complexity. What's the benefit of allowing buildpacks support SPDX and CycloneDX in the initial pass?

samj1912:
I added some more motivation for supporting both SPDX and CycloneDX and also fixed the implementation details. cc: @sclevine @ekcasey @hone

nishakm:
> I agree with @sclevine and @ekcasey that supporting multiple BOM formats adds a lot of complexity. What's the benefit of allowing buildpacks support SPDX and CycloneDX in the initial pass?

Expect the BOM landscape to change in the future. Inherently supporting one format makes it difficult to migrate tooling to support another format that may be incompatible. Also, expect BOM workflows to change. For example, a producer may not want to share some components in their BOM, and a vulnerability found against a BOM may not be exploitable.

rnjudge:
> > I agree with @sclevine and @ekcasey that supporting multiple BOM formats adds a lot of complexity. What's the benefit of allowing buildpacks support SPDX and CycloneDX in the initial pass?
> 
> Expect the BOM landscape to change in the future. Inherently supporting one format makes it difficult to migrate tooling to support another format that may be incompatible. Also, expect BOM workflows to change. For example, a producer may not want to share some components in their BOM, and a vulnerability found against a BOM may not be exploitable.

I will also add that cycloneDX does not compare "apples to apples" with SPDX, i.e. you can't always convert a cdx document to spdx and vice versa.

sclevine:
I think the disadvantage of supporting multiple formats is that each and every buildpack used in a build will need to output the same format for the resulting image to have a complete SBoM. This means that developers who choose to include a utility buildpack that only supports CycloneDX might miss SBoM entries if the rest of their buildpacks only support SPDX.

Additionally, since it isn't possible for the lifecycle to merge SPDX SBoMs automatically, all buildpacks will need to support CycloneDX for the app image to end up with a single SBoM at the end of the build. I'm not convinced that SPDX support will be useful given this limitation.

That said, my preference for CycloneDX is practical and (mostly) based on the current state of tooling. I agree that we can't pick a winner right now, and I think we should proceed with support for both formats.

sclevine (approved):

hone:
@samj1912 sorry, I thought I posted a comment but it might have gotten lost in the shuffle. I've talked with @sclevine last week and he convinced me we need to support more than 1 BOM format. With that in mind, my concern that is addressable is being able to expose as a consumer or platform what BOM formats a buildpack supports. What are your thoughts on making this as part of the `buildpack.toml` or something in the registry?

hone:
@ekcasey will sync up with @samj1912 to maybe open up a PR on handling this during the `restore` phase.

coderpatros:
Just revisiting some of the comments/concerns in this thread.

IMHO, having SBOMs in different formats for different parts of a system might not be ideal. But, it's not a show stopper either. And it's probably going to be the norm across different ecosystems and industries. Unless, of course, there is only one globally accepted format. Which is unlikely anytime soon. The different formats exist because of different use cases.

That's one of the reasons we support external BOM references in CycloneDX. When it comes to generating the SBOM for your final assembled software/system you can indicate that the SBOM for a particular component is external to the SBOM you have in your "hands".

So it's quite feasible to have a CycloneDX SBOM generated as part of the final build that includes references to SPDX SBOMs for parts. Although, it does offload additional work to the consumer.

jromero (approved):

text/0000-sbom.md
```diff
@@ -0,0 +1,155 @@
...
+# This can be an array of supported SBOM formats by the buildpack.
+# Valid array values are "CycloneDX" and "SPDX" as of this RFC.
+sbom = ["CycloneDX"]
+```
+
+This information can be displayed on the registry and `pack inspect-buildpack`.
+
+The lifecycle of the respective `bom` files during rebuilds would be exactly the same as the `bom` table currently. 
+
+The lifecycle would be responsible for taking all of the above `bom` files, and putting them in `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` for the layer `bom` files where `type` call be `build` or `launch` or `both`. Buildpack level BOM files will reside at `/layers/config/sbom/<buildpack-id>/build.bom.<ext>` and `/layers/config/sbom/<buildpack-id>/launch.bom.<ext>` for the `launch` and `build` `bom` respectively.
```
natalieparellano, line 86: Just to call it out explicitly, `type` would be determined by what is the value under the `[types]` table in the `<layer>.toml` file?

text/0000-sbom.md
```diff
@@ -0,0 +1,155 @@
...
+[how-it-works]: #how-it-works
+
+The lifecycle would be responsible for reading, merging and restoring the appropriate `bom` files (the legacy CNB format and the CycloneDX format currently). The lifecycle of these `bom` files would be tied to their respective metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details. 
+
+For merging the CycloneDX `bom` files, the lifecycle could replicate or use tooling from [CycloneDX-cli](https://github.com/CycloneDX/cyclonedx-cli) which has a merge operation. The only additional piece of information that the lifecycle would inject are `CycloneDX` [`metadata`](https://cyclonedx.org/use-cases/#properties--name-value-store) the following property keys -
+
+- `io.buildpacks.bom.buildpack.id` - Buildpack ID for the buildpack that created the BOM
+- `io.buildpacks.bom.layer.name` (Optional) - Set to the name of the layer if the `bom` was associated with a specific layer. 
+
+The lifecycle will put all the `build` related entries in the `/layers/config/sbom/build.bom.cdx.json` file and `launch` entries in `/layers/config/sbom/launch.bom.cdx.json`
```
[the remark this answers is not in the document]
natalieparellano: > Do these end up in `/layers/config/sbom/launch.bom.cdx.json`?

If the answer to this question is yes, then I would think that any layer that has a `<layer>.bom.<ext>` file but is `launch = false` would have the bom end up in `/layers/config/sbom/build.bom.cdx.json`. Is that right?

text/0000-sbom.md
natalieparellano: Are we still putting bom information in `io.buildpacks.build.metadata`? If not, would it make sense for each `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` to be its own layer, to aid in layer re-use?

Could we translate this "[how it works](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works)" from the bom-in-layer-metadata RFC into this new world?

A couple of thoughts:
* If the layer is `cache = true` and there is a `<layer>.bom.<ext>` then the bom would be saved in the cache alongside the layer and restored to the buildpack's `<layers>` directory on the next build
* If the layer is `launch = true`, `cache = false`, `build = false` and there is a `<layer>.bom.<ext>` then normally we are only restoring the layer metadata, but would the buildpack also want the bom to be available, so that it could decide whether to re-use the layer or not? In this case we would need to copy the `/layers/config/sbom/<buildpack-id>/<layer-id>/<type>.bom.<ext>` from the previous image as well (unless this is going in a label?

JeroenKnoops:
my 2 cents.. 

I've worked a lot on SBOM and feel SPDX is a better choice. Also In-toto tends to move towards SPDX and Purls in context of SLSA and attestations: https://github.com/in-toto/attestation/issues/63

text/0000-sbom.md
ForestEckhardt: I feel like in the case of `launch = true`, `cache = false`, `build = false` I would want the `<layer>.bom.<ext>` or at least the information therein to be available so that I have the option to reuse the BOM and not generate a new one in the case that I am reusing that layer.

That is my two cents on the matter at least.

hone (approved):
I'm overall in favor of this. I know @samj1912  and @ekcasey  need to sync up still. I'd also like to see @natalieparellano's recent questions addressed.

text/0000-sbom.md
samj1912: This is the part that @ekcasey and I need to sync up about.

text/0000-sbom.md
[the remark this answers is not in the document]
samj1912: Yup that is correct as of now. I wonder if we should allow buildpack authors more control over this and let them create the the appropriate files with the extension for clarity + control?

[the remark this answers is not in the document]
samj1912: @natalieparellano clarified this a bit more. LMK if it makes sense.

ekcasey (approved):

text/0000-sbom.md
```diff
@@ -0,0 +1,151 @@
...
+[prior-art]: #prior-art
+
+- [#RFC 53](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
+- [#RFC 87](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Interactions with the stack SBOM and how to represent that and merge it with the Buildpack SBOM.
+- Restore/export logic for SBOM files. This will be covered in a subsequent RFC.
```
hone, line 136: With the recent changes from talking with @ekcasey, is this still true @samj1912?

hone:
@samj1912 what's the process for supporting a new SBOM format?

text/0000-sbom.md
samj1912: Nope - should remove that.

samj1912:
> @samj1912 what's the process for supporting a new SBOM format?

If we don't want to generate a merged SBOM (like SPDX), then it would largely involve a trivial change adding it to an allow list of supported format names <> extension mappings in the lifecycle.

hone:
FCP with a close on 9/29.

natalieparellano (approved):

jromero (approved):

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently the SBOM is defined in the `<layer>.toml`, `launch.toml` and `build.toml` respectively under the `[bom]` table.
+It may not be the most user-friendly way for buildpack authors to create SBOM documents in the above format. This RFC proposes that the `bom` be moved to `<layer>.bom.<ext>`, `launch.bom.<ext>` and `build.bom.<ext>` instead where `<ext>` will be `cdx.json` for CycloneDX documents and `spdx.json` for SPDX documents. A buildpack can also output the `bom` in multiple formats i.e. both CycloneDX and SPDX.
```
natalieparellano, line 65: /queue-issue buildpacks/lifecycle "Builder should warn if newer buildpacks write a bom in *.toml" type/enhancement epic/sbom

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+id = "<buildpack ID>"
+name = "<buildpack name>"
+# This can be an array of supported SBOM formats by the buildpack.
+# Valid array values are "CycloneDX" and "SPDX" as of this RFC.
+sbom = ["CycloneDX"]
+```
+
+This information can be displayed on the registry and `pack inspect-buildpack`.
+
+The `lifecycle` binaries would be responsible for taking all of the above `bom` files, and copying them in `/layers/config/sbom/<type>/<buildpack-id>/<layer-id>/bom.<ext>` for the layer `bom` files where `type` could be `build` or `launch`. Buildpack level BOM files will reside at `/layers/config/sbom/<type>/<buildpack-id>/bom.<ext>` where the type could be `launch` and `build` respectively.
```
natalieparellano, line 84: /queue-issue buildpacks/lifecycle "Builder should copy bom files to /layers/config/sbom" type/enhancement epic/sbom

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+
+Additionally the `lifecycle` binaries will be responsible for creating merged `bom` files, merging `bom` formats of the same type and outputting them at `/layers/config/<type>/sbom/bom.<ext>`. Initially this will only be supported for CycloneDX as it has a well defined and efficient way of merging multiple `bom` files. A reference implementation can be found at the [`cyclonedx-cli`](https://github.com/CycloneDX/cyclonedx-cli) project. In the future we may add support for merging SPDX `bom` files as well.
```
natalieparellano, line 86: /queue-issue buildpacks/lifecycle "Lifecycle should merge CycloneDX bom files" type/enhancement epic/sbom

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+
+Merging different formats (CycloneDX/SPDX/Legacy CNB format) is a non-goal and will not be supported. Since we will be making the individual `bom` files available as well, if users/platforms wish, they can choose to write conversion and merging logic but this will not be supported in the lifecycle.
+
+Since we will be preserving the original output files and if a buildpack chooses to output in both SPDX and CycloneDX, the lifecycle may eventually add support for SPDX merging in a future version if merging SPDX documents becomes better defined/efficient.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The `lifecycle` binaries would be responsible for reading, merging and restoring the appropriate `bom` files (the legacy CNB format and the CycloneDX format currently). The lifecycle of these `bom` files would be tied to their respective metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details.
```
natalieparellano, line 96: /queue-issue buildpacks/lifecycle "Exporter should cache bom files for cached layers" type/enhancement epic/sbom

natalieparellano: /queue-issue buildpacks/lifecycle "Exporter should export bom files for launch layers" type/enhancement epic/sbom

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The `lifecycle` binaries would be responsible for reading, merging and restoring the appropriate `bom` files (the legacy CNB format and the CycloneDX format currently). The lifecycle of these `bom` files would be tied to their respective metadata `toml` files. See [RFC 0087](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#how-it-works) for more details. 
+
+For merging the CycloneDX `bom` files, the `lifecycle` binaries could replicate or use tooling from [CycloneDX-cli](https://github.com/CycloneDX/cyclonedx-cli) which has a merge operation. The only additional piece of information that the `lifecycle` binaries would inject are `CycloneDX` [`metadata`](https://cyclonedx.org/use-cases/#properties--name-value-store) the following property keys -
+
+- `io.buildpacks.bom.buildpack.id` - Buildpack ID for the buildpack that created the BOM
+- `io.buildpacks.bom.layer.name` (Optional) - Set to the name of the layer if the `bom` was associated with a specific layer.
```
natalieparellano, line 101: /queue-issue buildpacks/lifecycle "Exporter should add io.buildpacks.bom.* labels" type/enhancement epic/sbom

text/0000-sbom.md
```diff
@@ -0,0 +1,150 @@
...
+
+For merging the CycloneDX `bom` files, the `lifecycle` binaries could replicate or use tooling from [CycloneDX-cli](https://github.com/CycloneDX/cyclonedx-cli) which has a merge operation. The only additional piece of information that the `lifecycle` binaries would inject are `CycloneDX` [`metadata`](https://cyclonedx.org/use-cases/#properties--name-value-store) the following property keys -
+
+- `io.buildpacks.bom.buildpack.id` - Buildpack ID for the buildpack that created the BOM
+- `io.buildpacks.bom.layer.name` (Optional) - Set to the name of the layer if the `bom` was associated with a specific layer. 
+
+The `lifecycle` binaries will put the `bom` files from layers set to `launch = true` inside the `/layers/config/sbom/launch/<buildpack-id>/<layer-id>/bom.<ext>` hierarchy otherwise it will put them in `/layers/config/sbom/build/<buildpack-id>/<layer-id>/bom.<ext>`.
+
+## BOM Restoration
```
natalieparellano, line 105: /queue-issue buildpacks/lifecycle "Restorer should restore bom files from app and cache" type/enhancement epic/sbom

samj1912:
/queue-issue buildpacks/lifecycle "Builder should warn if newer buildpacks write a bom in *.toml"

samj1912:
/queue-issue buildpacks/lifecycle "Restorer should restore bom files from app and cache" type/enhancement epic/sbom

samj1912:
/queue-issue buildpacks/lifecycle "Lifecycle should inject io.buildpacks.bom.* metadata when merging SBOMs" type/enhancement epic/sbom

/queue-issue buildpacks/lifecycle "Exporter should export bom files for launch layers" type/enhancement epic/sbom

/queue-issue buildpacks/lifecycle "Exporter should cache bom files for cached layers" type/enhancement epic/sbom

/queue-issue buildpacks/lifecycle "Lifecycle should merge CycloneDX bom files" type/enhancement epic/sbom

/queue-issue buildpacks/lifecycle "Builder should copy bom files to /layers/config/sbom" type/enhancement epic/sbom

PR #168, opened 2021-06-17
Remove Shell-Specific Logic

ekcasey:
[Readable](https://github.com/buildpacks/rfcs/blob/no-bash/text/0000-remove-shell-processes.md)

### DIFF
--- a/text/0000-remove-shell-processes.md
+++ b/text/0000-remove-shell-processes.md
@@ -207,7 +207,9 @@ EOF
   /usr/sbin/sshd -f $ssh_dir/sshd_config -o "Port ${ssh_port}"
 fi
 ```
-This script does not require any changes and would work as an `exec.d` or `profile.d`.
+This script does not require any changes and would work as an `exec.d` helper. To implement the new buildpack API the buildpack author would have to:
+* Move the above script from `<layer>/profile.d/sshd` to `<layer>/exec.d/sshd`.
+* Ensure script is executable by run-image user (if it isn't already).
 
 **Example 2: paketo-buildpacks/node-engine **
 
--- a/text/0000-remove-shell-processes.md
+++ b/text/0000-remove-shell-processes.md
@@ -100,7 +100,7 @@ The Paketo .Net Execute Buildpack may generates shell processes similar to the f
 ```
 [[processes]]
 type = "web"
-command = "dotnet my-app.dll --url http://0.0.0.0:${PORT:-8080}"
+command = "dotnet my-app.dll --urls http://0.0.0.0:${PORT:-8080}"
 direct = false
 ```
 NOTE: the buildpack API used by this buildpack (`0.5`) predates the introduction of `default`.
--- a/text/0000-remove-shell-processes.md
+++ b/text/0000-remove-shell-processes.md
@@ -1,0 +1,401 @@
+# Meta
+[meta]: #meta
+- Name: Remove Shell Processes
+- Start Date: 2021-05-14
+- Author(s): @ekcasey
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: [RFC 0045](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes changes to the structure of a process type in order to make the interface more similar to k8s, docker and other familiar tools and simultaneously simplify the implementation.
+
+# Definitions
+[definitions]: #definitions
+
+*process type*: A named process definition, contributed by a buildpack at build-time and executed by the launcher at run-time.
+
+*direct process*: A process that is executed directly.
+
+*shell process*: A process that is executed by a shell.
+
+*script process*: A special type of shell process where there are zero `args` and `command` is a shell script.
+
+*build-provided profiles scripts*: Profile scripts that are provided by buildpacks at build-time. When the launcher executes a shell process these scripts will be sourced in the shell, prior to process execution.
+
+*user-provided profile script*: A profile scripts that is provided by the user in the app root of the app dir. When the launcher executes a shell process this script will be sourced in the shell, after buildpack-provided profile scripts and prior to process execution.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC aims to:
+1. Reduce **complexity** in the CNB Specification
+2. Enable new **use cases** by supporting overridable arguments
+3. Improve **interoperability** between buildpacks and minimal or distroless stacks
+
+We will achieve these goals be addressing the following problems with our current process type model:
+
+## Argument Handling
+
+In [RFC 0045](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md) we introduced a change to the launcher interface that allows users to append additional arguments to a process type at runtime. This was a major improvement to the runtime interface for app images, as users no longer needed to respecify the entire command in order to append a single argument. However, in the Launcher Arguments proposal we made two mistakes:
+1. **complexity** - We tried to be clever and support appending argument appending to shell processes. This "works" but can result in surprising behavior such as the need to escape literal `"` characters to prevent Bash from removing them during evaluation. Users must know whether the process is a direct process or a shell process in order to provide arguments correctly, making process types a leaky abstraction.
+2. **use cases** - Arguments can serve two different purposes. Some are required and should always be included in the command. Some are merely defaults that may be overridden by the user. The current spec does not support the latter case.
+
+## Shell-Specific Logic
+
+**interoperability** - Currently, unless a buildpack author specifically indicates that the given process type is a `direct` process, a shell process with a dependency on Bash (linux) or Cmd (windows) is created. This is not ideal, as many in the industry are moving towards minimal images that do not include a shell, in order to reduce surface attack area. Buildpack authors may inadvertantly create process types that depend on a shell even when no such dependency is necessary.
+
+**complexity** - Having a direct dependency between the launcher and specific shells is inelegant. It makes the spec more complex and end users and buildpack author are forced to understand that complexity in order to understand the behavior of the resulting usage or debug issues. For example end users and buildpacks authors must understand:
+- The difference between a direct and shell process
+- The nuances of argument handling in direct vs. Bash vs. Cmd cases
+- That buildpack-provided profile scripts will not apply to direct processes
+- That a user-provided `.profile` script will not apply to a direct processes
+
+ Removing this special behavior and requiring buildpacks to explicitly include any required shell in the command itself creates a simpler, more comprehensible model for buildpack authors and end users alike.
+
+# What it is
+[what-it-is]: #what-it-is
+
+## Process Schema
+The existing schema for the processes table in `launch.toml` is
+```
+# Old Schema
+
+[[processes]]
+type = "required"
+command = "required"
+args = ["optional"]
+direct = false
+default = false
+```
+
+The new proposed schema is:
+```
+# New Schema
+
+[[processes]]
+type = "required"
+command = ["required"]
+args = ["optional"]
+default = false
+```
+The following changes have been made:
+- **`direct` has been removed** - all processes are executed directly. If `bash` or `cmd.exe` is required it should be included in `command`. No surprises.
+- **`command` is now an array** - Arguments in `command` *will not* be overwritten if a user provides additional arguments at runtime. `args` are default arguments that *will* be overwritten if a user provides additional arguments at runtime. The makes `command` analogous to `Entrypoint` in the OCI spec and `command` in a Kubernetes PodSpec. `args` is analogous to `cmd` and `args` in docker and Kubernetes respectively.
+
+## Using Environment Variables in a Process
+
+One upside to our previous execution strategy was that it enable users to include environment variable references in arguments that were later evaluated in the container. To preserve this feature we can instead adopt a variabtion on the Kubernetes strategy for [environment variables interpolation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config). If a buildpack or user includes `${<env>}` in the `command` or `args` and `{env}` is the name of an environment variable set in the launch environment, the launcher will replace this string with the value of the environment variable after apply buildpack-provided env modifications and before launching the process.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Buildpack-provided process types
+### Example 1 - A Shell Process
+The Paketo .Net Execute Buildpack may generates shell processes similar to the following:
+```
+[[processes]]
+type = "web"
+command = "dotnet my-app.dll --url http://0.0.0.0:${PORT:-8080}"
+direct = false
+```
+NOTE: the buildpack API used by this buildpack (`0.5`) predates the introduction of `default`.
+
+Using the new API this process could look like:
+```
+[[processes]]
+type = "web"
+command = ["dotnet", "my-app.dll", "--urls", "http://0.0.0.0:${PORT}"] # the default value of PORT would need to be provided in a layer
+default = true
+```
+Things to note:
+* In the above example I have eliminated the dependency on Bash instead of explicitly adding it to the command, because it is likely unnecessary.
+* If the buildpack authors believed that `--urls` should be overridable they could set move the last two arguments from `command` to `args`.
+
+### Example 2 - A Script Process
+When a buildpack upgrade to the new buildpack API, it must convert any existing shell processes into direct processes.
+
+The [Paketo Yarn Start Buildpack](https://github.com/paketo-buildpacks/yarn-start) currently may generate a script processes similar to the following:
+
+```
+[[processes]]
+type = "web"
+command = "pre-start.sh && nodejs server.js && post-start.sh"
+direct = false
+default = false
+```
+
+Using the new API this process look like:
+```
+[[processes]]
+type = "web"
+command = ["bash", "-c", "pre-start.sh && nodejs server.js && post-start.sh"]
+default = false
+```
+
+## User Provided Processes
+
+Currently if the user can specify a custom process dynamically at runtime by setting the container entrypoint to `launcher` directly rather than using a symlink to the launcher, the providing a custom `cmd`. This custom command is executed directly if `cmd` is an array and the first element is `--`. Otherwise the custom command is assumed to be a shell process. In the interest of removing complexity we should do away with the special `--` argument and execute all custom commands directly. 
+
+### Example 1 - A Direct process
+The follow direct commands:
+```
+docker run --entrypoint launcher <image> -- env
+docker run --entrypoint launcher <image> -- echo hello '$WORLD' 
+```
+will become the following, using the new platform API
+```
+docker run --entrypoint launcher <image> env
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+```
+Previously, in the second command in this example `${WORLD}` would not have been interpolated because this is a direct process; instead the output would include the literal string `$WORLD`. With the changes proposed, `${WORLD}` will now be evaluated, even though the process is direct.
+
+### Example 2 - A Shell Process
+The follow custom shell command:
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> echo hello '${WORLD:-world}'
+```
+will become the following, using the new platform API
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> bash -c 'echo hello "${WORLD:-world}"'
+```
+The first command in this example did not need to change to behave as expected with the new API. Previously it was necessary to use a shell process in order to evaluate `${WORLD}`. Now, the shell is unescessary. However, if the user wants to use more features of Bash interpolation they will need to explicitly invoke a shell, as shown in the second command in this example.
+
+### Example 3 - A Script Process
+The follow custom script command:
+```
+docker run --entrypoint launcher <image> 'for opt in $JAVA_OPTS; do echo $opt; done'
+```
+will become the following, using the new platform API
+```
+docker run --entrypoint launcher <image> bash -c 'for opt in $JAVA_OPTS; do echo $opt; done'
+```
+
+## What About Profile Scripts?
+
+### `<layer>/profile.d/*`
+When a buildpack upgrades to the new buildpack API, buildpack-provided profile scripts will no longer be supported. Instead buildpacks can use `exec.d`. Most existing profiles scripts can be easily converted.
+
+**Example 1: `jkutner/sshd` **
+
+The following profile script example was taken from https://github.com/jkutner/sshd-buildpack/blob/master/sbin/sshd.sh
+```
+#!/usr/bin/env bash
+
+if [[ "${SSH_DISABLED:-}" != "true" ]]; then
+  ssh_layer=$(realpath $(dirname ${BASH_SOURCE[0]})/..)
+  ssh_dir=$(realpath $HOME/.ssh)
+
+  mkdir -p $ssh_dir
+  cat $ssh_layer/id_rsa.pub >> $ssh_dir/authorized_keys
+
+  cat << EOF >> $ssh_dir/sshd_config
+HostKey $ssh_layer/id_rsa
+AuthorizedKeysFile $ssh_dir/authorized_keys
+EOF
+
+  chmod 600 /workspace/.ssh/*
+
+  ssh_port=${SSH_PORT:-"2222"}
+  echo "at=sshd state=starting user=$(whoami) port=${ssh_port}"
+  /usr/sbin/sshd -f $ssh_dir/sshd_config -o "Port ${ssh_port}"
+fi
+```
+This script does not require any changes and would work as an `exec.d` or `profile.d`.
+
+**Example 2: paketo-buildpacks/node-engine **
+
+The following profile script example was taken from https://github.com/paketo-buildpacks/node-engine/blob/main/environment.go
+```
+if [[ -z "$MEMORY_AVAILABLE" ]]; then
+		memory_in_bytes="$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)"
+		MEMORY_AVAILABLE="$(( $memory_in_bytes / ( 1024 * 1024 ) ))"
+fi
+export MEMORY_AVAILABLE
+```
+
+The following revised version of this script would implement the `exec.d` interface
+```
+if [[ -z "$MEMORY_AVAILABLE" ]]; then
+		memory_in_bytes="$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)"
+		MEMORY_AVAILABLE="$(( $memory_in_bytes / ( 1024 * 1024 ) ))"
+fi
+
+cat << EOF >&3
+MEMORY_AVAILABLE = "${MEMORY_AVAILABLE}" 
+EOF
+```
+
+### `<app>/.profile`
+
+When a platform upgrades to the new platform API, user-provided profile scripts will no longer be supported. A buildpack could provide an identical or similar interface by detecting a `.profile` file in the app dir, wrapping it so that it implements the `exec.d` interface and add creating an executable with that layer. A platform that wishes to always provide that functionality could add a buildpack like this to every build. By making dynamic runtime environment modification a buildpack concern and not duplicating similar behavior in the platform API, we reduce the surface area and complexity of the platform API and thus improve normalization of the spec.
+
+## API compat
+
+### Overriding Args for Shell Processes
+We should convert args in `launch.toml` to `command` in `metadata.toml` to expose a consistent interface to users. For example, if a buildpack the following process type in `launch.toml` 
+```
+[[process]]
+type = "hi"
+command = "echo"
+args = ["hello", "world"]
+direct = true
+```
+
+we would convert it to the following in `metadata.toml`
+```
+[[process]]
+type = "hi"
+command = ["echo", "hello", "world"]
+args = []
+api = "0.5"
+direct = true
+```
+This spares users from having to learn about the differences between buildpack APIs in order to predict how additional arguments will behave.
+
+### Evaluating Args
+
+For shell processes provided by older buildpacks, we must continue to Bash evaluate buildpack-provided args to avoid breaking older buildpacks
+
+For example, if a buildpack using API `0.5` creates the following entry in `launch.toml`
+```
+[[process]]
+type = "hi"
+command = "echo"
+args = ["hello", "${WORLD:-world}"]
+```
+
+It should be converted to the following in `metadata.toml`
+
+```
+[[process]]
+type = "hi"
+command = ["echo", "hello", "${WORLD:-world}"]
+args = []
+api = "0.5"
+direct = false
+```
+
+and the launcher should Bash evaluate each entry in `command` to avoid breaking changes.
+
+However, any additional user-provided args should NOT be Bash evaluated to reduce the amount of complexity exposed to end users.
+
+### profiles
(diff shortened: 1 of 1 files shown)


sclevine (approved):

text/0000-remove-shell-specific-processes.md
```diff
@@ -0,0 +1,204 @@
...
+### `<layer>/profile.d/*`
+When a buildpack upgrades to the new buildpack API, buildpack-provided profile scripts will no longer be supported. Instead buildpacks can use `exec.d`. Most existing profiles scripts can be easily converted.
+
+**Example 1: **
+
+**Example 2: **
+
+### `<app>/.profile`
+
+When a platform upgrades to the new platform API, user-provided profile scripts will no longer be supported. A buildpack could provide an identical or similar interface by detecting a `.profile` file in the app dir, wrapping it so that it implements the `exec.d` interface and add creating an executable with that layer. A platform that wishes to always provide that functionality could add a buildpack like this to every build. By making dynamic runtime environment modification a buildpack concern and not duplicating similar behavior in the platform API, we reduce the surface area and complexity of the platform API and thus improve normalization of the spec.
```
natalieparellano, line 150: > A platform that wishes to always provide that functionality could add a buildpack like this to every build.

I recall something similar came up in discussions around prepare phase & project.toml (using buildpacks to perform actions instead of the platform - e.g., setting build time env) and I wonder if maybe that idea deserves its own RFC. It seems like a sticking point in these conversations has been "who is responsible" for creating these generic buildpacks, how would they be supported & distributed, etc.

ekcasey:
Not sure why the github-actions bot assigned three people ... ?

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+
+## Using Environment Variables in a Process
+
+One upside to our previous execution strategy was that it enable users to include environment variable references in arguments that were later evaluated in the container. To preserve this feature we can instead adopt a variabtion on the Kubernetes strategy for [environment variables interpolation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config). If a buildpack or user includes `${<env>}` in the `command` or `args` and `{env}` is the name of an environment variable set in the launch environment, the launcher will replace this string with the value of the environment variable after apply buildpack-provided env modifications and before launching the process.
```
hone, line 91:
```suggestion
One upside to our previous execution strategy was that it enable users to include environment variable references in arguments that were later evaluated in the container. To preserve this feature we can instead adopt a variation on the Kubernetes strategy for [environment variables interpolation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config). If a buildpack or user includes `${<env>}` in the `command` or `args` and `{env}` is the name of an environment variable set in the launch environment, the launcher will replace this string with the value of the environment variable after apply buildpack-provided env modifications and before launching the process.
```

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+command = "dotnet my-app.dll --url http://0.0.0.0:${PORT:-8080}"
+direct = false
+```
+NOTE: the buildpack API used by this buildpack (`0.5`) predates the introduction of `default`.
+
+Using the new API this process could look like:
+```
+[[processes]]
+type = "web"
+command = ["dotnet", "my-app.dll", "--urls", "http://0.0.0.0:${PORT}"] # the default value of PORT would need to be provided in a layer
```
hone, line 112:
```suggestion
command = ["dotnet", "my-app.dll", "--url", "http://0.0.0.0:${PORT}"] # the default value of PORT would need to be provided in a layer
```

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+default = true
+```
+Things to note:
+* In the above example I have eliminated the dependency on Bash instead of explicitly adding it to the command, because it is likely unnecessary.
+* If the buildpack authors believed that `--urls` should be overridable they could set move the last two arguments from `command` to `args`.
```
hone, line 117:
```suggestion
* If the buildpack authors believed that `--url` should be overridable they could set move the last two arguments from `command` to `args`.
```

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+[drawbacks]: #drawbacks
+
+In exchange for a reduction in complexity and cognitive overhead buildpack-authors and end users lose certain conveniences like the more intuitive profile interface. This drawback could be remediated by tooling for buildpack authors (to make creation of `exec.d` helpers easy) or by buildpacks (to support `.profile`). However, in the `.profile` case, consistency across ecosystems will be a matter of convention (like Procfile) rather than a guarantee.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## `$(<env>)` syntax
+
+We could use `$(<env>)` for environment variable replacements instead of `${<env>}`. urrently, Kubernetes only replaces `$(FOO)` if `FOO` matches the name of an enivornment variable defined in the `env` section of the PodSpec, otherwise `$(FOO)` is passed through literally. The would allow thelauncher to pick up where kubernetes leaves off and evaluate an remaining references.
```
jabrown85, line 302:
```suggestion
We could use `$(<env>)` for environment variable replacements instead of `${<env>}`. Currently, Kubernetes only replaces `$(FOO)` if `FOO` matches the name of an environment variable defined in the `env` section of the PodSpec, otherwise `$(FOO)` is passed through literally. The would allow the launcher to pick up where kubernetes leaves off and evaluate any remaining references.
```

text/0000-remove-shell-processes.md
jkutner, line 91: Is there a risk that in doing this we might somehow expose secrets? For example if the process type is:

```
./bin/server --key $MY_SECRET
```

If we substitute `MY_SECRET` in the process could it end up in the logs or somewhere that it might not when Bash substitutes it? 

Are they known libraries/mechanisms we can use to make sure we're on the right-side of the law here?

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> echo hello '${WORLD:-world}'
+```
+will become the following, using the new platform API
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> bash -c 'echo hello "${WORLD:-world}"'
+```
+The first command in this example did not need to change to behave as expected with the new API. Previously it was necessary to use a shell process in order to evaluate `${WORLD}`. Now, the shell is unescessary. However, if the user wants to use more features of Bash interpolation they will need to explicitly invoke a shell, as shown in the second command in this example.
```
jkutner, line 168:
```suggestion
The first command in this example did not need to change to behave as expected with the new API. Previously it was necessary to use a shell process in order to evaluate `${WORLD}`. Now, the shell is unnescessary. However, if the user wants to use more features of Bash interpolation they will need to explicitly invoke a shell, as shown in the second command in this example.
```

jkutner: I'm not sure I can complete sense of what "more features of Bash interpolation" includes. In particular, does this include things like `;`, `&`, and `&&` in the command? Those are the most common patterns I can think of (aside from env vars). What about different methods of quoting (`"` versus `'`)?

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Example 3 - A Script Process
+The follow custom script command:
+```
+docker run --entrypoint launcher <image> 'for opt in $JAVA_OPTS; do echo $opt; done'
```
jkutner, line 173: One of the things I worry about in this example is how it might unexpectedly fail. I expect that most users will do the following:

1. upgrade
2. try the exact same command they used before (i.e. no `bash -c`)
3. watch it fail
4. realize they need to add `bash -c`

In step 3, is it possible that something very bad happens if their process kinda works but not as they expected?

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+EOF
+
+  chmod 600 /workspace/.ssh/*
+
+  ssh_port=${SSH_PORT:-"2222"}
+  echo "at=sshd state=starting user=$(whoami) port=${ssh_port}"
+  /usr/sbin/sshd -f $ssh_dir/sshd_config -o "Port ${ssh_port}"
+fi
+```
+This script does not require any changes and would work as an `exec.d` or `profile.d`.
```
jkutner, line 210: I'd still need to move the script in the `exec.d` though right? are there any restrictions like filename or permissions that might change? (I'm sure these are all fine but would like to capture)

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+fi
+
+cat << EOF >&3
+MEMORY_AVAILABLE = "${MEMORY_AVAILABLE}" 
+EOF
+```
+
+### `<app>/.profile`
+
+When a platform upgrades to the new platform API, user-provided profile scripts will no longer be supported. A buildpack could provide an identical or similar interface by detecting a `.profile` file in the app dir, wrapping it so that it implements the `exec.d` interface and add creating an executable with that layer. A platform that wishes to always provide that functionality could add a buildpack like this to every build. By making dynamic runtime environment modification a buildpack concern and not duplicating similar behavior in the platform API, we reduce the surface area and complexity of the platform API and thus improve normalization of the spec.
```
jkutner, line 237: Is this true even when the use `bash -c` in their command?

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+## `$(<env>)` syntax
+
+We could use `$(<env>)` for environment variable replacements instead of `${<env>}`. urrently, Kubernetes only replaces `$(FOO)` if `FOO` matches the name of an enivornment variable defined in the `env` section of the PodSpec, otherwise `$(FOO)` is passed through literally. The would allow thelauncher to pick up where kubernetes leaves off and evaluate an remaining references.
+pros:
+* Users can continue to use a syntax they are already familiar with
+cons:
+* User provided variables may be evaluated too soon, before buildpack modifications are applied.
+
+## Lifecycle support for `<app>/.exec`
+When we remove support for `<app>/.profile` we could add support for `<app>/.exec` or similar where `<app>/.exec` must implement the `exec.d` interface.
```
jkutner, line 309: I'd like to bring this into scope

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+However, things get very complicated if we explicitly source profiles or shell evaluated args in `command`. The complexity of the resulting commands probably makes this strategy untenable.
+
+# Prior Art
+[prior-art]: #prior-art
+
+The Paketo Java buildpacks have already converted an extensive collection of profile script to exec.d binaries and converted all processes to direct processes in order to support minimal stacks like `io.paketo.stacks.tiny` and to provide users with more intuitive argument handling.
+
+The Procfile interface is supported by Paketo, Heroku, and Google buildpacks, demonstrating that it is possible to have a consistent interface across buildpack ecosystems without building direct support for that interface in the lifecycle.
```
jkutner, line 352: I'm not sure this is winning example. There are 3 different impls of something that should (IMO) be a native feature. The fact that these haven't converged on a single Procfile buildpacks is worrisome and not a pattern I'd like to repeat.

text/0000-remove-shell-specific-processes.md
ekcasey: @natalieparellano I agree that an RFC for this sounds like a good idea.

If we, as a project, wanted to provide buildpacks to make it easy for platforms or buildpck ecosystems to adopts extensions or conventions, that would deserve its own RFC. However, I don't think we need to block this RFC on that. Whether or not we take the extra step of providing common buildpacks (and I suspect we should) an app->lifecycle API is fundamentally untenable and I would argue that we need to remove this feature either way.

text/0000-remove-shell-processes.md
ekcasey: I'll admit to knowing precious little about the `dotnet` CLI, but I copied this command directly from the buildpack. And it seems to agree with [this blog post](https://andrewlock.net/5-ways-to-set-the-urls-for-an-aspnetcore-app/#command-line-arguments) I just found.

text/0000-remove-shell-processes.md
ekcasey: https://github.com/buildpacks/rfcs/pull/168#discussion_r659030633

text/0000-remove-shell-processes.md
ekcasey: I think we are right that there is some cause for concern here. I think we should change this RFC and switch the interpolation notation to something like `$((MY_SECRET))` so we never interpolate environment variables that are intended for Bash to interpolate in a command like `["bash", "-c", "./bin/server --key ${MY_SECRET}"]. It's a little sad that we need yet another notation but it seems good to remove any ambiguity with regards to what gets interpolated by k8s, vs the launcher, vs Bash (if it is explicitly included in a command).

In this example are you imagining that it is the user you provides `$MY_SECRET` dynamically at runtime? I would argue that this is not a safe thing to do with secrets because they end up in the container config. Secrets should be provided via a safe mechanism like bindings and buildpacks should be responsibility for protecting the secret, which hopefully will not require passing it directly on the command line (even as a reference for Bash to interpolate) in the vast majority of use cases.

text/0000-remove-shell-processes.md
ekcasey: To avoid this whole question I think we should switch from `${NAME}` to `$((NAME))` for our env var notation. Therefore nobody is tempted to try things like `${NAME:-default-name}`. We should keep it as simple as possible and avoid surprises.

text/0000-remove-shell-processes.md
ekcasey: I agree that they might upgrade and fail but I don't think there is any danger of the process even kinda working. In this case the launcher would be trying to directly invoke an executable called `'for opt in $JAVA_OPTS; do echo $opt; done'`

text/0000-remove-shell-processes.md
ekcasey: Good point. We should capture two things:
1. The script must move to `<layer>/exec.d/`
2. The runtime user must have permission to execute this file (e.g. executable bit should be set).

text/0000-remove-shell-processes.md
ekcasey: Yes I believe this is still true. `exec.d` helpers currently run for both direct and shell processes.

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,401 @@
...
+* User provided variables may be evaluated too soon, before buildpack modifications are applied.
+
+## Lifecycle support for `<app>/.exec`
+When we remove support for `<app>/.profile` we could add support for `<app>/.exec` or similar where `<app>/.exec` must implement the `exec.d` interface.
+pros:
+* Users may still dynamically modify the runtime environment without requiring a specific buildpack
+cons:
+* The exec.d interface must be duplicated in both the buildpack and platform API
+* If we modify the interface the same app might behave differently or fail to run on specific platforms depending on which version of the platform API they are using
+* Users must directly implement our less-than-perfectly-intuitive exec.d interface instead of whatever better UX buildpack authors invent.
```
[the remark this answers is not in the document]
ekcasey: > Often, those modifications aren't even buildpack specific, but app specific

Not sure I follow why this is relevant.

What I am trying to say is that a single buildpack could continue to provide support for .profile scripts without forcing users to migrate to the exec.d interface, if that was desirable.

text/0000-remove-shell-processes.md
ekcasey: After thinking long and hard about it I am very opposed to bringing this into scope. I don't think contracts between the app source and the lifecycle are supportable. The platform API should be a contract between a platform and the lifecycle, to enable predictable behavior and a clean upgrade story. Otherwise the same app will behave different on different platforms depending on the platform API in use. I realize the user must have some awareness of the "launch" api to understand argument handling for the resulting image but at least they won't need to change their source code to run on different platforms. This RFC is trying hard to shrink the surface area of platform API, that the app developer needs to be aware of.

If we wanted to respect config like this in the app dir, a buildpack should provide that functionality. App developers can included the exact buildpack they want to get the exact outcome they want without having to keep track of the versions of the platform API currently implemented by the platforms they use (which is entirely outside their control).

text/0000-remove-shell-processes.md
ekcasey: I love to see an RFC to provide utility buildpacks as part of the CNB project, that implement extension specs or common conventions.

text/0000-remove-shell-processes.md
ekcasey: Sorry, I think I found the source of confusion - https://github.com/buildpacks/rfcs/pull/168/commits/c97e7f22e989bb6aa45570d8f7da1ca86631ffa8

text/0000-remove-shell-processes.md
ekcasey: https://github.com/buildpacks/rfcs/pull/168#discussion_r661799403

text/0000-remove-shell-processes.md
samj1912: `$(ENV)` should work? (the k8s convention). It only works in k8s for env vars directly defined  in the spec, the rest are left untouched.

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,407 @@
...
+### Example 2 - A Shell Process
+The follow custom shell command:
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> echo hello '${WORLD:-world}'
+```
+will become the following, using the new platform API
+```
+docker run --entrypoint launcher <image> echo hello '${WORLD}'
+docker run --entrypoint launcher <image> bash -c 'echo hello "${WORLD:-world}"'
```
samj1912, line 166: Could we support these use cases by default? GIven that most envsubst libs already support it - 

https://github.com/a8m/envsubst
https://github.com/drone/envsubst

This would also be useful as it would end up in the metadata and users could see default values for these variables as set by processes.

text/0000-remove-shell-processes.md
ekcasey: Added in https://github.com/buildpacks/rfcs/pull/168/commits/b49861f13dd11b080099b0da7ac8b49e9611f113

text/0000-remove-shell-processes.md
ekcasey: I _just_ decided to take this in a different direction - https://github.com/buildpacks/rfcs/pull/168/commits/827cdd5782eceed7ba367941af03dd8d519bb3d3

ekcasey: I found some of the comment from the original k8s implementation of this illuminating - https://github.com/kubernetes/kubernetes/pull/7936#issuecomment-100950516

samj1912: Was going through the same thing.


On a separate note - I guess it should be a trivial change in the above libs to add support for a different marker instead of `{`. WIll see if they are open to it. OTOH I guess we could always start simple and add it later if needed. I would've like a way to specify a default value at the very least though.

samj1912: FWIW - just needs a change at https://github.com/drone/envsubst/blob/7bf45dbf5372da60a12eab36b9c8a63c99b92ea5/parse/scan.go#L174

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,451 @@
...
+# New Schema
+
+[[processes]]
+type = "required"
+command = ["required"]
+args = ["optional"]
+default = false
+```
+The following changes have been made:
+- **`direct` has been removed** - all processes are executed directly. If `bash` or `cmd.exe` is required it should be included in `command`. No surprises.
```
dmikusa-pivotal, line 86: I think there could be some nuance here that we need to document when switching from a shell process to direct regarding signals. Many platforms support the concept of a graceful shutdown, which is a SIGTERM followed by a pause, followed by SIGKILL on your process.

When you run as a shell process, pid 1 is your shell which generally traps and handles signals as you'd expect. When you switch to running your process as direct, it becomes pid 1 and it may or may not handle all the signals correctly (you can generally tell if it's not handling them correctly by running a container and trying to interrupt it with ctrl+c, if that works it's handling signals correctly). 

Anyway, I think we need to call out this consideration so that when buildpack authors are migrating their buildpacks they think about this nuance and ensure it's handled correctly.

text/0000-remove-shell-processes.md
hone: @ekcasey isn't this already true by us removing support fro `<app>/.profile` since this changes the contract? I get you're trying to simplify/minimize the scope here, but bringing this into scope or not is still changing that contract.

text/0000-remove-shell-processes.md
jkutner: https://github.com/buildpacks/rfcs/pull/175

text/0000-remove-shell-processes.md
ekcasey: updated to use k8s style notation in https://github.com/buildpacks/rfcs/pull/168/commits/827cdd5782eceed7ba367941af03dd8d519bb3d3

text/0000-remove-shell-processes.md
ekcasey: While you are right that removing `<app>/.profile` modifies this contract, it shrinks it in a way that should reduce future occurrences of this problem. Adding `<app>/.exec` modified this contract without helping to mitigate this problem in the future.

text/0000-remove-shell-processes.md
ekcasey: After some reading I suspect the default signal handling behavior is actually better for a direct process than a shell process. Bash does not forward the SIGTERM to the child process by default and therefore there is no opportunity for a truly graceful shutdown even if the server process has signal handling logic. Therefore a buildpack author that is not thinking about this explicitly might prevent the appropriate handling of a sigterm if they default to creating a shell process.

https://veithen.io/2014/11/16/sigterm-propagation.html

cc @hone, @dmikusa-pivotal

sclevine: > therefore there is no opportunity for a truly graceful shutdown even if the server process has signal handling logic. 

Surprisingly, it's even worse than this. A simple bash invocation (e.g., `bash -c "tail -f /dev/null"`) is treated differently by bash than a complex one (e.g., `bash -c "true && tail -f /dev/null"`). In the former case, bash `exec`s itself away, resulting in a single process. So from the user's perspective, whether this works or not is quite arbitrary.

dmikusa-pivotal: Hmm, why isn't anything ever easy?  Lol.

The behavior I'd observed is with apps that don't handle signals at all. If you run them as PID1, you can't terminate the container. It's SIGKILL or the app keeps running. However, if you wrap that in `bash -c` it'll terminate as expected. I guess I didn't dig deep enough to check how Bash was in turn stopping that app, based on what's linked here it is just killing it as well and not propagating signals.

Regardless, there seems to be enough complexity here that I still think this warrants a mention in this RFC document. It's nuanced behavior that could easily be missed by a buildpack author, if we call attention to it, that gives authors a chance to investigate what it means to them. 

Even if it's as simple as "When running a process as direct, your process will always run as PID1. Buildpack authors, please be aware this may change the behavior of your application at shutdown and you should review accordingly.", I think that give's authors a heads up.

dmikusa-pivotal: One more "fun" point on this subject. If you use `sh -c` instead of `bash -c` you get more different behavior because at least on Ubuntu, `/bin/sh` is not `/bin/bash`, it's `/bin/dash`. I was running a process this morning with `sh -c` and it completely ignored SIGTERM.

ekcasey: @sclevine Are you aware of any documentation regarding when bash does or doesn't exec the child process. When I tried out you example commands I observed Bash replacing itself in memory, even for the more complex command:
```
docker run --rm --name test --entrypoint bash ubuntu:bionic -c 'true && tail -f /dev/null'
```
```
docker exec test ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 14:41 ?        00:00:00 tail -f /dev/null
root           7       0  0 14:41 ?        00:00:00 ps -ef
```

ekcasey: Okay, if I am understanding correctly, I think there are relevant two cases here:
1. the command is a complicated Bash invocation
2. the command is a simple Bash invocation

My understanding of the behavior (based on @sclevine 's description and my experimental verification):

**case 1**: A complicated command cannot be executed directly (`true && tail -f /dev/null` is not a valid direct process) and therefore `bash -c` must be explicitly included in the process. Therefore, wrt signal handling, **nothing changes**.
**case 2**: Bash is already `exec`ing itself away leaving the process as PID1. If this is converted to a direct process the same process will be PID1. Therefore, wrt signal handling, **nothing changes**.

**Summary**: With regards to signal handling nothing changes. App developers and buildpacks authors do not need to worry about changes to signal handling.

text/0000-remove-shell-processes.md
ekcasey: @samj1912 can this comment be resolved? or do you still want to advocate for more env substitution functionality, including defaults?

Personally I think we should go with the simple version for now.

text/0000-remove-shell-processes.md
ekcasey: I'm gonna assume this can be resovled

text/0000-remove-shell-specific-processes.md
ekcasey: @natalieparellano can this comment be resolved given the discussion has moved to https://github.com/buildpacks/rfcs/pull/175? Or do you feel that this RFC need further language to address this before approval?

text/0000-remove-shell-processes.md
jkutner: No, i mean if they use `bash -c` will `.profile` still be not be run? I feel like that might be unexpected.

text/0000-remove-shell-processes.md
ekcasey: @dmikusa-pivotal & @hone do you feel that this comment can be resolved given the explanation above describing how this does not affect end users or buildpack authors?

text/0000-remove-shell-processes.md
jkutner: I've created a an RFC that proposes a new mechanism to replace the `.profile` behavior (thus moving the contract to be between a builder<>app). https://github.com/buildpacks/rfcs/pull/175

jkutner:
I'm supportive of this proposal, but I'd like to see us figure out how we'll replace the gaps left behind by not including shell-specific logic in the lifecycle. I've started a proposal in https://github.com/buildpacks/rfcs/pull/175, which combined with a forthcoming proposal for "Default Buildpacks in a Builder" may accomplish this. But I'd like to get alignment on these proposals before finalizing this one (#168)

text/0000-remove-shell-processes.md
ekcasey: Yes, `.profile` will still not be run even when the command includes `bash -c`. Part of the goal here is to avoid doing anything complicated or magic. In order to run `.profile`s we would need to change the provided command from something like `['bash', '-c', 'echo hi']` to `['bash', '-c', 'source <app>/.profile && echo hi']`. Modifying the provided command if and only if it starts with `bash -c` feel like too much instrusive magic to me.

text/0000-remove-shell-processes.md
dmikusa-pivotal: I think this discussion has been good around signals. It seems unlikely that it's going to cause issues, so I think we're good there.

Maybe one final suggestion with this sentence...

>If `bash` or `cmd.exe` is required it should be included in `command`.

I think "required" might be nebulous and assumes you know what is required. It's easy to overlook restrictions of running as direct/PID1. Perhaps we could have something like:

>If currently using `direct = false`, the recommended migration path for buildpack authors is to include `bash` or `cmd.exe`  in `command`. This will result in the same behavior as `direct = false`. Before switching to `direct = true`, it is recommended you confirm environment variable interpolation is not required and that your process can function as PID1.

I guess it doesn't necessarily have to be in this line, but I think we should offer some migration advice and list the concerns an author should review before switching to `direct = true`.

nebhale (approved):

samj1912 (approved):

ekcasey:
@jkutner 
> I'm supportive of this proposal, but I'd like to see us figure out how we'll replace the gaps left behind by not including shell-specific logic in the lifecycle. I've started a proposal in #175, which combined with a forthcoming proposal for "Default Buildpacks in a Builder" may accomplish this. But I'd like to get alignment on those proposals before finalizing this one (#168)

I added language to this RFC, making its implementation contingent on a .profile buildpack. Please see https://github.com/buildpacks/rfcs/pull/168/commits/f6cb6f4be2001c9f5fecd953e05fa36cc8bc19f0. I am hoping that with this caveat we can approve this RFC so we can start making progress while we figure out the nitty-gritty of the utility buildpacks. I assume that the spec and implemenntation of this RFC will take some time, while implementation of a `.profile` buildpack can proceed very quickly once we are done bikeshedding on #175.

jkutner (approved):

text/0000-remove-shell-processes.md
ekcasey: @dmikusa-pivotal please see https://github.com/buildpacks/rfcs/pull/168/commits/14066e50a3ee999fe0ff5249362e43c52169cac5

dmikusa-pivotal: Looks great, thanks!

sclevine:
FCP ending 8/4

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,470 @@
...
+command = "dotnet my-app.dll --urls http://0.0.0.0:${PORT:-8080}"
+direct = false
+```
+NOTE: the buildpack API used by this buildpack (`0.5`) predates the introduction of `default`.
+
+Using the new API this process could look like:
+```
+[[processes]]
+type = "web"
+command = ["dotnet", "my-app.dll", "--urls", "http://0.0.0.0:${PORT}"] # the default value of PORT would need to be provided in a layer
```
ryanmoran, line 113:
```suggestion
command = ["dotnet", "my-app.dll", "--urls", "http://0.0.0.0:$(PORT)"] # the default value of PORT would need to be provided in a layer
```

Should this follow the k8s interpolation format since this is no longer a shell process?

hone (approved):
@ekcasey Do we want to standardize "Depends On" for future RFCs?

text/0000-remove-shell-processes.md
hone: @ekcasey please take a look.

ekcasey:
/queue-issue buildpacks/spec "Remove Shell-Specific Logic from Buildpack API" api/buildpack
/queue-issue buildpacks/spec "Update handling of user-provided command and args" api/buildpack
/queue-issue buildpacks/lifecycle "Remove Shell-Specific Logic"
/queue-issue buildpacks/pack "Update process types in inspect-image output to support command slice"
/queue-issue buildpacks/docs "Remove Shell-Specific Logic"

text/0000-remove-shell-processes.md
```diff
@@ -0,0 +1,468 @@
...
+
+In exchange for a reduction in complexity and cognitive overhead buildpack-authors and end users lose certain conveniences like the more intuitive profile interface. This drawback could be remediated by tooling for buildpack authors (to make creation of `exec.d` helpers easy) or by buildpacks (to support `.profile`). However, in the `.profile` case, consistency across ecosystems will be a matter of convention (like Procfile) rather than a guarantee.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## `$((<env>))` syntax
+
+We could use our own explicit syntax like `$((<env>))` or environment variable replacements instead of `$(<env>)`.
+pros:
```
jabrown85, line 359: The pros/cons look funny in the PR markdown preview. I think you may need an additional line break between lists

jabrown85 (approved):

PR #171, opened 2021-06-24
Writable Asset Cache

ekcasey:


text/0000-writable-asset-cache.md
```diff
@@ -0,0 +1,91 @@
...
+
+The way caching currently works in CNB, the layer cache is scoped to a particular image and the cached layers are only restored during a rebuild of that exact image. There are good reasons for this. First of all, on the first build of a new app there is no way of knowing which layers might be helpful to restore. Because layers affect the behavior of subsequent buildpacks and may be directly included in the final app image (launch layers), the potential consequences of deliberate cache poisoning or accidental leakage of unexpected changes between images are high. Buildpacks general cannot predict the diffID of a given layer and therefore have no way to check the integrity of a restored layer. Therefore it is essential that layer caching remains scoped to rebuilds.
+
+However, caching of assets, particularly if they are not restored to a layer, presents much less risk. Buildpacks generally known the hash of the asset they are attempting to download and can therefore check the integrity of a cached asset before using it in a layer.
```
jabrown85, line 33:
```suggestion
However, caching of assets, particularly if they are not restored to a layer, presents much less risk. Buildpacks generally know the hash of the asset they are attempting to download and can therefore check the integrity of a cached asset before using it in a layer.
```

text/0000-writable-asset-cache.md
```diff
@@ -0,0 +1,91 @@
...
+Platforms that wish to enable asset caching should mount a persistent volume to the container when running either the `builder` or `creator` phases and set `$CNB_ASSET_CACHE` in the container environment, to the path where the volume is mounted.
+
+## Buildpack API
+If the platform has provided an asset cache the lifecycle will set `$CNB_ASSET_CACHE` in the buildpack execution environment.
+
+When downloading an asset, an optimally written buildpack should follow the following flowchart:
+
+![flowchart](https://docs.google.com/drawings/d/1cY91fi9DGSAK9BKO8kcjQSXaNnE-9suk2gDfRgxeHlY/export/png)
+
+
```
jabrown85, line 57: An example buildpack using this new `$CNB_ASSET_CACHE` would be very valuable in addition to this awesome flowchart.

shresthaujjwal:
@elcasey I am very much interested in this RFC. I have a similar concern too. We are still on design phase to start implementation of buildpacks in our enterprise and we use Jenkins for our CiCd and I have the same concern. When we run our pack build there is no guarantee that the build will run on same Jenkins node as we have lot of slave nodes and any one node can pick up the build process. Hopefully this RFC would solve our concerns as we need to be able to reuse the Docker cached layers across the slave nodes. 

Thanks

PR #172, opened 2021-06-30
RFC: Remove Stacks & Mixins

sclevine:
[Readable](https://github.com/buildpacks/rfcs/blob/remove-stacks-mixins/text/0000-remove-stacks-mixins.md).

This proposal is part of a larger initiative to reduce complexity originally outlined in https://github.com/buildpacks/rfcs/pull/167

### DIFF
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -13,11 +13,11 @@
 
 *NOTE: This proposal is part of a larger initiative to reduce complexity originally outlined in https://github.com/buildpacks/rfcs/pull/167*
 
-This RFC proposes that we remove the stack ID concept from the project and replace them with existing constructs in the container image ecosystem such as operating system name, operating system version and architecture.
+This RFC proposes that we remove the stack ID concept from the project and replace it with existing constructs in the container image ecosystem such as operating system name, operating system version, and architecture.
 
-This RFC also proposes that we remove the "mixin" concept from the project entirely, eliminating strict validation of all facets of base image, buildpack compatibility, in favor of a a more optimistic approach.
+This RFC also proposes that we remove the "mixin" concept from the project entirely, eliminating strict validation of all facets of base-image-to-buildpack compatibility in favor of a more optimistic approach.
 
-This proposal prioritizes simplicity and flexibility over the avoidance of all failure scenarios, by allowing more potentially successful build to execute.
+This proposal prioritizes simplicity and flexibility over the avoidance of all failure scenarios by allowing more potentially successful builds to execute.
 
 # Motivation
 [motivation]: #motivation
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -149,24 +149,30 @@ All existing labels and environment variables for stacks and mixins may be prese
 These labels will be deprecated (but allowed) for the forseeable future.
 If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user. More concretely, when using a stack with `io.buildpacks.stacks.bionic` lifecycle will assume the following metadata is set if these fields are empty:
 
-config.os = Linux
-config.architecture = x86_64
+```
+config.os = "linux"
+config.architecture = "x86_64"
 io.buildpacks.distribution.name = "ubuntu"
 io.buildpacks.distribution.version = "18.04"
+```
 
 Moving forward it's encouraged for buildpack authors to support both `[[stacks]]` and `[[targets]]` sections in `buildpack.toml` for maximum compatibility. In order to ease this process for those using the `io.buildpacks.stacks.bionic`, lifecycle will translate any section that sets this as on of the `stacks`:
 
+```toml
 [[stacks]]
 id = "io.buildpacks.stacks.bionic
+```
 
 to 
 
+```toml
 [[targets]]
 os = "linux"
 arch = "x86_64"
 [[targets.distributions]]
 name = "ubuntu"
 versions = ["18.04"]
+```
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -32,16 +32,18 @@ Summary of changes:
 
 ## Base Image Metadata
 
-Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
-- OS (e.g., "linux", `$GOOS`)
-- Architecture (e.g., "x86_64", `$GOARCH`)
-- Distribution (optional) (e.g., "ubuntu", `$ID`)
-- Version (optional) (e.g., "18.04", `$VERSION_ID`)
+Instead of a stack ID, runtime and build-time base images must contain the following canonicalized metadata:
+- OS (e.g., "linux", `$GOOS`), specified as `os` in the base image `config` 
+- Architecture (e.g., "arm", `$GOARCH`), specified as `architecture` in the base image `config`
+- Architecture Variant (optional) (e.g., "v6", `$GOARM`), specified as `variant` in the base image `config`
+- Distribution (optional) (e.g., "ubuntu", `$ID`), specified as a label `io.buildpacks.distribution.name`
+- Version (optional) (e.g., "18.04", `$VERSION_ID`), specified as a label `io.buildpacks.distribution.version`
 
-OS and Architecture must be valid identifiers as defined [here](https://golang.org/doc/install/source#environment).
+OS, Architecture, and Architecture Variant must be valid identifiers as defined in the [OCI Image specification](https://github.com/opencontainers/image-spec/blob/main/config.md).
 For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
+The `os.version` field in an base image `config` may contain combined distribution and version information, but it is not used by the lifecycle.
 
-The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture.
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
 
 App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
 
@@ -53,17 +55,36 @@ When an app image is rebased, `pack rebase` will fail if the new run image and p
 [[targets]]
 os = "linux"
 arch = "x86_64"
-[[targets.distros]]
+[[targets.distributions]]
 name = "ubuntu"
 versions = ["18.04", "20.04"]
 
 [[targets]]
 os = "linux"
 arch = "x86_64"
-[[targets.distros]]
+[[targets.distributions]]
+name = "ubuntu"
+versions = ["14.04", "16.04"]
+
+[[targets]]
+os = "linux"
+arch = "arm"
+variant = "v6"
+[[targets.distributions]]
 name = "ubuntu"
 versions = ["14.04", "16.04"]
 ```
+
+## Runtime Metadata
+
+To allow different runtime base images to be used, and to support cross-compilation in the future, buildpacks may need access to the runtime base image's target metadata.
+The following environment variables will be available to buildpacks directly in the build-time environment (not via `/platform/env`):
+- `CNB_TARGET_OS`
+- `CNB_TARGET_ARCH`
+- `CNB_TARGET_VARIANT`
+- `CNB_TARGET_DISTRO_NAME`
+- `CNB_TARGET_DISTRO_VERSION`
+- `CNB_TARGET_ID` (optional ID, if present on runtime base image `io.buildpacks.id` label)
 
 ## Mixins
 
@@ -104,10 +125,17 @@ If an SBoM is available, `pack rebase` will fail if packages are removed from th
 This check may be skipped by passing a new `--force` flag to `pack rebase`.
 However, this validation is not enforced by the specification.
 
+### Migration
+
+All existing labels and environment variables for stacks and mixins may be preserved on a base image until all users have migrated to the new format.
+These labels will be deprecated (but allowed) for the forseeable future.
+If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Involves breaking changes.
+- Involves breaking changes when existing metadata is removed.
+- `CNB_TARGET_*` env vars assume a single target -- may require more breaking changes to support parallel, single-container cross-compilation in the future.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -45,7 +45,9 @@ The `os.version` field in an base image `config` may contain combined distributi
 
 The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).
 Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
-If the `targets` list is empty, a single target with `os = "linux"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+If the `targets` list is empty and `/bin/build` is present, a target with `os = "linux"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+If the `targets` list is empty and `/bin/build.bat` or `/bin/build.exe` is present, a target with `os = "windows"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+
 
 App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images).
 An image without a specified Distribution is compatible with images specifying any Distribution.
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -167,6 +167,7 @@ If the newly-specified field values are missing, the lifecycle and pack may used
 
 - How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
 - What should builder.toml (and similar stack-dependent config files) look like? What should assets look like? Note: these could be decided in subsequent subteam RFCs / spec PRs.
+- How will this affect the UX for pack buidpack package? For example package.toml contains a conflicting [platform table](https://buildpacks.io/docs/reference/config/package-config/#platform-_optional_). This should be resolved in a separate subteam RFC. 
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -98,6 +98,8 @@ Buildpacks may specify a list of package names (as PURL URLs without versions or
 [alternatives]: #alternatives
 
 - Keep stacks.
+- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
+- Continue to allow for static definition of runtime package requirements in `buildpack.toml`.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -24,7 +24,7 @@ The "stack" and "mixin" concepts add unnecessary complexity to the project and m
 [what-it-is]: #what-it-is
 
 Summary of changes:
-- Replace mixins with a [CycloneDX](https://cyclonedx.org)-formatted list of packages.
+- Remove mixins
 - Replace stack metadata (including stack IDs) with canonical OS metadata.
 
 # How it Works
@@ -41,7 +41,7 @@ Instead of a stack ID, runtime and build-time base images are labeled with the f
 OS and Architecture must be valid identifiers as defined [here](https://golang.org/doc/install/source#environment).
 For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
 
-The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture. Each entry may also contain a list of package names (as PURL URLs without versions or qualifiers) that specify detect-time and build-time (but not runtime) OS package dependencies. Packages may be specified for all Versions of a distribution, or for a specific Version. Buildpacks may express **runtime** OS package dependencies during detection (see "Runtime Base Image Requirements" below).
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture.
 
 App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
 
@@ -55,30 +55,21 @@ os = "linux"
 arch = "x86_64"
 [[targets.distros]]
 name = "ubuntu"
-[[targets.distros.versions]]
-version = "18.04"
-packages = ["pkg:deb/ubuntu/curl"]
-[[targets.distros.versions]]
-version = "20.04"
-packages = ["pkg:deb/ubuntu/curl2"]
+versions = ["18.04", "20.04"]
 
 [[targets]]
 os = "linux"
 arch = "x86_64"
 [[targets.distros]]
 name = "ubuntu"
-packages = ["pkg:deb/ubuntu/curl"]
-[[targets.distros.versions]]
-version = "14.04"
-[[targets.distros.versions]]
-version = "16.04"
+versions = ["14.04", "16.04"]
 ```
 
 ## Mixins
 
-The mixins label on each base image is replaced by a layer in each base image containing a single file consisting of a [CycloneDX](https://cyclonedx.org)-formatted list of packages. Each package entry has a [PURL](https://github.com/package-url/purl-spec)-formatted ID that uniquely identifies the package.
+Mixins are no longer used. If an SBoM is available, platforms may warn when, e.g., a rebase operation would change the available packages.
 
-### Example: CycloneDX package list
+### Example: CycloneDX SBoM
 
 ```json
 {
@@ -97,22 +88,21 @@ The mixins label on each base image is replaced by a layer in each base image co
       "name": "libcurl",
       "version": "4.5.6",
       "purl": "pkg:deb/ubuntu/libcurl@4.5.6"
-    }
+    },
+    ...
   ]
 }
 ```
 
 ### Validations
 
-Buildpack base image metadata and packages specified in `buildpack.toml`'s `targets` list are validated against the runtime and build-time base images.
+Buildpack base image metadata specified in `buildpack.toml`'s `targets` list are validated against the runtime and build-time base images.
 
 Runtime and build-time base image packages are no longer validated against each other.
 
-When an app image is rebased, `pack rebase` will fail if packages are removed from the new runtime base image. This check may be skipped by passing a new `--force` flag to `pack rebase`.
-
-## Runtime Base Image Requirements
-
-Buildpacks may specify a list of package names (as PURL URLs without versions or qualifiers) in a `packages` table in the build plan. The build will fail if these packages are not present in the runtime base image.
+If an SBoM is available, `pack rebase` will fail if packages are removed from the new runtime base image.
+This check may be skipped by passing a new `--force` flag to `pack rebase`.
+However, this validation is not enforced by the specification.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -124,12 +114,11 @@ Buildpacks may specify a list of package names (as PURL URLs without versions or
 
 - Keep stacks.
 - Make the above changes but keep some existing terminology: stacks, build-image, run-image.
-- Continue to allow static definition of runtime package requirements in `buildpack.toml`.
+- Continue to allow buildpacks to specify package requirements (e.g., by PURL instead of mixins)
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- Should runtime packages be determined during the detect or build phase? Opinion: detect phase, so that (in a later RFC), a runtime base image's app-specified Dockerfiles may by applied in parallel to the buildpack build process.
 - How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
 - What should builder.toml (and similar stack-dependent config files) look like? What should assets look like? Note: these could be decided in subsequent subteam RFCs / spec PRs.
 
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -43,7 +43,13 @@ Instead of a stack ID, runtime and build-time base images must contain the follo
 - Distribution (optional) (e.g., "ubuntu", `$ID`), specified as a label `io.buildpacks.distribution.name`
 - Version (optional) (e.g., "18.04", `$VERSION_ID`), specified as a label `io.buildpacks.distribution.version`
 
+Additionally, the runtime base may contain the following metadata:
+- Target ID (optional) (e.g., "minimal"), specified as a label `io.buildpacks.id`
+
 OS, Architecture, and Architecture Variant must be valid identifiers as defined in the [OCI Image specification](https://github.com/opencontainers/image-spec/blob/main/config.md).
+
+Target ID is an identifier specified on the runtime base image that must be provided to buildpacks as `CNB_TARGET_ID` during the build process.
+This allows buildpacks to change their behavior if a run image is selected (e.g., distroless) that has special properties outside of OS, architecture, etc.
 
 For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
 The `os.version` field in an base image `config` may contain combined distribution and version information, but it is not used by the lifecycle.
@@ -54,7 +60,6 @@ The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where eac
 Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
 If the `targets` list is empty and `/bin/build` is present, a target with `os = "linux"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
 If the `targets` list is empty and `/bin/build.bat` or `/bin/build.exe` is present, a target with `os = "windows"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
-
 
 App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images).
 An image without a specified Distribution is compatible with images specifying any Distribution.
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -105,7 +105,7 @@ Buildpacks may specify a list of package names (as PURL URLs without versions or
 [unresolved-questions]: #unresolved-questions
 
 - Should packages be determined during the detect or build phase? Opinion: detect phase, so that (in a later RFC), a runtime base image's app-specified Dockerfiles may by applied in parallel to the buildpack build process.
-
+- How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -13,7 +13,11 @@
 
 *NOTE: This proposal is part of a larger initiative to reduce complexity originally outlined in https://github.com/buildpacks/rfcs/pull/167*
 
-This RFC proposes that we remove the "stack" and "mixin" concepts from the project and replace them with existing constructs in the container image ecosystem such as base images, Dockerfiles, and OS packages.
+This RFC proposes that we remove the stack ID concept from the project and replace them with existing constructs in the container image ecosystem such as operating system name, operating system version and architecture.
+
+This RFC also proposes that we remove the "mixin" concept from the project entirely, eliminating strict validation of all facets of base image, buildpack compatibility, in favor of a a more optimistic approach.
+
+This proposal prioritizes simplicity and flexibility over the avoidance of all failure scenarios, by allowing more potentially successful build to execute.
 
 # Motivation
 [motivation]: #motivation
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -40,11 +40,38 @@ Instead of a stack ID, runtime and build-time base images are labeled with the f
 
 For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
 
-The `stacks` list in `buildpack.toml` is replaced by a `platforms` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture. Each entry may also contain a list of package names (as PURL URLs without versions or qualifiers) that specify detect-time and build-time (but not runtime) OS package dependencies. Buildpacks may express runtime OS package dependencies during detection (see "Runtime Base Image Requirements" below).
+The `stacks` list in `buildpack.toml` is replaced by a `platforms` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture. Each entry may also contain a list of package names (as PURL URLs without versions or qualifiers) that specify detect-time and build-time (but not runtime) OS package dependencies. Packages may be specified for all Versions of a distribution, or for a specific Version. Buildpacks may express **runtime** OS package dependencies during detection (see "Runtime Base Image Requirements" below).
 
-App image builds fail if the build image and selected run image have mismatched metadata. We may consider introducing a flag to skip this validation.
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
 
 When an app image is rebased, `pack rebase` will fail if the new run image and previous run image have mismatched metadata. This check may be skipped for Distribution and Version by passing a new `--force` flag to `pack rebase`.
+
+#### Example: buildpack.toml `platforms` table
+
+```toml
+[[platforms]]
+os = "linux"
+arch = "x86_64"
+[[platforms.distros]]
+name = "ubuntu"
+[[platforms.distros.versions]]
+version = "18.04"
+packages = ["pkg:deb/ubuntu/curl"]
+[[platforms.distros.versions]]
+version = "20.04"
+packages = ["pkg:deb/ubuntu/curl2"]
+
+[[platforms]]
+os = "linux"
+arch = "x86_64"
+[[platforms.distros]]
+name = "ubuntu"
+packages = ["pkg:deb/ubuntu/curl"]
+[[platforms.distros.versions]]
+version = "14.04"
+[[platforms.distros.versions]]
+version = "16.04"
+```
 
 ## Mixins
 
--- a/text/0000-remove-stacks-mixins.md
+++ b/text/0000-remove-stacks-mixins.md
@@ -147,7 +147,26 @@ However, this validation is not enforced by the specification.
 
 All existing labels and environment variables for stacks and mixins may be preserved on a base image until all users have migrated to the new format.
 These labels will be deprecated (but allowed) for the forseeable future.
-If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user.
+If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user. More concretely, when using a stack with `io.buildpacks.stacks.bionic` lifecycle will assume the following metadata is set if these fields are empty:
+
+config.os = Linux
+config.architecture = x86_64
+io.buildpacks.distribution.name = "ubuntu"
+io.buildpacks.distribution.version = "18.04"
+
+Moving forward it's encouraged for buildpack authors to support both `[[stacks]]` and `[[targets]]` sections in `buildpack.toml` for maximum compatibility. In order to ease this process for those using the `io.buildpacks.stacks.bionic`, lifecycle will translate any section that sets this as on of the `stacks`:
+
+[[stacks]]
+id = "io.buildpacks.stacks.bionic
+
+to 
+
+[[targets]]
+os = "linux"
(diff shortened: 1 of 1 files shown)


text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Base Image Metadata
+
+Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
+- OS (e.g., "linux", `$GOOS`)
+- Architecture (e.g., "x86_64", `$GOARCH`)
+- Distribution (optional) (e.g., "ubuntu", `$ID`)
+- Version (optional) (e.g., "18.04", `$VERSION_ID`)
+
+For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
```
samj1912, line 41: Do we also need to define the windows equivalent?

samj1912, line 41: Would the lifecycle verify this against `/etc/os-release` if present? Or would it just accept what is in the appropriate labels?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,83 @@
...
+
+App image builds fail if the build image and selected run image have mismatched metadata. We may consider introducing a flag to skip this validation.
+
+When an app image is rebased, `pack rebase` will fail if the new run image and previous run image have mismatched metadata. This check may be skipped for Distribution and Version by passing a new `--force` flag to `pack rebase`.
+
+## Mixins
+
+The mixins label on each base image is replaced by a layer in each base image containing a single file consisting of a [CycloneDX](https://cyclonedx.org)-formatted list of packages. Each package entry has a [PURL](https://github.com/package-url/purl-spec)-formatted ID that uniquely identifies the package.
```
samj1912, line 51: Would it be easier just to replace mixins with `purls` in this RFC and leave CycloneDx out for a separate RFC that introduces stack/base image BOMs and how they intereact with the final image?

The rest of this RFC should still be valid with this change.

So we replace stacks with the above information and mixins defined by the CNB project for official stacks with `purls`?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,83 @@
...
+- Replace stack metadata (including stack IDs) with canonical OS metadata.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Base Image Metadata
+
+Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
+- OS (e.g., "linux", `$GOOS`)
+- Architecture (e.g., "x86_64", `$GOARCH`)
```
imjasonh, line 37: Off-topic: Is there any plan/idea for buildpacks to support building multi-platform images, such that a single image manifest can describe an image for `linux/arm64` and `linux/arm64` and `windows/amd64`, etc.? Or is the intention that someone wanting to build a multi-platform image should use buildpacks to produce multiple platform-specific images, then stitch them together with something like [`docker manifest create`](https://docs.docker.com/engine/reference/commandline/manifest/#manifest-create) ?

text/0000-remove-stacks-mixins.md
sclevine: I would not expect the lifecycle to be aware of `/etc/os-release`. Usage of the values from `/etc/os-release` would be a convention between Linux buildpacks and Linux base images that is suggested by the spec.

sclevine: Yes, but I'm not sure if it needs to be decided for each OS in this RFC vs. the spec PR.

text/0000-remove-stacks-mixins.md
sclevine: I worry that would leave us supporting three possible formats instead of two. Maybe we should block this RFC on SBoM format consensus?

text/0000-remove-stacks-mixins.md
sclevine: No current RFC, but I definitely agree that we should eventually support exporting multi-platform manifest indexes.

The BuildKit frontend might help us with this, via buildx: https://github.com/buildpacks/pack/issues/768

For non-local builds, we could extend the current exporter to accept a set of layers for each OS/ARCH combination.

In either case, we'd probably want to pass standardized env vars (e.g., `$CNB_TARGET_ARCH`) to the builder and buildpacks to support cross-compilation.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
+
+When an app image is rebased, `pack rebase` will fail if the new run image and previous run image have mismatched metadata. This check may be skipped for Distribution and Version by passing a new `--force` flag to `pack rebase`.
+
+#### Example: buildpack.toml `platforms` table
```
sclevine, line 49: @ekcasey as requested. Looks very noisy, but I think it meets the requirements.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
```
ekcasey, line 45: > App image builds fail if the build image and selected run image have mismatched metadata

Haven't thought this through all the way but would it make sense to this in the build plan as well. E.g. a buildpack could specify in `buildpack.toml` that it needs to run on a particular distribution but could express at detect time that the resulting layers could run on any-linux (distroless images).

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+[[platforms.distros.versions]]
+version = "18.04"
+packages = ["pkg:deb/ubuntu/curl"]
+[[platforms.distros.versions]]
+version = "20.04"
+packages = ["pkg:deb/ubuntu/curl2"]
+
+[[platforms]]
+os = "linux"
+arch = "x86_64"
```
ekcasey, line 66:
```suggestion
os = "linux"
arch = "arm64"
```
I think this needs to have a different os/arch combo than the other entries right? Otherwise selecting the correct buildpackage from the manifest list will require extra logic.

text/0000-remove-stacks-mixins.md
ekcasey: I would personally like to decouple base image package requirements from the sBOM. I.E I think platforms should be able to validate againt a flat list of packages in a label or file, without having to parse CycloneDx

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+
+Buildpack base image metadata and packages specified in `buildpack.toml`'s `platforms` list are validated against the runtime and build-time base images.
+
+Runtime and build-time base image packages are no longer validated against each other.
+
+When an app image is rebased, `pack rebase` will fail if packages are removed from the new runtime base image. This check may be skipped by passing a new `--force` flag to `pack rebase`.
+
+## Runtime Base Image Requirements
+
+Buildpacks may specify a list of package names (as PURL URLs without versions or qualifiers) in a `packages` table in the build plan. The build will fail if these packages are not present in the runtime base image.
```
ekcasey, line 90: Example please :)

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Involves breaking changes.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep stacks.
```
ekcasey, line 100:
```suggestion
- Keep stacks.
- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
- Continue to allow for static definition of runtime package requirements in `buildpack.toml`.
```

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should packages be determined during the detect or build phase? Opinion: detect phase, so that (in a later RFC), a runtime base image's app-specified Dockerfiles may by applied in parallel to the buildpack build process.
+
```
ekcasey, line 106:
```suggestion
- How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
```
I think this is a question that we need an answer to before we can move forward with this RFC.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
+
+When an app image is rebased, `pack rebase` will fail if the new run image and previous run image have mismatched metadata. This check may be skipped for Distribution and Version by passing a new `--force` flag to `pack rebase`.
+
+#### Example: buildpack.toml `platforms` table
+
+```toml
```
ekcasey, line 51: I also think we need to show what would happens to assets here. Assets current [include a list of compatible stacks](https://github.com/buildpacks/rfcs/blob/main/text/0073-offline-buildpackages.md#general-format)

text/0000-remove-stacks-mixins.md
dmikusa-pivotal: `An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.`

Is this what would replace wildcard stack support, i.e. `stack="*"`.

text/0000-remove-stacks-mixins.md
dmikusa-pivotal: Could you explain the sample, please? If I have this example in my buildpack's buildpack.toml, what does that mean for the buildpack & compatibility with images?

text/0000-remove-stacks-mixins.md
sclevine: > E.g. a buildpack could specify in buildpack.toml that it needs to run on a particular distribution but could express at detect time that the resulting layers could run on any-linux (distroless images).

I'm not sure this needs to be dynamic. I wonder if this could be part of the multi-platform interface? See: https://github.com/buildpacks/rfcs/pull/172#discussion_r661899153

> Is this what would replace wildcard stack support, i.e. stack="*".

The requirement is just describing the contract between the build and run images. Wildcard stack support is achieved by leaving off Distribution and Version entirely. (There is no wildcard support for OS/Architecture.)

text/0000-remove-stacks-mixins.md
sclevine: The intent was to allow different buildpacks to be built for different distribution with the same os/arch. Manifest Indexes allow this. We need custom logic to use buildpackages anyways.

text/0000-remove-stacks-mixins.md
sclevine: Will add example minimal CycloneDX JSON for feedback

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,112 @@
...
+- Replace mixins with a [CycloneDX](https://cyclonedx.org)-formatted list of packages.
+- Replace stack metadata (including stack IDs) with canonical OS metadata.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Base Image Metadata
+
+Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
+- OS (e.g., "linux", `$GOOS`)
```
jabrown85, line 36: Are these strings able to be anything? Could I, if I wanted to, create a `JesseOS` OS string to tightly control the buildpacks that have explicitly updated to work on my build/run images?

text/0000-remove-stacks-mixins.md
samj1912:
```json

{
  "bomFormat": "CycloneDX",
  "specVersion": "1.3",
  "version": 1,
  "components": [
    {
      "type": "library",
      "name": "tomcat-catalina",
      "version": "9.0.14",
      "purl": "pkg:maven/org.apache.tomcat/tomcat-catalina@9.0.14"
    },
    {
      "type": "library", 
      "name": "card-verifier",
      "version": "1.0.2",
      "purl": "pkg:maven/org.acme/card-verifier@1.0.2?repository_url=repo.acme.org/maven"
    }
  ]
}
```

This is the minimal required set of fields. purl is optional for cdx but we need it.

sclevine: Looks quite minimal to me.

@ekcasey Given that a separate list may end up as a list of objects anyways, and given that we'd likely need to put both lists in files in a layer, do you still feel like a separate list is worth it?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,110 @@
...
+- Architecture (e.g., "x86_64", `$GOARCH`)
+- Distribution (optional) (e.g., "ubuntu", `$ID`)
+- Version (optional) (e.g., "18.04", `$VERSION_ID`)
+
+For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
+
+The `stacks` list in `buildpack.toml` is replaced by a `platforms` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture. Each entry may also contain a list of package names (as PURL URLs without versions or qualifiers) that specify detect-time and build-time (but not runtime) OS package dependencies. Packages may be specified for all Versions of a distribution, or for a specific Version. Buildpacks may express **runtime** OS package dependencies during detection (see "Runtime Base Image Requirements" below).
```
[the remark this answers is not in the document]
sclevine: I'm not sure I understand the proposed example. Some questions:

- Where does the build-time base image come from for the JDK and Maven buildpacks?
- Is a different build-time base image selected when the runtime base image OS changes?
- Are the buildpack IDs optional?
- How does the `packages` list work?
- Are you proposing that we re-define stack to only refer to the runtime base image?

Would be great to dig into this during the weekly WG meeting.

I'm not attached to `platforms` for the table name, but I would like to remove unnecessary, project-specific terms to make the project more accessible.

text/0000-remove-stacks-mixins.md
sclevine: OS and Architecture would be restricted to `GOOS` and `GOARCH` values. Distribution and Version could technically be anything, but should match `/etc/os-release` for Linux distros.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,112 @@
...
+- Involves breaking changes.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep stacks.
+- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
+- Continue to allow for static definition of runtime package requirements in `buildpack.toml`.
+
+# Unresolved Questions
```
jkutner, line 104: How does this change the `[stack]` in `builder.toml`?

sclevine: Added to unresolved questions

text/0000-remove-stacks-mixins.md
sclevine: Added to unresolved questions, since assets likely needs other significant changes

sclevine:
Idea I'd like feedback on: what if we dropped the `packages` list from both buildpack.toml and detect, at least for now? Builder authors could be responsible for reading buildpack documentation and installing any extra packages. We could re-introduce this functionality, if necessary, as part of #174.

dmikusa-pivotal:
> Idea I'd like feedback on: what if we dropped the `packages` list from both buildpack.toml and detect, at least for now? Builder authors could be responsible for reading buildpack documentation and installing any extra packages. We could re-introduce this functionality, if necessary, as part of #174.

I like that this change makes things simpler and I'm not sure I would have used this `packages` list often.

I do think it would have been helpful for the Paketo PHP buildpack, given that PHP and its many C-based extensions have lots of library requirements. At the same time, it would have been very tedious to spell out all of the packages to really take advantage of this feature (same with mixins). I would also be OK with having a builder specific for PHP, where the requirements are spelled out, especially the way that #173 is shaping up.

nebhale (approved):

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,139 @@
...
+[what-it-is]: #what-it-is
+
+Summary of changes:
+- Replace mixins with a [CycloneDX](https://cyclonedx.org)-formatted list of packages.
+- Replace stack metadata (including stack IDs) with canonical OS metadata.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Base Image Metadata
```
ekcasey, line 33: Can we add the schema for the base image metadata to this proposal? I am assuming we want this metadata in one or more labels?

The OCI spec includes similar fields in the platform section of an image index, but not in the image config itself, and therefore it may not be sufficient for our purposes given we still support daemon-only workflows.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,139 @@
...
+
+Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
```
ekcasey, line 35: Will any or all of this data be provided to buildpacks during build or detect? If so:
1. in what format?
2. will run image metadata be provided somehow, or just build image?

We should provide explicit answers to the above in this RFC.

Thoughts on 1:
It seems like we will have to provide this data so that buildpacks, if they support multiple targets, can download and/install the correct assets. We could theoretically force buildpacks to read `/etc/os-release` for some of this, but that doesn't include architecture and it would be nice to provide everything in a consistent format.

Thoughts on point 2:
Right now there are a couple examples in Paketo Java buildpacks where build logic is conditional on the stack ID:
- [native image buildpack stactically compiles](https://github.com/paketo-buildpacks/native-image/blob/main/native/native_image.go#L72-L74) more libraries on the tiny stack
- [Profile buildpack creates direct process for the tiny stack](https://github.com/paketo-buildpacks/procfile/blob/3a6c8fdc14796db671d3697b916792b468307e5e/procfile/build.go#L57)

Given how many Java buildpacks there are, the fact that I am only aware of two examples suggests this isn't _super_ important. The Procfile example in particular is ugly and mostly just demonstrates how trying to fit that existing Procile interface to a distroless world is a bad idea. The native image example is more interesting though, without this data we would need to statically compile always and forgo the ability to patch CVEs in these libraries with rebase.

It's probably not a dealbreaker, but if we aren't providing run image metadata to buildpacks in any way I think this is a change worth calling out explicitly in the RFC.

cc @dmikusa-pivotal

text/0000-remove-stacks-mixins.md
ekcasey: > Is there any plan/idea for buildpacks to support building multi-platform images

This is hard to do well given that (some exceptions wrt docker-for-mac notwithstanding) the build takes place in a docker container and a given daemon only supports a single os/arch.

You could theoretically have a buildpacks execute on one target while producing launch layers that are valid for a different target but this puts a lot of responsibility on the buildpack author and comes with a lot of gotchas. e.g
- a single layer is unlikely to be valid as both a `build` and `launch` layers
- ecosystem tools that buildpacks delegate to will make assumptions based on the build environment

imho this is unlikely to end well

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,139 @@
...
+
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images). An image without a specified Distribution is compatible with images specifying any Distribution. An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
+
+When an app image is rebased, `pack rebase` will fail if the new run image and previous run image have mismatched metadata. This check may be skipped for Distribution and Version by passing a new `--force` flag to `pack rebase`.
```
ekcasey, line 48: I think the fact that you went with "run image" instead of "run base image" here illustrates my point about the merits of keeping some of the existing terminology ;)

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,139 @@
...
+Buildpack base image metadata and packages specified in `buildpack.toml`'s `targets` list are validated against the runtime and build-time base images.
+
+Runtime and build-time base image packages are no longer validated against each other.
+
+When an app image is rebased, `pack rebase` will fail if packages are removed from the new runtime base image. This check may be skipped by passing a new `--force` flag to `pack rebase`.
+
+## Runtime Base Image Requirements
+
+Buildpacks may specify a list of package names (as PURL URLs without versions or qualifiers) in a `packages` table in the build plan. The build will fail if these packages are not present in the runtime base image.
+
```
ekcasey, line 116: I think we need to address the migration path in this RFC

text/0000-remove-stacks-mixins.md
sclevine: For 1: Operating systems generally provide standardized ways of retrieving this information. E.g., `uname`, `lsb_release`. I'm not convinced that it's necessary to create a new interface for it, but I'm also not opposed to it.

For 2: Given that the run image isn't know until immediately before the build process, it seems reasonable to pass an identifier from a run image label to the buildpacks. Could this be a separate (future) RFC?

text/0000-remove-stacks-mixins.md
sclevine: > You could theoretically have a buildpacks execute on one target while producing launch layers that are valid for a different target

I'm not opposed to this. Cross-compilation seems like a reasonable feature request to me. We would probably want to pass "target" metadata (e.g., from a run image label) to the buildpacks. When multiple run images are provided, buildpacks could produce multiple versions of the same layer for different architectures.

text/0000-remove-stacks-mixins.md
sclevine: I think "run image" as a shorthand for "runtime base image" is reasonable. Less convinced about "stack" -- I'd like to move away from tight build image / run image contracts.

text/0000-remove-stacks-mixins.md
sclevine: Resolved by getting rid of mixins entirely.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,139 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Involves breaking changes.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep stacks.
+- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
```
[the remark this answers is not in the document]
sclevine: >  using stack to refer to a set of images that are **intended to work together** is useful

This is actually part of the reason I'd like to get rid of the term "stack." I don't think the project should enforce the contract between the images. Platforms and image authors can do this when it's important.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,128 @@
...
+Summary of changes:
+- Remove mixins
+- Replace stack metadata (including stack IDs) with canonical OS metadata.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Base Image Metadata
+
+Instead of a stack ID, runtime and build-time base images are labeled with the following canonicalized metadata:
```
samj1912, line 35: Could you also add the label keys to be used for all of this? Also, do we need labels for arch and os given that the manifest should have them? Could we re-use OCI index keys instead? https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions

text/0000-remove-stacks-mixins.md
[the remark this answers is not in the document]
jkutner: I think we could keep "stack" without continuing to enforce a contract between images. The stack concept would still be useful to encapsulate the various properties represented by the propose `target` table.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,128 @@
...
+- OS (e.g., "linux", `$GOOS`)
+- Architecture (e.g., "x86_64", `$GOARCH`)
+- Distribution (optional) (e.g., "ubuntu", `$ID`)
+- Version (optional) (e.g., "18.04", `$VERSION_ID`)
+
+OS and Architecture must be valid identifiers as defined [here](https://golang.org/doc/install/source#environment).
+For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
+
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS and Architecture.
```
jkutner, line 44: Would we support a wildcards in the `target`?

frenzymadness: Is the `target` mandatory for buildpacks. What about a very universal buildpack compatible with all supported images?

text/0000-remove-stacks-mixins.md
sclevine: Done (and moved some labels to `config` fields)

text/0000-remove-stacks-mixins.md
sclevine: Let me know if the last set of changes addresses these points.

text/0000-remove-stacks-mixins.md
sclevine: Added, and switched to OCI config keys (not index keys) + labels for remaining fields.

text/0000-remove-stacks-mixins.md
sclevine: Added extra arm example

text/0000-remove-stacks-mixins.md
sclevine: Addressed, let me know if you think I covered the relevant migration scenarios

text/0000-remove-stacks-mixins.md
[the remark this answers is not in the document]
sclevine: I added an optional "target ID" on the runtime base image that brings back some of the functionality without implying that the images are coupled. Any thoughts?

text/0000-remove-stacks-mixins.md
jabrown85: Does the omission of `[targets]` in `buildpack.toml` mean `* <any stack>`?

sclevine: At least one OS/arch combination is necessary, but it doesn't need to include a distro/version.

I'm hesitant to make OS/arch (e.g., linux x86) optional, because OCI images must contain these fields.

jkutner: If we were to support wildcards, could the builder or platform provide the OS/arch labels for the image? Surely it can infer this from the runtime base image being targeted.

sclevine: I just meant for the buildpack image. If we supported a single buildpack image that could run on multiple architectures, it's `architecture` field would incorrect. Do you have use cases for multi-OS or multi-architecture buildpacks? The correct OS/architecture will get selected automatically when the buildpack image is pulled.

jkutner: My goal here is to eliminate boiler plate, and to make buildpack authorship simpler. Not necessarily to support multiple archs

jkutner: With `stack` these details where encapsulated and hidden behind IDs, while the builder owner managed what those IDs mapped to. By removing stack, we're pushing that onto the buildpack author (who now need to ensure that the `targets` they put in their `buildpack.toml` align with all the targets in the other buildpacks they want to combine with.

jkutner: Actually, I think there are plenty of multi-OS or multi-architecture buildpacks. Maybe Maven, Gradle, etc. Most buildpacks that run something atop a language VM. Maybe https://github.com/jkutner/no-secrets-buildpack, which just scans a repos for secrets (doesn't even provide or install anything)

sclevine: Added default values (matching OS & x86_64) to reduce boilerplate. Now buildpack authors only need to care about targets when they want to restrict usage of their buildpack to a specific distro/version, or when they need support for less common architectures. The OS is detected automatically (`/bin/build` vs. `/bin/build.bat`).

OCI images must have a single OS and single architecture field, and I don't think we should create buildpack images for more than one architecture automatically. Can we iterate on this in the future?

sclevine:
All outstanding feedback addressed. Looking for additional feedback or approval @buildpacks/core-team

jabrown85 (approved):

text/0000-remove-stacks-mixins.md
jkutner: Thanks for adding this! Would we also continue to allow `[[stacks]]` in `buildpack.toml`, and ignore it at some point?

text/0000-remove-stacks-mixins.md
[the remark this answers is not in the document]
jromero: > > using stack to refer to a set of images that are **intended to work together** is useful
> 
> This is actually part of the reason I'd like to get rid of the term "stack." I don't think the project should enforce the contract between the images. Platforms and image authors can do this when it's important.

I agree with @sclevine on this one. The value of the term `stack` just creates unnecessary indirection. When giving presentations, doing write ups, etc the first thing we need to do when we mention `stacks` is break them up into two images. The two images become instantly clear to understand when we mention them as "a runtime image" and "a build time image".

Not to mention that the term stack is more commonly used to mean things layered on top of each other working together; not two disjointed things that are of the same type but used in different contexts.

text/0000-remove-stacks-mixins.md
sclevine: Definitely, I can add that as well

text/0000-remove-stacks-mixins.md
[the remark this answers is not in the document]
jkutner: Yea, I agree you on that point @jromero (it's a confusing concept to learn). I'm comfortable with the changes @sclevine has made to `target` now, and am ok with dropping stack as a result.

jkutner (approved):

text/0000-remove-stacks-mixins.md
ekcasey: In general I lean to including as much as possible in the RFC so that the spec PR feedback can focus on the wording more than the content, the the greatest extent possible.

Also, so that the RFC document itself is as useful as possible to readers.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,162 @@
...
+
+- Keep stacks.
+- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
+- Continue to allow buildpacks to specify package requirements (e.g., by PURL instead of mixins)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
+- What should builder.toml (and similar stack-dependent config files) look like? What should assets look like? Note: these could be decided in subsequent subteam RFCs / spec PRs.
```
ekcasey, line 157:
```suggestion
- What should builder.toml (and similar stack-dependent config files) look like? What should assets look like? Note: these could be decided in subsequent subteam RFCs / spec PRs.
- How will this affect the UX for pack buidpack package? For example package.toml contains a conflicting [platform table](https://buildpacks.io/docs/reference/config/package-config/#platform-_optional_). This should be resolved in a separate subteam RFC. 
```

text/0000-remove-stacks-mixins.md
ekcasey:
```suggestion
When an app image is rebased, `rebaser` must fail if the new run image and previous run image have mismatched metadata. This check can be skipped for Distribution and Version by passing a new `--force` flag to `rebaser`.
```

I would rather frame this in terms of the lifecycle rebaser and leave pack UX for a pack RFC.

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+### Validations
+
+Buildpack base image metadata specified in `buildpack.toml`'s `targets` list are validated against the runtime and build-time base images.
+
+Runtime and build-time base image packages are no longer validated against each other.
+
+If an SBoM is available, `pack rebase` will fail if packages are removed from the new runtime base image.
```
ekcasey, line 130:
```suggestion
Platforms may choose to add additional validation. For example, if an SBoM is available, `pack rebase` may fail if packages are removed from the new runtime base image.
```
I think there are enough missing details here (b/c we have not finalized the merge BOM RFC) that I would rather leave this as a strong possibility that should be fleshed out by the pack team, rather than a requirement per this RFC.

E.g. I don't want implementation of this RFC to depend on a not-yet-approved RFC

ekcasey (approved):

text/0000-remove-stacks-mixins.md
hone: Is there even a `$ID` for windows is that like "Windows 10" and the version is the "build number"?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,162 @@
...
+- Architecture Variant (optional) (e.g., "v6", `$GOARM`), specified as `variant` in the base image `config`
+- Distribution (optional) (e.g., "ubuntu", `$ID`), specified as a label `io.buildpacks.distribution.name`
+- Version (optional) (e.g., "18.04", `$VERSION_ID`), specified as a label `io.buildpacks.distribution.version`
+
+OS, Architecture, and Architecture Variant must be valid identifiers as defined in the [OCI Image specification](https://github.com/opencontainers/image-spec/blob/main/config.md).
+For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
+The `os.version` field in an base image `config` may contain combined distribution and version information, but it is not used by the lifecycle.
+
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).
+Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
```
hone, line 47: What's the reasoning behind the limitation on OS/Arch/Variant?

text/0000-remove-stacks-mixins.md
hone: > These labels will be deprecated (but allowed) for the forseeable future.

@jkutner This makes me think `[[stacks]]` and those labels would go away at the same time.

text/0000-remove-stacks-mixins.md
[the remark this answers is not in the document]
hone: @jkutner @sclevine what the intention of the `CNB_TARGET_ID` changes are for?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,162 @@
...
+## Runtime Metadata
+
+To allow different runtime base images to be used, and to support cross-compilation in the future, buildpacks may need access to the runtime base image's target metadata.
+The following environment variables will be available to buildpacks directly in the build-time environment (not via `/platform/env`):
+- `CNB_TARGET_OS`
+- `CNB_TARGET_ARCH`
+- `CNB_TARGET_VARIANT`
+- `CNB_TARGET_DISTRO_NAME`
+- `CNB_TARGET_DISTRO_VERSION`
+- `CNB_TARGET_ID` (optional ID, if present on runtime base image `io.buildpacks.id` label)
```
hone, line 93: is this the "stack id"? I'm looking through the spec and can't actually find anything that is `io.buildpacks.id` except for [Stack ID](https://github.com/buildpacks/spec/blob/main/platform.md#stack-id)?

hone: I just talked with @sclevine and this is NOT "Stack ID".

This instead is a run image ID that isn't needed for a build image.

The use case Stephen explained to me:

> Just an optional target ID, so that buildpacks built on a builder know if the build will use a weird run image (like tiny), without needing a totally different builder.

He's going to add some more detail when he's back from vacation on Tues.

A question I have from this, since this a new concept is this just an unofficial contract b/t the buildpack and run image?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,162 @@
...
+
+Platforms may choose to add additional validation. For example, if an SBoM is available, `pack rebase` may fail if packages are removed from the new runtime base image.
+This check may be skipped by passing a new `--force` flag to `pack rebase`.
+However, this validation is not enforced by the specification.
+
+### Migration
+
+All existing labels and environment variables for stacks and mixins may be preserved on a base image until all users have migrated to the new format.
+These labels will be deprecated (but allowed) for the forseeable future.
+If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user.
```
hone, line 138: I think this could be more concrete (we only have 1 standardized stack afaik?).

```suggestion
If the newly-specified field values are missing, the lifecycle and pack may used existing, standardized stack IDs (i.e., `io.buildpacks.stacks.*`) to determine the values of the missing fields, as long as the lifecycle and pack provide a warning for the user. More concretely, when using a stack with `io.buildpacks.stacks.bionic` lifecycle will assume the following metadata is set if these fields are empty:

config.os = Linux
config.architecture = x86_64
io.buildpacks.distribution.name = "ubuntu"
io.buildpacks.distribution.version = "18.04"

Moving forward it's encouraged for buildpack authors to support both `[[stacks]]` and `[[targets]]` sections in `buildpack.toml` for maximum compatibility. In order to ease this process for those using the `io.buildpacks.stacks.bionic`, lifecycle will translate any section that sets this as on of the `stacks`:

[[stacks]]
id = "io.buildpacks.stacks.bionic

to 

[[targets]]
os = "linux"
arch = "x86_64"
[[targets.distributions]]
name = "ubuntu"
versions = ["18.04"]
```

I think the bar for creating base images is slightly higher, should we add stack image templates in `pack` for creating build/run images where these labels and things get set?

hone (commented):
I'd like to propose closing this FCP by a week or two after reviewing it in more depth. I've talked with @sclevinebefore he left for vacation and he said he'd make some changes when he gets back next Tuesday. For me I'd like to see the following issues resolved:

* Resolving @samj1912 comments on Windows.
* more clarity on `$CNB_TARGET_ID`'s purpose since it's a net new concept
* more concrete plans for helping both buildpack authors and stack creators. This is a major breaking change and we should do as much as we can to ease this for everyone. I took a pass on expanding on Stephen's suggestion and would like to see some discussion around it.

text/0000-remove-stacks-mixins.md
sclevine: Clarified in RFC

sclevine:
@hone addressed in RFC, let me know if you have more questions/concerns

natalieparellano (approved):

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+[motivation]: #motivation
+
+The "stack" and "mixin" concepts add unnecessary complexity to the project and make it difficult for new users and contributors to understand how buildpacks work. Compatibility guarantees that are strongly enforced by the stack contract could be replaced with metadata validations and warnings.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Summary of changes:
+- Remove mixins
+- Replace stack metadata (including stack IDs) with canonical OS metadata.
```
jromero, line 32: /queue-issue buildpacks/spec "Remove stacks from platform spec"
/queue-issue buildpacks/samples "Remove stacks and replace them with base images"
/queue-issue buildpacks/docs "Remove stacks and replace them with base images"

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+[how-it-works]: #how-it-works
+
+## Base Image Metadata
+
+Instead of a stack ID, runtime and build-time base images must contain the following canonicalized metadata:
+- OS (e.g., "linux", `$GOOS`), specified as `os` in the base image `config` 
+- Architecture (e.g., "arm", `$GOARCH`), specified as `architecture` in the base image `config`
+- Architecture Variant (optional) (e.g., "v6", `$GOARM`), specified as `variant` in the base image `config`
+- Distribution (optional) (e.g., "ubuntu", `$ID`), specified as a label `io.buildpacks.distribution.name`
+- Version (optional) (e.g., "18.04", `$VERSION_ID`), specified as a label `io.buildpacks.distribution.version`
```
jromero, line 44: /queue-issue buildpacks/spec "Define base image in distribution spec"

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+The `os.version` field in an base image `config` may contain combined distribution and version information, but it is not used by the lifecycle.
+
+For Windows-based images, Distribution should be empty. Version should be the [suggested value of `os.version`](https://github.com/opencontainers/image-spec/blob/main/config.md#properties) in the OCI spec (e.g., `10.0.14393.1066`).
+
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).
+Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
+If the `targets` list is empty and `/bin/build` is present, a target with `os = "linux"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+If the `targets` list is empty and `/bin/build.bat` or `/bin/build.exe` is present, a target with `os = "windows"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
```
jromero, line 62: /queue-issue buildpacks/spec "Replace stacks with targets in buildpack api"

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+      "purl": "pkg:deb/ubuntu/libcurl@4.5.6"
+    },
+    ...
+  ]
+}
+```
+
+### Validations
+
+Buildpack base image metadata specified in `buildpack.toml`'s `targets` list are validated against the runtime and build-time base images.
```
jromero, line 138: /queue-issue buildpacks/pack "validate buildpacks' targets to based images"

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+- Architecture Variant (optional) (e.g., "v6", `$GOARM`), specified as `variant` in the base image `config`
+- Distribution (optional) (e.g., "ubuntu", `$ID`), specified as a label `io.buildpacks.distribution.name`
+- Version (optional) (e.g., "18.04", `$VERSION_ID`), specified as a label `io.buildpacks.distribution.version`
+
+Additionally, the runtime base may contain the following metadata:
+- Target ID (optional) (e.g., "minimal"), specified as a label `io.buildpacks.id`
+
+OS, Architecture, and Architecture Variant must be valid identifiers as defined in the [OCI Image specification](https://github.com/opencontainers/image-spec/blob/main/config.md).
+
+Target ID is an identifier specified on the runtime base image that must be provided to buildpacks as `CNB_TARGET_ID` during the build process.
```
natalieparellano, line 51: /queue-issue buildpacks/lifecycle "Analyzer should read io.buildpacks.id from run image" epic/dockerfiles type/enhancement
/queue-issue buildpacks/lifecycle "Builder should provide CNB_TARGET_ID in buildpack env" epic/dockerfiles type/enhancement

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+This allows buildpacks to change their behavior if a run image is selected (e.g., distroless) that has special properties outside of OS, architecture, etc.
+
+For Linux-based images, each field should be canonicalized against values specified in `/etc/os-release` (`$ID` and `$VERSION_ID`).
+The `os.version` field in an base image `config` may contain combined distribution and version information, but it is not used by the lifecycle.
+
+For Windows-based images, Distribution should be empty. Version should be the [suggested value of `os.version`](https://github.com/opencontainers/image-spec/blob/main/config.md#properties) in the OCI spec (e.g., `10.0.14393.1066`).
+
+The `stacks` list in `buildpack.toml` is replaced by a `targets` list, where each entry corresponds to a different buildpack image that is exported into a [manifest index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).
```
natalieparellano, line 59: > that is exported into a manifest index

Are we planning to export multiple images? Or is this a list of "potential" targets?

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,200 @@
...
+Each entry may contain multiple valid values for Distribution and/or Version, but only a single OS, Architecture, and Variant.
+If the `targets` list is empty and `/bin/build` is present, a target with `os = "linux"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+If the `targets` list is empty and `/bin/build.bat` or `/bin/build.exe` is present, a target with `os = "windows"` and `arch = "x86_64"` is assumed by tools reading `buildpack.toml`.
+
+App image builds fail if the build image and selected run image have mismatched metadata. We may introduce flags or additional labels to skip this validation (e.g., for cross-compilation or minimal runtime base images).
+An image without a specified Distribution is compatible with images specifying any Distribution.
+An image specifying a Distribution without a Version is compatible with images specifying any Versions of that Distribution.
+
+When an app image is rebased, `rebaser` must fail if the new run image and previous run image have mismatched metadata. This check can be skipped for Distribution and Version by passing a new `--force` flag to `rebaser`.
+
```
natalieparellano, line 69: /queue-issue buildpacks/lifecycle "Rebaser should validate new run image metadata against old run image metadata" epic/dockerfiles type/enhancement

hone (approved):

samj1912 (approved):

text/0000-remove-stacks-mixins.md
natalieparellano: Relaying discussion - the `targets` in `buildpack.toml` are for the purpose of creating buildpackages.

text/0000-remove-stacks-mixins.md
dmikusa-pivotal: Looks OK to me. I like the `CNB_TARGET_*` env variables, as opposed to needing to fish that out of the image manually.

dmikusa-pivotal (approved):

hone:
Is there anything else outstanding to be resolved on this?

text/0000-remove-stacks-mixins.md
natalieparellano: So I believe the answer to my question is "no, not yet". Marking this as resolved...

hone:
@sclevine can you add to unresolved so I can merge this:

> Are we planning to export multiple images? Or is this a list of "potential" targets?

hone:
@ekcasey kindly requests when this gets merged, @buildpacks/platform-maintainers starts with a sub-team RFC to discuss the UX around removing stacks, etc.

sclevine:
@hone added, ready to merge

hone:
I think the bot may be broken. :( @buildpacks/implementation-maintainers @buildpacks/platform-maintainers are there any issues that need to be opened before I merge this in?

text/0000-remove-stacks-mixins.md
jromero: @natalieparellano The labels are expected to be in the following format `[epic/dockerfiles][type/enhancement]`.

jromero:
> I think the bot may be broken. :( @buildpacks/implementation-maintainers @buildpacks/platform-maintainers are there any issues that need to be opened before I merge this in?

It's not broken. The issues only get created when this gets merged via [`merge-rfc.sh`](https://github.com/buildpacks/rfcs#automation)

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,202 @@
...
+variant = "v6"
+[[targets.distributions]]
+name = "ubuntu"
+versions = ["14.04", "16.04"]
+```
+
+## Runtime Metadata
+
+To allow different runtime base images to be used, and to support cross-compilation in the future, buildpacks may need access to the runtime base image's target metadata.
+The following environment variables will be available to buildpacks directly in the build-time environment (not via `/platform/env`):
```
jromero, line 99: This part is a little unclear to me. This appears to be talking about a set of environment variables that are given to the buildpacks during **build-time**. 

1. It's not talking about these environment variable being set on the run image right?
2. What determines what these environment variables should be set to?
3. What does happen if the platform provides these environment variables via `/platform/env` as a way to configure the desired outcome? Is it explicitly something we don't want to allow?

text/0000-remove-stacks-mixins.md
jromero: /queue-issue buildpacks/pack "Add --force to rebase operation"

text/0000-remove-stacks-mixins.md
sclevine: 1. Correct: "directly in the build-time environment"
2. "runtime base image's target metadata" -- same metadata as above, but for the runtime base image
3. All `CNB_` env vars are set directly so that they are unaffected by "strict" mode that forces buildpacks to read the env vars from `/platform/env`

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,202 @@
...
+
+- Keep stacks.
+- Make the above changes but keep some existing terminology: stacks, build-image, run-image.
+- Continue to allow buildpacks to specify package requirements (e.g., by PURL instead of mixins)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How will migration work? Can we make new base images compatible with older buildpacks? Can we make newer buildpacks compatible with older stacks?
+- What should builder.toml (and similar stack-dependent config files) look like? What should assets look like? Note: these could be decided in subsequent subteam RFCs / spec PRs.
```
jromero, line 194: /queue-issue buildpacks/pack "Remove 'stacks' from builder.toml"

text/0000-remove-stacks-mixins.md
```diff
@@ -0,0 +1,202 @@
...
+- How will this affect the UX for pack buidpack package? For example package.toml contains a conflicting [platform table](https://buildpacks.io/docs/reference/config/package-config/#platform-_optional_). This should be resolved in a separate subteam RFC.
```
jromero, line 195: /queue-issue buildpacks/pack "Update package commands with targets in mind"

edmorley:
The current buildpack API spec includes `CNB_STACK_ID` being set in the environment of buildpack detect/build:
https://github.com/buildpacks/spec/blob/buildpack/v0.8/buildpack.md#provided-by-the-platform

That env var is commonly used by buildpacks when determining which binary to download/install for those that are compiled for specific OSes/OS versions.

This RFC doesn't mention what happens to that env var. Is it now considered deprecated too? Or are only the `buildpack.toml` `stacks` parts being deprecated?

If it is being deprecated, could the RFC be updated to clarify this + explain what buildpacks should now use instead? (I would presume they now need to construct the stack name/version/... from the new `CNB_TARGET_*` env vars?)

text/0000-remove-stacks-mixins.md
joe-kimmel-vmw: just updating this thread in case someone comes along in two more years, but anyway it turned out while implementing we did make an explicit extension for "*" OS/Architecture.

PR #175, opened 2021-07-01
RFC: Officially supported utility buildpacks

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/jkutner/official-buildpacks/text/0000-official-utility-buildpacks.md)


text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+- `cnb/profile` - Evaluating `.profile` scripts in the app repo during launch
+- `cnb/inline` - Executing [Inline buildpacks](https://github.com/buildpacks/rfcs/blob/main/text/0048-inline-buildpack.md)
+- `cnb/processes` - Setting app specific process types from a configuration file like `project.toml`
+- `cnb/env` - Setting customer environment variables defined in a configuration file like `project.toml`
+- `cnb/purge` - Purging files from the app repo before export (similar to [`.slugignore`](https://devcenter.heroku.com/articles/slug-compiler#ignoring-files-with-slugignore))
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
```
samj1912, line 47: Should this be the BAT team?

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+- New components to support
+- We have to ensure that each utility buildpack is portable and works in conjunction with all stacks
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Do Nothing - leave the capabilities in the lifecycle, and continue to add more native capabilities
+
+# Prior Art
+[prior-art]: #prior-art
```
samj1912, line 62: Some other relevant buildpacks https://github.com/samj1912/project-descriptor-buildpack

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal to adopt utility buildpacks as officially supported components of the Cloud Native Buildpacks project.
+
+# Definitions
+[definitions]: #definitions
+
```
samj1912, line 18: We have also talked about project descriptor related parts that can be implemented as buildpacks should be owned by cnb as that would ensure that the project only provides implementation for different parts of the spec. I think that would be a good boundary for the kind of utility buildpacks that the project maintains i.e. it should only provide implementations for things that are spec-ed out.

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+However, defering to the community and vendors to implement these kinds of utility buildpacks has not necessarily had good results. For example, there are three implementations of the [Procfile buildpack](https://registry.buildpacks.io/buildpacks/paketo-buildpacks/procfile). None of these implementations add significant (if any) advantages over the other. Yet buildpack users are force to decide which one to select. This is an unneccessary burden for a capability that could be provided by default.
+
+We seek to reduce complexity in the lifecycle and provide well maintained utility buildpacks for users by supporting them as official components of the Cloud Native Buildpacks project. This also aligns with our "everything's a buildpack" mantra.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Reserve the `cnb` namespace for officially supported buildpacks. Then, on a case by case basis, we adopt utility buildpacks under this
```
aemengo, line 34: I'm a little worried that this opens a bit of a pandora's box.

We might have a clear distinction of utility buildpacks and _other buildpacks_. But newcomers might likely just be seeing **"officially supported buildpacks"**. Once we start doing this, I'm worried that we'll be pressured to have more of these officially supported buildpacks. If it were me, I'd associate the `cnb/` namespace with the `library/` namespace on DockerHub (_which is what I think cnb namespace should be called, btw_).

I'm sure that newcomers are a bit thrown when they find that they have to go to the [paketo](https://paketo.io/) project to find all of their buildpacks. Which leads me to worry that we will be pit in a (_losing_) competition with the paketo project. If it's going be ours vs theirs, we'll have a hard time covering the ground that they already have.

And finally, "owning, supporting, and maintaining" and "testing buildpacks across stacks" _seems like_ a lot of responsibilities on top of what Implementation team probably already has to do.

But I like the idea 🙂

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+
+Examples of potential utility buildpacks include:
+
+- `cnb/profile` - Evaluating `.profile` scripts in the app repo during launch
+- `cnb/inline` - Executing [Inline buildpacks](https://github.com/buildpacks/rfcs/blob/main/text/0048-inline-buildpack.md)
```
natalieparellano, line 39: I'm a bit confused how this would work - it seems that the inline script would need to be copied to the proper directory under `/cnb/buildpacks` ?? That doesn't feel like a buildpack concern...

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Bubbling these capabilities up to the user layer (as buildpacks) may add cognitive burden for users and errode the DX. We can mitigate this with new features like "default buildpacks".
```
natalieparellano, line 52: Should these buildpacks ship with the builder, or the lifecycle?

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+- New components to support
+- We have to ensure that each utility buildpack is portable and works in conjunction with all stacks
```
natalieparellano, line 54: Also all buildpack APIs...

dmikusa-pivotal:
A few concerns/thoughts come to mind about this:

1. I'm not sure that it will be less complex/confusing to have an official set of utility buildpacks. First, utility buildpacks by themselves don't do anything. I think that could create a point of confusion for users. They will see "official" and assume that is what they should be using, but the "official" set of buildpacks would not be complete and users would still need to find their way to one of the communities producing buildpacks to get a complete working set.

   Plus, I'm sure everyone is familiar with [this](https://xkcd.com/927/), so what's to say this attempt at making an "official" set of utility buildpacks doesn't just end up with the creation of a fourth Procfile buildpack?

    My $0.02 is that the Buildpacks project should be **strictly spec + tools + libraries + samples**, and no actual, production buildpacks. I think one of the smartest things that Kubernetes did to encourage a strong ecosystem (and achieve world domination) was to leave space for others to come in and fill in the gaps. I worry that by providing official buildpacks, you're making that space for others to come in and participate smaller (who's going to try and compete against "official" buildpacks?), which I think is a net negative for the community. 

    Sure this might result in having some duplicate work (how many ingress controllers are there for K8s?), and that results in users having to make some choices, but I think in the end it results in a stronger ecosystem. Again, just my $0.02, take it for nothing more than that.

2. I worry that taking three versions of already established buildpacks, even ones doing roughly the same thing, and turning out one "official" buildpack will not be as easy as it sounds and possibly full of bikeshedding. 

    For example, I know Paketo has conventions about how its buildpacks operate to define a consistent user experience across them. If Paketo were to use an upstream buildpack, it would need to still follow those conventions. What if someone else has different conventions? Do we now have two ways to do the same thing in the buildpack? Is that any more clear than having two different buildpacks?

3. `Many capabilities of lifecycle and pack do not require a native implementation within these components`. I think that's great. I'm all for keeping things manageable and not pushing too many features into the lower layers. I think it's quite feasible for buildpacks or libraries used to build buildpacks to implement the things you listed. 

    Again, just my $0.02, but I think that the Buildpacks projects should declare these items as off-limits or things they won't implement and call it a day (It's OK to say no to features if it fits the desired vision/architecture). By doing this, you're intentionally leaving space for the community to fill in these needs. I believe that's a good thing and it will promote a healthy ecosystem.

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: +1 to the Pandora's box comment. I think that having some "official" buildpacks will absolutely prompt people to request more. So where do you draw the line? Technically you only need one buildpack that provides a language runtime binary, should that become an official buildpack too?

In regards to newcomers, I think that can be mitigated with good docs & collaboration with the overall ecosystem. The Buildpacks projects can and should direct users to community resources (in a fair, taking no favorites way). Inform users that there are choices and allow them to pick what makes the most sense for their needs.

text/0000-official-utility-buildpacks.md
jkutner: I have no strong opinion

text/0000-official-utility-buildpacks.md
jkutner: These are a very real concerns, and they are the reasons we never included buildpacks in the project in the past.

What makes the buildpacks in this proposal a little different is that they _replace_ native capabilities. In some sense, the support burden is equal to the existing support burden.

I'm not worried about being pitted against Paketo. They should be a different set of buildpacks. If Paketo has utility buildpacks that need to be contributed upstream we can consider accepting them.

text/0000-official-utility-buildpacks.md
jkutner: I'm not sure I follow. You can see an implementation of this works in https://github.com/hone/inline-buildpack

text/0000-official-utility-buildpacks.md
jkutner: I think it would interesting to ship them with the lifecycle. But I had Builder in mind.

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: >What makes the buildpacks in this proposal a little different is that they replace native capabilities. In some sense, the support burden is equal to the existing support burden.

Isn't the idea to reduce burden though? If the burden is equal, then doesn't that make a case for implementing these features in the lifecycle/pack?

IMHO, you can reduce the burden by making it clear that those features won't be part of the buildpack and allowing the community time to fill in the features through buildpacks (see my other comment below).

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Do Nothing - leave the capabilities in the lifecycle, and continue to add more native capabilities
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [inline-buildpack](https://registry.buildpacks.io/buildpacks/hone/inline)
+- [Procfile buildpack](https://registry.buildpacks.io/buildpacks/paketo-buildpacks/procfile)
```
ekcasey, line 65:
```suggestion
- [Procfile buildpack](https://registry.buildpacks.io/buildpacks/paketo-buildpacks/procfile)
- [Paketo Image Labels Buildpack](https://github.com/paketo-buildpacks/image-labels)
- [Paketo Environment Variables Buildpack)(https://github.com/paketo-buildpacks/environment-variables)
```

ekcasey (approved):

text/0000-official-utility-buildpacks.md
hone: If these are things that extensions of the spec/lifecycle I can see the implementation team owning them. Thoughts on this @ekcasey @natalieparellano ?

text/0000-official-utility-buildpacks.md
hone: On the plus side this means we'll have to dogfood our breaking changes/migrations as a project.

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,75 @@
...
+We seek to reduce complexity in the lifecycle and provide well maintained utility buildpacks for users by supporting them as official components of the Cloud Native Buildpacks project. This also aligns with our "everything's a buildpack" mantra.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Reserve the `cnb` namespace for officially supported buildpacks. Then, on a case by case basis, we adopt utility buildpacks under this
+
+Examples of potential utility buildpacks include:
+
+- `cnb/profile` - Evaluating `.profile` scripts in the app repo during launch
```
hone, line 38: This isn't meant to replace the `.profile.d` scripts proposed to be removed by #168, but just in the `app` dir?

sclevine (approved):

text/0000-official-utility-buildpacks.md
ekcasey: I think it is okay for the utility buildpacks to implement a single buildpack API?

text/0000-official-utility-buildpacks.md
jabrown85: I would vote not to have these maintained by the implementation team to be honest. I think BAT makes most sense.

text/0000-official-utility-buildpacks.md
jabrown85: I like shipping with a builder

text/0000-official-utility-buildpacks.md
hone: @jabrown85 it just seems somewhat of an awkward fit since these buildpacks are more for platform operators b/c we're pulling out functionality from lifecycle afaict.

text/0000-official-utility-buildpacks.md
hone: > Should these buildpacks ship with the builder, or the lifecycle?

I think it depends on how we're positioning them. If we want them to be default things that are taken from lifecycle, I think it makes sense to ship with lifecycle. That of course is totally net new and doesn't fit into an existing mechanism. If builders have to build them in, we're essentially saying by default they're off.

hone:
The biggest unanswered question for me is the guidelines / criteria of what makes a utility buildpack. I like @samj1912's suggestion of utility buildpacks must be in the spec including extensions. This way we can limit scope initially. We can always open it up more later.

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: Paketo has several utilities style buildpacks, see here: https://github.com/paketo-buildpacks?q=utilities

I can't speak for the team, but if this proposal goes through I would be up for helping to upstream the utilities Paketo CNBs that make sense.

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: IMHO, you have to look more at the skills/interests of the team to be maintaining these buildpacks. BAT seems to make the most sense in that light given the BAT team has folks already producing buildpacks and those working on libcnb and trying to make the buildpack development experience as smooth as possible. To me their interests/skills line up very well with this task.

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: +1 for builders. It doesn't require any new features or work, plus it provides the most flexibility. The idea of "default" behavior is nice in theory, but downstream buildpack teams need a way to control what buildpacks are run. 

Speaking from the Paketo and VMware proprietary buildpacks standpoint, I need to be able to control the user experience, which means taking out upstream buildpacks that are not needed/wanted or that conflict with conventions we use in our buildpacks. If they are in the lifecycle that would seem to be harder or we'd end up needing to have controls similar to what you get with builders, so why not just use builders.

dmikusa-pivotal:
> The biggest unanswered question for me is the guidelines / criteria of what makes a utility buildpack. 
 +1

>I like @samj1912's suggestion of utility buildpacks must be in the spec including extensions. This way we can limit scope initially. 

I feel like even this needs to be flushed out further. Procfile was mentioned as an example in this RFC, but Procfile isn't part of our spec. Would so would that qualify? 

Beyond using these buildpacks to implement spec-related things that we don't want in the lifecycle, there seems to be the use case of "we have X almost duplicate buildpacks that do Y in the community". Should these be included so we can reduce duplication of work and confusion of users? I feel like the answer is yes, but it makes it difficult to draw a line as to what should be upstream vs not.

My suggestion would be for two things:

1. For the RFC to include a list of criteria for acceptance as an upstream buildpack. It seems like we have at least two criteria: a.) functionality that would otherwise be in the lifecycle or b.) a buildpack with at least X functionally equivalent implementations (where X is maybe 3?). Plus I would assume majority approval on an RFC to add any new buildpacks.

2. This RFC should include a list of the initial list of buildpacks to be accepted. That way it's clear what is being accepted in advance and there's a clear starting point for what's going to be upstreamed. We can certainly add more later, but it'll take time & resources to get the initial set going so this can help to focus on what is most important for the initial set.

text/0000-official-utility-buildpacks.md
jkutner: Our goal is to reduce burden by changing the architecture (i.e. moving logic out of the lifecycle, and reducing coupling). I do not have a goal of removing capabilities/features from the project.

text/0000-official-utility-buildpacks.md
jkutner: @dmikusa-pivotal 

> I need to be able to control the user experience, which means taking out upstream buildpacks that are not needed/wanted

Is the builder a good way to do this? That is, if you can create a builder w/o the default buildpacks does that solve the problem. I think that's what you're saying, but wanted to be clear.

If we ship these with the lifecycle, I still imagine there'd be a mechanism to turn them off.

text/0000-official-utility-buildpacks.md
jkutner: yes

jkutner:
@dmikusa-pivotal 

> Procfile was mentioned as an example in this RFC, but Procfile isn't part of our spec. So would that qualify?

I think we would replace Procfile with a spec for process types in the `project.toml`, and thus use some project-descriptor buildpack.

> For the RFC to include a list of criteria for acceptance as an upstream buildpack. It seems like we have at least two criteria: a.) functionality that would otherwise be in the lifecycle or b.) a buildpack with at least X functionally equivalent implementations (where X is maybe 3?). Plus I would assume majority approval on an RFC to add any new buildpacks. We could also include a list of criteria that would exclude a buildpack from being upstreamed, such as buildpacks that are specific to a language or that provide server/language dependencies. Excluding things would help further define a line of what is acceptable.

Yea, I'll try to work some of this in. It might be difficult to define exclusion requirements, but we can always loosen them later if need be.

> This RFC should include a list of the initial list of buildpacks to be accepted.

If we do this I would want to limit it to only those buildpacks that replace functionality removed by https://github.com/buildpacks/rfcs/pull/168. We could explore process-types, inline, etc later on.

text/0000-official-utility-buildpacks.md
natalieparellano: Ohh I see (thank you for linking to the source!). The "inline wrapper" buildpack when run looks in project.toml and runs the script there. I was confusing it with pack's implementation, where the script is copied (I think?).

text/0000-official-utility-buildpacks.md
natalieparellano: Is someone who wants to use utility buildpacks going to have to run `pack build <thing> --buildpack add-my-env-var --buildpack run-my-inline-buildpack --buildpack set-my-process` etc.? That seems cumbersome (vs. just configuring project.toml).

text/0000-official-utility-buildpacks.md
natalieparellano: Would we have multiple versions of each utility buildpack (one per buildpack API)?

text/0000-official-utility-buildpacks.md
dmikusa-pivotal: > I do not have a goal of removing capabilities/features from the project.

Not suggesting that specifically. You mentioned Procfile in the RFC, so that is what I'm thinking of here. There's no need for buildpacks to directly have a Procfile buildpack because it was a need identified by the downstream communities and they have provided support for it.

All I'm saying is that if there's a clear delineation between what the buildpacks will and won't do, that provides guides and leaves room for the community to come in and implement any features that are marked out of scope but are important to the downstream communities.

text/0000-official-utility-buildpacks.md
jkutner: @natalieparellano that's definitely not what I have in mind, but this RFC doesn't propose a mechanism for "default" buildpacks (i.e. buildpacks that would run without the user specifying those verbose args). I intend to create another RFC for that.

text/0000-official-utility-buildpacks.md
jkutner: I think it would make sense to bump the utility buildpacks API compat and cut a release each time the lifecycle adds a new api version. I also think that makes a good case for Implementation team owning these buildpacks.

natalieparellano: I'll be honest, I'm a bit wary of the implementation team owning these (we'd be using Buildpack Author Tooling to maintain them...) but at the same time, I could see the argument for it. We could use the utility buildpacks as test fixtures when adding support for a new api. They should ship together (or at least, concurrently). That said, coordination between the two teams to enable this should be possible!

text/0000-official-utility-buildpacks.md
```diff
@@ -46,6 +46,12 @@ Examples of potential utility buildpacks include:
 
 The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
 
+The criteria for accepting a utility buildpack into the project is as follows:
+- The behavior provided by the buildpack MUST be defined in the spec or an extension spec.
+- The buildpack MUST support arbitrary stacks
+- The buildpack MUST support the latest API version
+- The buildpack MUST include detection logic (i.e. it does not always pass)
```
dmikusa-pivotal, line 53: 👍 

Question regarding the last item. It seems a little strict. Can you elaborate on the rationale for this? Is this requirement to only allow buildpacks that can be disabled in some way?

I have seen a number of buildpacks that follow the pattern of always pass, but restrict behavior in other ways (based on bindings, env variables, or buildplan entries). 

Would it be sufficient to say `The buildpack MUST include a way to be disabled` instead?

text/0000-official-utility-buildpacks.md
ekcasey: > I think it would make sense to bump the utility buildpacks API compat and cut a release each time the lifecycle adds a new api version. I also think that makes a good case for Implementation team owning these buildpacks.

By the same logic the implementation team should maintain `pack` as well, given `pack` should also upgrade to use the latest platform API after the lifecycle cuts a new release. Maybe we need to rename the implementation team because 98% of the code in the project is an implementation of at least one of the spec :p

text/0000-official-utility-buildpacks.md
```diff
@@ -46,6 +46,12 @@ Examples of potential utility buildpacks include:
 
 The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
 
+The criteria for accepting a utility buildpack into the project is as follows:
+- The behavior provided by the buildpack MUST be defined in the spec or an extension spec.
+- The buildpack MUST support arbitrary stacks
+- The buildpack MUST support the latest API version
```
hone, line 52: Do we care about it support multiple APIs or only one?

text/0000-official-utility-buildpacks.md
hone: yeah, I'm curious about the thought process behind this one.

nebhale (approved):

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,83 @@
...
+- `cnb/purge` - Purging files from the app repo before export (similar to [`.slugignore`](https://devcenter.heroku.com/articles/slug-compiler#ignoring-files-with-slugignore))
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- The behavior provided by the buildpack MUST be defined in the spec or an extension spec.
+- The buildpack MUST support arbitrary stacks
```
samj1912, line 51: Would this involve both linux and windows stacks? Would this also mean all the buildpacks would have to be compiled for all architectures?

text/0000-official-utility-buildpacks.md
ekcasey: @jkutner Can we update this to give ownership to the BAT team. It seems like the BAT team wants to maintain these and the Implementation team doesn't, which seems like the most important consideration to me.

text/0000-official-utility-buildpacks.md
ekcasey: If we want the `.profile` buildpack to be the first utility buildpack, then I think this requirement is impossible. A `.profile` buildpack will never work on a stack that doesn't include Bash.

text/0000-official-utility-buildpacks.md
ekcasey: I don't think we should support multiple APIs. We should support the latest and folks can use older versions if they require an older APIs, or update the lifecycle :)

text/0000-official-utility-buildpacks.md
hone: For the record I'd like to voice it's a bit out of scope for the BAT team especially as we continue to focus on buildpack tooling which I believe this is out of scope. The overlap is that it'd be nice to dogfood buildpacks with our own tooling, but feels a bit of a miss on the charter. I'm not sure there's a perfect subteam for it.

text/0000-official-utility-buildpacks.md
hone: Are we going to bugfix or maintain multiple versions then for older Buildpack APIs?

text/0000-official-utility-buildpacks.md
hone: @ekcasey i think we should rename the team. :)

> They should ship together (or at least, concurrently). That said, coordination between the two teams to enable this should be possible!

@natalieparellano I feel like some joint maintenance makes sense but there isn't really that structure today in the project.

text/0000-official-utility-buildpacks.md
jkutner: I think we may need to let #172 finalize in order to address this. If `[[targets]]` and `os` under that becomes a thing, this will be much easier to deal with.

text/0000-official-utility-buildpacks.md
jkutner: Maybe we should have the same policy as we would for libcnb

text/0000-official-utility-buildpacks.md
jkutner: I didn't pull this from a specific example. But I suspect that something that _always_ passes detection is something that we might want to have as a native behavior (not a buildpack). `The buildpack MUST include a way to be disabled` seems like a good alternative though.

jkutner:
I think we need to block this on #172, as that will determine how we deal with utility buildpacks like `.profile` that can't support `*` stacks.

text/0000-official-utility-buildpacks.md
ForestEckhardt: Throwing my hat into the ring as well https://github.com/ForestEckhardt/source-removal

text/0000-official-utility-buildpacks.md
hone: If we're calling them system buildpacks maybe we can use `cnb-system/` as the namespace?

text/0000-official-utility-buildpacks.md
hone: What would that change L51 if that goes through?

jkutner: It depends on what we decide with #172. If wildcard targets are allowed, then these buildpacks probably use that. Otherwise, they'll have to specific all the targets they support explicitly.

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,83 @@
...
+- `cnb/env` - Setting customer environment variables defined in a configuration file like `project.toml`
+- `cnb/purge` - Purging files from the app repo before export (similar to [`.slugignore`](https://devcenter.heroku.com/articles/slug-compiler#ignoring-files-with-slugignore))
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- The behavior provided by the buildpack MUST be defined in the spec or an extension spec.
```
jkutner, line 50: In the working group we discussed having this target an "API" rather than a spec, but given that the current spec is not decomposed that way, I don't think it makes sense to change the wording this proposal. If we refactor the spec, I think we can change the criteria for accepting utility buildpacks. cc @sclevine @ekcasey

hone:
@jkutner will update to remove spec restriction, but every new buildpack will require a project level RFC to keep the bar high + very clear documentation for these buildpacks.

jkutner:
The criteria for accepting a utility buildpack into the project has been updated. I think there's still a discussion to be had about the namespace used though. Currently, `cnb` is proposed, but we should also consider:

* `buildpacksio`
* `library`

I'm very comfortable changing this to `buildpacksio`.

hone:
Since @ekcasey is on leave, I'm unassigning her from this RFC.

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,87 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Implementation team](https://github.com/buildpacks/community/blob/main/TEAMS.md#implementation-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- The behavior provided by the buildpack MUST be defined in an API specification co-located with the buildpack
+- The buildpack MUST support arbitrary stacks
+- The buildpack MUST support the latest API version
+- The buildpack MUST include detection logic (i.e. it does not always pass)
```
hone, line 53: Trying to roll up this comment thread.

```suggestion
- The buildpack MUST include detection logic so it can be disabled.
```

jkutner:
@hone @samj1912 please re-review. I've added the RFC requirement, and changed default ownership to the BAT team in 0e1c749

hone (approved):

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,88 @@
...
+
+The utility buildpacks will be owned, supported, and maintained by the [Buildpack Authors' Tooling Team](https://github.com/buildpacks/community/blob/main/TEAMS.md#buildpack-authors-tooling-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- An [RFC](https://github.com/buildpacks/rfcs) describing the buildpack MUST be approved.
+- The behavior provided by the buildpack MUST be defined in an API specification
+- The buildpack MUST support arbitrary stacks
+- The buildpack MUST support the latest API version
+- The buildpack MUST include detection logic so it can be disabled.
+- The buildpack MUST NOT install any dependencies
```
samj1912, line 55:
```suggestion
- The buildpack MUST NOT install any dependencies in the output image
```

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,88 @@
...
+- The buildapck MUST NOT be specific to a language/framework/ecosystem/toolchain
+- The buildpack behavior MUST be well-documented on buildpacks.io
+- The buidpack SHOULD only provide functionality that we would added to `pack` or the platform API absent the existence of utility buildpacks
```
samj1912, line 58:
```suggestion
- The buidpack SHOULD only provide functionality that we would have added to `pack` or the platform API absent the existence of utility buildpacks
```

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,88 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Buildpack Authors' Tooling Team](https://github.com/buildpacks/community/blob/main/TEAMS.md#buildpack-authors-tooling-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- An [RFC](https://github.com/buildpacks/rfcs) describing the buildpack MUST be approved.
+- The behavior provided by the buildpack MUST be defined in an API specification
+- The buildpack MUST support arbitrary stacks
+- The buildpack MUST support the latest API version
```
samj1912, line 53:
```suggestion
- The buildpack SHOULD support the latest API version
```

text/0000-official-utility-buildpacks.md
```diff
@@ -0,0 +1,88 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The utility buildpacks will be owned, supported, and maintained by the [Buildpack Authors' Tooling Team](https://github.com/buildpacks/community/blob/main/TEAMS.md#buildpack-authors-tooling-team).
+
+The criteria for accepting a utility buildpack into the project is as follows:
+- An [RFC](https://github.com/buildpacks/rfcs) describing the buildpack MUST be approved.
+- The behavior provided by the buildpack MUST be defined in an API specification
+- The buildpack MUST support arbitrary stacks
```
samj1912, line 52:
```suggestion
- The buildpack SHOULD support arbitrary stacks
```

samj1912 (commented):
Left some minor suggestions. LMK what you think.

jkutner:
Final Comment Period closing 11/24/21

PR #179, opened 2021-07-29
RFC: System Buildpacks

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/jkutner/system-buildpacks/text/0000-system-buildpacks-in-builder.md)

This RFC supplements #175 and #168


text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,155 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+Unless otherwise stated, system buildpacks conform to the [buildpack API](https://github.com/buildpacks/spec/blob/main/buildpack.md).
+
+The `system.pre.buildpacks` will be transformed into a new table in `order.toml`. The `[system]` table in `order.toml` will be processed by the lifecycle, and each `pre`/`post` buildpack will run during the detect phase. Those that pass detection will run during the build phase.
+
+## Detection
+
+The results of detection by system buildpacks MUST NOT influence the selected buildpack group. If no system buildpacks pass detection, any buildpack group MAY pass detection. If a system buildpack pass detection and no buildpack groups pass detection, then detection MUST fail.
```
sclevine, line 56: > The results of detection by system buildpacks MUST NOT influence the selected buildpack group.

I'm not sure I understand this requirement. If system buildpacks can augment the build plan, they will also influence the selected buildpack group. 

> If no system buildpacks pass detection, any buildpack group MAY pass detection.

Similar issue to the above -- certain groups may require a system buildpack to include build plan entries.

> If a system buildpack pass detection and no buildpack groups pass detection, then detection MUST fail.

Is this just to prevent groups with all optional buildpacks from turning into groups that only contain system buildpacks?

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,155 @@
...
+
+System buildpacks may require/provide in the build plan following standard buildpack API specification.
+
+A new flag to the lifecycle `detector`, `--disable-system-buildpacks`, will disable system buildpacks
+
+## Build
+
+System buildpacks that have passed detection will be executed during the build phase. All `pre` buildpacks must execute before the detected buildpack group. All `post` buildpacks must execute after the detected buildpack group. However, the system buildpacks MUST NOT be added to the buildpack group, and their execution may be hidden from the end user.
```
sclevine, line 64: Why do we want to hide system buildpacks from the user? Shouldn't we be transparent about what features are activated?

> the system buildpacks MUST NOT be added to the buildpack group

Not sure I understand what this means.

text/0000-system-buildpacks-in-builder.md
jkutner: >>  The results of detection by system buildpacks MUST NOT influence the selected buildpack group.

> I'm not sure I understand this requirement. If system buildpacks can augment the build plan, they will also influence the selected buildpack group.

I really meant for this to be about exit code. With regard to the build plan, maybe the system buildpack should have either not have that, or have a separate build plan.

text/0000-system-buildpacks-in-builder.md
jkutner: > Why do we want to hide system buildpacks from the user? Shouldn't we be transparent about what features are activated?

I went back and forth on this. But the fact that they are buildpacks is a bit of an implementation detail. In verbose logging you would want to see this, but in the normal flow of things, people want to see the buildpacks they care about "node, java, etc". This may be a difference in the Heroku/Paketo stance too.

>> the system buildpacks MUST NOT be added to the buildpack group

> Not sure I understand what this means.

This is really about exit code again, and not wanting passing system buildpacks to allow groups with only optional buildpacks that fail to continue. I'll reword

text/0000-system-buildpacks-in-builder.md
ForestEckhardt: I think that I would like to add a +1 to @sclevine point. I don't understand why these wouldn't just be treated as buildpacks that you want to run before and after every other group. Maybe I am missing something bigger here but I feel like the concept to me was simpler if it is essentially just matrix expansion if you will. And with the ability to mark the buildpacks optional I think you still have a very wide range of flexibility.

jkutner: I'm hoping to avoid having our decision to decouple certain behaviors (like `.profile`) from the implementation of the lifecycle manifest itself in the user experience. 

I think buildpack _users_ shouldn't need to understand or care how things like "optional" or groups, or any other buildpack concepts work. They should run `pack build` (or whatever for their platform) and it just detects a python app, and builds it. Adding buildpacks like `.profile` or whatever into the build breaks the abstraction provided by `pack`. (also maybe this decision can be scoped to pack and not lifecycle).

Ultimately though, I think this comes down to _how much_ we decide to push into the system/utility buildpacks. If it's really just `.profile` in most cases, then I probably don't care as much. But I worry about a situation where it used to be as simple as "the `heroku/python` buildpack passed detect" to "why are there 8 buildpacks running?" Again I think that comes back to a difference b/w Heroku/Paketo buildpack design.

nebhale (approved):

text/0000-system-buildpacks-in-builder.md
hone: Does this mean the system buildpack can't write to the build plan and only participates with the exit code? I think that'd make sense in the `.profile` case.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,155 @@
...
+
+Unless otherwise stated, system buildpacks conform to the [buildpack API](https://github.com/buildpacks/spec/blob/main/buildpack.md).
+
+The `system.pre.buildpacks` will be transformed into a new table in `order.toml`. The `[system]` table in `order.toml` will be processed by the lifecycle, and each `pre`/`post` buildpack will run during the detect phase. Those that pass detection will run during the build phase.
+
+## Detection
+
+The results of detection by system buildpacks MUST NOT influence the selected buildpack group. If no system buildpacks pass detection, any buildpack group MAY pass detection. If a system buildpack pass detection and no buildpack groups pass detection, then detection MUST fail.
+
+System buildpacks may require/provide in the build plan following standard buildpack API specification.
```
hone, line 58: Do you have use cases for system buildpacks you're thinking of that would use the Build Plan?

hone:
@jkutner is it out of scope where the system buildpacks come from? They don't live on the registry do they?

text/0000-system-buildpacks-in-builder.md
jkutner: I have no specific examples/use-cases. I only wanted to default to the having them be the "same" as normal buildpacks

jkutner:
@hone i suppose it's out of scope, but I could see them on the registry.

hone:
From last WG, this is blocked on the resolution of #175

jkutner:
@hone @sclevine please review

sclevine (approved):

text/0000-system-buildpacks-in-builder.md
samj1912: I am not so sure on including system buildpacks in the buildplan if they are not interacting with the normal buildpacks in a group. Could we either limit the system buildpacks plan resolution to their respective groups? `pre` system buildpacks need to match up all their plan entries separately, normal buildpacks separately and `post` buildpacks separately.

Given that we want to discourage people from using system buildpacks as hidden buildpacks and given that these are mainly supposed to be utility buildpacks - would a simplifying assumption be to remove system buildpacks from contributing to the build plan entirely? And limit them to just detect `pass` or `fail`?

hone: Should we just have buildpacks not be part of the buildplan for now if we don't have a use case for it?

text/0000-system-buildpacks-in-builder.md
hone: Would system buildpacks be a part of https://github.com/buildpacks/spec/blob/main/platform.md#iobuildpacksbuildmetadata-json?

hone (commented):
It looks like there are two remaining issues that still need to be resolved before we can move this forward:
* system buildpacks' Build Plan - https://github.com/buildpacks/rfcs/pull/179/files#r686954199
* system buildpacks being hidden from user - https://github.com/buildpacks/rfcs/pull/179/files#r679613177

jkutner:
@hone please review the stance taken in the text:

> system buildpacks' Build Plan

system buildpacks use the build plan as normal buildpacks do

> system buildpacks being hidden from user

They "may" be hidden. I think we should leave this up to platforms.

hone (approved):

jkutner:
From working group:
* Enforce that system buildpacks are always optional. Remove `optional` from schema.
* spec changes will require a new <system-cnb>.toml or similar in the builder
* lifecycle will need a new flag `--system-buildpacks`

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+  version = "<buildpack version>"
+  optional = true
+
+[[system.post.buildpacks]]
+  id = "<buildpack ID>"
+  version = "<buildpack version>"
+  optional = true
+```
+
+The fields in the `system.pre.buildpacks` table and `system.post.buildpacks` table match the fields in the existing [`order.group` table](https://buildpacks.io/docs/reference/config/builder-config/#order-_list-required_). However, `optional` in this case is required and the only acceptable value is `true`. Non-optional buildpacks will cause the builder creation to fail.
```
ekcasey, line 43: > However, `optional` in this case is required and the only acceptable value is `true`. Non-optional buildpacks will cause the builder creation to fail.

I know I originally suggested this but I thought the outcome of the discussion in WG was that we do not want to enforce this? I remember being convinced at least.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+
+When a builder includes one or more `system.*.buildpacks` entry, each build phase that uses the builder will run the `pre` buildpacks before ALL other buildpacks (whether they are defined by a default `order.group` or if they are provided as an explicit order to the lifecycle) and the `post` buildpack after ALL other buildpacks.
```
ekcasey, line 45:
```suggestion
When a builder includes one or more `system.*.buildpacks` entry, the detect phase will prepend and append all `pre` and `post` buildpacks to each detection group in the provided order, respectively.
```
I am trying to make two pedantic clarifications here:

1. We can explicitly call out the detect phase here because once the results of detection are written to `group.toml` (presumably containing any system buildpacks that passed detection) other phases shouldn't need to be aware of the existence of system buildpacks.

2. There is no distinction between an `order.toml` order and a order "provided as an explicit order to the lifecycle". The only way to provide an order to the lifecycle is with an `order.toml` file.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Unless otherwise stated, system buildpacks conform to the [buildpack API](https://github.com/buildpacks/spec/blob/main/buildpack.md).
+
+The `system.pre.buildpacks` will be transformed into a new table in `order.toml`. The `[system]` table in `order.toml` will be processed by the lifecycle, and each `pre`/`post` buildpack will run during the detect phase. Those that pass detection will run during the build phase.
```
ekcasey, line 52: I think we agreed in WG that we needed a new file for these, so that the platform can specify the order with `order.toml` without affecting the system buildpacks.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Detection
+
+The exit code of detection by system buildpacks MUST NOT influence the selected buildpack group. If no system buildpacks pass detection, any buildpack group MAY pass detection. If a system buildpack passes detection and no buildpack groups pass detection, then detection MUST fail.
+
```
ekcasey, line 57:
```suggestion
```

As discussed I think we should avoid special behavior with regard to system vs regular buildpack detection. We should add them to each group but treat them identically after that.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+System buildpacks may require/provide in the build plan following standard buildpack API specification.
+
+A new flag to the lifecycle `detector`, `--disable-system-buildpacks`, will disable system buildpacks.
```
ekcasey, line 60: Instead of having a `--disable-system-buildpacks` flag, once system buildpacks are in a separate file could an empty `CNB_SYSTEM_PATH=` or `-system=""`or something like that indicate the same thing without requiring another flag. Therefore we wouldn't run into weird situations where the platform is both explicitly passing the path to a system buildpack file and also disabling those same buildpacks.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Build
+
+System buildpacks that have passed detection will be executed during the build phase. All `pre` buildpacks must execute before the detected buildpack group. All `post` buildpacks must execute after the detected buildpack group. However, the system buildpacks MUST NOT be added to the buildpack group, and their execution may be hidden from the end user.
```
ekcasey, line 64:
```suggestion
System buildpacks that have passed detection will be added to `group.toml` and treated like any other buildpack for the remainder of the build.
```

I feel strongly that we should not treat system buildpacks as a special case except when strictly necessary (how they are added to the order). I also don't think it is realistic or desirable for the lifecycle to hide the existence of these buildpacks.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+
+If a system buildpack exits with a status of `100`, the build will fail.
+
+A new flag to the lifecycle `builder`, `--disable-system-buildpacks`, will disable system buildpacks.
+
```
ekcasey, line 69:
```suggestion
```

I don't think this is necessary. See comments above.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Default system buildpacks are hidden from the user before the build and their execution may be unexpected.
```
ekcasey, line 73: See comments above re hiding buildpacks.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+## Do Nothing
+
+End users would have to add buildpacks like the `profile-buildpack` or other buildpacks that implement system/spec behaviors themselves.
+
+## Use the `[[order]]` table
+
+Instead of a new `[system]` table, we could put `pre` and `post` in the `[[order]]` table. However, this could imply that there is a interaction/override/etc between these buildpacks and the `pre`/`post` buildpacks in `project.toml`. But there is not.
```
ekcasey, line 84: I kind of like this alternative `builder.toml` schema? But it's not a strong preference.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,134 @@
...
+```
+/cnb/lifecycle/detector \
+  [--pre-buildpacks <group>]\
+  [--post-buildpacks <group>]\
+```
+
+The lifecycle:
+
+* SHALL merge the `<pre-buildpacks>` group with each group from `<order>` such that the `pre` buildpacks are placed at the beginning of each group before running detection.
+* SHALL merge the `<post-buildpacks>` group with each group from `<order>` such that the `post` buildpacks are placed at the end of each group before running detection.
```
ekcasey, line 132: This is pretty different from what is described and what I reviewed above?

ekcasey (changes requested):

text/0000-system-buildpacks-in-builder.md
jkutner: I don't have a strong preference

text/0000-system-buildpacks-in-builder.md
jkutner:
```suggestion
- If system buildpacks are hidden from the user before the build and their execution may be unexpected.
```

text/0000-system-buildpacks-in-builder.md
jkutner: I'm split

jkutner: If we allow non-optional, it could cause a group with all optional non-system buildpacks to pass though

text/0000-system-buildpacks-in-builder.md
jkutner: Yea, I'm onboard with that, but i think we still need to have "If a system buildpack passes detection and no buildpack groups pass detection, then detection MUST fail" somehow. That might mean optional is required to be true.

text/0000-system-buildpacks-in-builder.md
jkutner: addressed in f8f776d

text/0000-system-buildpacks-in-builder.md
jkutner: I changed my mind. I think we decided that making a system buildpack non-optional should be a platform/builder owner's decision. That does create the possibility that a group with all optional buildpacks could pass detection when it otherwise would not, but we should leave that up to the platform/builder owner. As long as the platform has a mechanism to disable system buildpacks (and `pack` will), then there's an escape valve.

jkutner: I'm revising this paragraph:
>The exit code of detection by system buildpacks MUST influence the selected buildpack group. If a system buildpack is non-optional and fails detection, then detection MUST for that group fail. If a system buildpack is optional and passes detection, then detection MAY pass for that group.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,145 @@
...
+The fields in the `system.pre.buildpacks` table and `system.post.buildpacks` table match the fields in the existing [`order.group` table](https://buildpacks.io/docs/reference/config/builder-config/#order-_list-required_).
+
+When a builder includes one or more `system.*.buildpacks` entry, the detect phase will prepend and append all `pre` and `post` buildpacks to each detection group in the provided order, respectively.
+
+**Note:** A non-`optional` system buildpack creates the possibility that a user provided group with all optional buildpacks could pass detection when it otherwise would not. We leave that up to the platform/builder owner. As long as the platform has a mechanism to disable system buildpacks (and `pack` will), then there is an escape valve for this situation.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Unless otherwise stated, system buildpacks conform to the [buildpack API](https://github.com/buildpacks/spec/blob/main/buildpack.md).
```
ekcasey, line 52:
```suggestion
System buildpacks conform to the [buildpack API](https://github.com/buildpacks/spec/blob/main/buildpack.md).
```

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,145 @@
...
+## Build
+
+System buildpacks that have passed detection will be added to `group.toml` and treated like any other buildpack for the remainder of the build.
+
+If a system buildpack exits with a status of `100`, the build will fail.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Default system buildpacks are hidden from the user before the build and their execution may be unexpected.
```
ekcasey, line 71:
```suggestion
- If system buildpacks are hidden from the user before the build, their execution may be unexpected.
```

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,145 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [RFC-0078: Group additions to Builder order](https://github.com/buildpacks/rfcs/blob/main/text/0078-group-additions.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the system buildpacks that pass detection be added to the buildpack group? This might make it difficult to hide them.
+    - disposition: yes. hiding is optional and up to the platform
```
ekcasey, line 110:
```suggestion
```
If I am understanding correctly we resolved this in favor of adding them to the group.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,145 @@
...
+- Should the system buildpacks output to stdout/stderr and should that be displayed?
+    - disposition: it may be
```
ekcasey, line 112:
```suggestion
```
Given that we have agreed not to do anything "special" beyond adding these buildpacks to the order I think this is also resolved. The system buildpacks will print to stdout/stderr just like all the other buildpacks.

text/0000-system-buildpacks-in-builder.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+## `detector` in Platform specifiction
+
+This proposal introduces a `--pre-buildpacks` and `--post-buildpacks` option on the `detector`.
```
ekcasey, line 119:
```suggestion
This proposal introducing `--system` option on the `detector`.
```

ekcasey (approved):

sclevine (approved):

jkutner:
Moving this into FCP, closing Feb 11. (next Friday)

hone:
@buildpacks/implementation-maintainers are there any issues that should be created for this?

PR #181, opened 2021-08-13
Additional exportable layers

samj1912:


text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,117 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+N/A.
+
+# Alternatives
+[alternatives]: #alternatives
+
+This proposal is useful as it relies of existing OCI image/Docker conventions to add a much needed extension point to the API.
```
hone, line 100: I'm not sure I understand this alternative?

samj1912: Was missing a section.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,117 @@
...
+The platform would be responsible for mounting appropriate volumes based on the `Volumes` key in the OCI image config. For a platform like `pack` which relies on a daemon this should be fairly straight-forward to achieve since `docker run` automatically mounts appropriate volumes. For other platforms like `kpack` this would involve inspecting the builder image beforehand and modifying the build pod spec to accommodate the specified volumes.
+
+The lifecycle changes would involve exporting the files present in the above locations which should be similar to the logic that currently exists for exporting application workspace. Buildpacks could also take advantage of `slices` to specify paths in these additional directories that should exist as separate layers.
+
+Changes to the Buildpack API would be minimal and would mostly involve the Buildpacks being passed the additional `CNB_EXPORT_VOLUMES` variable for detection and build logic.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+N/A.
```
hone, line 95: I actually think one of the drawbacks is any buildpack using this feature is highly tied to the underlying stack image. With #167, does this make it harder for buildpacks to have that tight coupling? How does one know what `VOLUME`s are required before building?

hone:
What happens if the `VOLUME` is not set in the stack image and the buildpack wants to export it?

hone:
I do think the use cases are compelling, but my concerns are around exposing this information upfront and I think this impacts #167.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,123 @@
...
+FROM base as build
+
+# Volume instruction in the build image that the lifecycle
+# will export out. This must match the volumes declared in the run image.
+VOLUME ["/app", "/opt/extensions", "/home/cnb/.config"]
+
+# Optionally users can override the default application workspace
+# by specifying it through the WORKDIR directive in the build
+# and run images.
+WORKDIR /app
```
ekcasey, line 73: Should we instead use `CNB_APP_DIR` for this? Although workdir=appdir in the buildpack API, in the platform API we don't use the working directory to indicate this and instead use an environment variable.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,123 @@
...
+
+# Set user and group (as declared in base image)
+USER ${CNB_USER_ID}:${CNB_GROUP_ID}
+```
+
+Buildpacks will be passed a list of volumes through the `CNB_EXPORT_VOLUMES` variable and this variable would be a JSON list. Buildpacks may read this environment variable during `detect` or `build`.
```
ekcasey, line 79:
```suggestion
Buildpacks will be passed a list of volumes through the `CNB_EXPORT_DIRS` variable and this variable would be a JSON list. Buildpacks may read this environment variable during `detect` or `build`.
```
Given we are exporting the contents of these build-time volumes into real layers in the resulting app image maybe this makes more sense from the buildpack perspective? It is irrelevent to the buildpack that these are build time volumes.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,123 @@
...
+
+The main motivation behind this RFC is to unlock exporting application images that require libraries or software to be installed in certain directories apart from the default application directory and these installations are rebase safe and self-contained. This unlocks use cases where buildpacks may install software in `/opt` directories (for AWS Lambda extensions or other common standalone software) for example or preserve some settings in a `/home/$USER/.config` for the output application. This provides a rebase safe extension point without major changes to the buildpacks API.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Users would be able to define stack images with `VOLUME` directives in Dockerfiles.
+
+
+For example a build/run base image may look like -
```
ekcasey, line 34: Is it confusing to add these volumes to the run image, given that when we extend the run image to create the app image these directories will have buildpack-contributed contents and thus will not behave as volumes?

If did wanted these as volumes on the run image (to prevent folks from extending the run image in a way that breaks rebase) I think the exporter should remove these volumes from the final app image.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,123 @@
...
+
+The platform would be responsible for mounting appropriate volumes based on the `Volumes` key in the OCI image config. For a platform like `pack` which relies on a daemon this should be fairly straight-forward to achieve since `docker run` automatically mounts appropriate volumes. For other platforms like `kpack` this would involve inspecting the builder image beforehand and modifying the build pod spec to accommodate the specified volumes.
+
+The lifecycle changes would involve exporting the files present in the above locations which should be similar to the logic that currently exists for exporting application workspace. Buildpacks could also take advantage of `slices` to specify paths in these additional directories that should exist as separate layers.
+
+Changes to the Buildpack API would be minimal and would mostly involve the Buildpacks being passed the additional `CNB_EXPORT_VOLUMES` variable for detection and build logic.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
```
ekcasey, line 94: It would be hard for a simple platform like tekton to implement this.

hone:
@samj1912 will update with an alternative proposal from the last WG.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,81 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+This RFC would require changes to the lifecycle and platform API.
+
+The platform would be responsible for mounting appropriate volumes based on the `io.buildpacks.export_dirs` label in the OCI image config. For a platform like `pack` which relies on a daemon this should be fairly straight-forward to achieve. For other platforms like `kpack` this would involve inspecting the builder image beforehand and modifying the build pod spec to accommodate the specified volumes.
+
+The lifecycle changes would involve exporting the files present in the above locations which should be similar to the logic that currently exists for exporting application workspace. Buildpacks could also take advantage of `slices` to specify paths in these additional directories that should exist as separate layers.
+
+The lifecycle `analysis` phase would also be responsible for validating that the above list of `export_dirs` is valid for the provided `run` image.
```
jabrown85, line 53: How would `analyzer` validate this? We would need to be able to do this quickly, so some sort of pre-processing would need to occur.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,81 @@
...
+[prior-art]: #prior-art
+
+- [RFC 72](https://github.com/buildpacks/rfcs/blob/main/text/0072-image-workdir.md)
+- [RFC PR #172](https://github.com/buildpacks/rfcs/pull/172)
+- [RFC PR #173](https://github.com/buildpacks/rfcs/pull/173)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+TBD.
```
jabrown85, line 76: Are there any security ramifications for certain paths? Do platforms or builders need a way to deny paths?

text/0000-additional-exported-layers.md
natalieparellano, line 53: How?

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,81 @@
...
+
+For example a a buildpack may look like - 
+
+```toml
+api = "<buildpack API version>"
+
+[buildpack]
+id = "<buildpack ID>"
+name = "<buildpack name>"
+export_dirs = ["/opt", "/home/cnb/.config"]
```
jabrown85, line 39: Are we worried about this coupling between a buildpack and the stack? Does this hurt portability of buildpacks between stacks?

text/0000-additional-exported-layers.md
samj1912: It would make the analysis phase fairly expensive. There are a couple of ways -

1. During the builder creation, the platform add a label to the run images with a flag making sure that the buildpacks are compatible with the run image.
2. The other option is to download the run images and check all the paths across the layers to confirm that the specified folders are empty.

Not ideal, but we could skip the validation as discussed in the WG if the operator trusts the builder to be valid.

text/0000-additional-exported-layers.md
samj1912: It probably will, but it is an optional feature to implement. If a buildpack is aiming for portability it could choose not to do things this way.

text/0000-additional-exported-layers.md
samj1912: It might be good to have a block list of paths.

text/0000-additional-exported-layers.md
samj1912: Answered below. It would make it fairly slow if the platform has not put additional metadata in the run image.

jabrown85: Could validation be _just_ the Deny path list that is set at builder creation time and stored as a label? Or is that not validating the image in the way you would expect?

hone:
@samj1912 @sclevine @hone will meet to discuss the multiple input directory proposal before moving forward with this one.

hone:
@samj1912 will update with single input / multiple output proposal from the discussion.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,86 @@
...
+```toml
+api = "<buildpack API version>"
+
+[buildpack]
+id = "<buildpack ID>"
+name = "<buildpack name>"
+exportable-volumes = ["aws-extensions", "user-config-dir"]
+```
+
+The above would only be valid for normal buildpacks and not meta-buildpacks. Based on the `exportable-volumes` field, a platform would be responsible for validating that the `build` and `run` image do not have any content on these paths. By default these volume names will map to `/workspaces/<exportable-volume-name>` but a user may specify a label `io.buildpacks.exportable-volumes` on the `build` image to map a volume name to a different location. This label would look like -
```
jkutner, line 42: Is `workspaces` (with a `s`) just a typo?

jkutner: it's not a typo, i see now

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+- `exportable-volumes`: A set of volumes describing where the buildpack process is likely to write data specific and should be exported alongside the application workspace.
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation behind this RFC is to unlock exporting application images that require libraries or software to be installed in certain directories apart from the default application directory and these installations are rebase safe and self-contained. This unlocks use cases where buildpacks may install software in `/opt` directories (for AWS Lambda extensions or other common standalone software) for example or preserve some settings in a `/home/$USER/.config` for the output application. This provides a rebase safe extension point without major changes to the buildpacks API.
```
aemengo, line 24: If the motivation is to allow for "libraries or software that install in certain directories", it goes at odds with _requiring_ that "these exportable volumes be empty". After all, we can't guarantee that these "libraries or software" aren't installing to `/` or something.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,86 @@
...
+id = "<buildpack ID>"
+name = "<buildpack name>"
+exportable-volumes = ["aws-extensions", "user-config-dir"]
+```
+
+The above would only be valid for normal buildpacks and not meta-buildpacks. Based on the `exportable-volumes` field, a platform would be responsible for validating that the `build` and `run` image do not have any content on these paths. By default these volume names will map to `/workspaces/<exportable-volume-name>` but a user may specify a label `io.buildpacks.exportable-volumes` on the `build` image to map a volume name to a different location. This label would look like -
+
+```json
+[
+    {"name": "user-config-dir", "value": "/home/cnb/.config"},
```
jabrown85, line 46: I like this mapping concept being done by the base image author. 

Could this list be something buildpack authors could get at and then name layers that match `user-config-dir` that gets exported to `/home/cnb/.config` during export phase? Or perhaps a key in the `layer.toml` has to be set so it is more explicit.

The requirement for an empty dir and symlinking for builder could potentially be avoided. A buildpack author could put a new binary in the resulting app image's `/usr/bin` by putting the binary in `/layers/<bp-dir>/usr-bin/`. If a new stack is used that doesn't support it, the layer would still exist on the resulting app image but not in the overridden path.

Giving the buildpack a dictionary so they can easily check if "usr/bin" is available to use would allow buildpack authors to either put the binary in "usr/bin" or do more work like altering PATH and whatnot to get it to work on other base images.

```
{
  "usr/bin" : "usr-bin",
  "~/.config": "user-config-dir"
}
```

text/0000-additional-exported-layers.md
jkutner: Is it possible this could open a new attack vector? I can't think of anything specific, but i wonder what's possible if a buildpack can drop files in `/etc/` or `/usr/bin/` for example.

text/0000-additional-exported-layers.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation behind this RFC is to unlock exporting application images that require libraries or software to be installed in certain directories apart from the default application directory and these installations are rebase safe and self-contained. This unlocks use cases where buildpacks may install software in `/opt` directories (for AWS Lambda extensions or other common standalone software) for example or preserve some settings in a `/home/$USER/.config` for the output application. This provides a rebase safe extension point without major changes to the buildpacks API.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Buildpack authors would be able to buildpacks with `buildpack.exportable-volumes` keys in their `buildpack.toml` files.
```
jkutner, line 29:
```suggestion
Buildpack authors would be able define a set of `buildpack.exportable-volumes` keys in their `buildpack.toml` files.
```

PR #182, opened 2021-08-20
RFC: Project Descriptor(project.toml) Converter

haliliceylan:
[Readable](https://github.com/haliliceylan/rfcs/blob/project-toml-converter/text/0000-project-toml-converter.md)

Hello everyone, 

This is my first RFC as a GSOC Mentee at this summer. We prepared this RFC with @natalieparellano. We spent some time in my own fork to get ready this RFC. You can access previous comment and feedbacks from [here](https://github.com/haliliceylan/rfcs/pull/2).

### DIFF
--- a/text/0000-project-toml-converter.md
+++ b/text/0000-project-toml-converter.md
@@ -25,7 +25,7 @@ The idea is to ship a binary with the lifecycle that would be responsible for tr
 
 - This feature has similarities with the "prepare" phase that has been discussed previously (see #555). But it is much smaller in scope. Potentially, the binary described in this RFC could eventually take on more responsibilities and look more like "prepare". But that is out of scope for this RFC.
 
-- It is non-trivial for platform operators to upgrade support for different versions of project.toml. Project Descriptor is a file which describe behavior related to buildpacks. Platform operators need to give these behaviors according to developer app requests. So with this feature we are giving to developers the ability to use any version of project.toml knowing that the platform can always support it. At the same time we are giving to operators the convenience of only having to know about one schema at a time for project.toml.
+- Project Descriptor (aka project.toml) is a file which describes behavior related to buildpacks. It is non-trivial for platform operators to upgrade support for different versions of project.toml. So with this feature we are giving to developers the ability to use any version of project.toml knowing that the platform can always support it. At the same time we are giving to operators the convenience of only having to know about one schema at a time for project.toml.
 
 - What is the expected outcome?
 
--- a/text/0000-project-toml-converter.md
+++ b/text/0000-project-toml-converter.md
@@ -67,7 +67,7 @@ value = "<string>"
 
 
 Example Project.TOML with 0.2 schema
-````
+```
 [_]
 id = "io.buildpacks.my-app"
 version = "0.1"
--- a/text/0000-project-toml-converter.md
+++ b/text/0000-project-toml-converter.md
@@ -25,7 +25,7 @@ The idea is to ship a binary with the lifecycle that would be responsible for tr
 
 - This feature has similarities with the "prepare" phase that has been discussed previously (see #555). But it is much smaller in scope. Potentially, the binary described in this RFC could eventually take on more responsibilities and look more like "prepare". But that is out of scope for this RFC.
 
-- Sometimes Operators has some issues with different project.toml files. As you know system operator can not easily upgrade their system packages, pack and buildpacks. So with this feature we are giving to developers able to use always latest version of project.toml. In same time we are giving to operators able to stable update scheduling.
+- It is non-trivial for platform operators to upgrade support for different versions of project.toml. Project Descriptor is a file which describe behavior related to buildpacks. Platform operators need to give these behaviors according to developer app requests. So with this feature we are giving to developers the ability to use any version of project.toml knowing that the platform can always support it. At the same time we are giving to operators the convenience of only having to know about one schema at a time for project.toml.
 
 - What is the expected outcome?
 
@@ -66,7 +66,7 @@ Discuss prior art, both the good and bad.
 - What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
 
 # Spec. Changes (OPTIONAL)
-[spec-changes]: #spec-changes
+[spec-changes]: #spec-changesf
 Does this RFC entail any proposed changes to the core specifications or extensions? If so, please document changes here.
 Examples of a spec. change might be new lifecycle flags, new `buildpack.toml` fields, new fields in the buildpackage label, etc.
 This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
--- a/text/0000-project-toml-converter.md
+++ b/text/0000-project-toml-converter.md
@@ -42,27 +42,43 @@ Example Project.TOML with 0.1 schema
 [project]
 id = "io.buildpacks.my-app" # machine readable
 name = "my app" # human readable
-version = "0.1"
-authors = ["<string>"]
-documentation-url = "<url>"
-source-url = "<url>"
+version = "1.2.3"
+authors = ["Great Developer"]
+documentation-url = "my-site.com/docs"
+source-url = "my-site.com/source"
 
 [[project.licenses]]
-type = "<string>"
-uri = "<uri>"
+type = "some-license-type"
+uri = "my-site.com/license"
+
 
 [build]
-include = ["<string>"]
-exclude = ["<string>"]
+include = [
+    "cmd/",
+    "go.mod",
+    "go.sum",
+    "*.go"
+]
+exclude = ["secrets.yaml"]
+
 [[build.buildpacks]]
-id = "<string>"
-version = "<string>"
-uri = "<string>"
+id = "io.buildpacks/java"
+version = "1.0"
+uri = "myreg.io/myspace/java"
+
+[[build.buildpacks]]
+id = "example/post-build"
+
+    [build.buildpacks.script] # does this make sense?
+    api = "0.5"
+    inline = "./post-build.sh"
+
 [[build.env]]
-name = "<string>"
-value = "<string>"
+name = "SOME_VAR"
+value = "some-val"
+
+
 [metadata]
-# additional arbitrary keys allowed
 cdn = "https://cdn.example.com"
 ```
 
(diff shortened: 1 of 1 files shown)


haliliceylan:
for DCO I force pushed.

jromero:
> for DCO I force pushed. and fixed conflict errors because of force push.

Looks like you might have accidently signed too many commits back. The commits tab should only include your changes.

haliliceylan:
> > for DCO I force pushed. and fixed conflict errors because of force push.
> 
> Looks like you might have accidently signed too many commits back. The commits tab should only include your changes.

Thanks for making me realize, To fix problem, I runned these commands one by one; 
```
git checkout -b backup_3 origin/haliliceylan:project-toml-converter # backup for better way
git rebase main/main # i discard any conflict except rfc file
git reset main/main # Keep files but delete all previous commits
git add . && git commit --signoff -m "blabla" # create one commit with signed off
```

I hope problem is fixed. Please let me know if there is any better way to protect branch history.

aemengo (commented):
> The Project TOML Converter is a CLI tool that will ship alongside the lifecycle.

Why are we deciding to make this a separate binary? As opposed to another code path that gets run? Is this a preemptive move for a `./prepare` phase binary?

--

I guess... I'm not sure that this is a problem that has to be solved. While I agree with the underlying premise:

>  We think it's too hard for platform operators to support multiple versions of project.toml.

the behavior that I'm expecting is that the **platform complains/rejects about deprecated project schema versions**. My thoughts:

- There's plenty of other tooling that exhibit this behavior.
- Having the internal component of the platform silently upgrade my input not only feels like _black magic_, but a bit over-engineered.
- A future schema version might indeed be a _breaking change_, which would require user input before making a schema upgrade
- And finally, keeping in mind that the application developer is writing these things, I'm of the opinion that _they should be prompted_ to upgrade by the platform. Otherwise, this RFC encourages this non-best-practice of using `schema v0.1` forever.

But this is a good discussion that we should be having. Thanks for raising this RFC, @haliliceylan

aemengo:
RE @jromero's [comment](https://github.com/buildpacks/rfcs/pull/182#issuecomment-906440138)

These are great points in response to my original comment. Thank you for raising them. 

In general, once more, perhaps _I_ don't see the "the need for platforms to all be in sync". To me, a parallel to the **project.toml** is the [AWS Cloudformation Template](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/format-version-structure.html). The template itself has a _Format Version_, with _only one valid_ entry. We can assume that AWS will not convert my template for me. This is the behavior I've come to expect - even if other platforms accepted AWS Cloudformation templates.

I just wanted to point my predilection based on conditioning as a cloud tool user. It looks like this project, and maintainers, has a better perspective on how platforms should behave. And that's completely fine.

jkutner (approved):

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+
+* __CNB_PLATFORM_API__ The Project TOML Converter will use the `CNB_PLATFORM_API` environment variable to decide the target schema. The `CNB_PLATFORM_API` environment variable is set by the platform in the lifecycle's execution environment. Version 0.3 is assumed if it is not set.
+
+# Motivation
+[motivation]: #motivation
+
+- This feature has similarities with the "prepare" phase that has been discussed previously (see #555). But it is much smaller in scope. Potentially, the binary described in this RFC could eventually take on more responsibilities and look more like "prepare". But that is out of scope for this RFC.
+
+- Project Descriptor (aka project.toml) is a file which describes behavior related to buildpacks. It is non-trivial for platform operators to upgrade support for different versions of project.toml. So with this feature we are giving to developers the ability to use any version of project.toml knowing that the platform can always support it. At the same time we are giving to operators the convenience of only having to know about one schema at a time for project.toml.
```
jromero, line 28: What operators are we talking about here? I have the sense that this should say platform implementers. If that is so then I have a predicament with this  statement. Only a subset of the `project.toml` wouldn't be translated due to the fact that the spec allows for undefined keys/tables. It seems like platforms may very well still have to care about the initial input format to some degree.

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+metadata = metadata
+```
+
+For conversion to/from v0.2:
+
+```
+project = _
+build = io.buildpacks
+build.buildpacks = io.buildpacks.group
+metadata = _.metadata
```
jromero, line 170: what happens to overlapped properties? ie. translating the following from `v0.2` to `v0.1`:

```toml
[_]
name = "My App"

# in this schema version this is consider arbitrary/user data
[project]
name = "Client 2 Project"
```

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+version = "0.1"
+
+[_.metadata]
+cdn = "https://cdn.example.com"
+
+[[_.metadata.assets]]
+url = "https://cdn.example.com/assets/foo.jar"
+checksum = "3b1b39893d8e34a6d0bd44095afcd5c4"
+
+buzz = ["a", "b", "c"]
```
jromero, line 82: If I recall correctly, arbitrary data would be retained. Could we get an example of a 1-to-1 real complex example to get a better understanding on how the translation would work?

jromero (changes requested):

text/0000-project-toml-converter.md
haliliceylan: @jkutner @hone, do you have any good examples for that ?

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+
+For conversion to/from v0.1
+```
+project = project
+build = build
+build.buildpacks = build.buildpacks
+metadata = metadata
+```
+
+For conversion to/from v0.2:
```
natalieparellano, line 164: We should call out explicitly that we only intend to "translate" data under `_` and `io.buildpacks`. Then there is the question of what to do with data under other domains - I believe that 
* the original project.toml in 0.2 schema
* a separate file for "other" stuff
* a separate file for each domain - but this would be hard because we wouldn't know which `.` to split on

were all discussed. I don't think there is much substantive difference between the first two, and my preference is for the first one for its simplicity.

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+[definitions]: #definitions
+
+* __project descriptor__ - a file that is included in the application directory to configure various behavior related to buildpacks; for more information, see the [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) extension specification
+
+* __CNB_PLATFORM_API__ The Project TOML Converter will use the `CNB_PLATFORM_API` environment variable to decide the target schema. The `CNB_PLATFORM_API` environment variable is set by the platform in the lifecycle's execution environment. Version 0.3 is assumed if it is not set.
+
+# Motivation
+[motivation]: #motivation
+
+- This feature has similarities with the "prepare" phase that has been discussed previously (see #555). But it is much smaller in scope. Potentially, the binary described in this RFC could eventually take on more responsibilities and look more like "prepare". But that is out of scope for this RFC.
```
natalieparellano, line 26:
```suggestion
- This feature has similarities with the "prepare" phase that has been discussed previously (see #555). But it is much smaller in scope. Potentially, the binary described in this RFC could eventually take on more responsibilities and look more like "prepare". But that is out of scope for this RFC. If the binary were to eventually take on more responsibility, all of its functions should be easily turned on and off, so that platform operators could choose which functionality to expose to users.  
```

text/0000-project-toml-converter.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Project Descriptor (aka project.toml) is a file which describes behavior related to buildpacks. It is non-trivial for platform operators to upgrade support for different versions of project.toml. So with this feature we are giving to developers the ability to use any version of project.toml knowing that the platform can always support it. At the same time we are giving to operators the convenience of only having to know about one schema at a time for project.toml.
+
+- What is the expected outcome?
+
+# What it is
+[what-it-is]: #what-it-is
+
+The Project TOML Converter is a CLI tool that will ship alongside the lifecycle. It has only one responsibility at this time, to support conversion between schemas.
```
natalieparellano, line 35:
```suggestion
The Project TOML Converter is a CLI tool that will ship alongside the lifecycle. It has only one responsibility at this time, to support conversion between schemas. The tool will not need to be executed in a containerized environment (similar to `rebase`). Though it can be invoked as a separate binary it may end up being part of the lifecycle which itself is a multi-call binary (see https://github.com/buildpacks/rfcs/blob/main/text/0024-lifecycle-multicall-binary-for-build.md). 
```

text/0000-project-toml-converter.md
haliliceylan: maybe we can add a global prefix or postfix for keys which not included this list ? (just an idea)

text/0000-project-toml-converter.md
natalieparellano: @jromero is this what you had in mind? https://github.com/buildpacks/rfcs/pull/182/files#diff-d6828a110979d409ba620d1b80593932c1cea332756b3cc7f7c14922dfc0b16aR132-R133

text/0000-project-toml-converter.md
natalieparellano: 5c2075d should help clarify the mapping from 0.1 <-> 0.2 ...though I wasn't sure what to do about `[build.buildpacks.script]`.

text/0000-project-toml-converter.md
natalieparellano: https://github.com/buildpacks/rfcs/commit/5c2075de14bd73f6a972e4b6441c8de9b1b57ae6 may help

text/0000-project-toml-converter.md
jromero: IIRC, we were going to ignore (not map) previously non-existent properties. It doesn't make sense for platforms to expect properties that aren't associated with their specification.

hone:
@natalieparellano do you mind if I move this blocked dependent on the resolution of #191?

natalieparellano:
Let's draft this as we can safely say it is superseded by https://github.com/buildpacks/rfcs/pull/202

PR #186, opened 2021-10-26
Add run image SBOM

aemengo:
[Readable](https://github.com/aemengo/rfcs/blob/add-run-image-sbom/text/0000-run-image-sbom.md)

### DIFF
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -1,0 +1,76 @@
+# Meta
+[meta]: #meta
+- Name: Run Image Structured SBOM
+- Start Date: 2021-10-25
+- Author(s): aemengo
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following -
+
+- A run image can contain a single layer holding an SBOM (in format of CycloneDX or SPDX or Syft) of installed packages at `/cnb/sbom/bom.<ext>.json`, where `<ext>` will be `cdx` for CycloneDX documents, `spdx` for SPDX documents, and `syft` for Syft documents. These will initially be the only 3 supported SBOM types.
+- The diffID of this layer is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.base.sbom`.
+- For the sake of parity, the diffID of buildpack SBOM layers (created by lifecycle) is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.app.sbom`.
+
+# Definitions
+[definitions]: #definitions
+
+- SBOM (Software Bill Of Materials) / BOM: A list of components in a piece of software. Software vendors often create products by assembling open source and commercial software components. The SBOM describes the components in a product. In case of buildpacks the SBOM describes the contents of the various layers, buildpacks, stacks and the output app container.
+- Run Image: A container image that serves as a base image for application images in the buildpack toolchain.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC serves as an addendum to [#RFC 95: SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md) which introduces the SBOM for buildpack installed packages. For a full depiction of packages within a container image, container scanning tools would _also_ need an SBOM of the installed OS packages. These multiple SBOM files would be merged as part of the lifecycle build process, for easy consumption by an automated scanning tool.
+
+# What it is
+[what-it-is]: #what-it-is
+
+A stack creator is permitted to specify a SBOM for a given run image, in one discrete layer, at the following file path:
+
+```shell
+/cnb/sbom
+└── bom.<ext>.json
+```
+
+And a label with key of `io.buildpacks.base.sbom`, and value of the diffID of this layer, stored as metadata on the run image.
+
+With this label specified, the rebase operation can be augmented to reference the new sBOM layer for the intended application image.
+
+Merging BOM files, whether files of the same format or files of different formats, is out of scope for this RFC. It would require its own separate RFC.
+
+## Tooling
+
+To facilitate adherence to this specification, the CNB project will create tooling that consumes a given run container image and a given SBOM json file, to create a new image with the SBOM information embedded at the correct location and diffID metadata attached. This allows for generation of SBOM files both within the image and outside it.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This RFC tries to thrust a paradigm which is an even greater burden on stack operators, in light of the recently approved [#RFC 96: Remove Stacks](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md).
+
+# Alternatives
+[alternatives]: #alternatives
+
+    -
+
+# Prior Art
+[prior-art]: #prior-art
+
+- [#RFC 95: SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+    -
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+- The [platform.md#run-image](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) would need an addition for the `io.buildpacks.base.sbom` key, referencing the layer diffID that holds the run-image BOM.
+- The [platform.md#labels](https://github.com/buildpacks/spec/blob/main/platform.md#labels) would need an addition for the `io.buildpacks.app.sbom` key, referencing the app image layer diffID that holds the buildpack BOM.


text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,70 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following -
+
+- A run image can contain a layer holding an SBOM (in format of CycloneDX or SBOM) of installed packages at `/sbom/bom.<ext>.json`, where `<ext>` will be `cdx` for CycloneDX documents and `spdx` for SPDX documents. These will initially be the only 2 supported SBOM types.
```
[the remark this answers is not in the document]
aemengo: Thanks! _Fixed._

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,70 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following -
+
+- A run image can contain a layer holding an SBOM (in format of CycloneDX or SBOM) of installed packages at `/sbom/bom.<ext>.json`, where `<ext>` will be `cdx` for CycloneDX documents and `spdx` for SPDX documents. These will initially be the only 2 supported SBOM types. 
+- The digest of this layer is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.sbom`.
```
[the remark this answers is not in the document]
aemengo: During our last discussion, we centered around:
> `io.buildpacks.app.sbom` for buildpack SBoMs label
> `io.buildpacks.base.sbom` for run-image SBoM label

https://buildpacks.slack.com/archives/C94UJCNV6/p1635433855037000?thread_ts=1635433432.033600&cid=C94UJCNV6

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,70 @@
...
+A stack creator is permitted to specify a SBOM for a given run image, in one discrete layer, at the following file path:
+
+```shell
+/sbom
+└── bom.<ext>.json
+```
+
+And a label with key of `io.buildpacks.sbom`, and value of the digest of this layer, stored as metadata on the run image.
+
+When the lifecycle comes across a run image with this label and valid digest, it will attempt to merge this SBOM with any SBOMs delivered by buildpacks (during the build phase) and output them at `/layers/config/<type>/sbom/bom.<ext>`, per [#RFC 95: SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md). Initially this merging will only be supported for CycloneDX as it has a well-defined and efficient way of merging multiple `bom` files. A reference implementation can be found at the [`cyclonedx-cli`](https://github.com/CycloneDX/cyclonedx-cli) project. In the future we may add support for merging SPDX `bom` files as well.
```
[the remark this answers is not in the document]
aemengo: During our last discussion, we decided to keep _merging_ out of scope for this RFC. As of right now, 2 sBOM layers (and 2 labels) are proposed: 1 for run-image, and 1 buildpack. Downstream tooling can do the "merging", if desired.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,72 @@
...
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following -
+
+- A run image can contain a layer holding an SBOM (in format of CycloneDX or SPDX or Syft) of installed packages at `/cnb/sbom/bom.<ext>.json`, where `<ext>` will be `cdx` for CycloneDX documents, `spdx` for SPDX documents, and `syft` for Syft documents. These will initially be the only 3 supported SBOM types.
```
sclevine, line 17: I would suggest that we use `/cnb/sbom/launch/bom.<ext>.json`, so that both the build-time and runtime SBoM could be included in the future, but that would make it difficult to use the same base image as both a runtime and build-time base image.

Ideally, the build-time and runtime base image SBoM could be included in the final image without needing to modify either of their corresponding layers.

Instead, maybe we could `/cnb/sbom/<id>.<ext>.json`, where `<id>` is the first 8 characters of the original base image digest. E.g., `/cnb/sbom/ef3945a2.cdx.json`.

Uniquely labeling the SBoM with a digest means that we could also include unextended SBoMs when the Dockerfiles RFC is implemented.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,72 @@
...
+- The digest of this layer is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.base.sbom`.
+- For the sake of parity, the digest of buildpack SBOM layers (created by lifecycle) is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.app.sbom`.
+
```
sclevine, line 20: In the future, we may want to add a flag to lifecycle to include build-time SBoMs in the app image. We could use labels like `io.buildpacks.build.base.sbom` and `io.buildpacks.build.app.sbom`.

I **don't** think this means we should rename `io.buildpacks.base.sbom` to `io.buildpacks.launch.base.sbom` -- just thought it might be worth pointing this out.

sclevine (changes requested):

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,72 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the following -
+
+- A run image can contain a layer holding an SBOM (in format of CycloneDX or SPDX or Syft) of installed packages at `/cnb/sbom/bom.<ext>.json`, where `<ext>` will be `cdx` for CycloneDX documents, `spdx` for SPDX documents, and `syft` for Syft documents. These will initially be the only 3 supported SBOM types.
+- The digest of this layer is written as a `LABEL` on the corresponding run image with the key: `io.buildpacks.base.sbom`.
```
sclevine, line 18: I think we should use the diffID instead of the digest, to match `io.buildpacks.app.sbom`

text/0000-run-image-sbom.md
sclevine: I put together a POC for attaching base image SBoMs in this format: https://github.com/sclevine/cnb-sbom
(You can also use it to read both base image and app layer SBoMs.)

text/0000-run-image-sbom.md
aemengo: Sure.

text/0000-run-image-sbom.md
aemengo: Yes, we _may_. I personally find it hard to base decisions purely on what may happen. I'd be curious if other explicitly have this use-case.

sophiewigmore:
Hey @aemengo! What's the status of this RFC? I haven't been able to attend the CNB working group meeting. I ask because I am working on a Paketo RFC that relies on this RFC, and wondering if you plan to make more changes and what the timeline is if so? Thanks!!

aemengo:
> Hey @aemengo! What's the status of this RFC? I haven't been able to attend the CNB working group meeting. I ask because I am working on a Paketo RFC that relies on this RFC, and wondering if you plan to make more changes and what the timeline is if so? Thanks!!

@sophiewigmore  Thank you for asking. Unfortunately, I'm afraid that more changes will be required on this RFC. At the very least, more discussion needs to happen. During our last working group meeting, we had concerns about the workflow of attaching a BOM to an already-built base image, and thereby _polluting_ the image. As of right now, the timeline is unknown. If your Paketo RFC is pending Paketo concerns, it might be more expedient to move at your own pace. Finally, I want remind (_in a non-patronizing way_) that buildpack RFCs are soon to close when all the core team members have voted and the _"Final Comment Period"_ status is applied.

sclevine (approved):

text/0000-run-image-sbom.md
aemengo: As discussed in Office Hours, on Nov 18, w/ @samj1912. There's disagreement about the complexity that this (the `<id>`, 8 characters of original base image digest) would introduce.

text/0000-run-image-sbom.md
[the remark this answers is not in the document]
aemengo: Though not explicitly called `pack attach sbom`, the need for additional tooling has been called out in the RFC.

hone (approved):

hone:
@jkutner @nebhale can you please review? :heart:

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,76 @@
...
+This RFC serves as an addendum to [#RFC 95: SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md) which introduces the SBOM for buildpack installed packages. For a full depiction of packages within a container image, container scanning tools would _also_ need an SBOM of the installed OS packages. These multiple SBOM files would be merged as part of the lifecycle build process, for easy consumption by an automated scanning tool.
+
+# What it is
+[what-it-is]: #what-it-is
+
+A stack creator is permitted to specify a SBOM for a given run image, in one discrete layer, at the following file path:
+
+```shell
+/cnb/sbom
+└── bom.<ext>.json
```
samj1912, line 39: Now that we are doing an external attach - I think I am actually happy for it to refer to the digest of the manifest liek @sclevine proposed. Might ensure that the bom is at a unique location and is not overridden. But happy either way.

aemengo: Noted 👍🏾

samj1912:
@aemengo really sorry to be going back and forth on this, but @natalieparellano and I were discussing the genpkgs implementation (https://buildpacks.slack.com/archives/CCP60GJLS/p1639669644077300) and it might make the implementation easier if the output file name did not use the digest - especially for implementing it in pack where the digest for the output run image may not be available until the lifecycle actually goes ahead and exports it and in this case the lifecycle might have to export the output run image to the registry.

I think we should instead solve it this way - use the diffid in the label and have the lifecycle ensure that it loads the bom from that diffid only if it is the last layer in the output image.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+And a label with key of `io.buildpacks.base.sbom`, and value of the diffID of this layer, stored as metadata on the run image.
```
samj1912, line 42: Can we also add the condition here that this must be the last layer on the image for the lifecycle to consider it a valid sbom? This is to prevent further extensions from re-using the same sbom.

aemengo: Please clarify. How would other extensions use an SBOM depending on the layer order?

nebhale (approved):

hone:
@jkutner @ekcasey can you please take a look at this :pray:

jkutner (approved):

text/0000-run-image-sbom.md
samj1912: @aemengo I meant something like the following - 

let's say someone takes the above base run image with the SBOm and label+diffid and extends it as 


```dockerfile
FROM valid-run-image-with-sbom
USER root
RUN apt-get install my-package
```

This new image will still have the above specified SBOM and label with the appropriate diffid of the SBOM layer, but the SBOM will be incorrect/stale.

With my proposal, the above extension will cause other layers to be added after the SBOM layer which might be a good way to indicate to the lifecycle that the SBOM is potentially stale since the layer with the SBOM diffid won't be the last layer anymore.

ekcasey: > Can we also add the condition here that this must be the last layer on the image for the lifecycle to consider it a valid sbom?

@samj1912 If we go this route maybe we can loosen this a little to be  SHOULD? I think a warning might be sufficient here.

Also, I am not sure what "must" would imply here. Would the build/rebase fail? Would the lifecycle remove the io.buildpacks.base.sbom` entirely.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,76 @@
...
+
+With this label specified, the rebase operation can be augmented to reference the new sBOM layer for the intended application image.
+
+Merging BOM files, whether files of the same format or files of different formats, is out of scope for this RFC. It would require its own separate RFC.
+
+## Tooling
+
+To facilitate adherence to this specification, the CNB project will create tooling that consumes a given run container image and a given SBOM json file, to create a new image with the SBOM information embedded at the correct location and diffID metadata attached. This allows for generation of SBOM files both within the image and outside it.
```
ekcasey, line 50: Is this inteneded to be separate tooling or a part of `pack`. I would like to see the answer that question in this RFC.

Once we know which sub-team is responsible I think it's fine to punt the details of the interface to a sub-team specific RFC but it would be good to align generally about who this RFC is handing the baton to.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This RFC tries to thrust a paradigm which is an even greater burden on stack operators, in light of the recently approved [#RFC 96: Remove Stacks](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md).
+
+# Alternatives
+[alternatives]: #alternatives
+
```
ekcasey, line 59: A Label is one alternative albeit an unappealing one b/c of the containerd label size issue.

An annotation is another alternative. The drawbacks to an annotation are consistency with buildpack-generated SBOM and the impossibility of doing something sensible in the daemon case.

hone:
@natalieparellano volunteering to take this over.

ekcasey:
Summary of WG Discussion 02/17/22

There is lingering discomfort around storing this SBOM in the run image because:
1. special tooling is required to create the image
2. it may [fall out of date](https://github.com/buildpacks/rfcs/pull/186#discussion_r773246636)
3. there is widespread acknowledgment that the layer solution is temporary pending changes (removal of daemon support) that would allow us to store the SBOM in an annotation or attestation

Although we can solve problem 1 with custom tooling. Problem 2 cannot be completely solved. While the lifecycle can use the layer order as a hint about whether the SBOM is out of date, the lifecycle cannot guarantee accuracy. Efforts to solve for or mitigate problem 1 and mitigate problem 2 feel like a waste of effort given problem 3. It feels especially bad to bake new temporary interfaces into the spec, given that we would like to move towards spec stability.

We discussed moving the SBOM out of the run image, and making it a responsibility of the platform to supply an accurate run-image SBOM to the lifecycle for export, at which point it will be included in the sbom layer with the buildpacks generated SBOM. This free platforms up to be as strict or flexible as they want and to adapt to evolving specifications around sbom and attestation format/storage.

(cc @samj1912 @natalieparellano @jabrown85 @sclevine )

jromero:
@ekcasey based on your latest comment do we have someone adopting this RFC to make the necessary changes or creating a superseding RFC? It sounds like we would still need an RFC for the SBOM contribution to the exporter.

jromero (changes requested):
@ekcasey based on your latest comment do we have someone adopting this RFC to make the necessary changes or creating a superseding RFC? It sounds like we would still need an RFC for the SBOM contribution to the exporter.

natalieparellano:
@jromero please see https://github.com/buildpacks/rfcs/pull/211

PR #187, opened 2021-11-05
RFC: Add images table (project.toml)

pbusko:
[Readable](https://github.com/sap-contributions/buildpacks-rfcs/blob/project-descriptor-images/text/0000-project-descriptor-images.md)

### DIFF
--- a/text/0000-project-descriptor-images.md
+++ b/text/0000-project-descriptor-images.md
@@ -21,17 +21,30 @@ Add a new `images` key in the `project.toml` file.
 * **image**: the output of a running a buildpack(s) against a project
 * **image name**: fully qualified name of an OCI image in the form `<registry>/<name>:<tag>`.
 
-
 # Motivation
 [motivation]: #motivation
 
-Currently there is no support for specifying the output image in the `project.toml` although flags like `--tags` are already present.
+Currently, there is no support for specifying an output image in the `project.toml` although flags like `--tag` are already present.
 
 # What it is
 [what-it-is]: #what-it-is
 
-buildpack user will be able to specify the resulting images name in the `project.toml`
+We propose the following changes, so that buildpack user will be able to specify the resulting images in the `project.toml`:
 
+- Add an optional top level array of tables `[[images]]`, that describes resulting images.
+
+```toml
+[[images]]
+name = "<string>"
+registry = "<string>"
+tags = ["string"]
+```
+
+- `images.name`: mandatory.
+- `images.tags`: optional, defaults to `latest`.
+- `images.registry`: optional, defaults to `docker.io`.
+
+`pack`'s flag `--tag` should take precedence over `images` blocks specified in a `project.toml`.
 
 # How it Works
 [how-it-works]: #how-it-works
@@ -43,6 +56,8 @@ buildpack user will be able to specify the resulting images name in the `project
   - https://github.com/buildpacks/rfcs/blob/main/text/0084-project-descriptor-domains.md
   - Closed PR which includes some similar changes: https://github.com/buildpacks/rfcs/blob/project-descriptor-simple/text/0000-project-descriptor.md
 
+Example:
+
 ```toml
 [[images]]
 name = "spring-petclinic"
@@ -51,29 +66,29 @@ tags = ["latest", "v1"]
 
 [[images]]
 name = "spring-petclinic"
-registry = "hub.docker.com"
 tags = ["latest", "v1"]
+
+[[images]]
+name = "spring-petclinic"
+registry = ["private.registry.corp:8443"]
 ```
 
-- Tags:
-  - Could be read from the `--tags` flag
-  - Could be defaulted to `project.lversion` field of `project.toml`
-- Name: Could just be taken from `project.name` (if present)
+This example will produce the following images:
 
-# Drawbacks
-[drawbacks]: #drawbacks
-
-TBD
+- `gcr.io/spring-petclinic:latest`
+- `gcr.io/spring-petclinic:v1`
+- `docker.io/spring-petclinic:latest`
+- `docker.io/spring-petclinic:v1`
+- `private.registry.corp:8443/spring-petclinic:latest`
 
 # Alternatives
 [alternatives]: #alternatives
-
 
 - have `registries` as an array:
 
   ```toml
   [images]
-  registries = ["gcr.io", "hub.docker.com"]
+  registries = ["gcr.io", "docker.io"]
   name = "spring-petclinic"
   tags = ["latest", "v1"]
   ```
@@ -87,7 +102,7 @@ TBD
   tag = "latest"
 
   [[images]]
-  registry = "hub.docker.com"
+  registry = "docker.io"
   name = "spring-petclinic"
   tag = "v1"
 
@@ -100,8 +115,8 @@ TBD
   images = [
       "gcr.io/spring-petclinic:latest",
       "gcr.io/spring-petclinic:v1",
-      "hub.docker.com/spring-petclinic:latest",
-      "hub.docker.com/spring-petclinic:v1"
+      "docker.io/spring-petclinic:latest",
+      "docker.io/spring-petclinic:v1"
   ]
   ```
 
@@ -112,14 +127,10 @@ TBD
   images = [
       "gcr.io/{name}:latest",
       "gcr.io/{name}:v1",
-      "hub.docker.com/{name}:latest",
-      "hub.docker.com/{name}:v1"
+      "docker.io/{name}:latest",
+      "docker.io/{name}:v1"
   ]
   ```
-
-> Why is this proposal the best?
-
-Is it? Lets discuss.
 
 # Prior Art
 [prior-art]: #prior-art
@@ -130,9 +141,6 @@ Is it? Lets discuss.
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-Let us know.
-
-# Spec. Changes (OPTIONAL)
-[spec-changes]: #spec-changes
-
-TBD
+- Naming (`images` / `targets` / other)?
+- Should `project.name` be used instead of `images.name`?
+- Should `project.version` be used instead of `images.tags`?
--- a/text/0000-project-descriptor-images.md
+++ b/text/0000-project-descriptor-images.md
@@ -144,3 +144,33 @@ This example will produce the following images:
 - Naming (`images` / `targets` / other)?
 - Should `project.name` be used instead of `images.name`?
 - Should `project.version` be used instead of `images.tags`?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+The following will be added to the [Project Descriptor Extension Spec](https://github.com/buildpacks/spec/blob/extensions/project-descriptor%2F0.2/extensions/project-descriptor.md):
+
+### `io.buildpacks.images` (optional)
+
+This table MAY contain an array of image repository label definitions. The schema for this table is:
+
+```
+[[io.buildpacks.images]]
+name = "<string (optional, default=io.buildpacks.name)>"
+tags = ["<string (optional, default=latest)>"]
+registry = "<string (optional default=docker.io)"
+```
+
+* `name` - the name component of an image repository
+* `tags` - a list of tags used as the tag component of images in an image repository
+* `registry` - a registry server containing the image repository
+
+If no `name` component is provided, the default value MUST be derived from the `io.buildpacks.name` field.
+
+If no `tags` are provided, the default value MUST be `"latest"`.
+
+If no `registry` is provided, the default value MUST be `"docker.io"`.
+
+ALl of these values MAY be overridden by a platform (ex. `pack` may allow a `--tag` flag to overridden the provided `tag` value).
+
+Multiple `[[io.buildpacks.images]]` entries MUST result in the creation of multiple Docker image repositories. All images will have the same digest irrespective of the different names, tags, and registries that are provided.
--- a/text/0000-project-descriptor-images.md
+++ b/text/0000-project-descriptor-images.md
@@ -18,7 +18,7 @@ Add a new array of tables `[[images]]` to the `project.toml` file.
 
 * **project descriptor**: The `project.toml` file that describes a project.
 * **image**: the output of a running a buildpack(s) against a project
-* **image name**: fully qualified name of an OCI image in the form `<registry>/<name>:<tag>`.
+* **image name**: Global Unique Name (GUN) of an OCI image in the form `[REGISTRY_HOST[:REGISTRY_PORT]/]REPOSITORY[:TAG]`.
 
 # Motivation
 [motivation]: #motivation
@@ -44,17 +44,17 @@ Example:
 
 ```toml
 [[images]]
-name = "spring-petclinic"
 registry = "gcr.io"
+repository = "spring-petclinic"
 tags = ["latest", "v1"]
 
 [[images]]
-name = "spring-petclinic"
+repository = "spring-petclinic"
 tags = ["latest", "v1"]
 
 [[images]]
-name = "spring-petclinic"
 registry = "private.registry.corp:8443"
+repository = "spring-petclinic"
 ```
 
 This example will produce the following images:
@@ -73,7 +73,7 @@ This example will produce the following images:
   ```toml
   [images]
   registries = ["gcr.io", "docker.io"]
-  name = "spring-petclinic"
+  repository = "spring-petclinic"
   tags = ["latest", "v1"]
   ```
 
@@ -82,18 +82,18 @@ This example will produce the following images:
   ```toml
   [[images]]
   registry = "gcr.io"
-  name = "spring-petclinic"
+  repository = "spring-petclinic"
   tag = "latest"
 
   [[images]]
   registry = "docker.io"
-  name = "spring-petclinic"
+  repository = "spring-petclinic"
   tag = "v1"
 
   # ...
   ```
 
-- Simple array of strings containing all information i.e. `registry`, `name`, `tag`
+- Simple array of strings containing all information i.e. `registry`, `repository`, `tag`
 
   ```toml
   images = [
@@ -104,15 +104,15 @@ This example will produce the following images:
   ]
   ```
 
-- Simple array of images containing all information with `name` being a placeholder
+- Simple array of images containing all information with `repository` being a placeholder
 
   ```toml
-  name = "spring-petclinic"
+  repository = "spring-petclinic"
   images = [
-      "gcr.io/{name}:latest",
-      "gcr.io/{name}:v1",
-      "docker.io/{name}:latest",
-      "docker.io/{name}:v1"
+      "gcr.io/{repository}:latest",
+      "gcr.io/{repository}:v1",
+      "docker.io/{repository}:latest",
+      "docker.io/{repository}:v1"
   ]
   ```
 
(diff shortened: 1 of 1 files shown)


text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+registry = "gcr.io"
+tags = ["latest", "v1"]
+
+[[images]]
+name = "spring-petclinic"
+tags = ["latest", "v1"]
+
+[[images]]
+name = "spring-petclinic"
+registry = ["private.registry.corp:8443"]
```
jromero, line 57: This seems to conflict with line 48 and the proposed spec changes in line 149. I believe this is intended to be a string not a list of strings.

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+
+  [[images]]
+  registry = "docker.io"
+  name = "spring-petclinic"
+  tag = "v1"
+
+  # ...
+  ```
+
+- Simple array of strings containing all information i.e. `registry`, `name`, `tag`
```
jromero, line 96: I'm more of a fan of this alternative. It's versatile enough to allow for any possible desires that I can think of while maintaining simplicity. The only cost I see is the potential repetitiveness of the host/registry. Are there other drawbacks/benefits to consider between this and the main proposed solution?

text/0000-project-descriptor-images.md
pbusko: you're right, the typo is fixed

text/0000-project-descriptor-images.md
pbusko: One of the drawbacks is that we will loose the possibility to apply defaults to certain keys as described in the lines 141-155, e.g. to fall back to the default `docker.io` registry if none is specified, or to apply the `latest` tag.
Also apart from the potential repetitiveness of the host/registry combinations, the main proposed solution is probably easier to consume by the platform, which would simplify for example a validation of inputs.

jkutner: As long as the registry, name, and tag are structured in a url scheme (and not arbitrary) then wouldn't it still be possible for a value like `"me/myimage"` to be converted to `"docker.io/me/myimage:latest"`?

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+registry = "private.registry.corp:8443"
+```
+
+This example will produce the following images:
+
+- `gcr.io/spring-petclinic:latest`
+- `gcr.io/spring-petclinic:v1`
+- `docker.io/spring-petclinic:latest`
+- `docker.io/spring-petclinic:v1`
+- `private.registry.corp:8443/spring-petclinic:latest`
```
jkutner, line 66: In working group we discussed how having multiple registries in a single build could be problematic if `pack` is running with `--publish`.

It may also create problems on re-build, because one registry might have the previous layers, and another may not. Thus, where does it pull the layers from (or not pull from)

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+* **project descriptor**: The `project.toml` file that describes a project.
+* **image**: the output of a running a buildpack(s) against a project
+* **image name**: fully qualified name of an OCI image in the form `<registry>/<name>:<tag>`.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, there is no support for specifying an output image in the `project.toml` although flags like `--tag` are already present.
+
+# What it is
```
jkutner, line 28: .@sclevine said that even though this is a lifecycle proposal, it would be good to see the `pack` experience. This is somewhat alluded to with the mention of `--tag` option.

jromero: I believe the concerns come about the interoperability between config and provided inputs. How does any platform work with these values?

Take for example:

A `project.toml` has:

```toml
[[images]]
name = "spring-petclinic"
registry = "gcr.io"
tags = ["latest", "v1"]

[[images]]
name = "spring-petclinic"
tags = ["latest", "v1"]
```

and the user runs:

```bash
pack build my-app --tags my-app:v2
```

What is the expected outcome?

### Option 1

Order of precedence, any user-provided options override configuration. 

Following images are created:

1. `my-app` => `index.docker.io/my-app:latest`
1. `my-app:v2`  => `index.docker.io/my-app:v2`

### Option 2

Order of precedence, any configuration overrides user-provided options. 

Following images are created:

1. `gcr.io/spring-petclinic:latest`
1. `gcr.io/spring-petclinic:v1`
1. `spring-petclinic:latest`  => `index.docker.io/spring-petclinic:latest`
1. `spring-petclinic:v1`  => `index.docker.io/spring-petclinic:latest`

### Option 3

A union of user-provided and config.

Following images are created:

1. `my-app` => `index.docker.io/my-app:latest`
1. `my-app:v2`  => `index.docker.io/my-app:v2`
1. `gcr.io/spring-petclinic:latest`
1. `gcr.io/spring-petclinic:v1`
1. `spring-petclinic:latest`  => `index.docker.io/spring-petclinic:latest`
1. `spring-petclinic:v1`  => `index.docker.io/spring-petclinic:latest`

---

My vote would be Option 1 because it follows our current order of precedence strategy.

Something to consider is that arg1 (`<name>`) in `pack build` would now need to be optional to allow for `images[0]` value to take affect.

text/0000-project-descriptor-images.md
loewenstein: > In working group we discussed how having multiple registries in a single build could be problematic if `pack` is running with `--publish`.

Could you provide more details in how far this would be problematic?

> It may also create problems on re-build, because one registry might have the previous layers, and another may not. Thus, where does it pull the layers from (or not pull from)

Given we know the SHA sums of all layers, is it relevant from where `pack` would pull them?

text/0000-project-descriptor-images.md
pbusko: You're right, it's possible to do it also with the array of canonical names, but on the other hand image tags are frequently being accessed/changed by that CI systems which produces the actual build. Having this data structured makes it a lot easier to consume/modify programatically.

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+registry = "gcr.io"
+tags = ["latest", "v1"]
+
+[[images]]
+name = "spring-petclinic"
+tags = ["latest", "v1"]
+
+[[images]]
+name = "spring-petclinic"
+registry = "private.registry.corp:8443"
```
jkutner, line 57: I wonder if we'd eventually want to add things like a `processes` table under this, or at least allow a `default-process` field.

text/0000-project-descriptor-images.md
jkutner: >> In working group we discussed how having multiple registries in a single build could be problematic if pack is running with --publish.

> Could you provide more details in how far this would be problematic?

The problem is that `--publish` causes pack to write directly to the remote registry (as opposed to writing locally and pushing an image). It's not clear _if_ this is possible with multiple remote registries. If it is not, it may mean the `--publish` option is incompatible with having images on multiple registries.

> Given we know the SHA sums of all layers, is it relevant from where pack would pull them?

I think the problem may occur when different registries have different layers for the previous build. The lifecycle would need to know how to determine which is the one to use. 

I think think this might be solvable by adding something like:

```toml
[[io.buildpacks.images.previous]]
name = "..."
registry = "..."
```

text/0000-project-descriptor-images.md
jromero: The problem I foresee with structured format is that I don't think there's a convention in the OCI realm on what `registry` vs `name` vs `tag` is. We're getting there but it may still be confusing to newcomers.

For example, on `grc.io` with image `gcr.io/cncf-buildpacks-ci/demo/kpdemo:v1`: 

- What is the `registry`, `gcr.io` (hostname) or `gcr.io/cncf-buildpacks-ci/demo`?
- What is the `name`, `cncf-buildpacks-ci/demo/kpdemo` or `kpdemo`? 
- What is the tag, `gcr.io/cncf-buildpacks-ci/demo/kpdemo:v1` or `v1`?
   - (the answer sounds obvious, `tag` but many systems, even `pack` use `tag` to mean the full reference)

Do your answers change on `docker.io` with image `index.docker.io/library/bash:latest`?

text/0000-project-descriptor-images.md
pbusko: > It's not clear if this is possible with multiple remote registries. If it is not, it may mean the --publish option is incompatible with having images on multiple registries.

Currently the lifecycle will throw an error during the analyze stage, if multiple registries are provided together with `--publish` flag.

Another option is to split the images blocks into two separate: main target and additional targets. The main target will be used as a previous image (similarly to what pack is doing by default)

Example:
```toml
[image.main]
name = "spring-petclinic"
registry = "gcr.io"
tag = "latest"

[[image.additional]]
registry = "private-corp:8080"
name = "spring-petclinic"
tags = ["v3", "v1"]
```

text/0000-project-descriptor-images.md
pbusko: I would vote for the Option 1.
Option 2 is hard to defend, if the user-provided explicit arguments to the `pack` are completely ignored.
Option 3 won't follow the currently established precedence strategy and will potentially provide unwanted results.

text/0000-project-descriptor-images.md
pbusko: According to the Docker Container Trust [documentation](https://docs.docker.com/engine/security/trust/#image-tags-and-dct), the Global Unique Name is composed as following:
`[REGISTRY_HOST[:REGISTRY_PORT]/]REPOSITORY[:TAG]`

This structure is also used by Notary, Cosign and google/go-containerregistry. In such case the answers are:

* registry is `index.docker.io` / `gcr.io`
* repository (name) is `library/bash` / `cncf-buildpacks-ci/demo/kpdemo`
* tag is `latest` / `v1`

The main confusion could be coming from the usage of `--tag` argument (`pack` refers to the GUN, instead of the last part after the `:`)

jromero: Thank you for those learnings. I can concede to the use of a GUN. 😉 

I suppose another reason for the reservation has to do with considering other images that go into a Buildpacks build project. I do wonder if we should look at this holistically. For example, `run` image isn't hard to envision being part of the project descriptor. How would that look like side by side? Why isn't the [builder](https://github.com/buildpacks/rfcs/pull/138) property decomposed?

[Here's](https://github.com/buildpacks/rfcs/blob/e9f36ba8718926edb62652ccc1767ebfcd26eef1/text/0000-buildpacks-config.md?plain=1#L122-L140) a _very_ different format on this proposed configuration. I wonder how that would look if certain image references had a "decomposed" structure.

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+[definitions]: #definitions
+
+* **project descriptor**: The `project.toml` file that describes a project.
+* **image**: the output of a running a buildpack(s) against a project
+* **image name**: Global Unique Name (GUN) of an OCI image in the form `[REGISTRY_HOST[:REGISTRY_PORT]/]REPOSITORY[:TAG]`.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, there is no support for specifying an output image in the `project.toml` although flags like `--tag` are already present.
```
hone, line 26: Is the use case to do a single build but want to be able to push to one or registries with the same image?

text/0000-project-descriptor-images.md
hone: We could also treat the first entry in the `[[images]]` table as the default.

text/0000-project-descriptor-images.md
hone: I'd like to see a user provided processes table, but definitely think that's out of scope for now.

If the images have different processes would that impact the way the image is built?

text/0000-project-descriptor-images.md
sclevine: I vote Option (4): build fails.

I think `pack build my-app` should mean "build my-app from the images table" when the images table exists.

hone: @jromero for Option 1, is `latest` image created because it's always created?

jkutner:
The following questions need to be resolved in order to finalize this RFC:

1. [What should the `pack` behavior be when mixing `--tag` option and `[images]` table?](https://github.com/buildpacks/rfcs/pull/187/files#r747590406)
   * Option 1: user-provided options override configuration.
   * Option 2: configuration overrides user-provided options
   * Option 3: A union of user-provided and config
   * Option 4: build fails
2. [How will this work if there are multiple images in the table and `pack` is running with `--publish`?](https://github.com/buildpacks/rfcs/pull/187/files#r747588966)
   * Option 1: introduce `[[image.additional]]`
   * Option 2: use the first image in the list as the default
   * Option 3: build fails

I'd like to propose that "build fails" for both outstanding questions is the best way to move this forward. It would more easily allow us to add new behavior in the future (like an order of precedence) because users will not rely on it to work in any way.

/cc @pbusko @hone @jromero @sclevine

loewenstein:
> [What should the `pack` behavior be when mixing `--tag` option and `[images]` table?](https://github.com/buildpacks/rfcs/pull/187/files#r747590406)
> * Option 1: user-provided options override configuration.
> * Option 2: configuration overrides user-provided options
> * Option 3: A union of user-provided and config
> * Option 4: build fails

What about the positional argument to `pack build <my-image>`? One use case for specifying the image in the `pack` call is tagging with the release version in CI. I don't think that we should or want to force users into modifying the `project.toml` in their release jobs. 

> 2. [How will this work if there are multiple images in the table and `pack` is running with `--publish`?](https://github.com/buildpacks/rfcs/pull/187/files#r747588966)
> * Option 1: introduce `[[image.additional]]`
> * Option 2: use the first image in the list as the default
> * Option 3: build fails

Option 3 basically means with multiple images in `project.toml` publishing of images has to be done separately, i.e. `pack build` followed by potentially multiple `docker push` calls, right?
Question: What happens if I call `pack build my-image:0.1 --tags my-image:latest --publish`?

loewenstein:
One more general remark. On a Twitter conversation involving @jkutner, a third person and myself, the topic of producing multiple distinct images in one go (think a Java and a Node app container image).
I guess such a request would not clash with this one, because we would anyway need to push basically all configuration one level down then, right?

jromero:
> > [What should the `pack` behavior be when mixing `--tag` option and `[images]` table?](https://github.com/buildpacks/rfcs/pull/187/files#r747590406)
> > 
> > * Option 1: user-provided options override configuration.
> > * Option 2: configuration overrides user-provided options
> > * Option 3: A union of user-provided and config
> > * Option 4: build fails
> 
> What about the positional argument to `pack build <my-image>`? One use case for specifying the image in the `pack` call is tagging with the release version in CI. I don't think that we should or want to force users into modifying the `project.toml` in their release jobs.
> 

I completely agree with your sentiment here. I think it should work out-of-the-box via option 1.

> > 2. [How will this work if there are multiple images in the table and `pack` is running with `--publish`?](https://github.com/buildpacks/rfcs/pull/187/files#r747588966)
> > 
> > 
> > * Option 1: introduce `[[image.additional]]`
> > * Option 2: use the first image in the list as the default
> > * Option 3: build fails
> 
> Option 3 basically means with multiple images in `project.toml` publishing of images has to be done separately, i.e. `pack build` followed by potentially multiple `docker push` calls, right? Question: What happens if I call `pack build my-image:0.1 --tags my-image:latest --publish`?

I think the main concern is [cross-registry publishing](https://github.com/buildpacks/lifecycle/blob/b6408639c0b0daada589b7b59a8eda2631bc4738/cmd/lifecycle/analyzer.go#L108-L110). I don't know that this concern is specific to an images table in the project descriptor. Multiple images using publish should work as long as they are on the same registry (as discussed [here](https://github.com/buildpacks/rfcs/pull/187#discussion_r763041695)).

Final thoughts, these open questions seem less relevant to the spec and more a part of the implementation. I'd be happy to resolve these questions as part of the implementation. 

My only outstanding concern is about the structure. I'm still not convinced about the image URI breakdown to components as  [discussed here](https://github.com/buildpacks/rfcs/pull/187#discussion_r746075079).

text/0000-project-descriptor-images.md
jromero: @jkutner, I conceded to the idea of GUN but still not convinced that we should be doing the component level breakdown in the configuration. I'd like to have more discussion around this since it would have implications _anywhere_ we reference an image.

text/0000-project-descriptor-images.md
jromero: In my mind, I'm struggling to answer this because it greatly depends on whether we're trying to map it to `pack` arguments. 

If we were, @pbusko [suggestion](https://github.com/buildpacks/rfcs/pull/187#discussion_r764051852) makes the most sense. If we were not, either of @jkutner's [suggestion](https://github.com/buildpacks/rfcs/pull/187#discussion_r763041695) or @hone's [suggestion](https://github.com/buildpacks/rfcs/pull/187#discussion_r765160071) seem ideal to me.

text/0000-project-descriptor-images.md
jromero: To elaborate a little, I would like to request that whatever decision be made here that it apply to all image references. This would include making the same decomposition to `builder` and any future references to images. 

In my opinion, the decomposition doesn't make sense everywhere such as `builder`, `run-image` (very likely future parameter). The cost/benefit just doesn't add up for me.

text/0000-project-descriptor-images.md
jromero: > @jromero for Option 1, is `latest` image created because it's always created?

`latest` was created because the user provided `my-app` (without a tag) as a positional argument. The default when no tag is provided is `latest`.

jromero: > I vote Option (4): build fails.
> 
> I think `pack build my-app` should mean "build my-app from the images table" when the images table exists.

I don't think this is ideal. @loewenstein gave a [good example of a use case](https://github.com/buildpacks/rfcs/pull/187#issuecomment-1006340692): 

> One use case for specifying the image in the pack call is tagging with the release version in CI. I don't think that we should or want to force users into modifying the project.toml in their release jobs.

hone:
Outstanding question is how we want serialize image references. @jromero and @jkutner to hammer out details here.

text/0000-project-descriptor-images.md
```diff
@@ -0,0 +1,159 @@
...
+tags = ["<string (optional, default=latest)>"]
+```
+
+- `registry` - a `REGISTRY_HOST[:REGISTRY_PORT]` component of the image name.
+- `repository` - a `REPOSITORY` component of the image name.
+- `tags` - a list of tags used as the `TAG` component of the image name.
+
+If no `registry` is provided, the default value MUST be `"docker.io"`.
+
+If no `repository` component is provided, the default value MUST be derived from the `io.buildpacks.name` field.
```
loewenstein, line 153:
```suggestion
If no `repository` component is provided, the default value MUST be derived from the `io.buildpacks.id` field.
```

I just realized, that `name` is supposed to be human readable and hence likely unfit to serve as image name.

loewenstein (commented):
The most immediate use case I see is to enable users to specify image and tag across platforms without specifying it multiple times with platform specific means like the positional arg or `--tags` for `pack`.

With this as part of the project.toml a user can call `pack build` (without further, image specific arguments) and the CI build will use the same.

The separation of registry, image and tag makes it easy for local dev, CI and release builds to target different registries (or path prefixes) and use different tags ( latest locally and idk Git SHA in CI)

nebhale (approved):

jromero (changes requested):
Posting a few statements about _my_view of the current status of this RFC. I think a few people may be a little confused, rightfully so, due to all the ongoing project descriptor RFCs and conversations.

Some outcomes from those conversations may help this RFC but none of them effectively blocked this RFC. For example, through the preparer RFC there was a reasonable number of positive consensus on [what type of properties should go in the io.buildpacks namespace](https://github.com/buildpacks/rfcs/pull/202/files#diff-da0c0db1eef841f956d231dff1f0ad819daf2ee72f55b71032073d431b767277R59):

> this namespace would be based on whether they affect inputs to the lifecycle phases

I think images fall under that category hopefully making it less debatable as to whether images belong in the namespace.

For me, the only unresolved conversations in this RFC would be the final format:

- [Decomposed format](https://github.com/buildpacks/rfcs/pull/187#discussion_r746075079) - I stand by my last comment [here](https://github.com/buildpacks/rfcs/pull/187#discussion_r786886927).
- [Previous image](https://github.com/buildpacks/rfcs/pull/187/files#r747588966)

jromero:
@pbusko have you had any chance to talk more in depth with individuals about that outstanding concerns in this RFC? I'd love to discuss how we can move this forward.

loewenstein:
I think from our end, as providers of a platform, what's still left is that having a single source of information for the name of an image produced between `pack` and other platforms would be nice, but the complexity of defining multiple images and what that means for caches and other things might not be justified.

jkutner:
I'm going to close this for now. But let's revisit after we finalize `pack.toml`. There may be opportunity to do something more specific to pack (and avoid some of the complications uncovered here)

PR #189, opened 2021-11-16
RFC: Support for pack.toml

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/jkutner/pack-toml/text/0000-pack-toml.md)

### DIFF
--- a/text/0000-pack-toml.md
+++ b/text/0000-pack-toml.md
@@ -36,7 +36,7 @@ We desire a descriptor file that is:
 The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
 * An app MAY have a `pack.toml`
 * An app MAY have a `project.toml`
-* An app MAY have both a `pack.toml` and a `projec.toml`, in which case they will be merged.
+* An app MAY have both a `pack.toml` and a `project.toml`, in which case the `project.toml` will be ignored.
 
 The `pack.toml` will support an additional `[io.buildpacks.pack]` table that is NOT defined in the [project descriptor spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md). The `[io.buildpacks.pack]` table will have the following schema:
 
--- a/text/0000-pack-toml.md
+++ b/text/0000-pack-toml.md
@@ -1,0 +1,123 @@
+# Meta
+[meta]: #meta
+- Name: [Joe Kutner](http://github.com/jkutner)
+- Start Date: 2021-11-16
+- Author(s): @jkutner
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This is a proposal for a `pack.toml` file that is a special case of the `project.toml` descriptor file. The `pack.toml` will include configuration options specific to the Pack CLI, which may not be honored by other platforms.
+
+# Definitions
+[definitions]: #definitions
+
+- *project descriptor* - a extension spec defining the [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) file.
+
+
+# Motivation
+[motivation]: #motivation
+
+The `project.toml` file has been a useful and essential tool for buildpack users who need to codify certain arguments of their builds. However, there has been much debate around how to handle configuration options that may not be supported by all platform.
+
+We desire a descriptor file that is:
+* specific to pack (i.e. not platform-neutral)
+* advertises that the repo containing the file can be built with pack
+* adheres to the best practices and learning of [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
+
+# What it is
+[what-it-is]: #what-it-is
+
+The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
+* An app MAY have a `pack.toml`
+* An app MAY have a `project.toml`
+* An app MAY have both a `pack.toml` and a `projec.toml`, in which case they will be merged.
+
+The `pack.toml` will support an additional `[io.buildpacks.pack]` that is not defined in the [project descriptor spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md). The `[io.buildpacks.pack]` table will have the following schema:
+
+```toml
+[io.buildpacks.pack]
+builder = "<string (optional)>"
+```
+
+Where:
+
+* `builder` is a uri to a CNB Builder docker image
+
+## Example
+
+```
+[io.buildpacks.build]
+exclude = ["/README.md"]
+
+[[io.buildpacks.group]]
+id = "example/java"
+version = "1.0.0"
+
+[io.buildpacks.pack]
+builder = "example/my-builder-image"
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Pack CLI will honor both the `pack.toml` and `project.toml` file descriptors with the following rules:
+
+* If one or more descriptors are provided on the command line they will all be merged and used.
+* If a `pack.toml` file exists in the workspace and no descriptors are provided on the command line, it will be used.
+* If a `project.toml` file exists in the workspace and no descriptors are provided on the command line, it will be merged with any `pack.toml` that is present in the workspace and the result will be used.
+* Else no project descriptor will be used
+
+When multiple descriptors are _merged_ the following precedence is given:
+
+* Configuration in a file provided on the command line will take precedence over files provided on the command line after it.
+* Configuration in a `pack.toml` will take precedence over configuration in a `project.toml`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Using a `pack.toml` and `project.toml` descriptor together in a `pack` build may be confusing. The rules of merging two files may not be obvious.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Support `pack.toml` as a subset of `project.toml`
+
+* If a path to a `pack.toml` is provided on the command line it will be used.
+* Else if a `pack.toml` file exists in the workspace, it will be used.
+* Else no project descriptor will be used
+
+At the same time:
+
+* If a path to a `project.toml` is provided on the command line it will be used.
+* Else if a `project.toml` file exists in the workspace, it will be used.
+* Else no project descriptor will be used
+
+## Support `pack.toml` and `project.toml` as mutually exclusive descriptors
+
+* If a descriptor is provided on the command line it will be used.
+* Else If a `pack.toml` file exists in the workspace, it will be used. Only configuration in the `[io.buildpacks]` table and sub-tables will be honored. The root `[_]` will be optional.
+* Else if a `project.toml` file exists in the workspace, it will be used. Any configuration in the `[io.buildpacks.pack]` table will be ignored.
+* Else no project descriptor will be used
+
+# Prior Art
+[prior-art]: #prior-art
+
+## Terraform Manifests
+
+A repository may contain multiple [Terraform](https://www.terraform.io/) manifest, in which case the resources in those files will be merge into a single plan.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+N/A
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+N/A
--- a/text/0000-pack-toml.md
+++ b/text/0000-pack-toml.md
@@ -1,8 +1,8 @@
 # Meta
 [meta]: #meta
-- Name: [Joe Kutner](http://github.com/jkutner)
+- Name: Support for `pack.toml`
 - Start Date: 2021-11-16
-- Author(s): @jkutner
+- Author(s): [jkutner](http://github.com/jkutner)
 - Status: Draft
 - RFC Pull Request: (leave blank)
 - CNB Pull Request: (leave blank)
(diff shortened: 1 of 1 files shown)


loewenstein:
I understand that e.g. builder might be difficult to impossible to support. On the other hand side, introducing this special handling for the platform maintained as part of CNB core feels a bit like opening the box of Pandora.

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+* advertises that the repo containing the file can be built with pack
+* adheres to the best practices and learnings from [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
+
+# What it is
+[what-it-is]: #what-it-is
+
+The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
+* An app MAY have a `pack.toml`
+* An app MAY have a `project.toml`
+* An app MAY have both a `pack.toml` and a `projec.toml`, in which case they will be merged.
```
ekcasey, line 39:
```suggestion
* An app MAY have both a `pack.toml` and a `project.toml`, in which case they will be merged.
```

ekcasey, line 39: What is the use case for having both? Wasn't the purpose of https://github.com/buildpacks/rfcs/blob/main/text/0084-project-descriptor-domains.md to allow for different platform-specific fields to exist in a single master file?

I can't help but feeling like the existence of this proposal is a sign that we need to stop and reevaluate `project.toml`. Clearly being to commit config for `pack` is valuable. But requiring all platforms to conform to the same interface isn't working, even for the simplest platforms like pack.

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The `project.toml` file has been a useful and essential tool for buildpack users who need to codify certain arguments of their builds. However, there has been much debate around how to handle configuration options that may not be supported by all platforms.
+
+We desire a descriptor file that is:
+* specific to pack (i.e. not platform-neutral)
+* advertises that the repo containing the file can be built with pack
+* adheres to the best practices and learnings from [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
```
ekcasey, line 31: What do you consider to be the best practices or learnings from `project.toml`?

text/0000-pack-toml.md
loewenstein: I'd be interested in far it seems impossible to get to a common descriptor that all platforms can adhere to? Actually, I'd expect that at least a lot of those could even be provided by a shared implementation. Isn't that what the RFC discussion about a `prepare` lifecycle is about?

If we need platform specific configuration, and `builder` might be such a case as platforms like kpack or project Piper's `cnbBuild` won't read the `project.toml` until well after running in a process on a builder image, platform specific or optionally interpreted keys in a common file sound far superior to separate files that need merging.

hone:
@jkutner do you mind if I move this to "blocked" on #191?

loewenstein:
Is #191 the same thing that @jromero is currently drafting in https://hackmd.io/4LGBHOZAQ6GMOTeDf2U3_w?view#What-it-is? If yes, the existence of project.toml namespaces like io.buildpacks.pack.* leads me to the question if that would supersede the proposal for a pack specific configuration file...?

If I am right in this, then [section about merging pack specific config](https://hackmd.io/4LGBHOZAQ6GMOTeDf2U3_w?view#Overlay-Pack-specific-config) seems to build a circle with this rfc that might be worth resolving ;)

Thoughts? @jromero @jkutner

jkutner:
@loewenstein I think this RFC started the conversation that led to @jromero's hackmd draft RFC. I'm supportive of that one, and will likely close this once we turn his into a real RFC

loewenstein:
> @loewenstein I think this RFC started the conversation that led to @jromero's hackmd draft RFC. I'm supportive of that one, and will likely close this once we turn his into a real RFC

I was asking because the proposal of @jromero seems to try to accommodate this one ;)

jromero:
@jkutner, I submitted the RFC for a [Prepare Operation](https://github.com/buildpacks/rfcs/pull/202) and excluded the section referring to the `pack.toml` specifically in order to keep the RFC as concise as possible. I'd like to propose that we continue to make changes to this RFC separately with the other RFC in mind.

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
+* An app MAY have a `pack.toml`
+* An app MAY have a `project.toml`
+* An app MAY have both a `pack.toml` and a `projec.toml`, in which case they will be merged.
+
+The `pack.toml` will support an additional `[io.buildpacks.pack]` table that is NOT defined in the [project descriptor spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md). The `[io.buildpacks.pack]` table will have the following schema:
```
jromero, line 41: @jkutner IIRC, we discussed minimizing this by now using reverse name structure but instead having everything in a top-level `[pack]` dictionary. Is that your understanding as well?

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+[io.buildpacks.pack]
+builder = "example/my-builder-image"
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Pack CLI will honor both the `pack.toml` and `project.toml` file descriptors with the following rules:
+
+* If one or more descriptors are provided on the command line they will all be merged and used.
```
jromero, line 71: I wonder if this would be simpler if we only had one flag, say something like `--config` that could accept either `pack.toml` OR `project.toml`.

The same could then apply to an inferred discovery mechanism where if the `--config` flag is not presented then it's looked up by precedence and NOT merged.

1. Look up `pack.toml`
2. Look up `project.toml`

text/0000-pack-toml.md
jromero, line 39: The more I think about it, the more I feel like we shouldn't merge the contents. It's hard to implement things like "removals", "resets", and "append" vs "override". Additionally, by not doing a merging them I feel like it would also be easier for users to comprehend how they are applied. For example, "If `pack.toml` is present it uses that; otherwise it looks for a `project.toml`."

jromero (changes requested):

jromero:
@jkutner have you had a chance to review my latest comments?

@dfreilich your feedback would also be appreciated here since this seems rather impacting.

loewenstein:
Has there been any decision with regards to additional file vs. some flexibility and additional tables in the existing project descriptor file?

jkutner:
@loewenstein no decision, but I'm reopening this PR for discussion. My hope is that this will allow us to introduce pack specific attributes in the `pack.toml` (like for multiple images) that would not necessarily be honored by all platforms.  Is that something that might work for you?

loewenstein:
> @loewenstein no decision, but I'm reopening this PR for discussion. My hope is that this will allow us to introduce pack specific attributes in the `pack.toml` (like for multiple images) that would not necessarily be honored by all platforms. Is that something that might work for you?

I guess I'd still prefer to have "project" related configuration in a single file, i.e. *the* `project.yaml` file. With the sole exception of maybe entirely only `pack` concerning config, but I think `pack` has config for e.g. default builders already in the user home.

jkutner:
@loewenstein IIUC that matches the intent of this proposal. Any general project configuration can go in `project.toml` (including pack config if you want it there). Pack-only config can go in `pack.toml`. What's the gap from what you described?

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
+* An app MAY have a `pack.toml`
+* An app MAY have a `project.toml`
+* An app MAY have both a `pack.toml` and a `project.toml`, in which case the `project.toml` will be ignored.
+
+The `pack.toml` will support an additional `[io.buildpacks.pack]` table that is NOT defined in the [project descriptor spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md). The `[io.buildpacks.pack]` table will have the following schema:
+
+```toml
+[io.buildpacks.pack]
+builder = "<string (optional)>"
```
joshwlewis, line 45: `io.buildpacks.builder` already seems to be supported in `project.toml` here: https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuilder-optional. Is this different somehow?

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+- *project descriptor* - a extension spec defining the [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) file.
+
+
+# Motivation
+[motivation]: #motivation
+
+The `project.toml` file has been a useful and essential tool for buildpack users who need to codify certain arguments of their builds. However, there has been much debate around how to handle configuration options that may not be supported by all platforms.
+
+We desire a descriptor file that is:
+* specific to pack (i.e. not platform-neutral)
```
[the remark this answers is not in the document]
jkutner: Two things:
* ability to support pack specific configuration w/o putting it in the project.toml spec
* a file that advertises that a repo can be used with `pack` (like a `Dockerfile` does for docker)

text/0000-pack-toml.md
jkutner: I think that may have been added since I first wrote this. I'll update with a different example that shows something that's specific to pack (like `tag` or maybe `env-file`)

hone:
Moving this into FCP for next Thurs Dec 8.

hone (approved):

hone:
@jkutner is the goal to move the pack configuration into `pack.toml`?

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+* advertises that the repo containing the file can be built with pack
+* adheres to the best practices and learnings from [`project.toml`](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
+
+# What it is
+[what-it-is]: #what-it-is
+
+The Pack CLI will support a new descriptor file, `pack.toml`, that is a superset of the project descriptor.
+* An app MAY have a `pack.toml`
+* An app MAY have a `project.toml`
+* An app MAY have both a `pack.toml` and a `project.toml`, in which case the `project.toml` will be ignored.
```
loewenstein, line 39: This contradicts https://github.com/buildpacks/rfcs/pull/189/files#diff-9110fb10fa3ce5f9676262392e4b1cb227092e9df99be33edab24508dd1a81a7R73

loewenstein:
> @loewenstein IIUC that matches the intent of this proposal. Any general project configuration can go in `project.toml` (including pack config if you want it there). Pack-only config can go in `pack.toml`. What's the gap from what you described?

I guess my main concern is with things that are maybe `pack` only today, but will be picked up by other platforms in the future. From my point of view, it would make sense for most of the configuration you would possible want to support to go through the project descriptor spec and potentially align support across platforms.

jjbustamante:
I really like the motivation: **advertises that the repo containing the file can be built with pack**

Now, I have the following doubt:  Is there an idea to accomplish that? I mean, for example, if we think of developers that use vscode and they want some [extension](https://marketplace.visualstudio.com/search?term=docker&target=VSCode&category=All%20categories&sortBy=Relevance) for Docker, then they have a lot of options. Are we considering creating or maintaining our own IDEs extensions for CNB tooling?

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+
+* If one or more descriptors are provided on the command line they will all be merged and used.
+* If a `pack.toml` file exists in the workspace and no descriptors are provided on the command line, it will be used.
+* If a `project.toml` file exists in the workspace and no descriptors are provided on the command line, it will be merged with any `pack.toml` that is present in the workspace and the result will be used.
+* Else no project descriptor will be used
+
+When multiple descriptors are _merged_ the following precedence is given:
+
+* Configuration in a file provided on the command line will take precedence over files provided on the command line after it.
+* Configuration in a `pack.toml` will take precedence over configuration in a `project.toml`
```
schneems, line 79: I would like some outputs from `pack` to indicate the config being used (even better if it could trace it back to the origin and say why). That way users can know at a glance if their changes have been picked up or if they're not being picked up, get a hint that maybe another file is taking precedence. 

I'm unsure whether that desire can be encoded in the RFC in a sensible way.

text/0000-pack-toml.md
hone, line 39:
```suggestion
* An app MAY have both a `pack.toml` and a `project.toml` where they will be merged and used.
```

samj1912 (changes requested):
I still see multiple project descriptor related RFCs open. Can we consolidate them and figure out how each fits in the larger picture? I don't want there to be multiple ways of configuration as that just leads to complicated code and lots of corner cases. If we do decide to implement multiple related RFCs, I would also like to see how they interplay with each other and what precedence would look like amongst them.

text/0000-pack-toml.md
jkutner: @hone we've gone back and forth on this. Merging could actually be very complicated, and if we support merging we should only do it at the top level shared key (i.e. we're not going to try and merge lists of things for example)

loewenstein: @jkutner this is precisely why I have a problem with a platform specific descriptor covering platform agnostic configuration.

I would want to have as much as possible in an agnostic way, at least from the perspective of all platforms other than `pack`. Why? Because for 99% of the non `pack` platforms I can imagine, `pack` will be the platform for local execution. See kpack and our integration into project Piper as two examples and Tekton most likely as a third.

Also, as I pointed out in https://github.com/buildpacks/rfcs/pull/189#discussion_r1038109057 the “`project.toml` will be ignored” statement contradicts other parts of this proposal.

jkutner: fixed

text/0000-pack-toml.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The Pack CLI will honor both the `pack.toml` and `project.toml` file descriptors with the following rules:
+
+* If one or more descriptors are provided on the command line they will all be merged and used.
+* If a `pack.toml` file exists in the workspace and no descriptors are provided on the command line, it will be used.
+* If a `project.toml` file exists in the workspace and no descriptors are provided on the command line, it will be merged with any `pack.toml` that is present in the workspace and the result will be used.
```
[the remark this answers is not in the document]
loewenstein: I guess I have expressed multiple times, that I don’t like this resolution of the inconsistency in the RFC ;)

jkutner:
I'm going to close this RFC again. There is enough about merging pack.toml and project.toml that isn't clear, and there isn't enough interest in using pack.toml (even from me) to make me move it forward.

PR #190, opened 2021-11-18
RFC: Replace positional args to Buildpack executables with env vars

jkutner:
[Readable](https://github.com/buildpacks/rfcs/blob/jkutner/input-vars/text/0000-buildpack-input-vars.md)


samj1912 (approved):
I like it - might still be worth passing in the CLI args for b/w compat but happy with either.

natalieparellano (approved):

jkutner:
@samj1912 it's kind of buried, but I'm proposing something close to that:

> The positional arguments will be deprecated, but no warnings will be emitted if they are consumed. The lifecycle will continue to provide them to buildpack executables indefinitely, with no plan to remove them.

samj1912:
I feel like this is probably going to be the quickest rfc we have put in then XD

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,142 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+We will deprecate the positional arguments to `bin/detect` and `bin/build` with the following environment variables:
+
+### bin/detect
+
+* `CNB_PLATFORM_DIR` - replaces the first positional argument to `bin/build`. Uses the same env var name as the Platform spec.
+* `CNB_PLAN_PATH` - replaces the second positional argument to `bin/build`. Uses the same env var name as the Platform spec.
```
hone, line 42: as far as naming goes, do we use `DIR` for directories and `PATH` for files? Just wondering if we should take the time to standardize here (if needed).

jkutner: I believe @sclevine was pretty consistent with this in other places in the spec, which is why I adopted it

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,142 @@
...
+
+### bin/build
+
+* `CNB_BP_LAYERS_DIR` - replaces the first positional argument to `bin/build`. Uses `_BP_` to differentiate it from the `CNB_LAYERS_DIR` in the Plaform spec, which is a different value.
```
sclevine, line 46: I don't like that we're sacrificing a common env var name in the buildpack spec for a rarely-used env var in the platform spec. How do we feel about renaming the platform one?

Otherwise, buildpack authors will wonder why it's not `CNB_BP_PLAN_PATH`, etc.

sclevine (changes requested):

text/0000-buildpack-input-vars.md
hone: Does this require buildpack/platform APIs to change in lock step are we ok with the overlap?

natalieparellano: The lifecycle should be able to handle the overlapping case, so I think it's up to the team's discretion.

nebhale (approved):

text/0000-buildpack-input-vars.md
ekcasey, line 42: This file is a buildpack specific [build plan](https://github.com/buildpacks/spec/blob/main/buildpack.md#build-plan-toml). which is different from the [plan.toml](https://github.com/buildpacks/spec/blob/main/platform.md#plantoml-toml) which is referred to with `CNB_PLAN_PATH` in the platform spec.

This build plan is also different from the [buildpack plan](https://github.com/buildpacks/spec/blob/main/platform.md) we are using `CNB_PLAN_PATH` to reference in the builld interface below.

Using the same name for three different files with three different schemas might be confusing.

Maybe instead we could go with `CNB_BUILD_PLAN_PATH` and `CNB_BP_PLAN_PATH`? Or a better idea if anyone has one?

text/0000-buildpack-input-vars.md
ekcasey: > Does this require buildpack/platform APIs to change in lock step are we ok with the overlap?

Even if you release them both simultaneously most people will experience an overlap anyways as new buildpacks will run on old platforms.

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,142 @@
...
+
+* `/bin/detect <platform[AR]> <plan[E]>`
+* `/bin/build <layers[EIC]> <platform[AR]> <plan[ER]>`
+
+Using env vars helps make these inputs more easily accesible from language bindings like [libcnb.bash](https://github.com/jkutner/libcnb.bash).
+
+# What it is
+[what-it-is]: #what-it-is
+
+We will deprecate the positional arguments to `bin/detect` and `bin/build` with the following environment variables:
```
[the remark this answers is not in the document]
hone: While I think that's true, this is a pretty big shift change. How many of these kind of changes do we want to make?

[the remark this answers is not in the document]
ekcasey: If we want to make this change I think there are two options are
1. **option 1 - no deprecation** requires buildpack authors to migrate to the new usage when they adopt buildpack API 0.10 (or whatever API it lands in)
2. **option 2 - deprecation** requires buildpack authors to migrate to the new usage when they adopt buildpack API 0.17 (or whatever API it is removed in)

The way I see it the amount of total pain is the same in both cases and delaying the migration until we remove a bunch of deprecated features all at once, isn't necessarily better for buildpack authors. I would feel differently if users had to migrate to upgrade the lifecycle, but they can run the latest lifecycle and stay on the current API as long as they like (until we deprecate and remove the API).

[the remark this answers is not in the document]
jkutner: With any breaking change, I think we need to compare the burden of supporting backwards compat with the burden we put on our users/community to migrate. In this case, the burden to migrate can be very high (I would compare it to the addition of the `[types]` table in layer toml). At the same time, I don't see how continuing to support the env vars adds "complexity".

We only get to make so many breaking changes before people stop upgrading, or leave the community. I think #172 consumes our budget for that tolerance.

text/0000-buildpack-input-vars.md
jkutner: @sclevine any preference on the new platform env var name?

text/0000-buildpack-input-vars.md
jkutner: `CNB_BUILD_PLAN_PATH` seems good.

jkutner, line 42:
```suggestion
* `CNB_BUILD_PLAN_PATH` - replaces the second positional argument to `bin/build`. Uses the same env var name as the Platform spec.
```

text/0000-buildpack-input-vars.md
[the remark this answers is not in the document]
jkutner: @ekcasey i'm not proposing we remove the positional args at all

text/0000-buildpack-input-vars.md
jkutner: done

jkutner:
@sclevine please re-review

text/0000-buildpack-input-vars.md
sclevine: `CNB_PARENT_LAYERS_DIR`

sclevine (approved):
Approve conditional on platform env var changing to `CNB_PARENT_LAYERS_DIR`

hone (approved):

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,146 @@
...
+
+### bin/detect
+
+* `CNB_PLATFORM_DIR` - replaces the first positional argument to `bin/build`. Uses the same env var name as the Platform spec.
+* `CNB_BUILD_PLAN_PATH` - replaces the second positional argument to `bin/build`. Uses the same env var name as the Platform spec.
+
+### bin/build
+
+* `CNB_LAYERS_DIR` - replaces the first positional argument to `bin/build`. **Note:** `CNB_LAYERS_DIR` conflicts with the platform spec, which will beupdated to rename its `CNB_LAYERS_DIR` to `CNB_PARENT_LAYERS_DIR`.
```
hone, line 46:
```suggestion
* `CNB_LAYERS_DIR` - replaces the first positional argument to `bin/build`. **Note:** `CNB_LAYERS_DIR` conflicts with the platform spec, which will be updated to rename its `CNB_LAYERS_DIR` to `CNB_PARENT_LAYERS_DIR`.
```

hone:
Best on @ekcasey's comments on how we should be moving things into FCP with the votes, I'm going to move this into FCP.

jkutner:
A grievance has been raise about the env var names, their conflicts with existing env var names, and the complexity associated with renaming them. Here are some options we may consider:

## Option 1

Use a new prefix, `CNB_BP_` to disambiguate from any existing vars.

* `CNB_BP_BUILD_PLAN_PATH`
* `CNB_BP_LAYERS_DIR`
* `CNB_BP_PLATFORM_DIR `
* `CNB_BP_PLAN_PATH `

Drawback: doesn't match `CNB_BUILDPACKS_DIR` (we could also move to CNB_BP_ROOT_DIR` or simliar).

## Option 2

Just reuse the var names. It's a different context and people will know the difference

* `CNB_BUILD_PLAN_PATH` (unique)
* `CNB_LAYERS_DIR` (used in platform spec)
* `CNB_PLATFORM_DIR ` (unique)
* `CNB_PLAN_PATH ` (used in platform spec)

## Option 3

Use slightly less obvious/intuitive names that are unambiguous

* `CNB_BUILD_PLAN_PATH` (unique)
* `CNB_BUILD_LAYERS_DIR`
* `CNB_PLATFORM_DIR ` (unique)
* `CNB_BP_PLAN_PATH `

Drawback: this might make it difficult if we ever want to introduce new vars (we'd have to avoid collision every time)

ekcasey:
@jkutner Thank you kindly. I would be happy moving forward with any of these options. I think my preference order is 2, 1, 3, but I am happy to defer to others (looking at you @sclevine) who might strongly prefer one to the other.

Option 2b: Just like option 2 but we could still use `CNB_BP_PLAN_PATH` instead of `CNB_PLAN_PATH` b/c the spec calls this file the "Buildpack Plan".

jkutner:
I suppose an option 4 (building on @ekcasey's 2b) is:

* `CNB_BUILD_PLAN_PATH`
* `CNB_BP_LAYERS_DIR`
* `CNB_PLATFORM_DIR`
* `CNB_BP_PLAN_PATH`

All of these are unique, but how long can we get away with that without a special prefix? Is now the the time to assert `CNB_BP_` is the prefix for all buildpack API inputs (in an effort to future-proof)?

jkutner:
In the almost four years of the project, we've only added one new "input" (`CNB_BUILDPACK_DIR`). For that reason, I favornot worrying about a special prefix

jkutner:
Myself, @sclevine, and @ekcasey have agreed to allow the env var names to overlap with existing env var names in the platform spec, but we will update this RFC to include a statement about changing the platform env var names in aa larger platform spec refactoring.

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,149 @@
...
+Using env vars helps make these inputs more easily accesible from language bindings like [libcnb.bash](https://github.com/jkutner/libcnb.bash).
+
+# What it is
+[what-it-is]: #what-it-is
+
+We will deprecate the positional arguments to `bin/detect` and `bin/build` with the following environment variables:
+
+### bin/detect
+
+* `CNB_PLATFORM_DIR` - replaces the first positional argument to `bin/build`. Uses the same env var name as the Platform spec.
```
jabrown85, line 41: /queue-issue buildpacks/lifecycle "Add CNB_PLATFORM_DIR and CNB_BUILD_PLAN_PATH to bin/detect"

text/0000-buildpack-input-vars.md
```diff
@@ -0,0 +1,149 @@
...
+* `CNB_BUILD_PLAN_PATH` - replaces the second positional argument to `bin/build`. Uses the same env var name as the Platform spec.
+
+### bin/build
+
+* `CNB_LAYERS_DIR` - replaces the first positional argument to `bin/build`. **Note:** Uses the same env var name as the Platform spec, but refers to a different location.
```
jabrown85, line 46: /queue-issue buildpacks/lifecycle "Add CNB_LAYERS_DIR, CNB_PLATFORM_DIR and CNB_BP_PLAN_PATH to bin/build"

PR #191, opened 2021-11-18
RFC: Buildpacks config

jromero:
An alternative to project descriptor that is solely focused on Cloud Native Buildpacks.

### DIFF
--- a/text/0000-buildpacks-config.md
+++ b/text/0000-buildpacks-config.md
@@ -1,0 +1,208 @@
+# Meta
+[meta]: #meta
+- Name: (fill in the feature name: My Feature)
+- Start Date: (fill in today's date: YYYY-MM-DD)
+- Author(s): (Github usernames)
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes:
+    - [RFC#19](https://github.com/buildpacks/rfcs/blob/main/text/0019-project-descriptor.md)
+    - [RFC#54](https://github.com/buildpacks/rfcs/blob/main/text/0054-project-descriptor-schema.md)
+    - [RFC#54](https://github.com/buildpacks/rfcs/blob/main/text/0054-project-descriptor-schema.md)
+    - [RFC#80](https://github.com/buildpacks/rfcs/blob/main/text/0080-builder-key-in-project-descriptor.md)
+    - [RFC#84](https://github.com/buildpacks/rfcs/blob/main/text/0084-project-descriptor-domains.md)
+
+# Summary
+[summary]: #summary
+
+A configuration file intended for end-users to influence how a buildpack build occurs.
+
+# Definitions
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+- What use cases does it support?
+- What is the expected outcome?
+
+# What it is
+[what-it-is]: #what-it-is
+<!--
+This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+-->
+
+`./.buildpacks.<ext>`
+
+### Supported Formats
+
+Proposed supported extensions:
+
+ - `toml` => TOML
+ - `yaml`, `yml` => YAML
+
+> If multiple files exist, `toml` would be selected.
+
+#### TOML
+
+```toml
+[images]
+builder="cnbs/sample:builder"
+names=[
+    "cnbs/sample-app",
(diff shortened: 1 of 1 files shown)


text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,219 @@
...
+This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+-->
+
+`./.buildpacks.<ext>`
```
jkutner, line 46: I'm afraid that using `buildpacks` for something that is `pack`-specific will be confusing when it doesn't work on other buildpack platforms.

text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,219 @@
...
+### Supported Formats
+
+Proposed supported extensions:
+
+ - `toml` => TOML
+ - `yaml`, `yml` => YAML
+
+> If multiple files exist, `toml` would be selected.
+
+#### TOML
```
jkutner, line 57: For a dot-file, i would expect this file structure to map directly to the `pack` commands. So a table for each command, with 

```toml
[build]
builder = ""
clear-cache = true

[[buildpacks]]
id = "
```

and so on...

text/0000-buildpacks-config.md
jromero: The intent of this file is to be platform agnostic, in other words NOT pack specific.

text/0000-buildpacks-config.md
jromero: I thought about the `build` table but given that all of the properties only apply to build I didn't see the need to add a layer of depth that the users need to repeatedly add.

text/0000-buildpacks-config.md
jromero: I added a lot more details. Give it another read and see if it's any clearer where I'm trying to go with this proposal.

text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Configuration Properties
+
+To prevent unnecessary complexity for end-users (app developers), the `buildpacks.<ext>` file would only support properties associated with Cloud Native Buildpacks. See full list of properties in the [file](#file) section.
+
+### Cross-Platform Support
```
jkutner, line 254: This doesn't appear to improve on the status quo (`project.toml`). The `buildpacks.*` would. need to be converted into another format to become cross-platform supportable.

text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,314 @@
...
+[issue-tekton-33]: https://github.com/buildpacks/tekton-integration/issues/33
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposed solution to the issues listed in the [motivation](#motivation) is to create a **Cloud Native Buildpacks specific file** that **MUST be supported by platforms**.
+
+## File
+
+`buildpacks.<ext>` where `<ext>` corrolates to a supported format.
```
jkutner, line 104: this is way too easily confused with `buildpack.toml`

text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,314 @@
...
+By supporting multiple file formats, we lower the barrier to entry and allow for the flexibility desired by some app developers.
+
+Proposed supported extensions (in order of precedence):
+
+ - `toml` => TOML
+ - `yaml`, `yml` => YAML
+
+#### TOML
+
+```toml
```
jkutner, line 119: this doesn't look any less complex to me than `project.toml`

text/0000-buildpacks-config.md
```diff
@@ -0,0 +1,314 @@
...
+                                       Platform API
+```
+
+
+[rfc-pr-182]: https://github.com/buildpacks/rfcs/pull/182
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
```
jkutner, line 289: - `buildpacks.*` is easily confused with `buildpack.toml`
- `buildpacks*` is not inherently portable. We still need to implement a mechanism and more spec to convert it to something portable
- yet-another-config-file: I won't have a place to put config that's common across tools (license, include/exclude, version, etc)
- doesn't map well to pack commands/flags

text/0000-buildpacks-config.md
jromero: I agree but do we expect app developers to be using `buildpack.toml`? Alternatively, maybe something like `Buildpacks`, with no extension, similar to `Dockerfile`.

text/0000-buildpacks-config.md
jromero: It does in that Platforms become required to support `buildpacks.*`. That said, we could do the same to `project.toml`. This RFC tries to paint a _new_ picture on a CNB specific config file that becomes required for platforms to support.

jkutner: what are the advantages of requiring this file over `project.toml`?

text/0000-buildpacks-config.md
jkutner: > I agree but do we expect app developers to be using buildpack.toml?

If we don't, I think that makes the problem worse, because they won't even notice when they need to pay attention to the differences.

hone:
Should we close this?

PR #194, opened 2021-12-02
Add image history metadata

samj1912:

[Readable](https://github.com/samj1912/rfcs/blob/history/text/0000-history.md)

### DIFF
--- a/text/0000-history.md
+++ b/text/0000-history.md
@@ -1,0 +1,85 @@
+# Meta
+[meta]: #meta
+- Name: Buildpack layer metadata
+- Start Date: 2021-12-02
+- Author(s): [@samj1912](https://github.com/samj1912)
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Add layer history metadata to the output image for buildpack and app layers to improve visualization in tools like `DockerHub`, `dive` etc.
+
+# Definitions
+[definitions]: #definitions
+
+`history`: The history key in the `OCI` config. See [properties](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)
+
+# Motivation
+[motivation]: #motivation
+
+This will allow Buildpack built images to be better visualized by common container visualization and debugging tools.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Various container image introspection tools look at `history.created_by` for each layer to visualize the layer. We should populate this key with a value that describes where a layer came from, for buildpacks this can be -
+
+- `Layer: {{ buildpack.layer.name }}, Created by: {{ buildpack.id }}@{{ buildpack.version }}`
+
+for app layers, this can be - 
+
+- `Application Slice: {{ slice_number }} Created by: {{ buildpack.id }}@{{ buildpack.version }}`
+
+Where `slice_number` is just an integer number starting with `1` and the `buildpack.id` is the id of the buildpack that specified the application slice.
+
+for config layer, this can be -
+
+- `Buildpacks Launcher Config`
+
+for SBOM layer, this can be - 
+
+- `Software Bill-of-Materials`
+
+for launcher this can be - 
+
+- `Buildpacks Application Launcher`
+
+for the process types layer this can be -
+
+- `Buildpacks Process Types`
+
+For the base image, the lifecycle should copy the existing history to the output image.
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle adds the above metadata to the output config blob.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More complexity?
+
+# Alternatives
+[alternatives]: #alternatives
+
+Allow buildpacks to create image history metadata.
+
+# Prior Art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+N/A
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+Noted above.


jabrown85 (approved):

text/0000-history.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This will allow Buildpack built images to be better visualized by common container visualization and debuggin tools.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The following keys will be added to the following files -
```
natalieparellano, line 30: IIRC this would require both a buildpack and platform api bump. Is there a way to decouple the two? Can the lifecycle choose a "default" name if none is specified?

text/0000-history.md
```diff
@@ -0,0 +1,83 @@
...
+```
+[[slices]]
+# name is a new key which uniquely identitifies the app slice
+# required key
+name = "<name-of-the-slice>" # max 255 chars
+```
+
+All these tools look at `history.created_by` for each layer to visualize the layer. We should populate this key with a value that describes where a layer came from, for buildpacks this can be -
+
+- `{{ buildpack.id }}: Buildpack: {{ buildpack.name }} Layer: {{ buildpack.layer.name }}`
```
natalieparellano, line 43: `Layer: {{ buildpack.layer.name }}, Created by: {{ buildpack.id }}`

natalieparellano: The lifecycle stringifies buildpacks as "Name Version" and sometimes "ID@Version" ...those could be other options.

text/0000-history.md
```diff
@@ -0,0 +1,83 @@
...
+for app layers, this can be - 
+
+- `Application Slice: Name {{ slices.name }}, Created by: {{ buildpack.id }}`
+
+Where `slice.name` comes from `slices.name` and the `buildpack.id` is the id of the buildpack that specified the application slice.
+
+For the leftover, generic app layer, the `history.created_by` will be set to `Application Workspace`.
+
+The above application slice name should also be added to `io.buildpacks.lifecycle.metadata.app` struct.
+
```
ekcasey, line 54: I would love to see suggestions for the history fields for the remaining layers
* config
* launcher
* process types
* sbom

text/0000-history.md
```diff
@@ -0,0 +1,83 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle adds the above metadata to the output config blob.
+
```
ekcasey, line 59: Right now we drop the run image history on the floor and add an one empty history entry per layer. I would love to see us preserve the run image history as part of this proposal. However, we will need to add a field to image metadata denoting the end of the run-image history (similar to run-image top-layer) in order to rebase the history correctly.

text/0000-history.md
```diff
@@ -0,0 +1,83 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More complexity?
+
+# Alternatives
+[alternatives]: #alternatives
```
ekcasey, line 66: Not saying I prefer this alternative but for the `created_by` section for buildpack-provided layers we could let the buildpack set this value, e.g. with a new field in `layer.toml`

jabrown85 (approved):

text/0000-history.md
```diff
@@ -0,0 +1,81 @@
...
+
+- `Buildpack Launcher Config`
+
+for SBOM layer, this can be - 
+
+- `Software Bill-of-Materials`
+
+for launcher this can be - 
+
+- `Buildpacks Application Launcher`
```
natalieparellano, line 50:
```suggestion
- `Buildpacks Application Launcher`

for the process types layer this can be -

- `Buildpack Process Types`
```

text/0000-history.md
natalieparellano: @ekcasey can we not derive this by counting the number of layers up to and including the top layer of the base image?

natalieparellano (approved):

text/0000-history.md
```diff
@@ -0,0 +1,85 @@
...
+[motivation]: #motivation
+
+This will allow Buildpack built images to be better visualized by common container visualization and debugging tools.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Various container image introspection tools look at `history.created_by` for each layer to visualize the layer. We should populate this key with a value that describes where a layer came from, for buildpacks this can be -
+
+- `Layer: {{ buildpack.layer.name }}, Created by: {{ buildpack.id }}`
```
natalieparellano, line 32:
```suggestion
- `Layer: {{ buildpack.layer.name }}, Created by: {{ buildpack.id }}@{{ buildpack.version }}`
```
(for consistency with the below)

natalieparellano:
@samj1912 [technically](https://github.com/buildpacks/community/blob/main/TEAMS.md#maintainers), we need one more ;)

ekcasey (approved):

samj1912:
@natalieparellano we should be good now I think.

natalieparellano:
FCP closing Friday 3/4

hone (approved):

jromero:
@natalieparellano is there a reason why this isn't merged yet?

text/0000-history.md
```diff
@@ -0,0 +1,85 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+Various container image introspection tools look at `history.created_by` for each layer to visualize the layer. We should populate this key with a value that describes where a layer came from, for buildpacks this can be -
+
+- `Layer: {{ buildpack.layer.name }}, Created by: {{ buildpack.id }}@{{ buildpack.version }}`
+
+for app layers, this can be - 
+
+- `Application Slice: {{ slice_number }} Created by: {{ buildpack.id }}@{{ buildpack.version }}`
```
jromero, line 36: I'm assuming adding the `buildpack.version` here doesn't impact the concept of reproducible builds even when there's a minor buildpack version bump if the results of the buildpack execution yields the same outcome right?

jromero (approved):

text/0000-history.md
samj1912: We already store this information as a label in the output image.

text/0000-history.md
```diff
@@ -0,0 +1,85 @@
...
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+N/A
+
+# Spec. Changes (OPTIONAL)
```
natalieparellano, line 82: /queue-issue buildpacks/spec "Exporter should set image history metadata" api/platform

PR #195, opened 2021-12-03
Add RFC for cosign integration

samj1912:


dlorenc:
Cosign maintainer here! We'd love any feedback on the SBOM use case. This is really just a first draft based on how we guessed people might use it. If there's anything you don't like or we could change to make things easier we can do that! I'm excited to see it used here at all.

text/0000-cosign.md
```diff
@@ -0,0 +1,109 @@
...
+key-data = "<base64-encoded-private-key>"
+# optional value for COSIGN_REPOSITORY, optional, default:""
+repository = "<cosign-repository>"
+# value for COSIGN_DOCKER_MEDIA_TYPES, optional, default:false
+docker-media-types = true
+# optional path to password file. Contents will be used as COSIGN_PASSWORD
+password = "<path-to-password-file>"
+# Exclusive OR with `password`
+# optional base64 encoded value for COSIGN_PASSWORD
+password-data = "<base64-encoded-password>"
```
jabrown85, line 60: I really rather not put this here at all. I think we should require the `COSIGN_PASSWORD` to be provided by the platform.

text/0000-cosign.md
```diff
@@ -0,0 +1,109 @@
...
+[definitions]: #definitions
+
+- cosign: A sigstore project to sign container images - https://github.com/sigstore/cosign
+- SBOM: Software Bill-of-Materials
+
+
+# Motivation
+[motivation]: #motivation
+
+Supply chain security has been on top of everyone's minds in 2021. It is important to ensure the integrity and origin of software artifacts in order to have a secure supply chain. Sigstore's cosign has quikcly risen to become the defacto container signing solution. This RFC proposes that we natively add cosign support to the lifecycle to facilitate easy signed container artifacts as part of the build process.
```
jabrown85, line 26:
```suggestion
Supply chain security has been on top of everyone's minds in 2021. It is important to ensure the integrity and origin of software artifacts in order to have a secure supply chain. Sigstore's cosign has quickly risen to become the defacto container signing solution. This RFC proposes that we natively add cosign support to the lifecycle to facilitate easy signed container artifacts as part of the build process.
```

text/0000-cosign.md
natalieparellano: Is there a similar concern with the base64 encoded private key?

text/0000-cosign.md
```diff
@@ -0,0 +1,109 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The exporter can use `cosign` as a library to generate, attach and publish signatures for the output image. `cosign` exposes a GGCR compatible interface which should fit nicely with the lifecycle.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Increased complexity in lifecycle
```
natalieparellano, line 83: I recall we discussed this in sub sync, but is it possible to add the advantage of having the lifecycle do the signing vs leaving it up to the platform?

text/0000-cosign.md
nebhale: I have the same broad concern about _requiring_ (as an additional option to one that does not require that data to rest, is fine) any sensitive data to be encoded into files.  Can we please explore some additional styles of contributing all of this sensitive information without requiring it to be in a file owned by us, on the file system?

nebhale (changes requested):

text/0000-cosign.md
hectorj2f: Yes, I agree with @jabrown85 and @nebhale here, those should be injected by the platform.

text/0000-cosign.md
hectorj2f: How does it increase the complexity here ? Could someone add more details ?

text/0000-cosign.md
```diff
@@ -0,0 +1,109 @@
...
+# optional flag to sign sbom if exported, default: true
+sign-sbom = true
+```
+
+The `analyzer` will be responsible for verifying registry access to push the required signatures. The exporter will be responsible for generating and pushing the signatures defined by each of the configs above for each tag specified to the `exporter`. As a part of this process, the lifecycle will also strip out any environment variables that start with `COSIGN_` before analyzing or exporting cosign specific tasks.
+
+During the export phase, the lifecycle will add all the provided annotations to each exported signature. It will also add the corresponding image tag and build time of the image as tags to the signature.
+The lifecycle will also export and sign related SBOM artifacts from the build. The format for attaching SBOM to the image will follow the specification described at https://github.com/sigstore/cosign/blob/main/specs/SBOM_SPEC.md.
+
+The above will only be supported if `lifecycle` is run in `registry` mode. If we wish to support it in local mode, we may add support for this in case we drop daemon mode and introdue an OCI layout mode.
```
ekcasey, line 72: Similar to https://github.com/buildpacks/rfcs/pull/196 I think it would be bad for the health of the project to implement these registry-only features conditionally now. Instead we should bit the bullet and remove daemon mode in favor of OCI layout mode first.

jromero (changes requested):
If I recall correctly, this RFC falls under the same idea as other external operations such as `preparer`. Those of which I believe we would develop PoCs independently and try to incorporate back into the project via [these guidelines](https://github.com/buildpacks/rfcs/blob/main/text/0086-component-level-contrib.md). If so, should this be a draft?

samj1912:
> If I recall correctly, this RFC falls under the same idea as other external operations such as `preparer`. Those of which I believe we would develop PoCs independently and try to incorporate back into the project via [these guidelines](https://github.com/buildpacks/rfcs/blob/main/text/0086-component-level-contrib.md?rgh-link-date=2022-03-22T19%3A29%3A56Z). If so, should this be a draft or closed?

I would prefer if this was a repo under buildpacks org, potentially under the platform team since pack would be the first target usecase. It would make it easier to manage/review/depend on repositories that fall under the buildpacks org umbrella.

samj1912:
> Is the only outstanding issue of how we proceed with a PoC/where the work happens?

Pretty much

text/0000-cosign.md
```diff
@@ -0,0 +1,100 @@
...
+[definitions]: #definitions
+
+- cosign: A sigstore project to sign container images - https://github.com/sigstore/cosign
+- SBOM: Software Bill-of-Materials
+
+
+# Motivation
+[motivation]: #motivation
+
+Supply chain security has been on top of everyone's minds in 2021. It is important to ensure the integrity and origin of software artifacts in order to have a secure supply chain. Sigstore's cosign has quikcly risen to become the defacto container signing solution. This RFC proposes that we natively add cosign support to the lifecycle to facilitate easy signed container artifacts as part of the build process.
```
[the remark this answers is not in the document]
samj1912: Ah yes, it's a leftover from the old rfc. Will clean it up.

text/0000-cosign.md
```diff
@@ -0,0 +1,100 @@
...
+[unresolved-questions]: #unresolved-questions
+
+
+How to deal with daemon case? I am of the opinion that we should deprecate daemon flag in lifecycle and instead switch to OCI but that can come later.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+
+Noted above.
```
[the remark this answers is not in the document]
samj1912: Also a leftover.

jromero:
@samj1912 as we discussed, this RFC requires a few set of changes to align with the latest agreed upon strategy. Please let me know when it's updated and I'll review/start the voting period.

natalieparellano:
@samj1912 what is the status of this RFC? Is this still something we want to do?

text/0000-cosign.md
AidanDelaney: If this moves outside `lifecycle` to being a platform responsibility, does the binary belong within the lifecycle distribution?

natalieparellano: Probably not... but in this case `pack create builder` should be expanded to include optional platform-provided binaries within the builder at the desired location.

text/0000-cosign.md
```diff
@@ -0,0 +1,100 @@
...
+- Name: Cosign support
+- Start Date: 2021-12-03
+- Author(s): [@samj1912](https://github.com/samj1912)
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
```
cmoulliard, line 12: This RFC supports indeed to sign an existing SBoM but also to export the attestation or attachment. Summary text should be reviewed

PR #196, opened 2021-12-03
Add OCI recommended labels

samj1912:


imjasonh:
There's an important distinction between labels and annotations. OCI recommends these annotations, but AIUI images in the daemon can't (currently?) be annotated, only labeled.

Talking to @ekcasey about this a few months ago it sounded like buildpacks would have to become more remote-image-centric in order to propagate annotations through the build. Or maybe the annotations only if the build is doing `--push`? 

That being said I'm +1 on the change, and I'd love to help however I can.

samj1912:
> There's an important distinction between labels and annotations. OCI recommends these annotations, but AIUI images in the daemon can't (currently?) be annotated, only labeled.
> 
> Talking to @ekcasey about this a few months ago it sounded like buildpacks would have to become more remote-image-centric in order to propagate annotations through the build. Or maybe the annotations only if the build is doing `--push`?
> 
> That being said I'm +1 on the change, and I'd love to help however I can.

@imjasonh Thanks for your comments! 

> There's an important distinction between labels and annotations. OCI recommends these annotations, but AIUI images in the daemon can't (currently?) be annotated, only labeled.
> 
> Talking to @ekcasey about this a few months ago it sounded like buildpacks would have to become more remote-image-centric in order to propagate annotations through the build. Or maybe the annotations only if the build is doing `--push`?
> 
> That being said I'm +1 on the change, and I'd love to help however I can.

@imjasonh all your help is appreciated! I think I got a bit confused with the b/w compatibility clause at https://github.com/opencontainers/image-spec/blob/main/annotations.md#back-compatibility-with-label-schema

but reading it more closely it looks like it recommends not just changing the domain but also moving from labels to annotations. Happy to update the rfc to use annotations and making it conditional on registry pushes directly. 

That being said, we have recently been facing more and more restrictions because of the daemon use case. Especially with things like cosign/sboms etc. I have been thinking about writing an RFC about deprecating the daemon flag and instead moving to OCI layout and registry. I could really use some help there if you are interested in the RFC. Your experience/knowledge with GGCR and OCI would be really helpful!

samj1912:
@imjasonh - updated to annotations conditional on the registry.

imjasonh:
> @imjasonh - updated to annotations conditional on the registry.

Thanks, this looks great to me. 👍

text/0000-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+
+- `org.opencontainers.image.base.name`: Run image tag name. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
+- `org.opencontainers.image.base.digest`: Run image digest. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
+
+The platform may also provide appropriate keys inside `project-metadata.toml` to the lifecycle so that it exports the appropriate annotations listed below.
+
+- `org.opencontainers.image.source`: URL to the image source if the source is a source controlled repository, URL to a blob etc. Sourced from `source.metadata.repository`.
+- `org.opencontainers.image.revision`: Source controlled revision of application source code. Sourced from `source.version.commit`.
+- `org.opencontainers.image.authors`: Authors of hte image. Sourced from `source.metadata.authors`.
+- `org.opencontainers.image.version`: Authors of hte image. Sourced from `source.metadata.version`.
```
imjasonh, line 45:
```suggestion
- `org.opencontainers.image.authors`: Authors of the image. Sourced from `source.metadata.authors`.
- `org.opencontainers.image.version`: Version of the image. Sourced from `source.metadata.version`.
```

text/0000-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+
+Adds OCI recommended annotations to the output image.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Optionally adds the following annotations where applicable when lifecycle is used to publish images directly to the registry - 
+
+- `org.opencontainers.image.base.name`: Run image tag name. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
+- `org.opencontainers.image.base.digest`: Run image digest. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
```
imjasonh, line 38: Does the `run-image.reference` data include both the mutable tag reference and the immutable digest?

hone (approved):
This looks good to me.

text/0000-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+[what-it-is]: #what-it-is
+
+Adds OCI recommended annotations to the output image.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Optionally adds the following annotations where applicable when lifecycle is used to publish images directly to the registry - 
+
+- `org.opencontainers.image.base.name`: Run image tag name. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
```
natalieparellano, line 37: I think this is actually `io.buildpacks.lifecycle.metadata.runImage.reference`

text/0000-annotations.md
natalieparellano, line 38:
```suggestion
- `org.opencontainers.image.base.digest`: Run image digest. Currently stored in `io.buildpacks.lifecycle.metadata.runImage.reference`
```

natalieparellano (approved):

hone:
@sclevine @jkutner @ekcasey ready for review please :pray:

text/0000-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+- `org.opencontainers.image.licenses`: License(s) under which contained software is distributed as an SPDX License Expression. Sourced from `source.metadata.licenses`.
+
+A platform may choose to map the above fields to the fields defined in a project descriptor as noted at https://github.com/buildpacks/spec/blob/extensions/project-descriptor%2F0.2/extensions/project-descriptor.md#non-_-tables.
+
+Everything under `source.metadata.<key>` is mapped from `_.<key>` from the project descriptor in this case.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+In case of the daemon mode, the annotations cannot be applied and will be ignored.
```
ekcasey, line 57: Up until this point we have endeavored to make the output image identical regardless of export target (registry vs daemon). I am leary of introducing behavior differences between the daemon and remote cases. IMO it would be much better to remove the daemon as an export target first before continuing with this RFC.

text/0000-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Optionally adds the following annotations where applicable when lifecycle is used to publish images directly to the registry - 
+
+- `org.opencontainers.image.base.name`: Run image tag name. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
+- `org.opencontainers.image.base.digest`: Run image digest. Currently stored in `io.buildpacks.lifecycle.metadata.run-image.reference`
+
+The platform may also provide appropriate keys inside `project-metadata.toml` to the lifecycle so that it exports the appropriate annotations listed below.
```
ekcasey, line 40: In the ideal world I think these project metadata annotation would be better contributed by a buildpack vs by the lifecycle for the following reasons:
1. Buildpack contributed behavior more easily upgradable accross a variable of platforms
2. Keeping the surface area of the platform API as small as possible makes the spec and the conditional logic in lifecycle less complicated, thus making our lives easier and enabling us to actually make progress shipping things.

The way I imagine it:
* **step 1** remove daemon
* **step 2** allow buildpacks to contribute aribtrary annotations
* **step 3** provide a (utility?) buildpack that contributes these annotations based off the values in `project.toml`

ekcasey (changes requested):
I am of the opinion that we should table this and come back to it after implementing a proposal to fix the only source of complexity here and remove lifecycle->daemon support.

text/0000-annotations.md
imjasonh: That ordering makes sense to me. Is there a rough plan/timeline for step 1?

jkutner (approved):

text/0000-annotations.md
jromero: It ties into this RFC issue: https://github.com/buildpacks/rfcs/issues/197

@jjbustamante is on it!

text/0000-annotations.md
jromero: Based on the [spec](https://github.com/buildpacks/spec/blob/main/platform.md#iobuildpackslifecyclemetadata-json) this would be the immutable digest:

> A digest reference to a manifest stored in an OCI image registry

text/0000-annotations.md
jromero: I would agree if the registry and daemon cases were identical to begin with but they are already not. Things like `io.buildpacks.lifecycle.metadata.runImage.reference` change depending on whether it's built for daemon or registry. The same inputs for an image don't survey round-tripping because of that and the difference in compression of layers. In other words, lifecycle -> daemon -> registry != lifecycle -> registry. IMO, it doesn't sound reasonable to block this feature on the eventual _hope_ that we'll get rid of the daemon. At the end of the day, the core concern (discrepancy) will always exist but be pushed onto the platform.

jromero (approved):

text/0000-annotations.md
jromero: Based on my comment [here](https://github.com/buildpacks/rfcs/pull/196#discussion_r789017080), I'm back to say that I don't think step 1 is necessary. 

steps 2 and 3 can be an alternative approach to this RFC without step 1. I'd love to see it comparatively in the `Alternatives` seciton.

text/0000-annotations.md
ekcasey: > IMO, it doesn't sound reasonable to block this feature on the eventual hope that we'll get rid of the daemon.

In order to get rid of the daemon instead of merely hoping it goes away I think we need to hold the line somewhere and actually prioritize getting rid of it. Adding remote-only features seems like the right place to draw the line to me.

Yes there is a path forward without removing the daemon first, but it is uglier and requires more compromises in the lifecycle and imgutil and caveats in our docs and specs. Instead we could invest in removing the daemon and clean up some of the uglier parts of the project.

text/0000-annotations.md
ekcasey: https://github.com/buildpacks/rfcs/pull/196#discussion_r789142013

ekcasey:
@samj1912 I am convinced that we can approve a version of this without blocking on the daemon removal. However I still feel pretty strongly that we shouldn't make the lifecycle aware of these specific annotations, but should instead:
1. Provide a mechanism whereby a buildpack can set abitrary annotations
2. Implementing this as a utility buildpack

Are you open to revising this RFC to make it address buildpack-provided annotations instead of addressing OCI specified annotations specifically?

PR #199, opened 2021-12-13
Add Migration section to template

dmikusa-pivotal:
Initially [proposed here](https://buildpacks.slack.com/archives/C94UJCNV6/p1639075530135600).

This PR adds a Migration section to the RFC template. The purpose of this section is to ensure consideration is given to how an RFC might break existing APIs and workflows so that a discussion can be had in advance of an RFC being approved.

As we get more and more users, this seems like an appropriate step to take to show that stability and continuity are important considerations as we evolve buildpacks.

Signed-off-by: Daniel Mikusa <dmikusa@vmware.com>

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -44,6 +44,11 @@ This is the technical portion of the RFC, where you explain the design in suffic
 
 The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
 
+# Migration
+[migration]: #migration
+
+This section should document breaks to public API and breaks in compatibility due to this RFC's proposed changes. In addition, it should document the proposed steps that one would need to take to work through these changes. Care should be give to include all applicable personas, such as platform developers, buildpack developers, buildpack users and consumers of buildpack images.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 


aemengo (commented):
It's hard to know _beforehand_ that an RFC might introduce a breaking change. In fact, the breaking changes might not be introduced in the RFC, but be introduced as part of last minute implementation changes. Does this addition have merit if, say, we might miss a few of these?

dmikusa-pivotal:
> It's hard to know _beforehand_ that an RFC might introduce a breaking change. In fact, the breaking changes might not be introduced in the RFC, but be introduced as part of last minute implementation changes. Does this addition have merit if, say, we might miss a few of these?

I agree, no one can see the future. I don't want to imply everything has to be written in stone for the RFC. It's an RFC, not a spec. Things can evolve.

If others have suggestions, I'm open to wording suggestions for the placeholder text in the template. I gave it a shot, but am by no means a wordsmith.

The general idea of adding this section is that it ensures a.) the proposer thought about it and b.) we're making this a part of the discussions we have when evaluating new proposals. 

I think a side benefit might be that writing migration documentation for new features will be come easier as well, since at least some of this has been discussed and written down already.

hone (commented):
This is great and thanks for adding it. Should we mark this as optional since it may not apply to every RFC?

dmikusa-pivotal:
> This is great and thanks for adding it. Should we mark this as optional since it may not apply to every RFC?

Yes, that's true. My $0.02 would be that it's required in the sense that you need to add some content or add an explanation of why it's not relevant. I don't feel like it should ever be empty.

hone:
> > This is great and thanks for adding it. Should we mark this as optional since it may not apply to every RFC?
> 
> Yes, that's true. My $0.02 would be that it's required in the sense that you need to add some content or add an explanation of why it's not relevant. I don't feel like it should ever be empty.

I'm fine with playing it by ear for a bit and circling back if we want to change it.

hone (approved):

nebhale (approved):

jkutner (approved):

ekcasey (approved):

PR #200, opened 2022-01-31
RFC: Utility buildpack for .profile

mboldt:
As part of RFC 93, [`.profile` scripts will cease to be supported by the platform API](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md#appprofile). This RFC proposes developing a [utility buildpack](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md) to support `.profile` scripts to prevent regressions after RFC 93 is implemented.


mboldt:
I have also written a very quick-and-dirty proof-of-concept of the approach here: https://github.com/mboldt/dot-profile-buildpack. It's throwaway code, but it shows the approach.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+call .profile.bat
+set >&%CNB_EXEC_D_HANDLE%
+```
+
+# Migration
+[migration]: #migration
+
+This buildpack is net new, so has no inherent migration considerations.
```
aemengo, line 97:
```suggestion
This buildpack is not new, so has no inherent migration considerations.
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+memory_in_bytes="$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)"
+MEMORY_AVAILABLE="$(( $memory_in_bytes / ( 1024 * 1024 ) ))"
+export MEMORY_AVAILABLE
+```
+
+The wrapper should ensure that the `MEMORY_AVAILABLE` environment variable is set in the environment with the proper value.
+
+## Example 2 (file system side effects)
+
+With this contrived `.profile` script:
```
aemengo, line 59:
```suggestion
With this `.profile` script:
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+
+If we do nothing, we introduce a regression in functionality, and force application developers to rework their `.profile` scripts.
+
+- What other designs have been considered?
+- Why is this proposal the best?
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
```
aemengo, line 119:
```suggestion
-
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+# Summary
+[summary]: #summary
+
+As part of RFC 93, [`.profile` scripts will cease to be supported by the platform API](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md#appprofile).
+This RFC proposes developing a [utility buildpack](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md) to support `.profile` scripts to prevent regressions after RFC 93 is implemented.
+
+# Definitions
+[definitions]: #definitions
+
+*utility buildpack*: A buildpack officially supported by the Buildpack Authors' Tooling Team per [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md).
```
aemengo, line 21:
```suggestion
*utility buildpack*: A buildpack officially supported by the Buildpack Authors' Tooling Team per [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md) that provides a generic capability.
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+This buildpack is net new, so has no inherent migration considerations.
+
+For the related migration concerns for removing shell functionality in general, see the [Migration Path section of RFC 93](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md#migration-path).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a new component to maintain and support.
+That being said, this will enable spec and lifecycle simplifications noted in [RFC 93](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md) (where this approach was first suggested).
+Also, [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md) resolves to support and maintain utility buidpacks.
```
aemengo, line 106:
```suggestion
Also, [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md) resolves to support and maintain utility buildpacks.
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we do nothing, we introduce a regression in functionality, and force application developers to rework their `.profile` scripts.
+
+- What other designs have been considered?
+- Why is this proposal the best?
```
aemengo, line 114:
```suggestion
```

aemengo (approved):

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Detect a `.profile` file in the app dir
+- Wrap the `.profile` file so that it implements the `exec.d` interface
+- Add the `exec.d` executable to the `<layer>/exec.d` directory so the launcher will apply it
+
+We propose the ID `buildpacksio/profie` for this buildpack (cf. [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#what-it-is)).
+
+## Example 1 (environment variables)
+
+Here is an example of a `.profile` script, inspired by paketo-buidpacks/node-engine:
```
aemengo, line 47:
```suggestion
Here is an example of a `.profile` script, inspired by paketo-buildpacks/node-engine:
```

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,131 @@
...
+In short, the wrapper should do:
+
+```
+#!/bin/bash
+set -euo pipefail
+source .profile
+env >&3
+```
+
+We acknoledge that the `env` output will need to be properly quoted to be valid TOML.
```
aemengo, line 80:
```suggestion
We acknowledge that the `env` output will need to be properly quoted to be valid TOML.
```

text/0000-dot-profile-utility-buildpack.md
mboldt: Updated to just "new," which was what was intended.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,129 @@
...
+The target persona is a platform operator or implementor who wants to update to the latest platform API, while maintaining the `.profile` functionality for application developers.
+
+We propose developing and supporting a buildpack to provide an identical interface to the existing `.profile` functionality.
+It will:
+
+- Detect a `.profile` file in the app dir
+- Wrap the `.profile` file so that it implements the `exec.d` interface
+- Add the `exec.d` executable to the `<layer>/exec.d` directory so the launcher will apply it
+
+We propose the ID `buildpacksio/profie` for this buildpack (cf. [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#what-it-is)).
```
jkutner, line 43:
```suggestion
We propose the ID `buildpacksio/profile` for this buildpack (cf. [RFC 97](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#what-it-is)).
```

Or did you mean to have a cute name?

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,129 @@
...
+RFC 93 recommends supporting the `.profile` script functionality in a utility buildpack to avoid regressions.
+This proposal is to develop and support the `.profile` utility buildpack, allowing RFC 93 to be implemented without regression.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The target persona is a platform operator or implementor who wants to update to the latest platform API, while maintaining the `.profile` functionality for application developers.
+
+We propose developing and supporting a buildpack to provide an identical interface to the existing `.profile` functionality.
+It will:
```
jkutner, line 37: Can it also support a `.profile.d/` dir with arbitrary scripts?

I'm not sure we _need_ this (because in the past it's recommend for use by buildpacks only, which can use exec.d)

https://devcenter.heroku.com/articles/buildpack-api#profile-d-scripts

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,129 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+If we do nothing, we introduce a regression in functionality, and force application developers to rework their `.profile` scripts.
+
+
+# Prior Art
+[prior-art]: #prior-art
+
+-
```
jkutner, line 117: https://devcenter.heroku.com/articles/buildpack-api#profile-d-scripts

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Unresolved Questions
```
jkutner, line 119: Do we need to specify versions of Bash or shell that this will support?

text/0000-dot-profile-utility-buildpack.md
mboldt: Certainly seems technically feasible, but per the spec and the recommendation in the link you shared, I think it makes sense to stick with the principle of "`.profile.d/` is for buildpacks and `.profile` is for apps." I would expect a spec change to explicitly support `.profile.d/` scripts in apps, and that feels beyond the scope of this RFC.

mboldt: Thinking through a little more, maybe your point is less about apps using `.profile.d/` and more about supporting buildpacks that haven't moved from `.profile.d/` to `exec.d/`. I.e., could this unlock buildpacks that use `.profile.d/` to still work when the shell is removed from the launcher?

Again, certainly seems technically feasible.

I think it's a matter of (1) what is the state of buildpacks using `exec.d` vs `.profile.d` (if most are using `exec.d` it's less of an issue), and (2) do we want shell removal to be a forcing function for buildpack authors to migrate to `exec.d`? I don't have insight into these issues. What do you and others think?

text/0000-dot-profile-utility-buildpack.md
mboldt: > Do we need to specify versions of Bash or shell that this will support?

Yes we should. The [spec](https://github.com/buildpacks/spec#operating-system-conventions) requires bash 3+ on linux, and cmd.exe (no specific version) on windows. Will add to the text.

text/0000-dot-profile-utility-buildpack.md
jkutner: Let's move this RFC forward without `.profile.d` and revisit it later as an enhancement.

text/0000-dot-profile-utility-buildpack.md
mboldt: > How will this work if the user targets a `--run-image` that's FROM scratch (i.e. has no Bash installed). Should the exec.d error out, or should it just no-op?

It would seem less surprising to me if this buildpack would add a layer with bash.

jkutner: that's an interesting idea

text/0000-dot-profile-utility-buildpack.md
mboldt: Thank you. Started with this and fleshed out some prior art.

text/0000-dot-profile-utility-buildpack.md
mboldt: I did a little exploration with running bash on scratch. It was pretty straight-forward. Debian has a `bash-static` package. Extract the static bash binary from that, plop it down on a `FROM scratch` container, and Bob's your uncle.

Dockerfile and Makefile for this experiment here: https://github.com/mboldt/scratch-play/tree/main/bash

Details to figure out (keeping up with bash versioning, are we free license-wise to use the Debian package), but the add-layer-with-statically-linked-bash approach is feasible.

If the Debian package is any sort of blocker, there's also some prior art on compiling bash statically from source here: https://github.com/robxu9/bash-static

jkutner: That's great to know. fwiw, I don't think we need to answer all those questions in order to finish this RFC and ship the first version of the profile buildpack, but definitely something we will probably want in the future. 

Let's capture this info in the text of the RFC though.

dmikusa-pivotal: I would be in favor of declaring in the docs for the buildpack that bash is required in the run image, and if it's not present failing with a reasonable error message.

On one hand, it's convenient to install bash and make it work, on the other most folks are using an image w/out bash for the specific reason of not having bash. If we add bash, that defeats the purpose of using that image. I think the user should be presented with the decision, either use an image with bash or don't use `.profile` scripts that require bash.

This also keeps the buildpack implementation much simpler.

dmikusa-pivotal:
The RFC looks good to me.

Outside of the RFC, I feel like there are a lot of implementation details that are going to need to be worked out given this will be the first buildpack released through the project & BAT team. I suggest that we make this a topic at this Friday's BAT team meeting if folks are interested and available to join.

text/0000-dot-profile-utility-buildpack.md
jkutner: yea, @dmikusa-pivotal makes a good point. It would be surprising to get a bash when you wanted scratch.

Maybe we can split the bash install into it's own buildpack, then it can be used as desired.

mboldt: Thank you for the discussion. The "bash buildpack" approach certainly sounds reasonable to me - more explicit and modular. I will capture this in the text.

dmikusa-pivotal: For what it's worth, I'm not sure we could accept a bash buildpack. The RFC for utility buildpacks in the requirements section says:

>The buildpack MUST NOT install any dependencies in the output image

https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#how-it-works

I agree it seems like a useful buildpack. It could be provided through some other project though, like Paketo. Folks could still combine and use them together.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,156 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- Current CNB spec:
+  - `.profile` scripts are part of the buildpack app interface ([spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#app-interface))
+  - `.profile.d/` scripts are part of the buildpack build phase output ([spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#build)).
+
+- Profile scripts:
+  - Heroku supports [`.profile` scripts](https://devcenter.heroku.com/articles/dynos#the-profile-file) and [`.profile.d/` scripts](https://devcenter.heroku.com/articles/buildpack-api#profile-d-scripts).
+  - Cloud Foundry supports [`.profile and .profile.d/ scripts](https://docs.cloudfoundry.org/devguide/deploy-apps/deploy-app.html#profile).
```
natalieparellano, line 125: Does this mean that developers using Cloud Foundry will have a harder migration path if we don't do https://github.com/buildpacks/rfcs/pull/200/files#r797813213 ?

natalieparellano: Ah, nevermind - the v3 spec doesn't acknowledge profile.d in the app root directory: https://github.com/buildpacks/spec/blob/main/buildpack.md#app-interface

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Prior related RFCs:
+  - [Officially Supported Utility Buildpacks](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md)
+  - [Remove Shell Processes](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md)
+  - [Exec.d - Shell-Free Profile.d](https://github.com/buildpacks/rfcs/blob/main/text/0057-exec.d-shell-free-profile-d.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- **Are there things that `.profile` scripts do that will not be covered by the exec.d interface?**
```
ekcasey, line 135: This is definitely an edge case but a good call out. Hopefully this shouldn't be an issue. The app profile runs after the buildpack provided profiles and just before the process type is executed so the only thing that could depend on one of the functions in the process.

This will almost certainly never happen with a buildpack-provided process (the buildpack depending on the app to define a bash function would be very weird and brittle).

If the user is provided the process definition (e..g with a procfile) this is _slightly_ more plausible. But in this case there is a very easy workaround. They can just source a script containing the functions as part of the command provided in the procfile.

ekcasey (changes requested):

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,156 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+In short, the wrapper should do:
+
+```
+#!/bin/bash
+set -euo pipefail
+source .profile
+env >&3
```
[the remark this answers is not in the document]
mboldt: Thank you. I've updated the text to get more precise about this.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In short, the wrapper should do:
+
+```
+#!/bin/bash
+set -euo pipefail
```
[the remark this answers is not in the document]
mboldt: Good catch. Removed.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,156 @@
...
+The wrapper would not need to set any environment variables, but should maintain the side effect of creating the `$HOME/hello` file.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In short, the wrapper should do:
+
+```
+#!/bin/bash
```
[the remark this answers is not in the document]
mboldt: Was meant to be illustrative. I've updated it to describe a more clear and concrete approach. Feels much stronger - thank you for the feedback!

text/0000-dot-profile-utility-buildpack.md
mboldt: I've updated the text to discourage the idea of a bash buildpack for the sake of .profile scripts.

text/0000-dot-profile-utility-buildpack.md
mboldt: Any action needed here in the text?

text/0000-dot-profile-utility-buildpack.md
mboldt: I've included this color in the text. Thank you again!

text/0000-dot-profile-utility-buildpack.md
hone: @mboldt I don't think so, b/c we aren't adding `profile.d` scripts here.

text/0000-dot-profile-utility-buildpack.md
hone: we can just include this edge case as part of the documentation for the profile buildpack.

hone (approved):

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,161 @@
...
+The wrapper would not need to set any environment variables, but should maintain the side effect of creating the `$HOME/hello` file.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+We will write two small programs to implement the buildpack.
+
+- `dot-profile-wrapper` will be the main entry point.
+- `write-env-toml` will write all environment variables to the [`exec.d` output TOML](https://github.com/buildpacks/spec/blob/main/buildpack.md#execd-output-toml).
```
ekcasey, line 74: What is the rationale for separating this into two executables? Is there any reason the entrypoint executable can't also write the output?

mboldt: `write-env-toml` must be able to see environment variables set by the `.profile` script. This will need to look something like `bash -c 'source .profile && write-env-toml'`. `dot-profile-wrapper` will need to spawn that `bash` command, handle errors, etc.

I suppose we could do something like `bash -c 'source .profile && env'` and parse the output but that seems much messier than directly reading the environment in code.

We could combine these into a single executable file, using symlink tricks or subcommands, if that is important, e.g., for minimizing the footprint.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,161 @@
...
+echo 'hello' >> "$HOME/hello"
+```
+
+The wrapper would not need to set any environment variables, but should maintain the side effect of creating the `$HOME/hello` file.
+
+
+# How it Works
+[how-it-works]: #how-it-works
+
+We will write two small programs to implement the buildpack.
```
samj1912, line 71: There are some minor differences in how the lifecycle executes exec.d scripts vs `.profile` right now. The CWD for the `.profile` script is the process working directory where as it is the app directory for the exec.d scripts. How do we plan on dealing with this edge case?

text/0000-dot-profile-utility-buildpack.md
samj1912: It should be possible to do it in a single script, this is the entire exec.d script we need - 


```bash
#!/bin/bash

source .profile
for var in $(compgen -e); do
	value=$(echo ${!var} | sed 's/"/\\"/g')
    echo "${var} = \"${value}\"" 1>&3
done
```

dmikusa-pivotal: My preference would be for using `sh -c '. .profile && write-env-toml'`. It's POSIX compliant, so it doesn't depend on Bash. It's a small change and being POSIX compliant and not depending on Bash specifics should allow this to work across a larger set of images.

I would also much prefer to have the logic of parsing and reformatting the env variables in code that's testable because there will likely be edge cases, so it should give us more confidence in the implementation. 

I could see `sh -c '. .profile && env'` working too, but then `dot-profile-wrapper` has to parse the `env` output format instead of being able to iterate `os.Environ()` so it would be more work.

If we're worried about binary size, we could combine them into one binary and use arg[0] to determine the action to take, like we do with `bin/detect` and  `bin/main` in libcnb.

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,161 @@
...
+
+- `dot-profile-wrapper` will be the main entry point.
+- `write-env-toml` will write all environment variables to the [`exec.d` output TOML](https://github.com/buildpacks/spec/blob/main/buildpack.md#execd-output-toml).
+
+On Linux, `dot-profile-wrapper` will:
+
+1. Ensure `bash` exists on the run image, and error if it does not.
+1. Launch a `bash` shell to:
```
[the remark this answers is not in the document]
samj1912: It has to be bash since that is what the spec specifies and what the lifecycle currently implements https://github.com/buildpacks/spec/blob/main/buildpack.md#app-interface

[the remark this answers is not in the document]
samj1912: > When the specification denotes a "shell", Linux stacks MUST use the Bourne Again Shell (bash) version 3 or greater

text/0000-dot-profile-utility-buildpack.md
samj1912: I believe the current spec requires this to be BASH and not any posix shell. Given that this buildpack is being created for backwards compatibility we have an inherent dependency on bash.

dmikusa-pivotal: I'd also be curious for your thoughts on using something like https://pkg.go.dev/mvdan.cc/sh/shell?utm_source=godoc#SourceFile. It would allow us to capture shell env variables to set without actually running a shell, so it would have the maximum compatibility (i.e. no shell required). I didn't test but the library also reports the ability to allow-list access to files/commands, so it could possibly allow a more locked down environment in which to run the scripts.

The obvious downside is that it's not actually running a shell, so you could possibly have edge-case/compatibility issues. Plus it requires taking a 3rd party dependency, although the project does seem active.

samj1912: As for the binary size and just in terms of security issues/bugs, IMHO a simple bash script has lesser surface area than a go binary, given that we are requiring the output run image to contain bash. This way, the end user can also rebase their base image to mitigate any bash related security issues and we/they don't have to worry about security issues in our go binaries that are running in production images.

samj1912: > I'd also be curious for your thoughts on using something like https://pkg.go.dev/mvdan.cc/sh/shell?utm_source=godoc#SourceFile. It would allow us to capture shell env variables to set without actually running a shell, so it would have the maximum compatibility (i.e. no shell required). I didn't test but the library also reports the ability to allow-list access to files/commands, so it could possibly allow a more locked down environment in which to run the scripts.
> 
> The obvious downside is that it's not actually running a shell, so you could possibly have edge-case/compatibility issues. Plus it requires taking a 3rd party dependency, although the project does seem active.

I worry about any inconsistencies in parsing behavior between that library and bash. Given that the spec mandates bash and that the lifecycle itself currently uses bash in its implementation, I would prefer if we keep things closer to the current spec/implementation.

ekcasey: @mboldt @samj1912 @dmikusa-pivotal 

Should we just take some of the implementation details out of this RFC so we can get it through? I know I started this by asking about the two executables but I think we could work some of this out at the subteam level as long as we agree on the user interface and high level requirements.

dmikusa-pivotal: >I believe the current spec requires this to be bash and not any posix shell. 

This is something I dislike about the spec. It's unnecessarily coupled to bash when sh would probably do fine.

>Given that this buildpack is being created for backwards compatibility we have an inherent dependency on bash.

Yes, that makes sense. You could always `bash -c` in your script if you really need bash and your container has it, but that's not current behavior.

>As for the binary size and just in terms of security issues/bugs, IMHO a simple bash script has lesser surface area than a go binary, given that we are requiring the output run image to contain bash. This way, the end user can also rebase their base image to mitigate any bash related security issues and we/they don't have to worry about security issues in our go binaries that are running in production images.

I don't disagree with your statement here. Using a go binary puts the go runtime into the container and requires us to keep that up-to-date. Given the small scope of what the binary would do though, that seems like a relatively low risk. 

I think that because we are by function of this buildpack giving direct, unfiltered access to the container where this runs through the `.profile` script, security is pretty much wide open already. A user doesn't really need to worry about exploiting the go binary that writes the env variables. The user can already put anything into the env variables you want or write/update the filesystem and it'll get executed.

My $0.02. I'm still strongly of the opinion that we do not implement this in bash. I would disagree that the script proposed is simple, properly quoting things is always a pain and should be tested. There are going to be edge cases to that, but we can't test because it's in a bash script. 

I think any bash script like this is also a slippery slope argument. What happens when there is an issue and we have to add more complicated logic? What I've seen when doing this in the past is that the bash script just gets a little bigger, a little more complicated, and still remains untested.

I think we can make this easier on ourselves, in the long run, to maintain by making it fully testable and I don't think that has much cost or downside.

mboldt: My $0.02:

- Stick with `bash` as opposed to `sh`/posix/shell substitute. As I understand it, the main purpose of this effort is to avoid regressions, not to expand adoption. We should encourage new adopters to use exec.d instead.
- I agree that "read the environment and write TOML" will benefit from testability and using library code that has likely encountered and handled edge cases already.

That said, I will pull implementation details out and we can ~duke it out~ 😉  continue discussion in subteam.

samj1912 (approved):

dmikusa-pivotal (approved):

nebhale (approved):

jromero (approved):

samj1912:
/queue-issue buildpacks/spec Document the profile buildpack

/queue-issue buildpacks/community Create a profile buildpack repo and add appropriate BAT maintainers to it

/queue-issue buildpacks/docs Document the profile buildpack and add it to the migration guide

text/0000-dot-profile-utility-buildpack.md
```diff
@@ -0,0 +1,165 @@
+# Meta
```
samj1912, line 1: /queue-issue buildpacks/spec Document the profile buildpack

samj1912, line 1: /queue-issue buildpacks/spec "Document the profile buildpack"

samj1912:
Closed in https://github.com/buildpacks/rfcs/commit/5d06f94d729bd80669ca7fa9341ff87023df2b79

PR #202, opened 2022-02-07
Add prepare operation RFC

jromero:


text/0000-prepare-operation.md
```diff
@@ -0,0 +1,374 @@
...
+
+ - Apply the requested configuration.
+
+#### Inputs
+
+The `prepare` phase would take all the same inputs as `creator` plus the addition of a path to `project.toml`.
+
+#### Outputs
+
+The `prepare` phase may affect the file system, and mutation or create files that would be consumed be futher phases.
```
jabrown85, line 135: I think we should elaborate here a bit. What are the limits? What volumes should be mounted? Is this phase intended/allowed to happen in a different container?

Thinking out loud a bit on use cases I can see leveraging a prepare phase

- Download source code (from a signed url, for instance, provided by the platform)
- Downloading previous build cache (again, from a signed url or perhaps copying from a k8s style PVC)
- Generating a `project.toml` for projects that don't have one (for platform consistency, maybe)

To support those, we would need to mount `/workspace` and `/cache`.

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,374 @@
...
+
+#### Responsibilities
+
+At minimum, the expected reponsibility of the `preparer` would be to:
+
+ - Apply the requested configuration.
+
+#### Inputs
+
+The `prepare` phase would take all the same inputs as `creator` plus the addition of a path to `project.toml`.
```
jabrown85, line 131: Is this binary intended to run on builder images?

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,374 @@
...
+
+###### `preparer` executable
+
+The Cloud Native Buildpacks project will have a `preparer` it ships along with the existing `lifecyle` image.
+
+> This would be useful for builder providers or platform implementers to not have to develop standard functionality.
+
+The default implementation COULD take care of applying the following configuration:
+
+- `io.buildpacks.defaults.group` →
```
loewenstein, line 155: Hasn't there been a configuration parameter to replace buildpacks instead of augmenting those present? Was `[build.buildpacks] ` if I am not mistaken. Does that still exist? Should it be added here?

loewenstein (commented):
FYI We have implementations for some of those preparer functionality in the [Piper cnbBuild step](https://github.com/SAP/jenkins-library/blob/master/cmd/cnbBuild.go) and [utility package](https://github.com/SAP/jenkins-library/tree/master/pkg/cnbutils).

E.g. include/exclude handling, buildpack downloads (from url or via buildpack registry), etc.

I guess we could provide those if this RFC was accepted.

text/0000-prepare-operation.md
jromero: Possible, yes. Required no.

loewenstein: But it's running either in the builder image or not at all, right?

Potentially related question:would it be called as part of the `creator` lifecycle?

jromero: > But it's running either in the builder image or not at all, right?

I can't think of a reason why it _must_ run on the builder image. As an example, Tekton already does some preparing step [here](https://github.com/buildpacks/tekton-integration/blob/771bbd2ecec9b671889db48214f4e65eb3e9113e/task/buildpacks-phases/0.2/buildpacks-phases.yaml#L76-L127) using a completely different image. I can foresee platforms wanting to provide their own container for the execution of their preparation mechanism.

Hypothetically, given that ultimately preparing only impacts the build using the existing file system interface, platforms can apply changes "natively" by making changes to the "workspace" on the fly or to volumes that would later be mounted to the next phases.

> Would it be called as part of the creator lifecycle?

No. There's a light mention of that [here](https://github.com/buildpacks/rfcs/pull/202/files/99fc145c93f07e045e62b9c5d42405ecffc53afc#diff-da0c0db1eef841f956d231dff1f0ad819daf2ee72f55b71032073d431b767277R119). This phase is considered to be outside of the Build operation.

text/0000-prepare-operation.md
jromero: > What volumes should be mounted? 

Volumes aren't really part of the spec. It sounds like you may be asking from a specific PoV. We could elaborate through examples but volumes aren't referenced because they aren't really required from a platform -> lifecycle interface.

> Is this phase intended/allowed to happen in a different container?

[This answer here](https://github.com/buildpacks/rfcs/pull/202/files/99fc145c93f07e045e62b9c5d42405ecffc53afc#r801953808) may shed some light on the question. It is intended to run on a separate container although not technically required to do so. Most platforms would execute their own prepare binary/container but some may rely on the [builder-provided `preparer`](https://github.com/buildpacks/rfcs/blob/99fc145c93f07e045e62b9c5d42405ecffc53afc/text/0000-prepare-operation.md?plain=1#L147-L160).

text/0000-prepare-operation.md
jromero: I believe you are referring to project descriptor 0.1. This work is based on the [soon-to-release 0.2 descriptor](https://github.com/buildpacks/spec/pull/282) where it's be [relocated to `io.buildpacks.group`](https://github.com/buildpacks/spec/blob/extensions/project-descriptor/0.2/extensions/project-descriptor.md#iobuildpacksgroup-optional).

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,385 @@
...
+
+```toml
+[io.buildpacks]
+schema-version = "0.2"
+
+###
+# common buildpacks config
+##
+
+[io.buildpacks.defaults]
```
jromero, line 87: **"Naming" Vote: Should we use or not use a `defaults` table?**

It's been discussed that `defaults` in `io.buildpacks.defaults` is unnecessary and that the extra characters aren't worth the effort in trying to convey that these are _only_ defaults and it's really up to the platform to determine which properties get applied.

Please vote for your opinion using the following reactions:

👍 → keep `defaults`
👎 → drop `defaults`

jromero: RFC has been updated to reflect consensus.

jabrown85 (approved):

jromero:
Based on further conversations of additional "operations" such as prepare (this RFC), [publish](https://github.com/buildpacks/rfcs/pull/203), and [cosign](https://github.com/buildpacks/rfcs/pull/195) we are considering moving forward with an "implement first and standardize later" strategy. This approach was decided because it heavily relies on the willingness of individual platforms to incorporate such operations. As a project, we think that by providing the utilities and "example" implementation through platforms we maintain, such as Pack and Tekon Tasks, can prove the value of such individual reusable components and operations. The hope is that, with due, time truly desired functionality is naturally promoted from an example, to convention, to specification.

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+
+<!-- </details> -->
+
+---
+
+## Project Descriptor Spec
+
+#### Remove `io.buildpacks` namespace
+
+We'll want to remove the `io.buildpacks` namespace since it will be versioned seperate from the project descriptor.
```
natalieparellano, line 311: I guess I understand the need for this but I also find the idea of another spec somewhat confusing. Unfortunately I don't have any constructive suggestions to offer.

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+
+> This would be useful for builder providers or platform implementers that choose to provide their own [`preparer` executable](#preparer-executable).
+
+###### `preparer` executable
+
+The Cloud Native Buildpacks project will have a `preparer` it ships along with the existing `lifecyle` image.
+
+> This would be useful for builder providers or platform implementers to not have to develop standard functionality.
+
+The default implementation COULD take care of applying the following configuration:
```
natalieparellano, line 154: To link up some discussion re: [https://github.com/buildpacks/spec/pull/288](https://github.com/buildpacks/spec/pull/288#issuecomment-1057283717), it could also download a run-image sbom if provided as a label pointing to a layer or a cosign-attached image manifest.

jromero:
Update: a new repository with this work has been started here: https://github.com/jromero/cnb-prepare

SaschaSchwarze0:
Hi @jromero, in [Shipwright](https://shipwright.io/), [we are using the lifecycle methods and therefore bypass pack](https://github.com/shipwright-io/build/blob/v0.10.0/samples/buildstrategy/buildpacks-v3/buildstrategy_buildpacks-v3_cr.yaml), similar as the [Tekton catalog task](https://github.com/tektoncd/catalog/blob/main/task/buildpacks-phases/0.2/buildpacks-phases.yaml). Your prepare operation is quite interesting. Is this something you are still looking at?

jkutner:
I'd love to see this move forward

PR #203, opened 2022-02-23
RFC: Export to OCI format

### LINKED ISSUE #197: Idea: Add support for OCI layout / deprecate daemon support
As we integrate further into the cloud native ecosystem, with features like attached SBOMs, cosign etc. daemon is proving to be harder to support. Daemon support has also proven to be hard to support during our BuildKit poc https://github.com/buildpacks/pack/pull/1314 and https://github.com/EricHripko/cnbp

which requires us to reimplement exporter in order to use it with BuildKit. All of these things would be easier if we supported oci image layout instead. GGCR, the core image manipulation library behind lifecycle, already supports loading and saving to oci layout format. Tools like podman support oci layout natively as well.

Given all of the points above, supporting oci layout in favor of daemon seems to make a lot of sense.

We should write out an RFC detailing how this would work while preserving backwards compatibility for tools like pack.

jjbustamante:
Resolves #197 

[Readable](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/publish-operation/text/0000-export-to-oci.md)

Signed-off-by: Juan Bustamante <jbustamante@vmware.com>

## Note

This RFC was initially proposed as a mechanism for removing the Daemon in the near future, in order to do that, a high level plan I proposed is as follows.

- Approve this RFC and add the export to OCI layout format feature in Lifecycle
- Implement in Pack a new workflow when the `-daemon` flag is enabled based on what we did with the PoC. The workflow could be enabled by the user in **experimental mode**
  - Make noise in the community so people can help us using it
  - Collect feedback about it
  - Correct issues until we feel confortable
- Turn off the actual `-daemon` implementation and replace it with the `-oci` by default in Pack
  - Maybe we can also add a new flag to **re-use the old implementation** in case the user needs it
  - Keep doing noise in the community
  - Collect feedback about it
  - Correct issues until we feel confortable
- Schedule the removal of the Daemon support in Lifecycle version X.Y.Z.
- Remove the actual daemon support from the code base for Lifecycle version X.Y.Z

I am not 100% we will remove the daemon support, but at least, I want to left this plan here to do not lose it.


text/0000-publish-operation.md
```diff
@@ -0,0 +1,144 @@
...
+- An [image index](https://github.com/opencontainers/image-spec/blob/main/image-index.md) is a higher-level manifest which points to a list of manifests and descriptors.
+- An [Image manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md) provides a configuration and set of layers for a single container image for a specific architecture and operating system.
+- A [config](https://github.com/opencontainers/image-spec/blob/main/descriptor.md) is a property references a configuration object for a container, by digest. It must support the following media type `application/vnd.oci.image.config.v1+json`
+
+# Motivation
+[motivation]: #motivation
+
+Because the [Image.Digest](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1#Image) method from the GCR library would compute the Digest value before pushing it to a Registry, it allows us to create the image and add  metadata require like annotations or even sign it and save it to disk and be sure that when the Image is push to the registry the Digest will not change. (See the [thread](https://cloud-native.slack.com/archives/C033DV9EBDF/p1644523524402149) on Slack).
```
jromero, line 34: I'm not understanding the motivation here. What's the problem that this is trying to solve?

jjbustamante: Hi @jromero , I rewrote the whole idea, I hope is more clear now

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,192 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+When the `Exporter` phase is invoked passing the `-daemon` flag  besides writing into the Daemon also save the image to disk in [OCI Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format updating the [report.toml](https://github.com/buildpacks/spec/blob/main/platform.md#reporttoml-toml) file with all the metadata require to validate the consistency of the image when it is published to a Registry.
```
jabrown85, line 15: Is there any reason we can't leave this as a practical example, but add `-layout oci` as the primary interface? I don't see any reason lifecycle should restrict outputting to disk if platforms want to do that in addition to publishing to a registry.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,192 @@
...
+└── oci/
+    └── my-app-image/
+        ├── blobs/
+        │   └── sha256/
+        │       └── 01..
+        ├── index.json
+        └── oci-layout
+
+```
+As we can see there is a new folder called `oci` and inside that folder we can find our application image in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format
```
jabrown85, line 76: How would lifecycle handle folders that are already present from a previous `-layout` execution? If `my-app-image/blobs/sha256/01` exists but is no longer a layer in the new app image, does the folder get left around?

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,192 @@
...
+
+Attempts to use this feature when the `-daemon` flag is not used could be ignore or show some warnings messages.
```
jabrown85, line 78: I don't think we need this if we allow a platform to use this `-layout` with any other configuration.

text/0000-export-to-oci.md
jjbustamante: I think the main reason was that the daemon is the only use case where the loosing of metadata is actually an issue.  When the image is pushed to a registry, we don't have problems adding annotations or signing the image. But I agree, offering this capability for any workflow could potentially open new possibilities for Platform implementors.

text/0000-export-to-oci.md
jjbustamante: This is actually an interesting question. Maybe it could depend on the final Cache implementation? if we managed to reimplement the cache based on OCI layout, maybe, one solution is just to clean the `<image>` folder before exporting to disk.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+Implementing this new feature will help us to solve the problem of loosing information when the image is saved into the Daemon keeping the image on disk along with the metadata it can be used as input for other tools to offer more capabilities to the end users.
+
+This feature will help to unblock uses cases like
+- OCI annotations. See [RFC](https://github.com/buildpacks/rfcs/pull/196)
+- Cosign integration. See [RFC](https://github.com/buildpacks/rfcs/pull/195)
+
+# What it is
+[what-it-is]: #what-it-is
+
+Currently the *Exporter*  writes either in an OCI image registry or a docker daemon, the idea is to add the capability to write into disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format only when the `-daemon` flag is used as argument AND the feature is enable using a new flag `-layout` or the default environment variable `CNB_LAYOUT_DIR`.
```
natalieparellano, line 40: We'll need to get the run image digest from the daemon, which may or may not be available: https://github.com/buildpacks/imgutil/issues/32

Should we assume that the digest is always available, and fail otherwise?

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+
+| Input             | Environment Variable  | Default Value            | Description
+|-------------------|-----------------------|--------------------------|----------------------
+| `<layout>`      |  `CNB_LAYOUT_DIR` | "" | The root directory where the OCI image will be written. The presence of a none empty value for this environment variable will enable the feature. |
+
+
+- When the exporter is executed WITH the flag `daemon` it will check for the presence of the flag `layout` or the environment variable `CNB_LAYOUT_DIR`
+- IF any of the values are present THEN
+  - It will create a folder name `<image>` located at the path defined `<layout>` or `CNB_LAYOUT_DIR`
+  - In parallel of exporting the Image to the Daemon it will save the final Image in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format in the directory created before. The image will be saved using **uncompressed** layers
```
natalieparellano, line 101: From 3/3 working group: it was suggested to save the layers compressed, since we'll always need to compress them (to get the digest) and it is more expensive to compress than decompress.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+The new information to be  added into the `report.toml` file can be summarize as follows:
+
+```toml
+[export]
+[[export.oci]]
+digest = "<image digest>"
+manifest-size = "<manifest size in bytes>"
+compression-algorithm = "<used by gcr>"
+library-url = "https://github.com/google/go-containerregistry/"
+library-language="go"
```
natalieparellano, line 117: If we do https://github.com/buildpacks/rfcs/pull/203/files#r819905382 this would go away.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+[drawbacks]: #drawbacks
+
+- We could increase the disk space if we not managed the duplication of saving the layers on disk. Currently the Cache implementation (used when daemon is ON) saved the layers tarballs on disk, because the current proposal is exporting the whole image on disk it will also require more space to save the layers for the OCI format in the `blobs` folder.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Redesign the current Cache
+
+Another potential solution could be to export the OCI image along with the current Cache concept and expose some contract for `Platform` to interact with this Cache and extract the final OCI image. At high level, the solution can be represented with the following container diagram from C4 model
```
natalieparellano, line 145:
```suggestion
Another potential solution could be to export the OCI image along with the current Launch Cache concept and expose some contract for `Platform` to interact with this Cache and extract the final OCI image. At high level, the solution can be represented with the following container diagram from C4 model
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+
+![](https://i.imgur.com/xl4gL1G.png)
+
+Notice that on this solution, because the Cache is an internal component from the Lifecycle implementation we will have to expose some kind of specification for `Platforms` to understand its format and been able to read the OCI image.
```
natalieparellano, line 149:
```suggestion
Notice that on this solution, because the Launch Cache is an internal component from the Lifecycle implementation we will have to expose some kind of specification for `Platforms` to understand its format and been able to read the OCI image.
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+-->
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- We could increase the disk space if we not managed the duplication of saving the layers on disk. Currently the Cache implementation (used when daemon is ON) saved the layers tarballs on disk, because the current proposal is exporting the whole image on disk it will also require more space to save the layers for the OCI format in the `blobs` folder.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Redesign the current Cache
```
natalieparellano, line 143:
```suggestion
## Redesign the current Launch Cache
```

natalieparellano: I'd be in favor of this solution (if it's possible, I haven't looked into all the details!). It would be nice not to effectively triplicate every daemon image (1 in daemon storage, 1 in launch cache, 1 in oci layout dir 😭 ).

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+}
+
+class Image {
+    <<interface>>
+
+}
+```
+### Drawbacks
+
+- We will have to expose Cache implementation details to the outside world, probably spec those details, for other tools to interact with this exported data
```
natalieparellano, line 188:
```suggestion
- We will have to expose Launch Cache implementation details to the outside world, probably spec those details, for other tools to interact with this exported data
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,213 @@
...
+    <<interface>>
+
+}
+
+class Image {
+    <<interface>>
+
+}
+```
+### Drawbacks
```
natalieparellano, line 186: Another drawback: the lifecycle still knows about the daemon 😞 . Maybe it's worth circling around and describing the reasons we found the alternative proposals unsatisfying. From what I can recall:

* oci-layout as the interface: lookups (the run image would need to be present on disk)
  * potentially made worse by dynamic run image selection w/ Dockerfiles
* registry-only as the interface: the need for platforms to re-implement daemon loading (pack, maven plugin, others?)
* registry-only as the interface: performance impact of having to write the entire image before loading into the daemon
...probably I missed some things

natalieparellano: cc @jromero @samj1912

text/0000-export-to-oci.md
jjbustamante: I am updating the RFC allowing to export no matter if you are exporting to a Daemon or to Registry. The idea of using a URI schema to specify different targets, it's something I will not include in this RFC because I think it could be handle later.

text/0000-export-to-oci.md
jjbustamante: Yes, I will remove this section

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,261 @@
...
+> tree .
+.
+└── launch-cache/
+    ├── committed/
+    │   ├── io.buildpacks.lifecycle.cache.metadata
+    │   ├── sha256:65d9067f915e01...tar
+    │   ├── sha256:6905011516dcf4...tar
+    │   └── sha256:83d85471d9f8a3...tar
+    ├── staging
+    └── my-app-image/
```
natalieparellano, line 79: This could still be a child of `/oci`, no? Since we are using symlinks, I would think we don't really have to care.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,261 @@
...
+        │   └── sha256/
+        │       ├── 65d9067f915e01...tar -> /launch-cache/committed/sha256:65d9067f915e01...tar
+        │       ├── 6905011516dcf4...tar -> /launch-cache/committed/sha256:6905011516dcf4...tar
+        │       └── 83d85471d9f8a3...tar -> /launch-cache/committed/sha256:83d85471d9f8a3...tar
+        ├── index.json
+        └── oci-layout
+
+```
+
+### Exporting to Daemon without launch cache enabled
```
natalieparellano, line 90: Though it's technically possible to not use a launch cache, in practice using the daemon without it is so painfully slow that I doubt this scenario ever occurs.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,261 @@
...
+
+Notice that we are relying on the OCI format Specification to expose the data for `Platforms`
+
+The following new input is proposed to be added to this phase
+
+| Input             | Environment Variable  | Default Value            | Description
+|-------------------|-----------------------|--------------------------|----------------------
+| `<layout>`      |  `CNB_LAYOUT_DIR` | "" | The root directory where the OCI image will be written. The presence of a none empty value for this environment variable will enable the feature. |
+
+Let's see the propose flow
```
natalieparellano, line 169: This flow chart is amazing :)

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,261 @@
...
+  N --> I[Write image to $export-dir in OCI format **]
+  I --> J[Calculate manifest's digest]
+  J --> K[/Write digest into report.toml/]
+  K -->END((End))
+```
+
+Notes:
+  - WHEN `-launch-cache` flow is executed
+    - The content of `blobs/<alg>/<encoded>` MAY contain symbolic links to content saved in the launch cache to avoid duplicating files.  
+    - The content of `blobs/<alg>/<encoded>` MAY reference tar files in **uncompressed** format because that's how they are saved in the cache
```
natalieparellano, line 203: We should probably change this to be consistent with the layout/registry case. We could uncompress the blobs when we need to save them into the daemon. We should probably understand the performance impact of this and whether simplicity is worth the tradeoff.

text/0000-export-to-oci.md
jjbustamante: Yes, my doubt is:
-   if I am referencing files that are saved in the launch cache (mounted in a volume by tools like Pack) then the final exported images MUST be in the same folder to guarantee we have access to those files

text/0000-export-to-oci.md
jjbustamante: Should we show a warning message to alert the user?

text/0000-export-to-oci.md
jjbustamante: My concern here is:
- Currently, the tar files are saved **uncompressed** in the launch cache. Was it done by design for some reason? I don't want to propose a breaking change without having the whole context of why it was done like this in the first place.
- I agree with you, the main reason is performance, maybe playing a little bit with the code I will get some hint about performance impact.

text/0000-export-to-oci.md
jjbustamante: After today's WG, I think this RFC is going in the right direction regarding the approaches discussed. 

- Adding the capability to export to OCI format to the Lifecycle opens the door to the other tooling we are considering (prepare, publish, etc)
  - It will also allow other Platforms to prepare their implementation once the daemon support is removed because the mechanism Lifecycle is giving them is the OCI format, so they can actually implement the new *loading into daemon* when this feature is available
  - It will also give us the chance to work in the tooling to help platforms to accomplish the goal

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+- A **tag reference** refers to an identifier of form `<registry>/<repo>:<tag>` which locates an image manifest in an [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/master/spec.md) compliant registry.
+- A **digest reference** refers to a [content addressable](https://en.wikipedia.org/wiki/Content-addressable_storage) identifier of form `<registry>/<repo>@<digest>` which locates an image manifest in an [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/master/spec.md) compliant registry.
+- A **image Manifest** provides a configuration and set of layers for a single container image for a specific architecture and operating system.
+- An [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) is the directory structure for OCI content-addressable blobs and location-addressable references.
+
+# Motivation
+[motivation]: #motivation
+
+Exporting to [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) will enable new user's workflows in the top of this functionality. For example:
```
natalieparellano, line 33:
```suggestion
Exporting to [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) will enable new user's workflows on top of this functionality. For example:
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+    │       ├── 68224c8aa806b9ec8cecf2282b7b10cdf1c615785652bcc85f3a79bd06e60384
+    │       ├── 6977801a3b8ac0c4a76ea22fa4dc0541fe8fc6ba964883ee0a21b5736f8acee9
+    │       ├── 6b534265b45a4c09a5e05dde5ccf2450ea505481658a1c9ae9e0aae32362e941 -> ../../../cnbs/sample-stack-run/blobs/sha256/6b534265b45a4c09a5e05dde5ccf2450ea505481658a1c9ae9e0aae32362e941
+    │       ├── 7c83d7d24ac43dbdd419d15abb849b6d155d0bc361eca0908a8ae5aefcd55557
+    │       └── addd2357f3a0175410ab8f9303e747cc72af9aaeb0e7402d3fd3f144adb29db5
+    ├── index.json
+    └── oci-layout
+```
+
+Also, we are proposing to use symbolic links to reference the blobs from the `run-image` this will help to save some hard drive space for the end user.
```
natalieparellano, line 167:
```suggestion
Also, we are proposing to use symbolic links to reference the blobs from the `run-image` as this will help to save some hard drive space for the end user.
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+```=shell
+> /cnb/lifecycle/exporter -oci -daemon -run-image cnbs/sample-stack-run:bionic my-app-image
+
+ERROR: exporting to multiples target is not allowed
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The lifecycle phases affected by this new behavior are:
```
natalieparellano, line 182: Restore as well, since we'll need to read previous image metadata from the disk and also copy the SBOM layer (if present) from the disk.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+ - [Analyze](https://buildpacks.io/docs/concepts/components/lifecycle/analyze/)
+ - [Export](https://buildpacks.io/docs/concepts/components/lifecycle/export/)  
+ - [Create](https://buildpacks.io/docs/concepts/components/lifecycle/create/)
+
+At high level view the proposed solution can be summarize with the following system landscape diagram from the C4 model
```
natalieparellano, line 187:
```suggestion
At a high level view the proposed solution can be summarized with the following system landscape diagram from the C4 model
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+ | `<oci>`      |  `CNB_USE_OCI` | false | Analyze or Export image from/to OCI layout format on disk |
+ | `<oci-dir>` | `CNB_OCI_PATH` | /oci | Path to oci directory where the images are saved |
+
+- WHEN `the new flag -oci or the default environment variable CNB_USE_OCI are set to true` during the phase invocation THEN the feature will be enabled.
+The image look up will be done following these rules:
+  - WHEN `the image points to a tag reference`
+    - Lifecycle will load the image from disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `<oci-dir>/<registry>/<repo>/<tag>`
+  - WHEN `the image points to a digest reference`
+    - Lifecycle will load the image from disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `<oci-dir>/<registry>/<repo>/<digest>`
+  - A part from the look up, the logic for each phase should remain the same
```
natalieparellano, line 206:
```suggestion
  - Apart from the look up, the logic for each phase should remain the same
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Migration
+[migration]: #migration
+
+- No breaking changes were identified
```
natalieparellano, line 211: Ah, maybe my suggestion [here](https://github.com/buildpacks/rfcs/pull/203/files#r891354463) was wrong - does this RFC propose removing the daemon? I see daemon removal as one of the motivations. If we are going to propose removing the daemon, there are definitely breaking changes! Are we deferring that discussion to a later RFC? To me, export-to-oci and daemon-removal feel very intertwined.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- We could increase the disk space if we do not managed the duplication of saving the layers on disk. The proposal suggest to use symbolic links to reference layers on disk and avoid duplication.
```
natalieparellano, line 216:
```suggestion
- We could increase the disk space if we do not manage the duplication of saving the layers on disk. The proposal suggests to use symbolic links to reference layers on disk and avoid duplication.
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Doing nothing, just keep exporting only to Daemon or Registry
+- Why is this proposal the best? [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format is an standard from which other tools can create a [OCI Runtime Specification bundle](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/bundle.md) exporting to this format enables Platforms to implement any feature in the top of this format
```
natalieparellano, line 223:
```suggestion
- Why is this proposal the best? [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format is a standard from which other tools can create a [OCI Runtime Specification bundle](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/bundle.md) exporting to this format enables Platforms to implement any feature in the top of this format
```

text/0000-export-to-oci.md
```diff
@@ -225,6 +225,45 @@ The following Dynamic Diagram from the C4 model, can give a little idea of the p
 
 ![](https://i.imgur.com/SkY3l62.png)
 
+### Measuring of performance impact
+
+In order to have an idea on how much is affected the performance of exporting to the Daemon using the OCI layout format, the following metrics were took.
```
natalieparellano, line 230:
```suggestion
In order to have an idea on how much is affected the performance of exporting to the Daemon using the OCI layout format, the following metrics were taken.
```

text/0000-export-to-oci.md
```diff
@@ -225,6 +225,45 @@ The following Dynamic Diagram from the C4 model, can give a little idea of the p
...
+Legend:
+- † Exporting run-image from Daemon to disk in OCI layout format
+- ‡ Exporting application image to disk in OCI layout format
+- ⁜ Exporting application image from disk to Daemon
+
+Notes:
+- I had some issues with the second build, that's why I couldn't take any measure
+- We divide the results in two sections:
+  - Pre/Post Lifecycle execution: As mentioned, skopeo tool was used here and most of the time spent goes into this category. Open to optimal implementations here.
+  - Lifecycle execution: We can see there is an increment of +10%, but probably this can be improve during the formal implementation
```
natalieparellano, line 264:
```suggestion
  - Lifecycle execution: We can see there is an increment of +10%, but probably this can be improved during the formal implementation
```

text/0000-export-to-oci.md
jjbustamante: Added

text/0000-export-to-oci.md
```diff
@@ -225,6 +225,45 @@ The following Dynamic Diagram from the C4 model, can give a little idea of the p
...
+| Kotlin  | 305.98 | 2.39s † <br> 4.73s ‡ (+10.77%) <br> 3.38s ⁜ <br> =10.5s  | -  |
+| Ruby  | 100.52  |  2.59s † <br> 1.31s ‡ (+13.91%) <br> 0.89s ⁜ <br> =4.79s  | -  |
+
+Legend:
+- † Exporting run-image from Daemon to disk in OCI layout format
+- ‡ Exporting application image to disk in OCI layout format
+- ⁜ Exporting application image from disk to Daemon
+
+Notes:
+- I had some issues with the second build, that's why I couldn't take any measure
```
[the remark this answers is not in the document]
jjbustamante: I added some detail about the second build error

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,332 @@
...
+[drawbacks]: #drawbacks
+
+- We could increase the disk space if we do not manage the duplication of saving the layers on disk. The proposal suggests to use symbolic links to reference layers on disk and avoid duplication.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Doing nothing, just keep exporting only to Daemon or Registry
+- Why is this proposal the best? [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format is a standard from which other tools can create a [OCI Runtime Specification bundle](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/bundle.md) exporting to this format enables Platforms to implement any feature in the top of this format
```
jromero, line 309: Should we add some examples of what some of those use cases may be? For example, [the  containerd request](https://github.com/buildpacks/lifecycle/issues/829) comes to mind. Loading to containerd seems more achievable if we had an OCI exported format.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,246 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Add the capability to the `Exporter` phase to save the image to disk in [OCI Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format.
```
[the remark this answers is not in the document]
jjbustamante: No, that's why I intentionally avoid merging this suggestion 🙃

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,332 @@
...
+Or
+
+```=shell
+>  /cnb/lifecycle/exporter -oci -run-image cnbs/sample-stack-run:bionic my-app-image
+```
+
+The expected output is the application `<image>` exported in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format
+
+```=shell
+├── cnbs
```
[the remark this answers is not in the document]
jjbustamante: Sounds good to me, I will add it

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,332 @@
...
+  - Make noise in the community so people can help us using it
+  - Collect feedback about it
+  - Correct issues until we feel confortable
+- Turn off the actual `-daemon` implementation and replace it with the `-oci` by default in Pack
+  - Maybe we can also add a new flag to **re-use the old implementation** in case the user needs it
+  - Keep doing noise in the community
+  - Collect feedback about it
+  - Correct issues until we feel confortable
+- Schedule the removal of the Daemon support in Lifecycle version X.Y.Z.
+- Remove the actual daemon support from the code base for Lifecycle version X.Y.Z
```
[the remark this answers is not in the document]
jjbustamante: I added, as a suggestion from @natalieparellano, to have conversations about removing the Daemon.

text/0000-export-to-oci.md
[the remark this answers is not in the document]
jjbustamante: Added

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,332 @@
...
+
+```=shell
+> export CNB_OCI_PATH=true
+> /cnb/lifecycle/exporter -run-image cnbs/sample-stack-run:bionic my-app-image
+```
+
+Or
+
+```=shell
+>  /cnb/lifecycle/exporter -oci -run-image cnbs/sample-stack-run:bionic my-app-image
```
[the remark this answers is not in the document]
jjbustamante: I've been thinking on this, I have the following idea: All the conversion logic should be handled by the platform right? because they have to **prepare** the inputs for lifecycle. Now, from the Lifecycle side, I like the idea to consider *asking for paths only*, based on that,  I am wondering the following:

Checking the **creator** phase (because that one has all the parameters required by the other phases), these are the current arguments (relevant for this RFC):

```=bash
 -cache-dir string
    	path to cache directory
  -cache-image string
    	cache image tag name
  -previous-image string
    	reference to previous image
  -run-image string
    	reference to run image
```
Also the **image name** it's important as you mentioned. We can add the following new arguments, valid only if the flag `oci` is enabled:

```=bash
  -previous-image-dir string
    	path to the previous image in OCI layout format. Relative to <oci-dir>
  -run-image-dir string
    	path to the run image in OCI layout format. Relative to <oci-dir>
  -image-dir string
       path to save the image in OCI layout format. Relative to <oci-dir>
```
In this way, we can use the actual **image name** as reference to update `org.opencontainers.image.ref.name`

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,332 @@
...
+Legend:
+- † Exporting run-image from Daemon to disk in OCI layout format
+- ‡ Exporting application image to disk in OCI layout format
+- ⁜ Exporting application image from disk to Daemon
+
+Notes:
+- I had some issues with the second build, that's why I couldn't take any measure
+  - This an example of the second build error
+    ```=bash
+    [exporter] ERROR: failed to export: saving image: creating layer from /oci/sample-ruby-app/blobs/sha256:0d1ae00330e04bac2e7f6da6dadc8d234198a3422b6e3412103c8b379e00eaad: open /oci/sample-ruby-app/blobs/sha256:0d1ae00330e04bac2e7f6da6dadc8d234198a3422b6e3412103c8b379e00eaad: no such file or directory
```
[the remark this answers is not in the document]
jjbustamante: I did a fix for this issue, and update the RFC with the metrics.

text/0000-export-to-oci.md
```diff
@@ -252,22 +252,16 @@ I tried different approaches, the best results were found when using the [stream
 
 | Sample  | Image size (MB)|  First Build |  Second Build |
 |---|---|---|---|
-| Java  | 238.1  | 2.41s † <br> 4.76s ‡ (+10.96%) <br> 3.21 ⁜ <br> =10.38s | - |
-| Kotlin  | 305.98 | 2.39s † <br> 4.73s ‡ (+10.77%) <br> 3.38s ⁜ <br> =10.5s  | -  |
-| Ruby  | 100.52  |  2.59s † <br> 1.31s ‡ (+13.91%) <br> 0.89s ⁜ <br> =4.79s  | -  |
+| Java  | 238.1  | 2.41s † <br> 4.76s ‡ (+10.96%) <br> 3.21 ⁜ <br> =10.38s | 3.47s † <br> 4.98s ‡ (+55.63%) <br> 3.02 ⁜ <br> = 11.47s|
+| Kotlin  | 305.98 | 2.39s † <br> 4.73s ‡ (+10.77%) <br> 3.38s ⁜ <br> =10.5s  | 3.47s † <br> 5.11s ‡ (+61.20%) <br> 3.22s ⁜ <br> = 11.8s  |
+| Ruby  | 100.52  |  2.59s † <br> 1.31s ‡ (+13.91%) <br> 0.89s ⁜ <br> =4.79s  | 2.44s † <br> 1.51s ‡ (+122.06%) <br> 1.13s ⁜ <br> = 5.08s  |
```
natalieparellano, line 257: It's interesting to note that second builds are actually slower than the first. I think this must be due to the re-uploading of ALL the layers vs. just the changed ones (the imgutil hack).

natalieparellano: Ah, I should have been more clear - for second builds, I was expecting (if there aren't many changed layers) that ⁜ times would be significantly faster.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,339 @@
...
+  ![](https://i.imgur.com/vtOjxJP.png)
+
+   -  The results are actually very similar to exporting to OCI layout format for Java and Kotlin, but Ruby application is actually worst.
+
+  ![](https://i.imgur.com/FfbqfF6.png)
+
+   - Second build is actually better compared with the export to OCI in disk, Java and Kotlin increases the time just by **2%**, but Ruby again is worst
+
+
+- Why is this proposal the best? [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format is a standard from which other tools can create a [OCI Runtime Specification bundle](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/bundle.md) exporting to this format enables Platforms to implement any feature in the top of this format
```
jromero, line 316: I would add that, managing an ephemeral registry comes with the additional complications:

1. Process Management: Platforms must not manage a parallel process (registry in the daemon). This would entail ensuring that the registry is started and cleaned up appropriately.
2. Networking: There are additional network complications in order to route images to the ephemeral registry. For example, [network drivers](https://docs.docker.com/network/#network-drivers), [proxy](https://docs.docker.com/desktop/networking/#httphttps-proxy-support) and [DNS configuration](https://docs.docker.com/config/containers/container-networking/#dns-services), [host name resolution](https://docs.docker.com/desktop/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host), and [TLS certificates](https://betterprogramming.pub/deploy-a-docker-registry-using-tls-and-htpasswd-56dd57a1215a) to name a few.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,339 @@
...
+
+Times are expressed in **seconds** and the first thing we noticed is for Java and Kotlin the `build time` can be affected by the network and the availability of maven repositories, so I decided to take the `same build time` to compare both approaches.
+
+Here are my thoughts about these results:
+- Java and Kotlin behavior are very similar, exporting only to OCI format increases 5% the time compared to Daemon approach, and from user perspective it represents a 5 seconds increase of time.
+- On the other hand for the Ruby application, exporting to OCI format represents a  20% increase of the time but from user perspective it is only 1.5 seconds which is probably difficult to notice from user perspective.
+- When the time spent for Pack to prepare the environment for the lifecycle execution (downloading the run-image from a registry to OCI format) and loading the OCI image from disk to the Daemon (which is the expected behavior from Users) is added, then:
+  - The Java and Kotlin applications time increases was **13%**, representing **+13 seconds** from user perspective
+  - The Ruby application increases **82%** but from user's side it represents **7 seconds**
```
jromero, line 252: Do we know why this increase exists? Are there any potential solutions to improve the performance?

text/0000-export-to-oci.md
jjbustamante: Added. Thanks @jromero

text/0000-export-to-oci.md
jjbustamante: I think with the latest table where I summarized the results we can see the difference came from
- Skopeo -> OCI
- OCI -> Skopeo

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,343 @@
...
+
+The following Dynamic Diagram from the C4 model, can give a little idea of the pieces implemented during the Poc
+
+![](https://i.imgur.com/SkY3l62.png)
+
+### Measuring of performance impact
+
+In order to have an idea on how much is affected the performance of exporting to the Daemon using the OCI layout format, the following metrics were taken.
+
+Using a local workstation with the following specifications:
```
jabrown85, line 236: Can we get a linux workstation setup in a cloud to run these tests on? The disk virtualization layer in macOS is known for being pretty slow in certain situations. I'm curious if OCI as the default would be much faster on linux with docker daemon.

jabrown85 (approved):
I think the feature makes sense for a lot of use cases even if we never end up replacing the daemon usage for typical desktop `pack` users.

text/0000-export-to-oci.md
jjbustamante: Well, I have a very good personal machine running Pop OS, I will run there! thanks for the idea!

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,343 @@
...
+  │       │       ├── 63882e64890d1adea5824dbb9bc6a812140d6b85752bf74a0a26fca78e1baf5a
+  │       │       ├── 63cb781c9d22ed765584437fea6db568f79be248ce40fc6695017d3ef3e8caaa
+  │       │       └── 6b534265b45a4c09a5e05dde5ccf2450ea505481658a1c9ae9e0aae32362e941
+  │       ├── index.json
+  │       └── oci-layout
+  └── my-app-image
+      ├── blobs
+      │   └── sha256
+      │       ├── 14eaea7168b1fc4b8b30f7a20f7609335cc3dbcfb6d4c1afeb1e5daefd26cdf9
+      │       ├── 219480c73c20efd82d425207e7555eba09cb113c845250cef09ba376e1dd506e -> ../../../cnbs/sample-stack-run/blobs/sha256/219480c73c20efd82d425207e7555eba09cb113c845250cef09ba376e1dd506e
```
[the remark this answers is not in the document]
jjbustamante: No, the idea is the `lifecycle` should find all the required layers on disk in OCI and the platform must take care of that by downloading them. 

In your suggested scenario, the `exporter` should throw an error because the run-image will not be found on the disk, but I feel tempted to take the approach I described [here](https://github.com/buildpacks/rfcs/pull/203#discussion_r896144953), where the inputs for the phases are explicit with a `param-dir`

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,343 @@
...
+For each case, I will present two ways of invoking the new feature:
+
+- Using the environment Variable
+- Using the new flag
+
+For both ways the expected output is the same
+
+### Analyzing run-image from OCI layout format
+
+Let's suppose the `<oci-dir>` or `CNB_OCI_PATH` environment variable points to a  directory that contains the `run-image` with the following format
```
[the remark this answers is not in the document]
jjbustamante: An error should be thrown. I will add that

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,343 @@
...
+The following Dynamic Diagram from the C4 model, can give a little idea of the pieces implemented during the Poc
+
+![](https://i.imgur.com/SkY3l62.png)
+
+### Measuring of performance impact
+
+In order to have an idea on how much is affected the performance of exporting to the Daemon using the OCI layout format, the following metrics were taken.
+
+Using a local workstation with the following specifications:
+- **(MacOS 12.3.1 / 2,4 GHz 8-Core Intel Core i9 / 32 GB 2667 MHz DDR4)**
```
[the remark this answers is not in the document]
jjbustamante: Done.

text/0000-export-to-oci.md
jjbustamante: I think the new table summarized better the results

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,410 @@
...
+- A **digest reference** refers to a [content addressable](https://en.wikipedia.org/wiki/Content-addressable_storage) identifier of form `<registry>/<repo>@<digest>` which locates an image manifest in an [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/master/spec.md) compliant registry.
+- A **image Manifest** provides a configuration and set of layers for a single container image for a specific architecture and operating system.
+- The **layer diffID** is the hash of the uncompressed layer
+- The **layer digest** is the hash of the compressed layer.
+- An [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) is the directory structure for OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references.
+
+# Motivation
+[motivation]: #motivation
+
+### Why should we do this?
```
Choobz, line 35: (This is maybe not the correct place for this comment, feel free to delete it if it's outside the scope of this PR)

This RFC and its associated implementation is really huge for end users, even if it may not pick up that much traction yet.

Right now, the current implementation offer 2 mains way to handle the buildpack resulting image:
* Set up your daemon to be able to push to  a registry
* Use another tool (any kind of docker client) to transform it into an artifact ([usually tar ball](https://github.com/buildpacks/lifecycle/issues/423))

Which is fine and dandy on your own machine but become quite a nightmare once implemented on Corporate Entreprise™ CI/CD. For the rest of my post, assume I'm moslty talking about containerized CI/CD which are becoming ubiquitous.

**Solution 1: Set up your daemon to be able to push to  a registry**
During the build phase of projects, organisation want to be able to segregate these environnements from the rest. They tend to be untrustable by nature (way too many thing are required to build projects nowadays). 
As most companies tend to be siloed, analysis tools are not available at build stage, so no analysis of the resulting image is possible before being pushed. 
For the same reason, build fragment tend to be maintained by developers with less security knowledge than the ones maintained by infrastructure/security teams.
You need to set up your "build" daemon to push to production registries [with patch that may end up exposing secrets really fast](https://stackoverflow.com/questions/46418652/docker-pull-returns-denied-access-forbidden-from-private-gitlab-registry/46422186#46422186). 
Overral allowing "build daemons" to push to registries end up being problematic really fast.

**Solution 2: Use another tool (any kind of docker client) to transform it into an artifact**
That's another can of worm. Most base build image don't come with a docker client by default. Which is great because it reduce the surface of attack drastically. But when you suddenly need one, it becomes quite a challenge to do that in a secure way:
* Creating base image with build tools  AND docker client. That's an organisational challenge for most company. These images tend to become unsecured quite fast. Either by being not up to date or because they become God Like Image with access to everything everywhere. We are looking at buildpack as a way to reduce our surface of attack drastically both at runtime and build time.
* Using 3rd party client/plugin to do that for you. [Exemple for maven plugin](https://github.com/fabric8io/docker-maven-plugin). That's a lot of added complexity and risk. 

Being able to generate a local artifact to be pushed outside the build time to "analyse steps" and "registries handling steps" maintained by teams better equiped to handle security and infrastructure would be really great.

So I wholeheartedly approve this PR. Thumb up, emoji heart, keep up the good work 😄.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,410 @@
...
+```=shell
+> export CNB_USE_OCI=true
+> /cnb/lifecycle/exporter -run-image cnb/my-stack-run:bionic my-app-image
+
+# OR
+
+>  /cnb/lifecycle/exporter -oci -run-image cnb/my-stack-run:bionic my-app-image
+
+# expected output
+# the store folder will be updated with the application image as follows
```
natalieparellano, line 210: Did we talk about what level of compression (if any) is applied to the layers? Can this be configurable?

text/0000-export-to-oci.md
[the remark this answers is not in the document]
natalieparellano: I'd be okay removing this for now if it will just complicate the discussion. We could add it as a comment on the RFC to not lose the information.

text/0000-export-to-oci.md
jjbustamante: @Choobz Thanks a lot for your feedback, I really appreciate it! and the use cases you mentioned are very valuable! I will try to incorporate them in the RFC

text/0000-export-to-oci.md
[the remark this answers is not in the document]
jjbustamante: I just copied that section into this RFC description.

text/0000-export-to-oci.md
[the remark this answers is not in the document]
jjbustamante: Done

natalieparellano:
WRT to the new RFC process, I'd be interested in stewarding this one

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,410 @@
...
+  - Reduce the Lifecycle complexity removing the interaction with the Daemon.
+  - Solve the problem of losing information when the image is saved into the Daemon, keeping the image on disk along with the metadata generated by the Lifecycle. The OCI Image can be used as input for other tools to offer more capabilities to the end users.
+  - This feature will help to unblock uses cases like
+    - OCI annotations. See [RFC](https://github.com/buildpacks/rfcs/pull/196)
+    - Cosign integration. See [RFC](https://github.com/buildpacks/rfcs/pull/195)
+    - Export to tarball. See [issue](https://github.com/buildpacks/lifecycle/issues/423)
+
+### What is the expected outcome?
+
+Lifecycle will be capable of exporting application image into disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format.
```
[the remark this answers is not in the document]
jjbustamante: @natalieparellano could you take a look at what I just added?

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,576 @@
...
+
+The target personas affected by this change are:
+
+- **Platform implementors**: they will have to take care of the responsibility of creating a store resource on disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format and pass it through the lifecycle during the phases execution.  
+
+The process of writing any image on disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format could be expensive in terms of hard drive space or IO operation (compressing or uncompressing layers). In order to provide flexibility for the implementation, the `analyzer` or `exporter` binaries only require the *Image Manifest* and the *Image Config* to execute their operations, based on this, we proposed the Lifecycle can be configured to work with a partial representation of the images on disk, meaning that some blobs MAY be missing (which is ok according to the [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format). The missing blobs SHOULD be those that are already available in a daemon or registry.
```
natalieparellano, line 70:
```suggestion
The process of writing any image on disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format could be expensive in terms of hard drive space or IO operation (compressing or uncompressing layers). In order to provide flexibility for the implementation, the `analyzer` or `exporter` binaries only require the *Image Manifest* and the *Image Config* to execute their operations on the previous image and run image; based on this, we proposed the Lifecycle can be configured to work with a partial representation of the images on disk, meaning that some blobs MAY be missing (which is ok according to the [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format). The missing blobs SHOULD be those that are already available in a daemon or registry.  
```

natalieparellano: (for clarity)

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,576 @@
...
+![](https://i.imgur.com/y972lTD.png)
+
+Notice that we are relying on the OCI format Specification to expose the data for `Platforms`
+
+The following new inputs are proposed to be added to these phases
+
+ | Input | Environment Variable  | Default Value | Description
+ |-------|-----------------------|---------------|--------------
+ | `<oci>`      |  `CNB_USE_OCI` | false | Analyze or Export image from/to OCI layout format on disk |
+ | `<oci-dir>` | `CNB_OCI_PATH` | /oci | Path to oci directory where the images are saved |
```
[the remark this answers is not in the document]
jjbustamante: I think Javier started using `layout` and for some reason I can't remember I switched to `oci`. But I can come back to `layout` you are right about that the image on the registry in also OCI.

I also remembered I tried to keep just one flag, but also faced some edge case during the PoC implementation, but I am happy to simplify it!

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,576 @@
...
+
+- WHEN `the new flag -oci or the default environment variable CNB_USE_OCI are set to true` during the phase invocation THEN the feature will be enabled.
+The image look up will be done following these rules:
+  - WHEN `the image points to a tag reference`
+    - Lifecycle will load the image from disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `<oci-dir>/<registry>/<repo>/<tag>`
+  - WHEN `the image points to a digest reference`
+    - Lifecycle will load the image from disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `<oci-dir>/<registry>/<repo>/<digest>`
+  - Apart for the image look up, the logic for each phase should remain the same and the lifecycle will write the layers on disk according to the following rules:
+    - Any layer generated by a buildpack SHOULD always be written to disk
+    - Layers from `base-image` or `run-image` MAY be written on disk if they were provided
```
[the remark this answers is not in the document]
jjbustamante: Yeah, I think we can be more creative on platform side to make users' experience more simple

text/0000-export-to-oci.md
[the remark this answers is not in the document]
jjbustamante: I think I remembered why I added the two flags. I wanted to make life easier for platforms implementors.

- Having a default root `<oci-dir>` I wanted to avoid platforms implementors to mount different folders when lifecycle is invoked, something similar to `layers` root folder, it's like, this is my root OCI storage folder.
- Because  I defined  a default value to `<oci-dir>` (my root storage folder), the `CNB_USE_OCI` allows platforms implementors to use the feature just enabling this environment variable.

Thinking in the suggestion `oci:/some/layout/path` makes me think on how to defined inputs to the lifecycle like `run-image` will be something like `-run-image oci:/foo/bar/path` or only the output image having the format `oci:/some/layout/path` will be enough we will assume inputs will be resolved based on the rules I described in the RFC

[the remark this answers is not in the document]
natalieparellano: To summarize from Working Group, there is some concern that allowing the image to be specified with `oci:` prefix may lead users to assume they can "mix and match" inputs and outputs (e.g., run image in registry and app image on disk) which could be confusing since we are not planning to support that at this stage. I'm not wedded to the idea - would be happy to go with whatever is the consensus.

jkutner (approved):

samj1912:
https://twitter.com/tonistiigi/status/1612885639854460929?s=20 related new feature in moby about OCI support.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,715 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- Discussion around removing the Daemon support [RFC](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/deprecate-daemon/text/0000-deprecate-daemon.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+  - Tools like [umoci](https://umo.ci/) used to create a runtime bundle from an image in OCI layout format, requires the [annotation](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) `org.opencontainers.image.ref.name` to be present. Also tools like [skopeo](https://github.com/containers/skopeo) when an image is `copy` in oci format the annotation is included.
```
jabrown85, line 676: My vote would be to write this annotation. It feels weird to not be able to interact with other oci layout tooling.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,715 @@
...
+  We are not adding the annotation as part of the Buildpacks Specification, but in this case this could make our output incompatible with some other tooling.
+  - Depending on the previous question, the rules for mapping an image reference into a path could change, also validations that the Lifecycle could do to verify the `org.opencontainers.image.ref.name` against the value provided in the image reference.
```
jabrown85, line 678: Can you explain a bit more about what problems would arise? Do we have to validate anything at first?

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,715 @@
...
+  - Exporting to a tarball can be handle on this RFC or a new one must be created?
```
jabrown85, line 679: I don't think it needs to be tied to this. I think if we decide to write a tar there could be a pack specific RFC  (if even that is needed - is a PR enough for an additive feature?)

text/0000-export-to-oci.md
jjbustamante: Yes, right now the rules to map the `image.ref` to a path are:

- `<registry>/<repo>/<image>:<tag>` is converted into:
  -  `<layout-dir>/<registry>/<repo>/<image>/<tag>`

- `<registry>/<repo>/<image>@<algorithm>:<digest>` is converted into:
  - `<layout-dir>/<registry>/<repo>/<image>/<algorithm>/<digest>`

My main concern was if we handle the `org.opencontainers.image.ref.name` annotation then the first case could be converted into `<layout-dir>/<registry>/<repo>/<image>` and lifecycle could validate if the `<tag>` provided matches the `org.opencontainers.image.ref.name` annotation of the image at path `<layout-dir>/<registry>/<repo>/<image>` BUT thinking a little bit more, I think the validation can exists, but the mappings rules are fine, because if I want to use `ubuntu:bionic` or `ubuntu:jammy` if I do not include the bionic or jammy tag as part of the path, then it could be complicated to resolve the image. 

Don't worry about this statement! I think it is not going to change the mapping rules. 😄

text/0000-export-to-oci.md
jjbustamante: Would be fine to keep that as implementation detail and do not add it to the RFC ?

text/0000-export-to-oci.md
jjbustamante: Well my concern was because of this: 

If we want the image to be exported as a tarball, then the `Image.SaveFile()` must be called, so it means we need to expose in the `exporter` something to do that. I added the flag `-layout` to tell the export the image must be export to OCI layout format, so my concern was, if we want that for a tarball, then something else needs to be expose. That's why I was thinking on a RFC

text/0000-export-to-oci.md
natalieparellano: This would introduce a slight difference between layout images and remote images. Is that an acceptable difference? I'm not aware of any others.

jjbustamante: I created a [PR](https://github.com/buildpacks/imgutil/pull/175)  to add it

natalieparellano: Would it be reasonable to add this to remote images as well?

natalieparellano: FWIW, `crane pull <image> <path> --format=oci` doesn't add the annotation... I don't know what to make of that.

jjbustamante: It does, if you add ` --annotate-ref` flag

text/0000-export-to-oci.md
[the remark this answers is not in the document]
natalieparellano: > I think Javier started using layout and for some reason I can't remember I switched to oci

`crane` uses `--format=oci`, maybe that's why

text/0000-export-to-oci.md
natalieparellano: When exporting to layout we can derive `org.opencontainers.image.ref.name` from the target path, but maybe we could also add an optional `-annotate-ref` (or something) to the lifecycle so that when publishing to the registry we could optionally add the annotation if users want it. This would allow us to keep parity for remote & layout images which seems desirable.

text/0000-export-to-oci.md
natalieparellano: Maybe we could derive the format from the target path (e.g., if it has .tar or .tgz or whatever) or from some prefix (such as `oci-archive:`) but I agree that could be done separately from this RFC.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,707 @@
...
+# OR
+
+> /cnb/lifecycle/analyzer -layout -run-image cnb/cnb/my-partial-stack-run:bionic my-app-image
+```
+
+expected analyzed.toml output
+
+```=toml
+[run-image]
+  reference = "/layout-repo/index.docker.io/cnb/my-full-stack-run/bionic@sha256:fab3bb83de466ed29d7e9dcfdbee5b5fb2ff90e91bc849af85b261b4c2062a7a"
```
natalieparellano, line 163:
```suggestion
  reference = "/layout-repo/index.docker.io/cnb/my-partial-stack-run/bionic@sha256:fab3bb83de466ed29d7e9dcfdbee5b5fb2ff90e91bc849af85b261b4c2062a7a"
```

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,707 @@
...
+## How to map an image reference into a path in the layout repository
+
+In the previous examples one key element was how to translate an image reference into a path to look for in the `<layout-dir>`, let's define those rules.
+
+Considering an **image reference** refers to either a tag reference or digest reference. It could have the following formats
+- A tag reference refers to an identifier of form `<registry>/<repo>/<image>:<tag>`
+- A digest reference refers to a content addressable identifier of form `<registry>/<repo>/<image>@<algorithm>:<digest>`
+
+The image look up will be done following these rules:
+  - WHEN `the image points to a tag reference`
```
natalieparellano, line 356:
```suggestion
  - WHEN `the image points to a name reference`
```
Maybe also worth adding - when `<tag>` is not provided it defaults to `:latest`

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,707 @@
...
+
+The process of writing any image on disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format could be expensive in terms of hard drive space or IO operation (compressing or uncompressing layers). In order to provide flexibility for the implementation, the `analyzer` or `exporter` binaries only require the *Image Manifest* and the *Image Config* to execute their operations on the previous image and run image; based on this, we proposed the Lifecycle can be configured to work with a partial representation of the images on disk, meaning that some blobs MAY be missing (which is ok according to the [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format). The missing blobs SHOULD be those that are already available in a daemon or registry.  
+
+Let's see some examples of the proposed behavior
+
+## Examples
+
+### Requirements
+
+A folder on disk (accessible by the lifecycle) is required to execute the feature, this new folder works as a local registry and the content must be in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format.
```
natalieparellano, line 78: Perhaps worth noting, at least in the current implementation, the media type in index.json must be `application/vnd.docker.distribution.manifest.v2+json` or `application/vnd.oci.image.index.v1+json` (`skopeo copy -a docker://busybox:latest <dest>` will give you the latter, whereas `crane pull busybox:latest <dest> --format=oci` will give you `application/vnd.docker.distribution.manifest.list.v2+json` which will fail). This is at least worth calling out in our documentation in case users try to piece together inputs from their own tooling.

text/0000-export-to-oci.md
jjbustamante: @natalieparellano I have a couple of things regarding the way the annotation is going to be implemented. 
1. `creator` already have a `tag` flag which works in this case.
2. I am not 100% on deriving the *tag* from the path. For example, if I am pack, I will use the `layout` package exposed on imgutil to download the `run-image` and saving it on disk before calling the lifecycle. Pack can save the `run-image` on whatever path it wants but if I try to determine the `org.opencontainers.image.ref.name` base on the path I could be tagging wrong the image.

Suppose the `run-image` is `ubuntu:bionic` and pack wants to save the image on `/home/user/.pack/repo/ubuntu`. Pack could create the image as follows:

``` =go
# remote-image is v1.Image load from remote `ubuntu:bionic`
image, _ := layout.NewImage("/home/user/.pack/repo/ubuntu", layout.FromBaseImage(remote-image))
image.Save()
```
If we try to determine the `org.opencontainers.image.ref.name` from the path, we could tagging the image with 
``` =go
org.opencontainers.image.ref.name: "ubuntu"
```
**Proposed solution**

I think I prefer to delete the tagging when you call the `Save(additionalNames ...string)` method, so on my previous example:

#### example 1

```=golang
# remote-image is v1.Image load from remote `ubuntu:bionic`
image, _ := layout.NewImage("/home/user/.pack/repo/ubuntu", layout.FromBaseImage(remote-image))
image.Save()
```

output: 

No `org.opencontainers.image.ref.name` is added because we didn't pass any additional name

#### example 2

```=golang
# remote-image is v1.Image load from remote `ubuntu:bionic`
image, _ := layout.NewImage("/home/user/.pack/repo/ubuntu", layout.FromBaseImage(remote-image))
image.Save("bionic")
```

output:

```=json
org.opencontainers.image.ref.name: "bionic"
```

natalieparellano: This is just an opinion, but it is confusing to me that we are overloading the `Save` method in this way. It feels like `Save` has side effects and requires library consumers to know too much.

For the remote package, image.Save("other name") means "save my image and also put it here". The creator treats `-tag` this way (i.e., `-tag some.registry.io/my-image/my-tag` and not `-tag my-tag`. In the layout package, I would expect that image.Save("other path") would also put the image at "other path".

natalieparellano: What I would propose instead is a new method on the Image interface like Image.AnnotateRef("my-tag") that the lifecycle and others could call instead as needed.

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,691 @@
...
+
+In any case the expected output is the same.
+
+#### Analyze phase
+
+##### Analyzing run-image full saved on disk
+
+```=shell
+> export CNB_USE_LAYOUT=true
+> /cnb/lifecycle/analyzer -run-image cnb/my-full-stack-run:bionic my-app-image
```
natalieparellano, line 132: Shouldn't this contain a path?

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,691 @@
...
+
+[previous-image]
+  reference = "/index.docker.io/bar/my-previous-app/latest@sha256:aa0cf7fc8f161bdb96166c1644174affacd70d17f372373ca72c8e91116e2d43"
+
+```
+
+##### Analyzing run-image not saved on disk
+
+```=shell
+> export CNB_USE_OCI_LAYOUT=true
```
natalieparellano, line 191:
```suggestion
> export CNB_USE_LAYOUT=true
```
(this gives parity with `-layout` and `CNB_USE_DAEMON`

text/0000-export-to-oci.md
```diff
@@ -0,0 +1,691 @@
...
+- A name reference refers to an identifier of form `<registry>/<repo>/<image>:<tag>`
+- A digest reference refers to a content addressable identifier of form `<registry>/<repo>/<image>@<algorithm>:<digest>`
+
+The image look up will be done following these rules:
+  - WHEN `the image points to a name reference`
+    - Lifecycle will load/save the image from/to disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `/<registry>/<repo>/<image>/<tag>`
+  - WHEN `the image points to a digest reference`
+    - Lifecycle will load the image from disk in [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) format at `/<registry>/<repo>/<image>/<algorithm>/<digest>`
+  - WHEN `<registry>` is not provided default value will be **index.docker.io**
+    - IF `<repo>` is not also provided, then default value will be **library**
```
natalieparellano, line 357: Aren't we always expecting the lifecycle to be provided a path? And from there we can derive the name if needed?

Because otherwise we are always expecting the images to be in the `/` directory, which is doubtful.

jjbustamante: `<layout-dir>` was added again, without default value, now platforms MUST set this value and avoid mounting into `/` directory.

text/0000-export-to-oci.md
jjbustamante: I think this is a valuable comment when documenting the feature. I will add into the doc issue to do not forget about it

natalieparellano (approved):

natalieparellano:
Moving to status/voting closing next Wednesday 2/29

PR #204, opened 2022-02-24
Use SOURCE_DATE_EPOCH to configure image create time

natalieparellano:
[Readable](https://github.com/buildpacks/rfcs/blob/1a9018dcbadd888e25b716501a0ea85f00ada72d/text/0000-source-date-epoch.md)

### DIFF
--- a/text/0000-source-date-epoch.md
+++ b/text/0000-source-date-epoch.md
@@ -50,7 +50,7 @@ This [PR](https://github.com/buildpacks/imgutil/pull/137) to imgutil would read 
 
 Why should we *not* do this?
 
-It breaks build reproducibility. However platforms must explicitly opt-in to this feature.
+Platforms that choose to set SOURCE_DATE_EPOCH to real creation time will not have 100% reproducible builds.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-source-date-epoch.md
+++ b/text/0000-source-date-epoch.md
@@ -1,0 +1,84 @@
+# Meta
+[meta]: #meta
+- Name: Use SOURCE_DATE_EPOCH to configure image create time
+- Start Date: 2022-02-24
+- Author(s): natalieparellano
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+To allow for [build reproducibility](https://github.com/buildpacks/spec/blob/main/platform.md#build-reproducibility), images created by Cloud Native Buildpacks have a hard-coded create time of January 1, 1980. We have received requests from the community (see [here](https://github.com/buildpacks/pack/issues/1281) and [here](https://buildpacks.slack.com/archives/C94UJCNV6/p1643842965830459)) to allow for this value to be configurable. This RFC proposes making the value configurable by setting `SOURCE_DATE_EPOCH` in the lifecycle's execution environment during `export`.
+
+# Definitions
+[definitions]: #definitions
+
+* build reproducibility: identical inputs should produce identical outputs ([blog post](https://medium.com/buildpacks/time-travel-with-pack-e0efd8bf05db))
+* `SOURCE_DATE_EPOCH`: a "standardised environment variable that distributions can set centrally and have build tools consume this in order to produce reproducible output" (see [here](https://reproducible-builds.org/docs/source-date-epoch/))
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  Users have asked for it
+- What use cases does it support?
+  A meaningful image create time, e.g., the ability to determine which images are most recent
+- What is the expected outcome?
+  Images built with `SOURCE_DATE_EPOCH` set will not be reproducible, as the config blob will change
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Define the target persona: buildpack user, platform operator
+  * A buildpack user could pass a flag to `pack` to set `SOURCE_DATE_EPOCH` to the current time in the lifecycle's execution environment during `export`.
+  * A platform operator could choose to set `SOURCE_DATE_EPOCH` whenever `export` is run.
+
+- If applicable, describe the differences between teaching this to existing users and new users.
+  We should mention this feature in the lifecycle release notes and docs.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This [PR](https://github.com/buildpacks/imgutil/pull/137) to imgutil would read the environment variable at the point of saving the image. An alternative implementation could be to have the `exporter` read the variable and provide it via a new `SetCreatedAt()` method on the `imgutil.Image` interface. The latter might be safer as it would avoid unintended side effects for other consumers of imgutil (e.g., `pack`). 
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+It breaks build reproducibility. However platforms must explicitly opt-in to this feature.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  * Doing nothing
+  * Allowing layer timestamps to also be configurable, but this would effectively prohibit launch layer re-use and make builds slower
+- Why is this proposal the best?
+  * It is easy to implement, easy to use, and solves for the desired use case
+- What is the impact of not doing this?
+  * The lack of meaningful image create times makes it difficult to automate cleanup of images and could be a deal breaker for some users
+
+# Prior Art
+[prior-art]: #prior-art
+
+* See under "Reading the variable" [here](https://reproducible-builds.org/docs/source-date-epoch/)
+* [ko](https://github.com/google/ko#why-are-my-images-all-created-in-1970)
+* [jib](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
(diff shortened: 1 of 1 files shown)


text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+
+* See under "Reading the variable" [here](https://reproducible-builds.org/docs/source-date-epoch/)
+* [ko](https://github.com/google/ko#why-are-my-images-all-created-in-1970)
+* [jib](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+  * Should we require users to be on platform api 0.9 in order to use this feature? (Opinion: no, because this feature is invisible if the environment variable is not set. The motivation for spec'ing this in the platform api is to make it clear for platform operators how to use it.)
```
jabrown85, line 77: I agree - no need to be on newer platforms to use it.

jabrown85 (approved):

text/0000-source-date-epoch.md
samj1912: If we are not tying it to a specific platform API, I don't see why we should note this in the platform spec and not in our docs. It seems a weird that we are retroactively adding something to the past platform APIs but only documenting it for 0.9. I would be in favor of just documenting this in the lifecycle docs somewhere and implementing it in pack via a flag if we plan on introducing this for past platform APIs that have already been released.

samj1912: We follow something similar with our estargz support in lifecycle.

jromero: I disagree that this shouldn't be part of the specification. As platform implementors, we rely on the API to know what parameters we can use especially when it has specific side effects on the outcome. `estargz` support is an accident and can't be documented as supported because of the similar.

Essentially platforms that don't get to select their lifecycle implementations but rely on what's on the builder go from saying _"by doing X you get Y"_ to _"by doing X you MAY get Y"_.

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+To allow for [build reproducibility](https://github.com/buildpacks/spec/blob/main/platform.md#build-reproducibility), images created by Cloud Native Buildpacks have a hard-coded create time of January 1, 1980. We have received requests from the community (see [here](https://github.com/buildpacks/pack/issues/1281) and [here](https://buildpacks.slack.com/archives/C94UJCNV6/p1643842965830459)) to allow for this value to be configurable. This RFC proposes making the value configurable by setting `SOURCE_DATE_EPOCH` in the lifecycle's execution environment during `export`.
```
jromero, line 15: I dislike that this environment variable doesn't align with the `CNB_` prefix of all other environment variables. I get that it comes from a [convention](https://reproducible-builds.org/specs/source-date-epoch/) but I'm not sure that it justifies breaking away from our convention. Even `ko` has `KO_DATA_DATE_EPOCH `. What if we left `SOURCE_DATE_EPOCH` up to the platforms to decide support on but have `CNB_DATE_EPOCH` or something similar for our interface?

text/0000-source-date-epoch.md
jromero: To the original opinion regarding backward compatibility, nothing on the specification says a lifecycle implementation _CAN'T_ support an arbitrary environment variable.

samj1912: Yes, but it very much violates the whole point of having a platform API. If we start introducing new features for older platform APIs, the platform is now dependent on a specific lifecycle version to support a specific feature rather than knowing the feature set it will have available given a platform API.

Say you are a platform implementing this feature in an older platform api and you set this environment variable. You as a platform will never know whether or not the lifecycle actually respected the variable and set an appropriate date for the image or not as the lifecycle has no such contract for older apis.

If we implement this as part of a new platform api then there is an inherent contract that the platform can rely upon.

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  Users have asked for it
+- What use cases does it support?
+  A meaningful image create time, e.g., the ability to determine which images are most recent
+- What is the expected outcome?
+  Images built with `SOURCE_DATE_EPOCH` set will not be reproducible, as the config blob will change
```
hone, line 31: @ekcasey did bring up that if they use the same `SOURCE_DATE_EPOCH` it will be reproducible :)

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+* build reproducibility: identical inputs should produce identical outputs ([blog post](https://medium.com/buildpacks/time-travel-with-pack-e0efd8bf05db))
+* `SOURCE_DATE_EPOCH`: a "standardised environment variable that distributions can set centrally and have build tools consume this in order to produce reproducible output" (see [here](https://reproducible-builds.org/docs/source-date-epoch/))
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  Users have asked for it
+- What use cases does it support?
+  A meaningful image create time, e.g., the ability to determine which images are most recent
```
hone, line 29: Don't we have a slack thread that describes in detail a use case for the most recent image like pruning old images.

natalieparellano: [Spec comment](https://github.com/buildpacks/spec/pull/292#discussion_r804971942), [Slack thread](https://buildpacks.slack.com/archives/C94UJCNV6/p1643842965830459)

text/0000-source-date-epoch.md
natalieparellano: IIUC `KO_DATA_DATE_EPOCH` is used for a slightly different purpose: "used to set the modtime timestamp of the files in `KO_DATA_PATH`". However jib uses `USE_CURRENT_TIMESTAMP` to set image create time.

natalieparellano: What about an emoji vote? `SOURCE_DATE_EPOCH` ❤️  - `CNB_ SOURCE_DATE_EPOCH` 🎉

text/0000-source-date-epoch.md
natalieparellano: > the lifecycle has no such contract for older apis

Personally I think this is totally fine. The feature isn't mentioned in older platform apis, so it can't be relied upon. Use at your own risk.

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+  * A buildpack user could pass a flag to `pack` to set `SOURCE_DATE_EPOCH` to the current time in the lifecycle's execution environment during `export`.
+  * A platform operator could choose to set `SOURCE_DATE_EPOCH` whenever `export` is run.
+
+- If applicable, describe the differences between teaching this to existing users and new users.
+  We should mention this feature in the lifecycle release notes and docs.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This [PR](https://github.com/buildpacks/imgutil/pull/137) to imgutil would read the environment variable at the point of saving the image. An alternative implementation could be to have the `exporter` read the variable and provide it via a new `SetCreatedAt()` method on the `imgutil.Image` interface. The latter might be safer as it would avoid unintended side effects for other consumers of imgutil (e.g., `pack`).
```
ekcasey, line 46: > The latter might be safer as it would avoid unintended side effects for other consumers of imgutil (e.g., pack).

As convenient as it would be to merge the existing PR, I tend to agree.

text/0000-source-date-epoch.md
```diff
@@ -0,0 +1,84 @@
...
+[how-it-works]: #how-it-works
+
+This [PR](https://github.com/buildpacks/imgutil/pull/137) to imgutil would read the environment variable at the point of saving the image. An alternative implementation could be to have the `exporter` read the variable and provide it via a new `SetCreatedAt()` method on the `imgutil.Image` interface. The latter might be safer as it would avoid unintended side effects for other consumers of imgutil (e.g., `pack`). 
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+It breaks build reproducibility. However platforms must explicitly opt-in to this feature.
```
ekcasey, line 53:
```suggestion
Platforms that choose to set SOURCE_DATE_EPOCH to real creation time will not have 100% reproducible builds.
```

I don't think we need to use the word "breaks" here. We are still providing reproducible builds given identical inputs, the platform just isn't providing reproducible inputs :p

text/0000-source-date-epoch.md
ekcasey: I think we should spec the feature b/c it's existence determines whether a platform like `pack` can provide a new feature (a flag to use the real creation time) to users.

> the platform is now dependent on a specific lifecycle version to support a specific feature rather than knowing the feature set it will have available given a platform API.

Whatever decision we make here, I still think the platform API is the source of truth.

With my platform author hat on, I basically see the decision to support this in old APIs as irrelevant. I won't trust that I can use it in an old API, but also it doesn't hurt me. Maybe there are some edge cases where having it is helpful. E.g. I operate a very specific platform that uses a specific builder and I know my lifecycle version, maybe I go ahead and use this feature without checking the lifecycle version? But I am only tempted if I control the environment and know what I am doing.

That was a very long winded way of saying "whatever y'all can agree on is fine with me on this one".

ekcasey (approved):

text/0000-source-date-epoch.md
hone: :+1: on it being spec'd if we want it as part of the contract.

hone (approved):

natalieparellano:
With 3/5 implementation maintainers in favor, I think this can be in FCP

text/0000-source-date-epoch.md
natalieparellano: Adjusted the wording

natalieparellano:
@sclevine @ekcasey can this be merged please? I can confirm that issues have been created.

PR #211, opened 2022-03-10
Allow platforms to provide run image sboms

natalieparellano:
[Readable](https://github.com/buildpacks/rfcs/blob/npa/run-image-sbom/text/0000-run-image-sbom.md)

### DIFF
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -91,7 +91,7 @@ https://github.com/buildpacks/rfcs/pull/173 proposes allowing the run image to b
 Why should we *not* do this? Leaving the location of the run image sbom unspec'd (as opposed to baking it into the run image at a specific location) could make it harder for logic-less platforms like Tekton to use the creator.
 
 ## preparer binary
-* A possible mitigation to the drawback noted above would be a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
+* A possible mitigation to the drawback noted above would be a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)). For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -115,7 +115,8 @@ Why should we *not* do this? Leaving the location of the run image sbom unspec'd
 [unresolved-questions]: #unresolved-questions
 
 - Should the lifecycle run `genpkgs` if no run image SBOM is provided?
-- Do the SBOM formats need to be communicated in a label?
+- Do the available SBOM formats need to be communicated in a label?
+- How should `genpkgs` behave? Should it also attempt to locate a pre-existing SBOM?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -30,9 +30,6 @@ This RFC serves as an addendum to [#RFC 95: Structured SBOMs](https://github.com
 # What it is
 [what-it-is]: #what-it-is
 
-```suggestion
-# What it is
-
 This RFC proposes a mechanism for platforms to supply the SBOM for a run image at build time, leaving the method of storing and associating the SBOM to the run image unspec'd. The lifecycle would ingest the provided SBOM and place it in a spec'd location in the final app image. A label would be used to designate the layer containing the SBOM for the run image.
 
 - Define the target persona: platform operator, platform implementor, end user.
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -48,13 +48,7 @@ Example invocation:
 Following the build invocation above, the exported app image would contain:
 * `my-run-image-sbom.cdx.json` at `/layers/sbom/launch/base-image/sbom.cdx.json`. Note that this would make `base-image` a reserved buildpack ID.
 
-* An `io.buildpacks.base.sbom` label with the following schema:
-```json
-{
-  "sha": "run-image-layer-diffID",
-  "formats": [ "<string>" ]
-}
-```
+* An `io.buildpacks.base.sbom` label containing the diffID of the layer containing the run image SBOM.
 
 The accepted SBOM media types would be:
 * `application/vnd.cyclonedx+json` for files with extension `cdx.json`
@@ -63,18 +57,20 @@ The accepted SBOM media types would be:
 
 If `-run-image-sbom` is provided as a file:
 * The lifecycle will ensure that the file has a supported extension
-* The lifecycle will add the appropriate media type to `formats`
 
 If `-run-image-sbom` is provided as a directory:
 * The lifecycle will ensure that all files within the directory have supported extensions
-* The lifecycle will add the appropriate media type for each file to `formats`
+
+## Buildpack-provided SBOMs
+
+SBOM files output by buildpacks are currently exported in `/layers/sbom/launch/<buildpack id>`. The layer containing the buildpack-provided SBOM files is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`. For parity with `io.buildpacks.base.sbom`, a `io.buildpacks.app.sbom` label will also be added.
 
 ## Rebase
 
 Following the rebase invocation above, the exported app image would contain:
 * `my-new-run-image-sbom.cdx.json` at `/layers/sbom/launch/base-image/sbom.cdx.json`
 * The layer containing the old run image sbom would be removed
-* The `io.buildpacks.base.sbom` label would be updated to contain the diffID of the layer containing the new run image sbom and to ensure that `formats` is accurate for the new file
+* The `io.buildpacks.base.sbom` label would be updated to contain the diffID of the layer containing the new run image sbom
 
 ## When a run image has an sbom baked in
 
@@ -121,8 +117,8 @@ Why should we *not* do this? Leaving the location of the run image sbom unspec'd
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
-- What should be the path of the run image sbom within the image?
-- How should the layer containing the run image sbom be designated? This layer containing the buildpack-provided sbom is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`. #186 proposed `io.buildpacks.base.sbom` for the layer containing the run image sbom, and `io.buildpacks.app.sbom` for the layer containing the buildpack-provided sbom (in addition to `io.buildpacks.lifecycle.metadata`). Should we "future proof" the label name in case we might eventually want to attach build SBOMs to application images?
+- Should the lifecycle run `genpkgs` if no run image SBOM is provided?
+- Do the SBOM formats need to be communicated in a label?
 
 - What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
   - The existence or behavior of a preparer binary that knows how to download run image sboms.
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -30,11 +30,16 @@ This RFC serves as an addendum to [#RFC 95: Structured SBOMs](https://github.com
 # What it is
 [what-it-is]: #what-it-is
 
+```suggestion
+# What it is
+
+This RFC proposes a mechanism for platforms to supply the SBOM for a run image at build time, leaving the method of storing and associating the SBOM to the run image unspec'd. The lifecycle would ingest the provided SBOM and place it in a spec'd location in the final app image. A label would be used to designate the layer containing the SBOM for the run image.
+
 - Define the target persona: platform operator, platform implementor, end user.
 - If applicable, provide sample error messages, deprecation warnings, or migration guidance.
   - The lifecycle could warn if a usable run image sbom is not provided.
 
-Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom my-run-image-sbom.json my-app-image`
+Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom <path or directory - e.g., ./my-run-image-sbom.json> my-app-image`
 
 # How it Works
 [how-it-works]: #how-it-works
@@ -54,7 +59,8 @@ The accepted media types would be:
 # Drawbacks
 [drawbacks]: #drawbacks
 
-Why should we *not* do this? Leaving the location of the run image sbom unspec'd could make it harder for logic-less platforms like Tekton to use the creator. However, a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
+Why should we *not* do this? Leaving the location of the run image sbom unspec'd (as opposed to baking it into the run image at a specific location) could make it harder for logic-less platforms like Tekton to use the creator. 
+* Possible mitigation: a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-run-image-sbom.md
+++ b/text/0000-run-image-sbom.md
@@ -48,7 +48,7 @@ Example invocation:
 Following the build invocation above, the exported app image would contain:
 * `my-run-image-sbom.cdx.json` at `/layers/sbom/launch/base-image/sbom.cdx.json`. Note that this would make `base-image` a reserved buildpack ID.
 
-* An `io.buildpacks.sbom.base` label with the following schema:
+* An `io.buildpacks.base.sbom` label with the following schema:
 ```json
 {
   "sha": "run-image-layer-diffID",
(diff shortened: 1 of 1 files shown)


text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [#RFC 95: Structured SBOMs](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What should be the path of the run image sbom within the image?
+- Should the run image sbom be part of the layer containing the buildpack-provided launch sboms? This layer is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`. Is there a better layer / label?
```
natalieparellano, line 78: Thinking on this further, there should be a separate layer - because otherwise that would break rebase. We'll need a new label.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+- What should be all the accepted media types / extensions?
+
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+  - The existence or behavior of a preparer binary that knows how to download run image sboms.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+Does this RFC entail any proposed changes to the core specifications or extensions? New field on the exporter. Potentially a new label.
```
natalieparellano, line 86:
```suggestion
Does this RFC entail any proposed changes to the core specifications or extensions? New fields on the exporter and rebaser. A new label on the app image to designate the layer containing the run image sbom.
```

text/0000-run-image-sbom.md
natalieparellano, line 78:
```suggestion
- How should the layer containing the run image sbom be designated? This layer containing the buildpack-provided sbom is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`.
```

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+- [#RFC 95: Structured SBOMs](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What should be the path of the run image sbom within the image?
+- How should the layer containing the run image sbom be designated? This layer containing the buildpack-provided sbom is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`.
```
natalieparellano, line 78:
```suggestion
- How should the layer containing the run image sbom be designated? This layer containing the buildpack-provided sbom is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`. #186 proposed `io.buildpacks.base.sbom` for the layer containing the run image sbom, and `io.buildpacks.app.sbom` for the layer containing the buildpack-provided sbom (in addition to `io.buildpacks.lifecycle.metadata`).
```

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+This RFC serves as an addendum to [#RFC 95: Structured SBOMs](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md), which introduced a structured way for buildpacks to provide an sbom for dependencies installed at build time. For a full depiction of dependencies within a container image, container scanning tools would _also_ need an sbom for OS packages that were installed on the run image when it was created.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- Define the target persona: platform operator, platform implementor, end user.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+  - The lifecycle could warn if a usable run image sbom is not provided.
+
+Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom my-run-image-sbom.json my-app-image`
```
jromero, line 37: Would it be possible for a platform to provide the sbom in various formats? For example, say `syft` and `cyclonedx`.

I was sort of expecting run image sboms to be more of a directory of "stuff" that would be copied over to their appropriate location.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+* `application/vnd.syft+json` with ext `syft.json`
+* others?
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Leaving the location of the run image sbom unspec'd could make it harder for logic-less platforms like Tekton to use the creator. However, a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
+
+# Alternatives
+[alternatives]: #alternatives
```
jromero, line 60: Wasn't there an alternative discussed where we would leave it up to the platform to determine the method of attaching an sbom? For instance, if our platform want to use more of an associative artifact (similar to cosign) we could do that. It seems like this RFC only proposes one of what I see possibly being many ways for platforms to want to add this feature. Would it be better if we spec'd out the possible options and a way for further tools to be able to discover how the run image sbom is attached to the app image?

For example, some label that states a) "it's this layer here", or b) "it's located at this URI".

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+  - The lifecycle could warn if a usable run image sbom is not provided.
+
+Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom my-run-image-sbom.json my-app-image`
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Following the invocation above, the exported app image would contain:
+* `my-run-image-sbom.json` somewhere in `/layers/sbom`; suggestions:
```
jromero, line 43: This seems like an open question that should be resolved as part of this RFC.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+The accepted media types would be:
+* `application/vnd.cyclonedx+json` with ext `cdx.json`
+* `application/spdx+json` with ext `spdx.json`
+* `application/vnd.syft+json` with ext `syft.json`
+* others?
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Leaving the location of the run image sbom unspec'd could make it harder for logic-less platforms like Tekton to use the creator. However, a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
```
jromero, line 57: I'm a little confused by this statement to be honest. It starts off by sounding like a justification instead of a drawback but then goes into other possible solutions 🤔 ?

jromero (changes requested):

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+Following the invocation above, the exported app image would contain:
+* `my-run-image-sbom.json` somewhere in `/layers/sbom`; suggestions:
+  * `/layers/sbom/launch/run-image/sbom.<ext>` (this would make `run-image` a reserved buildpack id)
+  * `/layers/sbom/run-image/sbom.<ext>`
+  * ?
+
+The accepted media types would be:
+* `application/vnd.cyclonedx+json` with ext `cdx.json`
+* `application/spdx+json` with ext `spdx.json`
+* `application/vnd.syft+json` with ext `syft.json`
```
fg-j, line 51: Related to https://github.com/buildpacks/lifecycle/issues/828, these media types need to be accepted with parameters (e.g. version) so that end users / buildpack authors can maintain stronger contracts about the schema versions of the SBOMs that come out of builds.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+- Attestation: authenticated metadata about one or more software artifacts, as per the SLSA Attestation Model (see [here](https://github.com/in-toto/attestation) and [here](https://github.com/sigstore/cosign#in-toto-attestations)).
+- Cosign sbom attachment: an sbom object represented as an OCI Image Manifest V1 (see [here](https://github.com/sigstore/cosign/blob/main/specs/SBOM_SPEC.md)).
+
+# Motivation
+[motivation]: #motivation
+
+This RFC serves as an addendum to [#RFC 95: Structured SBOMs](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md), which introduced a structured way for buildpacks to provide an sbom for dependencies installed at build time. For a full depiction of dependencies within a container image, container scanning tools would _also_ need an sbom for OS packages that were installed on the run image when it was created.
+
+# What it is
```
hone, line 30: Could we get some prose here just describing what is being proposed in this RFC beyond just the example invocation?

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+[what-it-is]: #what-it-is
```
natalieparellano, line 31:
```suggestion
[what-it-is]: #what-it-is

```suggestion
# What it is

This RFC proposes a mechanism for platforms to supply the SBOM for a run image at build time, leaving the method of storing and associating the SBOM to the run image unspec'd. The lifecycle would ingest the provided SBOM and place it in a spec'd location in the final app image. A label would be used to designate the layer containing the SBOM for the run image.
```
```

File: text/0000-run-image-sbom.md (hunk 4 above)
Comment (natalieparellano) [inline text/0000-run-image-sbom.md:37]:
```suggestion
Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom <path or directory> my-app-image`
```

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,86 @@
...
+Example invocation: `/cnb/lifecycle/exporter -run-image my-run-image -run-image-sbom my-run-image-sbom.json my-app-image`
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Following the invocation above, the exported app image would contain:
+* `my-run-image-sbom.json` somewhere in `/layers/sbom`; suggestions:
+  * `/layers/sbom/launch/run-image/sbom.<ext>` (this would make `run-image` a reserved buildpack id)
+  * `/layers/sbom/run-image/sbom.<ext>`
+  * ?
```
natalieparellano, line 46: I like `/layers/sbom/launch/base` or `/layers/sbom/launch/base-image`.

text/0000-run-image-sbom.md
natalieparellano: Thinking about how buildpacks "declare" the media types they support, in buildpack.toml - would we want to do the same for the run image? For example, to put a label on the final app image alerting vulnerability scanners which formats are supported. We could put the supported types in a label on the run image. We'd need to take care during rebase that the label is still accurate.

text/0000-run-image-sbom.md
natalieparellano, line 57:
```suggestion
Why should we *not* do this? Leaving the location of the run image sbom unspec'd (as opposed to baking it into the run image at a specific location) could make it harder for logic-less platforms like Tekton to use the creator. 
* Possible mitigation: a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
```

text/0000-run-image-sbom.md
natalieparellano: I am assuming when a directory is provided, we'd like to validate that the files inside all have supported extensions?

text/0000-run-image-sbom.md
natalieparellano: It's noted under unresolved questions ;)

text/0000-run-image-sbom.md
natalieparellano: I think there are two problems to solve here - one is how to attach or associate an SBOM with the run image. The other is how to attach or associate an SBOM with a CNB-built app image. This RFC is essentially saying we shouldn't worry about the first issue. As for the second issue, this RFC proposes staying consistent with the pattern introduced in https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md and put the run image sbom somewhere in `/layers/sbom`.

text/0000-run-image-sbom.md
natalieparellano: Re-worded the first sentence to (hopefully) be less confusing.

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Build
+
+Following the build invocation above, the exported app image would contain:
+* `my-run-image-sbom.json` somewhere in `/layers/sbom`; suggestions:
+  * `/layers/sbom/launch/base/sbom.<ext>` (this would make `base` a reserved buildpack id)
+  * `/layers/sbom/launch/run-image/sbom.<ext>` (this would make `run-image` a reserved buildpack id)
+  * `/layers/sbom/run-image/sbom.<ext>`
+  * ?
+* An `io.buildpacks.sbom.base` label containing the diffID of the layer containing the run image sbom
```
natalieparellano, line 54: Though `io.buildpacks.base.sbom` was discussed previously, I like `io.buildpacks.sbom.base` because it is more consistent with the directory structure `/layers/sbom/launch/base` and thus easier to reason about.

text/0000-run-image-sbom.md
```diff
@@ -30,31 +30,68 @@ This RFC serves as an addendum to [#RFC 95: Structured SBOMs](https://github.com
...
+Why should we *not* do this? Leaving the location of the run image sbom unspec'd (as opposed to baking it into the run image at a specific location) could make it harder for logic-less platforms like Tekton to use the creator.
+
+## preparer binary
+* A possible mitigation to the drawback noted above would be a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
```
natalieparellano, line 94:
```suggestion
* A possible mitigation to the drawback noted above would be a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)). For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
```

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,123 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this? Leaving the location of the run image sbom unspec'd (as opposed to baking it into the run image at a specific location) could make it harder for logic-less platforms like Tekton to use the creator.
+
+## preparer binary
+* A possible mitigation to the drawback noted above would be a CNB-provided "prepare" operation ([currently under discussion](https://github.com/buildpacks/rfcs/pull/202)) could make this easier. For example, a preparer could look for a run image sbom in a file, attestation, attachment, or layer (picking the first one it finds) and provide the data as a file to the creator. The Tekton task that uses the creator currently already has a ["prepare" step](https://github.com/tektoncd/catalog/blob/4bf8b57aa105f0c7ce05fc122a11b1b0d5822fcd/task/buildpacks/0.3/buildpacks.yaml#L70-L121) which could be modified to invoke a preparer binary.
```
natalieparellano: The preparer could even run `genpkgs` if no SBOM is found

nebhale (approved):

text/0000-run-image-sbom.md
jromero: One of the things that have helped in the past is to look at the context of where this would exist. What do the other labels look like? Especially the ones related to other base image information or SBOM?

text/0000-run-image-sbom.md
```diff
@@ -0,0 +1,123 @@
...
+
+- What should be the path of the run image sbom within the image?
+- How should the layer containing the run image sbom be designated? This layer containing the buildpack-provided sbom is referenced in the `io.buildpacks.lifecycle.metadata` label with key `sbom`. #186 proposed `io.buildpacks.base.sbom` for the layer containing the run image sbom, and `io.buildpacks.app.sbom` for the layer containing the buildpack-provided sbom (in addition to `io.buildpacks.lifecycle.metadata`).
+- What should be all the accepted media types / extensions?
+
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+  - The existence or behavior of a preparer binary that knows how to download run image sboms.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
```
jromero, line 122: Should this section be filled in?

jromero (changes requested):
Great improvements. Overall I've bought into the idea. The only outstanding items I care about would be the final locations for the sbom on disk and the label name.

text/0000-run-image-sbom.md
natalieparellano: Added in latest commit

text/0000-run-image-sbom.md
natalieparellano: Added in latest commit - but since we'd be deriving the media type from the file extension, I'm not sure how to derive the parameter. Could we have more flexible file names? Could the platform provide the media type as a separate argument?

text/0000-run-image-sbom.md
natalieparellano: On further reflection, went with `/layers/sbom/launch/base-image` as this seems less likely to have buildpack ID collisions.

text/0000-run-image-sbom.md
natalieparellano: Good point. The closest I could find is the `io.buildpacks.stack.XXX` labels. Which I guess would mean `io.buildpacks.base.sbom` makes more sense. I'll change it.

natalieparellano:
> Platforms should use an industry standard method of associating a run image with an SBOM.

> There is less of a need to [include SBOM files in the application image] for run image SBOMs…

> We do like including the run image SBOM in the application image at a specified location.

This part doesn't sit very well with me. I feel like, if we're telling platforms "follow industry standards" then it makes sense to use the same standard for ALL images in a pipeline (run image, application image, etc.). It feels weird that there would be this special way to scan buildpacks-produced images when other practices are being adopted at different points in the process.

This makes me think that we should just bite the bullet and say that we'll support multiple ways to associate SOBMs: in-toto attestation, cosign attachment, or (if nothing can be found) we'll just run `genpkgs` and output a file alongside report.toml.

WDYT @ekcasey @samj1912 @sclevine

natalieparellano:
Circling back on this - the consensus from 4/7 Working Group seems to be:

* While it's convenient to have the run image SBOM in the image (so that the complete SBOM can be constructed if the attached SBOM is ever lost)
* This adds a fair bit of complexity, especially with respect to rebase

Therefore, the suggestion above is preferred:

> we should just bite the bullet and say that we'll support multiple ways to associate SOBMs: in-toto attestation, cosign attachment, or (if nothing can be found) we'll just run `genpkgs` and output a file alongside report.toml

Probably the most fruitful path forward would be to work out some of the complexities as part of the [Dockerfiles POC](https://github.com/buildpacks/lifecycle/pull/802) and then open another RFC describing the `genpkgs` component and the platform's use of it. I'll put this RFC in draft for now.

natalieparellano:
Cross post from other thread: https://github.com/buildpacks/spec/pull/298#issuecomment-1099543096

Closing this in favor of https://github.com/buildpacks/rfcs/pull/195.

PR #212, opened 2022-03-15
Propose alternative project scaffolding in pack

AidanDelaney:
Allow pack to promote libcnb project scaffolding and support 3rd
parties to provide their own buildpack scaffolding.

Readable - https://github.com/AidanDelaney/rfcs/blob/pack-buildpacks-new-alternatives/text/0000-pack-buildpack-new-alternatives.md


text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+- Start Date: 2022-03-10
+- Author(s): aidandelaney
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+The `pack buildpack new` subcommand creates a new buildpack based on a shell-script template.  This proposal extends `pack buildpack new` with a `--libcnb` option and a `--template URL` option to allow alternatives to the shell-script template.  The `--libcnb` option creates a new buildpack written in golang using `libcnb`.  The `--template URL` option allows buildpack creation from an arbitrary 3rd party URL.
```
jkutner, line 13: Rather than `--libcnb`, could we use something that is Golang specific, such as `libcnb.go` or even just `go`?

Eventually, I'd like the project adopt libcnb libraries for python, bash, and other langs.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+`pack buildpack new --libcnb` supports end-users who wish to scaffold a new `libcnb`-based buildpack.  The operation of `pack buildpack new --libcnb` does not require internet access to succeed.  `pack buildpack new --template URL` allows projects (such as Paketo, Heroku and others) to define skeleton projects for new buildpacks.
+
+Extending `pack buildpack new` allows buildpack authors to more-easily create new buildpacks in their chosen language and framework.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Project scaffolding is [very](https://cookiecutter.readthedocs.io/) [popular](https://yeoman.io/) [in](https://github.com/kowainik/summoner) [other](https://crates.io/crates/cargo-scaffold) [ecosystems](https://github.com/golang-standards/project-layout).  Scaffolding systems are employed to ease onboarding of new developers.  Within `pack` this feature is targeted at onboarding buildpack authors.
+
+- Explaining the feature largely in terms of examples.
```
jkutner, line 36: What should the behavior be when no opts are provided (i.e. just `pack buildpack new`)? Does it still default to bash? Should we prompt with a choice of language/template?

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: That makes sense.  I'll make that change.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I'm very open to other choices, but I was thinking we should preserve the current behaviour with `pack buildpack new`.  However, prompting the end-user for a choice of implementation language is something we should discuss.

jkutner: Yea, I'm open to a lot of options too. I like bash as the default because it's a kind of lowest common denominator (and we can even add https://github.com/jkutner/libcnb.bash if want). But you're right that it implies something about our project's recommendations.

text/0000-pack-buildpack-new-alternatives.md
jkutner: i can't imagine we'd have a default go template that didn't include `libcnb`, so i kinda like `--golang` or similar

jromero: Could we do without `--libcnb` or `--golang` and simply have `--template URL` where URL can be a libcnb template?

I'd personally opt-out of littering boolean flags for different possible permutations when a single parameter could  support multiple cases.

jromero (changes requested):
I'm a strong proponent of this RFC but would like to discuss the inclusion of libcnb as an embedded format. There seem to be more drawbacks than benefits from my perspective.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I think I'm coming round to that view too.  Maybe `pack buildpack new [shell, golang, ...]` for buildpacks provided templates, `pack buildpack --template URL` for externally provided?

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+- What other designs have been considered?
+
+We have considered a wider re-implementation of [cookiecutter](https://cookiecutter.readthedocs.io/).  However, the scope of a re-implementation was considered too wide.  In addition, we have considered a cookiecutter-lite implementation where project scaffolding is cloned from a repository.  This proposal recommends a project scaffolding that can be embedded in `pack` and, in addition, can be downloaded from an Internet source.
+
+We have also considered [springerle](https://github.com/carlmjohnson/springerle) which uses a single txtar file to describe skeleton project structure.  The use of a single txtar file requires template providers to write and maintain this format.
+
+[bare](https://github.com/bare-cli/bare) is a new project.  It assumes that all templates are stored on github.com.
+
+- Why is this proposal the best?
+To integrate with `pack` we want a pure-golang project scaffolding tool.  This proposal advocates an approach that embeds cnb-provided skeletons in `pack` and allows `pack` to clone 3rd party skeletons from a location chosen by the end-user.
```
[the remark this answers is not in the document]
AidanDelaney: Embedded cnb-provided skeletons support offline project creation.  However, I'm happy to assume internet connectivity for all project creation options.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+    └── main.go
+```
+
+Where `prompts.yaml` is of the form:
+
+```yaml
+prompts:
+    - variable: packageName
+      prompt: Package name (often github.com/username/repo)
+```
```
AidanDelaney, line 96:
```suggestion
```toml
packageName = { prompt = "Package name (often github.com/username/repo"}
```
```

File: text/0000-pack-buildpack-new-alternatives.md [hunk 5]
```diff
@@ -0,0 +1,210 @@
...
+pack buildpack new --libcnb examples/ruby \
+    --api 0.7 \
+    --path ruby-buildpack \
+    --version 0.0.1 \
+    --stacks io.buildpacks.samples.stacks.bionic
+Package name (often github.com/username/repo)? github.com/AidanDelaney/ruby
+Created project in ruby-buildpack
+```
+
+Templates are provided as a file tree.  A root-level `prompts.yaml` file contains prompts for the end-user.  For example, a partial file tree containing user prompts and golang skeleton code is:
```
AidanDelaney, line 81:
```suggestion
Templates are provided as a file tree.  A root-level `prompts.toml` file contains prompts for the end-user.  For example, a partial file tree containing user prompts and golang skeleton code is:
```

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+* `.Stacks` -- a slice containing the buildpack stacks obtained from the argument to the `--stacks` flag
+
+Template variables introduced in `prompts.yaml` are required to have a name unique within the `prompts.yaml` file and not redefine reserved template variables.  An optional default value may be provided.  Variables that do not provide a default value must 
+
+```yaml
+prompts:
+    - variable: packageName
+      prompt: Package name (often github.com/username/repo)
+      default: DefaultValue
+```
```
AidanDelaney, line 131:
```suggestion
```toml
packageName = { prompt = "Package name (often github.com/username/repo", default = "DefaultValue" }
```
```

File: text/0000-pack-buildpack-new-alternatives.md [hunk 7]
```diff
@@ -0,0 +1,210 @@
+# Meta
+[meta]: #meta
+- Name: Pack Buildpack New to Support Alternatives
+- Start Date: 2022-03-10
+- Author(s): aidandelaney
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
```
hone, line 10: does this technically supersede https://github.com/buildpacks/rfcs/blob/main/text/0077-pack-buildpack-create.md#future-work?

text/0000-pack-buildpack-new-alternatives.md
hone: In [RFC 0077](https://github.com/buildpacks/rfcs/blob/main/text/0077-pack-buildpack-create.md#future-work) we mention a curated repo for this. It looks like we're choosing to not do that?

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+[definitions]: #definitions
+
+* **project scaffolding**: minimal code and file system structure used to start a project.
+
+# Motivation
+[motivation]: #motivation
+
+The creation of buildpacks in languages other than `bash` is undocumented in the [Buildpack Author Guide](https://buildpacks.io/docs/buildpack-author-guide).  In particular, creating a buildpack using `libcnb` is undocumented because there is no mechanism to generate a scaffolded project.  This proposal adds a minimal `libcnb` project template to `pack` and allows 3d parties to provide templates to create buildpacks for their chosen technology stack.
+
+Extending `pack buildpack new` with a `--libcnb` flag supports buildpack creation using the existing documented workflow.  This results easier on-boarding of buildpack authors into the `libcnb` ecosystem.  Additionally extending `pack buildpack new` with `--target URL` allows other projects to re-use `pack` as their scaffolding tool.
```
hone, line 25:
```suggestion
Extending `pack buildpack new` with a `--libcnb` flag supports buildpack creation using the existing documented workflow.  This results easier on-boarding of buildpack authors into the `libcnb` ecosystem.  Additionally extending `pack buildpack new` with `--template URL` allows other projects to re-use `pack` as their scaffolding tool.
```

text/0000-pack-buildpack-new-alternatives.md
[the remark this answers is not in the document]
hone: I was thinking about this too, it'd be nice if BAT team could maintain something that doesn't involve editing pack directly, but I get the no internet use case. Would the BAT team own both the bash and libcnb scaffolding?

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+	)
+}
+```
+
+The following template variables, known as the _reserved template variables_ are available for substitution in all project files:
+
+* `.ID` -- the buildpack id that will be written to `buildpack.toml` obtained from a positional argument to `pack`
+* `.Version` -- the buildpack version obtained from the argument to the `--version` flag
+* `.API` -- the buildpack API version obtained from the argument to the `--api` flag
+* `.Stacks` -- a slice containing the buildpack stacks obtained from the argument to the `--stacks` flag
```
hone, line 122: are there defaults for these values? I could imagine `version` always being `0.1.0` or `0.0.1` depending on what we want to do there.

I also imagine for simplicity a language binding may only support a single buildpack API so it in `prompt.yaml` may want to provide this.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We explicitly do not want to re-implement cookiecutter in golang due to the scope of such an implementation.  Nor do we want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell and libcnb skeletons from an embedded file system.
+
+A full invocation to generate `libcnb` scaffolding is similar to the [documented project scaffolding](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/building-blocks-cnb/#using-the-pack-cli).
+
+```bash
+pack buildpack new --libcnb examples/ruby \
+    --api 0.7 \
```
Malax, line 63: To me, `--api` implies that any template has to support all valid(?) buildpack API versions. Can a template decide to only support one API version? For example, [libcnb.rs ](https://github.com/Malax/libcnb.rs) currently only supports one API version.

If we offer `--api`, it would be nice if the template at least can warn that it does not support that API version, but it might even be worth letting the template have an opinion which API version is best and not offer `--api`.

text/0000-pack-buildpack-new-alternatives.md
samj1912, line 13: I would prefer if we just deprecate `pack buildpack new` (and leave it as it is in the next version of pack with a warning and remove it in a couple of versions) and point them to `pack buildpack create` which is interactive and doesn't take any of the existing CLI flags that `pack buildpack new` does (api, stack etc.). This would allow the template to decide which prompts/flags it takes as inputs. We can have certain trusted/recommended templates similar to trusted builders and users may select one of those as `default` if they wish to.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,210 @@
...
+The following template variables, known as the _reserved template variables_ are available for substitution in all project files:
+
+* `.ID` -- the buildpack id that will be written to `buildpack.toml` obtained from a positional argument to `pack`
+* `.Version` -- the buildpack version obtained from the argument to the `--version` flag
+* `.API` -- the buildpack API version obtained from the argument to the `--api` flag
+* `.Stacks` -- a slice containing the buildpack stacks obtained from the argument to the `--stacks` flag
+
+Template variables introduced in `prompts.yaml` are required to have a name unique within the `prompts.yaml` file and not redefine reserved template variables.  An optional default value may be provided.  Variables that do not provide a default value must 
+
+```yaml
```
samj1912, line 126: We would also need to provide some way for the template to validate user inputs.

samj1912 (changes requested):

text/0000-pack-buildpack-new-alternatives.md
jromero: > In [RFC 0077](https://github.com/buildpacks/rfcs/blob/main/text/0077-pack-buildpack-create.md#future-work) we mention a curated repo for this. It looks like we're choosing to not do that?

@hone I'm have no strong opinion against a curated repo but I do think it's necessary to allow for _arbitrary_ urls.

> I think I'm coming round to that view too. Maybe pack buildpack new [shell, golang, ...] for buildpacks provided templates, pack buildpack --template URL for externally provided?

@AidanDelaney That's interesting. I sort of like it. :) Would it be worse if we had a single signature `pack buildpack new <name|URL>` where the first argument could be a keyword (`shell`, `golang`, etc) or URL?

AidanDelaney: I expect to have a set of curated templates that implement our best practice.  `pack buildpack new python` would use the curated `python` scaffolding.  There is a decision around whether the curated scaffolding should be embedded or if it should reach out to a `buildpacks/templates` style repo.  `pack buildpack new python` could be a synonym for `pack buildpack new http://github.com/buildpacks/templates/python`.

text/0000-pack-buildpack-new-alternatives.md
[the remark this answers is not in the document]
AidanDelaney: The BAT team should own the scaffolding.  Whether we embed the templates at `pack` build time, or reach out to GH at runtime is an open question.

text/0000-pack-buildpack-new-alternatives.md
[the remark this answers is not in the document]
AidanDelaney: I've updated the RFC with a toml example.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney, line 10: Yes, thank you.
```suggestion
- Supersedes: https://github.com/buildpacks/rfcs/blob/main/text/0077-pack-buildpack-create.md
```

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I've changed the proposal and now put this choice in the definition of the project template.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I'm going down that route now.  I've updated the RFC to reflect this.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I've included some basic input validation in an example implemetation.

samj1912:
@AidanDelaney I just found out about https://github.com/coveooss/gotemplate - have you investigated this?

samj1912:
I am specifically interested in things like https://coveooss.github.io/gotemplate/docs/functions_reference/runtime/#__raise__ and https://coveooss.github.io/gotemplate/docs/functions_reference/runtime/#__run__

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,216 @@
...
+Template variables introduced in `prompts.yaml` are required to have a name unique within the `prompts.yaml` file.  An optional default value may be provided.
+
+Answers to prompts can be provided in a toml file.  This supports programmatic use of `pack`:
+
+```
+pack buildpack create --overrides overrides.toml
+Created project in bash_buildpack
+```
+
+The format of `--overrides` parallels the `prompts.toml` format.  An example of which is:
```
jromero, line 153: Not sure that I agree with (or understand) this statement. The format / structure of the `overrides.toml` appears to be a simple key/value to the prompt names where as `prompts.yaml` is an array of tables.

jromero (changes requested):
@AidanDelaney this is 🤏 close to being complete IMO. 

I have a few outstanding questions. Also, it may be worth referring to [scafall](https://github.com/AidanDelaney/scafall) and this [example](https://github.com/AidanDelaney/scafall-python-eg) to better drive the intention on how this would be implemented.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: If no parameters are provided, end users are prompted to choose from the default template collection.  A prototype collection looks like https://github.com/AidanDelaney/cnb-buildpack-templates.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I have changed this to be a sequence of `--override` CLI flags.  Switching it to CLI flags was suggested on Slack.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,216 @@
...
+	libcnb.Main(
+		bp.Detect{},
+		bp.Build{},
+	)
+}
+```
+
+Template variables introduced in `prompts.yaml` are required to have a name unique within the `prompts.yaml` file.  An optional default value may be provided.
+
+Answers to prompts can be provided in a toml file.  This supports programmatic use of `pack`:
```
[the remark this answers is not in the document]
AidanDelaney: If no override is provided for a prompt, then the end-user is prompted to provide a value.  In a CI environment an invocation of `pack buildpack create` would be expected to override all of the possible choices.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+
+Replacing `pack buildpack new` with `pack buildpack create` allows buildpack authors to more-easily create new buildpacks in their chosen language and framework.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Project scaffolding is [very](https://cookiecutter.readthedocs.io/) [popular](https://yeoman.io/) [in](https://github.com/kowainik/summoner) [other](https://crates.io/crates/cargo-scaffold) [ecosystems](https://github.com/golang-standards/project-layout).  Scaffolding systems are employed to ease onboarding of new developers.  Within `pack` this feature is targeted at onboarding buildpack authors.
+
+`pack buildpack create` should prompt the end user to choose an implementation language.  Thereafter, generate skeleton code according to the best practices for that language.  For example, users choosing Go as an implementation language would generate a [golang standard project layout](https://github.com/golang-standards/project-layout) and include `libcnb` best practices.  The layout would look similar to the following:
```
dmikusa-pivotal, line 36: So does this mean that included with pack would be one generic template for each language we officially support? or would this just default `--template` to a location that's controlled by the buildpacks project?

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+├── cmd
+│   └── main.go
+├── go.mod
+├── go.sum
+└── pkg
+    ├── build.go
+    └── detect.go
+```
+
+This approach is not opinionated on topics such as which testing framework to use.  Only components under github.com/buildpacks/* are used in the generated project.  `pack buildpack create --template URL` allows for more opinionated project scaffolding.  As such, we support new buildpack authors with onboarding and support experienced buildpack authors to use the scaffolding of their choice.
```
dmikusa-pivotal, line 50: I'm not sure I agree with this approach for the default. I don't think you're going to get something very useful out of the box. I think we should try to keep this as close as possible to what we would use if we were adding a new buildpack to the Buildpacks project. I think having skeleton tests would be helpful as well for developers, they can get some examples of how to test things.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We do we want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell and libcnb skeletons from a cloned git repoistory.
```
dmikusa-pivotal, line 55:
```suggestion
The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We do not we want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell and libcnb skeletons from a cloned git repoistory.
```

I was a little confused by this, I think you mean "do not".

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+
+A full invocation to generate `bash` scaffolding prompts for the values [currently documented as flags](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/building-blocks-cnb/#using-the-pack-cli).
+
+```bash
+pack buildpack new
```
dmikusa-pivotal, line 60:
```suggestion
pack buildpack create
```

You were using `create` as the new command above (`new` being the old command).

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+? Choose a project template:
+  ▸ bash
+    go
+
+```
+
+A full session includes the terminal prompts that the project scaffolding tool asks of the end user:
+
+```bash
+pack buildpack new
```
dmikusa-pivotal, line 71:
```suggestion
pack buildpack create
```

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+Use the arrow keys to navigate: ↓ ↑ → ←
+? Choose a project template:
+  ▸ bash
+    go
+✔ Enter a directory in which to scaffold the project: bash_buildpack
+Use the arrow keys to navigate: ↓ ↑ → ←
+? Choose the buildpack API version (use the default if you are unsure):
```
dmikusa-pivotal, line 78: This might be better as a free form entry, but default to the latest current release, or have an "other" option. Otherwise, you'd have to update to be able to select newer versions. It would also me we have to bump this every time there's a new version.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+Where `prompts.toml` is of the form:
+
+```toml
+[[prompt]]
+name="ProjectDirectory"
+prompt="Enter a directory in which to scaffold the project"
+default="bash_buildpack"
+required=true
```
dmikusa-pivotal, line 110: Would it be possible to have certain prompts that are always displayed? Like project directory, buildpack API, etc.. 

It seems repetitive to have to set them in every template, when pretty much every template is going to need that info.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+
+```
+pack buildpack create --override 'ProjectDirectory="test_example"' --overrride='ModuleName="github.com/test/test"'
+Created project in bash_buildpack
+```
+
+All input files are expected to be normalized to Unix line endings.
+
+# Migration
+[migration]: #migration
```
dmikusa-pivotal, line 156: I assume that `pack buildpack new` will be deprecated when we add `pack buildpack create`. We should document that here and spell out the plan to deprecate and remove it.

My suggestion would be:

1. Immediately deprecate when `pack buildpack create` is added. Including with that a printed deprecation notice for anyone using it.
2. Give it either a couple of releases or a couple of months, then remove it.

dmikusa-pivotal (commented):
Overall, I'm very excited for this. I think it'll be a big help. Added some thoughts/questions for discussion.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: This proposal provides a default template collection.  The template collection allows the end-user to choose an implementation technology and scaffolds a CNB for them.  An example of this is https://github.com/AidanDelaney/cnb-buildpack-templates which allows the choice between an implementation of the (current) bash template and a Go template.

Other projects can provide a template or a template collection.  `pack buildpack create --template URL` supports end-users choosing to use buildpack scaffolding from 3rd parties.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: One possible implementation [scafall](https://github.com/AidanDelaney/scafall/blob/main/pkg/scafall.go#L18) allows the calling code to offer `DefaultValues`.  Using this, `pack` version `0.27` could call `scafall` passing in `BuildpackApi =[0.5, 0.6, 0.7, 0.8]` and some future version of pack could call `scafall` passing in `BuildpackApi =[1.0]`.  This give the calling code some control over the prompts presented to the end-user.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: The current `pack buildpack new` has two flags `--api` and `--path`.  We could maintain support for these flags in `pack buildpack create`?

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: Sounds reasonable to me.

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: We had this discussion recently for the profile buildpack that will be the first official buildpack under the project and the decision was to use Go std testing there. I would suggest that here as well. IMHO, it's hard to argue with using what's built into the language and if someone doesn't like that choice they can make their own template.

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: So I have pack installed through a package manager and it auto updates, but not everyone does that. Some people are running older versions of pack. It would not be the greatest user experience, In my opinion, if I had to update pack to be able to get different versions of the API. The same could be said about pack providing 1.0, but maybe I'm not ready for that yet and want to keep using 0.9. If pack removes that from the list, now I have to downgrade pack to select it or go and update my templates manually after creating a new buildpack. 

It would be my preference if this were a free-form field. Perhaps the description could include the most recent API at the time as a hint? Or at least have an "other" option that is free-form. Or is this something the template controls?

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: +1 for that. It'll just reduce some of the repetition for template authors.

text/0000-pack-buildpack-new-alternatives.md
jromero: Aren't the values provided to the end-user derived from the template chosen and therefore managed in a [separate repository](https://github.com/AidanDelaney/cnb-buildpack-templates/blob/main/Go/prompts.toml#L10)? If so, I'm having trouble understanding why `pack` would need to even know about these API versions.

text/0000-pack-buildpack-new-alternatives.md
jromero:
```suggestion
The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We do not want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell and libcnb skeletons from a cloned git repoistory.
```

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+This approach is not opinionated on topics such as which testing framework to use.  Only components under github.com/buildpacks/* are used in the generated project.  `pack buildpack create --template URL` allows for more opinionated project scaffolding.  As such, we support new buildpack authors with onboarding and support experienced buildpack authors to use the scaffolding of their choice.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We do we want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell and libcnb skeletons from a cloned git repoistory.
+
+A full invocation to generate `bash` scaffolding prompts for the values [currently documented as flags](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/building-blocks-cnb/#using-the-pack-cli).
```
jromero, line 57: I'm confused by this statement and the command following it. Can you explain what we're trying to convey?

text/0000-pack-buildpack-new-alternatives.md
jromero, line 50: Does the `--template` flag support subdirectories? I noticed [this repository you mentioned](https://github.com/AidanDelaney/cnb-buildpack-templates) has multiple templates. Would this repository be supported or would you need to create a repository per template?

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+* Python's [Cookiecutter](https://cookiecutter.readthedocs.io/): widely used to scaffold projects in many languages, including golang.  Requires a Python runtime to be available.
+* [springerle](https://github.com/carlmjohnson/springerle): golang re-think of cookiecutter.  Springerle uses a single txtar file augmented with a header containing user prompts.  This proposal prefers using a filesystem rather than a single txtar file as the filesystem approach extends to cloning repositories.
+* [cgapp](https://github.com/create-go-app/cli): This proposal heavily borrows from cgapp.
+* JavaScript's [Yeoman](https://yeoman.io/): widely used in the web ecosystem
+* [boilr](https://github.com/tmrts/boilr/): golang cookiecutter.  Moribund project.
+* [bare](https://github.com/bare-cli/bare): golang cookiecutter. Possible successor to boilr.  Both boilr and bare assume the templates are stored on github.com and use the zip downloading functionality of that specific service.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
```
jromero, line 201:
```suggestion
* Where do the template repositories live for `bash` and `golang`?
  * 
* What team maintains template repositories?
  * Buildpack Author Tooling team
```

jromero: > Where do the template repositories live for `bash` and `golang`?

I think depends on whether we can support subdirectories.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: We started this RFC by including the template collection in `pack`.  Based on feedback we've moved it to a git repo.  The `pack buildpack create` operation requires an internet connection to succeed.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: Great.  I'm happy with that.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: We had decided that we didn't want to use the existing `--api`, `--path` etc... flags in `pack buildpack create`.  Instead the buildpack should prompt for these values.  I think we're now leaning towards keeping the existing flags and only prompting the user for a required value if the flag is not provided.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: The `--template` flag supports both templates and template collections.  A template collection is a repository that contains templates in subdirectories.

text/0000-pack-buildpack-new-alternatives.md
samj1912: I think we decided that we would not want that. There can definitely be templates that don't need to prompt for the api version and just support one.

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: It won't let me for some reason, but feel free to resolve this conversation.

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: It won't let me for some reason, but feel free to resolve this conversation.

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: True. I see ways around that though.

1. Pack could look at the template and see if it's using that variable. If it's not, then skip the question.

2. It wouldn't be hard for pack to augment the question list though. If pack takes the list of input required from the template, adds to that any default questions, and filters out the ones the template has overridden you could without too much trouble allow a standard list of questions but let templates override them.

    The template that only wants to support one API version could do so and set the allowed list to one item. Pack could then ask the question with only a single answer or we could perhaps optimize and skip that question knowing the answer. Or we could get fancy and add a "hidden" property to allow the template to control if it wants that to be shown or not.

It's a little extra effort but I think it could be worth it.

Reasons:

1. Templates will get out of date. That'll never happen to the official templates ;) but community templates which I hope crop up will need to be maintained to have things like the API updated when new API versions come out. That could be annoying for users, especially if the API version is formatted as a select list with no "other" option.

2. If you have many templates to update that would be a pain, especially if they are across multiple repos (I think it was asked but haven't seen an answer regarding if you can have multiple templates in a single git repo, that would lessen this problem).

text/0000-pack-buildpack-new-alternatives.md
dmikusa-pivotal: It won't let me for some reason, but feel free to resolve this conversation.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I think the core problem here is coupling between `pack` binary and "official" scaffolding.  Previously the scaffolding was distributed with  `pack`.  We're somewhat decoupling `pack` and official scaffolding.  However, we see situations where an older `pack` that supports only (for example) API `[0.7, 0.8]` might call out to newer official scaffolding that _can_ generate projects spec'ed to `1.0`.

One way of solving this is for `pack` to provide to the scaffolding, the set of API versions that `pack` supports.  It would then be the responsibility of the scaffolding to generate compliant projects.  This is what the current proposal suggests.  `pack` can provide a list of `BuildpackAPI` versions to the scaffolding.  `pack` can either choose to use one of these versions (an override) or leave it to the end-user to be prompted for one of these versions.

I wonder if there are other ways to solve this?  Should we more tightly couple `pack` and the official scaffolding?  So `pack` would point at a specific release of the official scaffolding?  This way an older version of `pack` will never try to use a newer scaffolding.  Furthermore, rather than coupling `pack` releases to scaffolding releases, we could release scaffolding by API version?   Thus `pack` supporting  API `[0.7, 0.8]` would scaffold using the latest release of scaffolding for API version `0.7` or API version  `0.8`.  This allows bug-fixes in the `0.8` release to be available to all `pack` releases that support API version `0.8`.

The key issue appears to be `pack` telling the scaffolding the API versions that it supports.  We can discuss this more widely in a BAT meeting.

text/0000-pack-buildpack-new-alternatives.md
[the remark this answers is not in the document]
AidanDelaney: Yes. We can reduce the scope of this proposal.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,216 @@
...
+Created project in bash_buildpack
+```
+
+Templates are provided as a file tree.  A root-level `prompts.toml` file contains prompts for the end-user.  For example, a partial file tree containing user prompts and Go skeleton code is:
+
+```bash
+.
+├── prompts.toml
+└── {{.ProjectDirectory}}
```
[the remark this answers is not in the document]
AidanDelaney: I've taken your advice and provided examples of both templates and template collections.  There is more detail now provided that explains where `BuildpackAPI` comes form.

text/0000-pack-buildpack-new-alternatives.md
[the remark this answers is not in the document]
AidanDelaney: I've made sure to provide more examples that demonstrate this.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: We've taken the view that this is not an issue that needs to be addressed right now.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: The proposal now defines what the listed CLI flags do.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,206 @@
...
+* Project scaffolding could be delegated to a 3rd party tool.  The current `pack buildpack new` functionality could be extracted from `pack` and, instead, we could suggest another tool for end users to use for project scaffolding eg: `bare use buildpacks/bash my_buildpack`.
+* Including generalized project scaffolding in `pack` will increase size of `pack` binary.  The size of `pack` will increase by the size of an appropriate prompting package (such as [survey](https://pkg.go.dev/github.com/AlecAivazis/survey/v2)) and an appropriate package allowing `git clone` (such as [go-git](https://github.com/go-git/go-git) at ).
+* This proposal commits to support a specific project scaffolding format.  A migration path should be established if and when a de-facto standard golang template library becomes available.
+* In default project scaffolding we cannot be opinionated about test frameworks; this will result in scaffolded projects with no default test setup.
```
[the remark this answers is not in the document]
AidanDelaney: I have removed this from the spec.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,308 @@
...
+
+`pack buildpack create` creates the same buildpack scaffolding as `pack buildpack new`.  However, `pack buildpack create --template URL` can be used to access alternative project scaffolding.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The design is modeled on [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/) with reference to [springerle](https://github.com/carlmjohnson/springerle) -- a similar implementation in golang -- and [create-go-app](https://github.com/create-go-app/).  We do not want to `os.Exec` a python subprocess to run cookiecutter as this would require availability of a python runtime environment.  Instead we propose to borrow heavily from create-go-app, and generate the default shell scaffolding from a cloned git repoistory.
+
+The `pack buildpack new` project scaffolding accepts `--api`, `--path` and `--stacks` command line flags.  These command line flags will be preserved in `pack buildpack create`.  However, where a flag is omitted, the end-user will be prompted to provide an appropriate value.
```
dmikusa-pivotal, line 43: I don't think we should preserve the `--api` or `--stacks` flags. The purpose of these flags is to provide data to templates, right? So in my mind, that conflicts with Overrides which are also to override values from templates.

It also provides a weird experience, IMHO, in that we collect these values by default but do not provide these values automatically to the template. The template author still needs to add an entry to prompts.toml for them (at least that's how I'm reading it, please correct me if I'm misunderstanding that).

I favor just dropping these arguments, but I could see an alternative approach being that we automatically provide those two values to every template and the template author doesn't need to include them in prompts.toml.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,275 @@
...
+| prompt        | Yes      | The default prompt question to ask the user                                                                                              | string         |
+| prompt.locale | No       | A translation of the default prompt into the locale specified.                                                                           |                |
+| required      | No       | Evaluates to `false` if not provided.  Specifies whether the user **must** provide a value for the variable identified by `name`.        | boolean        |
+| default       | No       | A default value for the variable identified by `name`.  Mutually exclusive to `choices`                                                  | string         |
+| choices       | No       | A list of default values from which the user may choose a value for the variable identified by `name`.  Mutually exclusive to `default`. | list of stings |
+
+Prompt locales are specified following [IETF BCP 47](https://tools.ietf.org/rfc/bcp/bcp47.txt).  As such `prompt.en-US` is the translation of `prompt` into US English and `prompt.es-MX` is the translation of `prompt` into Mexican Spanish.
```
jromero, line 156: 1. Is there an anticipated use for locale or should we defer this added complexity until the need arises?
2. Can you give a concrete example of how locales would work? How is locale determined at execution time?

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: We've now dropped these CLI flags.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,308 @@
...
+
+If specified, the following `pack buildpack create` flags override the value of variables with the listed identifier:
+
+| command line flag | variable name      |
+|-------------------|--------------------|
+| `--path`          | `ProjectDirectory` |
+| `--api`           | `BuildpackApi`     |
+| `--stacks`        | `BuildpackStacks` |
+
+In addition, the required positional argument to `pack buildpack create`, eg: `pack buildpack create example/bash`, overrides the variable identified by `BuildpackID`.
```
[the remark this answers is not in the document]
AidanDelaney: We've renamed "overrides" to "arguments" and dropped the the flags.  This takes on board your feedback on flags and their naming.

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,308 @@
...
+* [springerle](https://github.com/carlmjohnson/springerle): golang re-think of cookiecutter.  Springerle uses a single txtar file augmented with a header containing user prompts.  This proposal prefers using a filesystem rather than a single txtar file as the filesystem approach extends to cloning repositories.
+* [cgapp](https://github.com/create-go-app/cli): This proposal heavily borrows from cgapp.
+* JavaScript's [Yeoman](https://yeoman.io/): widely used in the web ecosystem
+* [boilr](https://github.com/tmrts/boilr/): golang cookiecutter.  Moribund project.
+* [bare](https://github.com/bare-cli/bare): golang cookiecutter. Possible successor to boilr.  Both boilr and bare assume the templates are stored on github.com and use the zip downloading functionality of that specific service.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+* Where do the template repositories live for `bash` and `golang`?
+  *
```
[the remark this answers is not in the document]
AidanDelaney: Where any template collection is hosted is outside the scope of this RFC.

text/0000-pack-buildpack-new-alternatives.md
AidanDelaney: I mostly didn't want to assume that the world is centered around a single language.  I've now removed this from the RFC, but I think it's something that I'd like to support in the medium term.

samj1912 (approved):

samj1912:
@AidanDelaney could you squash all of the commits and DCO sign off the final one? Thanks the proposal lgtm!

text/0000-pack-buildpack-new-alternatives.md
```diff
@@ -0,0 +1,272 @@
...
+Use the arrow keys to navigate: ↓ ↑ → ←
+? Choose the buildpack API version (use the default if you are unsure):
+  ▸ 0.7
+    0.8
+✔ Enter a stack this buildpack will support by default: io.buildpacks.samples.stacks.bionic
+
+Created project in bash_buildpack
+```
+
+The user can skip prompts by providing `--arg key=value` as command line flags. Each of the prompts from the previous interactive invocation can be skipped by providing appropriate pairs of key and value.
```
samj1912, line 60: How will an end user discover these args without understanding the prompt.toml section? Should we ensure that the `--help` flag when prompted with a template prints out the available  args? Or should we display the name alongside the prompt text so that it is evident?

AidanDelaney: I have an implementation of `--help` support in scafall.  After playing around with it a bit, I'm comfortable updating the RFC in line with your suggestion.  Thanks!

samj1912:
Taking on the stewardship of this RFC. Given the approvals from @jromero and myself, I would like to put this RFC up for votes with a lazy consensus deadline of October 14.

dmikusa (approved):

AidanDelaney:
To merge this, do I `git mv` the file to `0212-pack-buildpack-new-alternatives.md` and then merge?

samj1912:
@AidanDelaney the assignee typically merges it using the merge script in the repository. I will do that this Monday.

PR #215, opened 2022-03-23
Add template for tracking issues

natalieparellano:
Assuming we like the tracking issue pattern, having a template would be nice.

### DIFF
--- a/ISSUE_TEMPLATE/tracking.md
+++ b/ISSUE_TEMPLATE/tracking.md
@@ -1,0 +1,25 @@
+---
+name: Tracking Issue
+about: Open an issue to track progress toward implementation of a merged RFC
+title: '[RFC #<INSERT RFC NUMBER - e.g., 0099>] - <INSERT RFC TITLE>'
+labels: type/tracking
+assignees: ''
+
+---
+
+[RFC #<INSERT RFC NUMBER - e.g., 0099>](https://github.com/buildpacks/rfcs/blob/main/text/<INSERT FILENAME>) - <INSERT RFC TITLE>
+
+Spec:
+- [ ] https://github.com/buildpacks/spec/issues/<INSERT ISSUE NUMBER>
+- [ ] Released in <INSERT API - e.g., platform> API version `<placeholder>`
+
+Lifecycle:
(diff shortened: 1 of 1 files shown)


hone (approved):

ISSUE_TEMPLATE/tracking.md
```diff
@@ -0,0 +1,25 @@
...
+name: Tracking Issue
+about: Open an issue to track progress toward implementation of a merged RFC
+title: '[RFC #<INSERT RFC NUMBER - e.g., 0099>] - <INSERT RFC TITLE>'
+labels: type/tracking
+assignees: ''
+
+---
+
+[RFC #<INSERT RFC NUMBER - e.g., 0099>](https://github.com/buildpacks/rfcs/blob/main/text/<INSERT FILENAME>) - <INSERT RFC TITLE>
+
```
hone, line 11: are we calling out `pack` / `lifecycle` explicitly or just want a generic thing + spec/documentation? For instance, I can imagine changes for `libcnb` too or really any repo we have.

hone:
Thanks for putting this together @natalieparellano

ISSUE_TEMPLATE/tracking.md
natalieparellano: Makes sense... we could add a placeholder for `<insert repo here>` and folks could just delete it if it's not relevant.

natalieparellano: WDYT of 5faf18e ?

natalieparellano: Or maybe... I feel like most changes will touch spec + lifecycle + ??? + docs - where ??? could be pack, libcnb, or some combination of the two. I'm probably biased toward pack because I've worked on that repo and not libcnb 🙃 

I am happy with whatever everyone feels is the most useful.

samj1912: I would say that any changes to buildpack API require changes in libcnb.

natalieparellano: Please see df8e81c - added libcnb so it's on equal footing with pack.

jromero (approved):

ekcasey (approved):

PR #216, opened 2022-03-31
Stop Deleting Cache Images

jabrown85:
[READABLE](https://github.com/buildpacks/rfcs/blob/5c1d82583b0baf1e1fbb6cd2d6eaa201aabaa276/text/0000-stop-deleting-cache-images.md)

Fixes: [buildpacks/lifecycle#803](https://github.com/buildpacks/lifecycle/issues/803)

### DIFF
--- a/text/0000-stop-deleting-cache-images.md
+++ b/text/0000-stop-deleting-cache-images.md
@@ -52,7 +52,7 @@ Lifecycle will document this behavior change in Release Notes/Changelog along wi
 
 Why should we *not* do this?
 
-Platform authors relying on this behavior will need to take additional measures to ensure cache image cleanup.
+Platform authors relying on this behavior will need to take additional measures to ensure cache image cleanup or the destination registry will continue to grow.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-stop-deleting-cache-images.md
+++ b/text/0000-stop-deleting-cache-images.md
@@ -1,0 +1,73 @@
+# Meta
+[meta]: #meta
+- Name: Stop deleting cache images
+- Start Date: 2022-03-31
+- Author(s): jabrown85
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: [buildpacks/lifecycle#803](https://github.com/buildpacks/lifecycle/issues/803)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Lifecycle will no longer try to delete previous cache images when publishing cache images.
+
+# Definitions
+[definitions]: #definitions
+
+* Cache Image: The cache stored between builds - stored in a registry.
+* ECR: Amazon's container registry product
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  As discussed in [buildpacks/lifecycle#803](https://github.com/buildpacks/lifecycle/issues/803), some registries (ECR) do not support `DELETE`. For platforms that work exclusively with such registries, the warning output by lifecycle and the time taken to fail is unavoidable and lifecycle is wasting time trying to complete an operation that will never succeed.
+
+- What use cases does it support?
+  All platforms that use cache images against registries that do not support delete will no longer see warning messages.
+
+- What is the expected outcome?
+  Platforms will need to handle cleanup of their cache images on their own, if they desire.
+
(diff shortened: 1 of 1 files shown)


text/0000-stop-deleting-cache-images.md
```diff
@@ -0,0 +1,73 @@
...
+[migration]: #migration
+
+Lifecycle will document this behavior change in Release Notes/Changelog along with the associated Platform API that enables the new behavior.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+Platform authors relying on this behavior will need to take additional measures to ensure cache image cleanup.
```
natalieparellano, line 55: Maybe it's worth calling out explicitly what happens if the cache image is never deleted - I guess it grows and grows indefinitely?

text/0000-stop-deleting-cache-images.md
```diff
@@ -0,0 +1,73 @@
...
+Platform authors relying on this behavior will need to take additional measures to ensure cache image cleanup or the destination registry will continue to grow.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  * Add regex or configuration to drive registry hosts to ignore during DELETE  *
+- Why is this proposal the best?
+  * Lifecycle is not currently cleaning up any other resources
+  * Deleting the cache images can hurt reproducibility
```
hone, line 64: Cache images are already not guaranteed to be restored right?

text/0000-stop-deleting-cache-images.md
```diff
@@ -0,0 +1,73 @@
...
+
+Why should we *not* do this?
+
+Platform authors relying on this behavior will need to take additional measures to ensure cache image cleanup or the destination registry will continue to grow.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  * Add regex or configuration to drive registry hosts to ignore during DELETE  *
```
hone, line 61: Could we flip a flag/configuration where not deleting is the default, but have a way for lifecycle to delete to preserve the existing functionality?

text/0000-stop-deleting-cache-images.md
jabrown85: Correct. Deleting them _ensures_ you can't reliable reproduce an issue with a specific cache image contents.

text/0000-stop-deleting-cache-images.md
jabrown85: I'll add that to the RFC as an alternative

jabrown85: Done in bda8d65.

text/0000-stop-deleting-cache-images.md
```diff
@@ -0,0 +1,74 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Lifecycle will no longer attempt to delete cache images during cache image export.
+
+# Migration
+[migration]: #migration
+
+Lifecycle will document this behavior change in Release Notes/Changelog along with the associated Platform API that enables the new behavior.
```
jromero, line 48: I'm not able to find any reference to the delete operation as part of the current Platform API spec. It seems like this is an implementation feature and should not have any impact on the Platform API.

If that is true, I would expect new versions of the lifecycle to stop deleting images regardless of Platform API version they are operating in.

text/0000-stop-deleting-cache-images.md
```diff
@@ -0,0 +1,74 @@
...
+* ECR: Amazon's container registry product
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  As discussed in [buildpacks/lifecycle#803](https://github.com/buildpacks/lifecycle/issues/803), some registries (ECR) do not support `DELETE`. For platforms that work exclusively with such registries, the warning output by lifecycle and the time taken to fail is unavoidable and lifecycle is wasting time trying to complete an operation that will never succeed.
+
+- What use cases does it support?
+  All platforms that use cache images against registries that do not support delete will no longer see warning messages.
```
natalieparellano, line 30: Another use case this supports - when iterating on an app, and possibly toggling back and forth between different dependencies, this can avoid having to re-download things on every build, saving time for developers.

matthewmcnew (approved):

matthewmcnew:
FWIW, kpack expects this behavior already and users are instructed to delete cache images externally.

jkutner (approved):

AidanDelaney (approved):

jjbustamante (approved):
LGTM

natalieparellano:
Moving this to "voting" with window closing next Thursday, December 22.

PR #217, opened 2022-04-15
Sub-team RFC for lifecycle package re-organization

natalieparellano:
[Readable](https://github.com/buildpacks/rfcs/blob/b92a8e2963227d58471a12e1f7f7a86ece7b1cda/text/0000-lifecycle-pkg-refactor.md)


text/0000-lifecycle-pkg-refactor
```diff
@@ -0,0 +1,174 @@
...
+│    ├── image
+│    ├── io
+│    ├── platform               # services for handling logic that branches on platform api - e.g., metadata restorer, sbom restorer, potentially a new sbom copier, etc.
+│    └── str
+├── pkg
+│    ├── api                    # old api package with CNB logic that indicates supported and deprecated apis moved elsewhere
+│    ├── archive
+│    ├── buildpack
+│    │   ├── env
+│    │   ├── launch
```
natalieparellano, line 75: Actually nothing in `buildpack` imports this code. Maybe it should live at `pkg/launch`.

text/0000-lifecycle-pkg-refactor
```diff
@@ -0,0 +1,174 @@
...
+│        ├── launch
+│        │   └── inputs         # handles platform <-> launcher interface (args, flags)
+│        └── priv
+```
+
+## Considerations
+
+* The structure of the lifecycle should mirror the spec as much as possible. However we should recognize that the spec is not perfect. Also, while there is a `lifecycle` package, there is no "lifecycle spec" - the responsibilities of the lifecycle are mixed between the platform and buildpack specs (see below).
+* In general the `lifecycle` package should be more of an orchestrator - see [example](https://github.com/buildpacks/lifecycle/blob/b6803be364429689a818a5d0a09db42bd21b9995/analyzer.go#L52-L54). We should push as much logic as possible into services.
+* CNB business logic should live inside the `buildpack`, `lifecycle`, and `platform` packages (and their nested packages) or their `internal` equivalents. All other packages that are direct children of `pkg` or `internal` should be CNB-free.
```
natalieparellano, line 95: I think having some sort of rule like this will help us keep our packages well scoped.

text/0000-lifecycle-pkg-refactor
```diff
@@ -0,0 +1,174 @@
...
+* Logic that branches on api version should be handled by an internal service nested under the appropriate package - e.g., logic that branches on buildpack api should live in `internal/buildpack` - see [example](https://github.com/buildpacks/lifecycle/blob/b6803be364429689a818a5d0a09db42bd21b9995/buildpack/bom.go#L14-L23).
```
natalieparellano, line 96: This to me feels like the most helpful organizing principle. Instead of having a bunch of `if api.MustParse(platformAPI).AtLeast("XXX")` sprinkled throughout the code, we can consolidate the branching logic in the constructor, by giving our orchestrator the right services for the api it implements.

text/0000-lifecycle-pkg-refactor
```diff
@@ -0,0 +1,174 @@
...
+
+Changing package import paths will be a breaking change for library consumers (GitHub [search](https://github.com/search?o=desc&q=%22github.com%2Fbuildpacks%2Flifecycle%22&s=indexed&type=Code)). However this is mitigated by version locking in `go.mod`.
+
+There have also (to this point) been no firm stability guarantees when using the lifecycle as a library. If we (hopefully, through this RFC) settle on a package structure that works for us, we might be able to offer some promises of stability in the future.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered? Doing nothing, just keep adding code as best we can.
+- Why is this proposal the best? Making it easier to add new code will allow us to deliver features faster.
```
natalieparellano, line 157: Another benefit of having a well-organized lifecycle is that by adding good README files and docs comments, we can turn the lifecycle into an easy-to-consume repository of information for further understanding the CNB spec.

text/0000-lifecycle-pkg-refactor.md
```diff
@@ -0,0 +1,174 @@
...
+
+  - Stacks
+  - Lifecycle
+    - Operations
+      - Build
+      - Rebase
+      - Launch
+    - Platform inputs and outputs (args, flags, env vars, mounts, exit codes)
+    - Registry auth
+  - Buildpacks directory layout
```
natalieparellano, line 114: It has always sort of bothered me that the `buildpack` package contains [`DirBuildpackStore`](https://github.com/buildpacks/lifecycle/blob/b6803be364429689a818a5d0a09db42bd21b9995/buildpack/store.go#L29-L38) and I think this is why!

text/0000-lifecycle-pkg-refactor.md
```diff
@@ -0,0 +1,174 @@
...
+│    │   ├── analyzer.go        # lifecycle logic moved to nested package
+│    │   ├── detector.go
+│    │   ├── rebaser.go
+│    │   ├── ...
+│    └── platform
+│        ├── auth
+│        ├── cache
+│        ├── inputs             # handles platform <-> lifecycle interface (args, flags)
+│        ├── launch
+│        │   └── inputs         # handles platform <-> launcher interface (args, flags)
```
natalieparellano, line 87: See also https://github.com/buildpacks/lifecycle/pull/805

natalieparellano:
Closing this RFC as ideas have evolved somewhat since its inception, though as an instrument for thinking things through it was still useful.

PR #223, opened 2022-06-19
Additional OCI Artifacts

agracey:
This proposal is about adding new functionality to allow publishing additional artifacts created during the build process.


text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+
+Allow buildpack authors to specify additional artifacts from their buildpacks to be uploaded to a registry along with the standard set of layers. 
+
+# Definitions
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
+
+- WebAssembly (WASM) -- Portable binaries to be run with a WebAssembly virtual machine.
+- OCI Artifact -- Any blob of data that can be stored in an OCI registry
```
natalieparellano, line 23:
```suggestion
- [OCI Artifact](https://github.com/opencontainers/artifacts/blob/main/definitions-terms.md#media-type) -- Any blob of data that can be stored in an OCI registry. From the spec: 
> An artifact has a type, which has a collection of layers. The Artifact is defined as unique by its `manifest.config.mediaType`. Layers are defined by their `layer.config.mediaType`.
```

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+For the WASM usecase, a buildpack author looking to also publish a node.js WASM bundle would write the build logic like any other buildpack. Create a directory in $LAYERS_DIR with the following `layer.toml`:
+
+```
+[types]
+publish = false
+cache = false
+
+[[additionalArtifact]]
+tag = "wasm"
+file = "mywasmprogram.wasm"    # Should be a file in $LAYERS_DIR
```
natalieparellano, line 53: It feels a bit weird to be specifying this at the layer-level (since an artifact could be a collection of layers).

Looking at the `[types]` table above, I wonder if the artifact case could be thought of as the addition of new layer types - `[[media_types]]` maybe?

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+[motivation]: #motivation
+
+Cloud Native Buildpacks are awesome for building container images but come with a draw back that they can only produce a single output. This is limiting to platform builders because there are many cases where you might want to have multiple outputs from the same build process. 
+
+# What it is
+[what-it-is]: #what-it-is
+
+This change allows for additional flexibility in what buildpack authors and platforms can build on top of buildpacks. Two initial use cases are WebAssembly and Helm.
+
+For the first, a buildpack would be able to also generate a wasm bundle that can be used by a downstream platform (such as the Krustlet or runwasi).
```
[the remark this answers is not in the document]
natalieparellano: Similarly, are there cases where you might want multiple different artifacts as an output from a single build? From the "How it Works" section, I gather this could be true (since there could be multiple layers with `additionalArtifacts` specified). Is it possible to give examples of those cases?

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+publish = false
+cache = false
+
+[[additionalArtifact]]
+tag = "wasm"
+file = "mywasmprogram.wasm"    # Should be a file in $LAYERS_DIR
+
+[[additionalArtifact.labels]]
+ConfigMediaType       = "application/vnd.wasm.config.v1+json"
+ContentLayerMediaType = "application/vnd.wasm.content.layer.v1+wasm"
```
natalieparellano, line 57: I also find it a bit strange to apply this at the layer level given that labels are currently specified in [launch.toml](https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml) at the buildpack level. It makes me think that you might want something like artifact.toml at the buildpack level to describe different artifacts (that would get aggregated across buildpacks, just like launch.toml).

natalieparellano: It should be noted that these values would be used by the lifecycle not just to apply labels but also to construct the output manifest.

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Migration
+[migration]: #migration
+
+This functionality should be purely additive and not break existing buildpacks. 
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This does bring in additional scope to the project and could potentially lead to confusion around what OCI image that got published should be used. I think this can be mitigated with good docs and platform output.
```
natalieparellano, line 69: There are also numerous points where the assumption that the output of a build is an "image" is baked in... these would need to be enumerated and thought through.

natalieparellano:
@agracey very interesting idea! This seems like a fairly significant change, but if done right could also deliver a lot of value. Looking forward to the discussion.

text/0000-additonal-oci-artifacts.md
agracey: That's a good point! I'll look into that this week.

text/0000-additonal-oci-artifacts.md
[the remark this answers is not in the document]
agracey: My thinking for "also" was to reduce the size of the engineering effort needed and to allow for this feature to be layered on top of the existing functionality.  Would it be preferred to remove "also" and call out the option in an additional call out?

text/0000-additonal-oci-artifacts.md
agracey: Fair, I'll add some time in my calendar later this week to think through additional drawbacks if we go with a less purely "bolt on top" approach.

agracey:
I apologize for the huge delay. Some stuff came up at work and I had to play engineer for a while. I have half of my time set aside this week explicitly for addressing comments and cleaning up. 

I might also take a first pass at an example implementation if I can 😀

text/0000-additonal-oci-artifacts.md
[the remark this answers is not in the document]
agracey: I was thinking about it and it seems like making the standard image optional is a larger change than needed. I'd suggest waiting until we see adoption of this feature before adding complexity.

text/0000-additonal-oci-artifacts.md
agracey: Updated to show artifacts.toml. I think that keeping it a file per layer makes sense for now as it allows for more control over what's published, cached, and ephemeral. Once there's adoption we can see what patterns authors need and adjust accordingly.

text/0000-additonal-oci-artifacts.md
agracey: Like I commented elsewhere, I think a purely additive approach for now gives us more flexibility in the future once we see what patterns make the most sense. It's also lower touch in implementation so there should be less chance of conflict with other changes. 

The worst case is that an buildpack produces an image that gets discarded, right? And a clever author could make that overhead fairly lightweight.

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Summary
+[summary]: #summary
+
+Allow buildpack authors to specify additional artifacts from their buildpacks to be uploaded to a registry along with the standard set of layers. 
+
+# Definitions
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
```
jkutner, line 20:
```suggestion
```

text/0000-additonal-oci-artifacts.md
```diff
@@ -0,0 +1,91 @@
...
+Cloud Native Buildpacks are awesome for building container images but come with a draw back that they can only produce a single output. This is limiting to platform builders because there are many cases where you might want to have multiple outputs from the same build process. 
+
+# What it is
+[what-it-is]: #what-it-is
+
+This change allows for additional flexibility in what buildpack authors and platforms can build on top of buildpacks. Three initial use cases are WebAssembly, test output, and Helm charts.
+
+For the first, a buildpack would be able to also generate a wasm bundle that can be used by a downstream platform (such as the Krustlet or runwasi). 
+
+Similarly, by allowing the buildpack to produce a helm chart, you could integrate in logic about how to set up an application into the build pipeline where you are able to do bits of code analysis. (NetworkPolicies that lock down an app, sidecars that need to be run, ConfigMaps to pass around out-of-band, etc...)
```
jkutner, line 38: Yes!

jkutner (approved):

jkutner:
Putting this RFC in FCP with the FCP ending on 28th November.

PR #230, opened 2022-08-29
Allow setting environment variables using the build image

samj1912:

[Readable](https://github.com/buildpacks/rfcs/blob/builder-config/text/0000-build-config.md)

[Reference Implementation](https://github.com/buildpacks/lifecycle/pull/899/files)

### DIFF
--- a/text/0000-build-config.md
+++ b/text/0000-build-config.md
@@ -1,0 +1,122 @@
+# Meta
+[meta]: #meta
+- Name: Build config
+- Start Date: 2022-08-29
+- Author(s): [samj1912](https://github.com/samj1912)
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes an easy way to configure build images to allow specifying a `/cnb/config/env.build` CNB environment directory that allows updating the Buildpack `detect` and `build` environment based on the directory.
+
+
+# Definitions
+[definitions]: #definitions
+
+- CNB environment directory: A directory that follows the conventions as defined [here](https://github.com/buildpacks/spec/blob/main/buildpack.md#provided-by-the-buildpacks)
+- Operator: Owner of builders and stacks. See [CNB Operator guide](https://buildpacks.io/docs/operator-guide/).
+
+
+# Motivation
+[motivation]: #motivation
+
+Often times, especially in enterprise settings, organizations often have to update the buildpacks to use internal mirrors, proxies and other settings which can be configured easily with environment variables.
+
+Some examples include -
+- `GOPROXY`
+- `PIP_INDEX_URL`
+- `npm_config_registry`
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in the Buildpacks largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may ideally also take precendence over any user provided values to ensure that the operators have full control over their builders.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The RFC proposes the introduction of the following directory `/cnb/config/env.build` in build images. The directory follows the same convention as a `CNB environment directory`. The notable difference is that the environment variables sourced from this directory are applied **AFTER** processing the user-provided platform environment variables i.e. they should have the highest precedence. These variables should be available during both `detect` and `build` phases (and the `generate` phase in the future).
+
+
+The operator can define this directory in the build image under `/cnb/config` or `CNB_CONFIG_DIR` if defined.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Reference implementation available at https://github.com/buildpacks/lifecycle/pull/899/files
+
+Examples - 
+
+Buildpack value: `FOO=test`
+Build config: `FOO.default=another-value`
+Final value: `FOO=test`
+
+
+Buildpack value: `FOO=test`
+Build config: `FOO.append=another-value, FOO.delim="`
+Final value: `FOO=test:another-value`
+
+Buildpack value: `FOO=test`
+Build config: `FOO.override=another-value`
+Final value: `FOO=another-value`
+
+Buildpack value: `FOO=test`
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO=another-value`
+Final value: `FOO=value`
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO.override=another-value`
+Final value: `FOO=another-value`
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO.prepend=another-value, FOO.delim=:`
+Final value: `FOO=another-value:value`
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO.append=another-value, FOO.delim=:`
+Final value: `FOO=value:another-value`
+
+In case a buildpack uses `clear-env=true` then it is up to the buildpack will not see user provided platform values unless it looks in the platform directory and resolve the user provided platform values against the values set by previous buildpacks and the build config.
+
+# Migration
+[migration]: #migration
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Increased complexity in sourcing of environment variables.
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+## Implementing at a Buildpack level
+
+See https://github.com/paketo-buildpacks/rfcs/pull/241
+
+# Prior Art
+[prior-art]: #prior-art
+
+- See https://github.com/paketo-buildpacks/rfcs/pull/241 and 
+- See the CNB BAT meeting. https://youtu.be/e8FgLwVN5VQ?t=1153
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+N/A
+
+# Spec. Changes (OPTIONAL)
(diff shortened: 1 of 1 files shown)


text/0000-build-config.md
```diff
@@ -0,0 +1,87 @@
...
+
+Some examples include -
+- `GOPROXY`
+- `PIP_INDEX_URL`
+- `npm_config_registry`
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in the Buildpacks largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may ideally also take precendence over any user provided values to ensure that the operators have full control over their builders.
```
jkutner, line 38: hmm, this feels unexpected

text/0000-build-config.md
```diff
@@ -0,0 +1,87 @@
...
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Increased complexity in sourcing of environment variables.
+
+# Alternatives
+[alternatives]: #alternatives
```
jkutner, line 63: what's this difference in this over a `ENV` directive the build image's Dockerfile?

samj1912: @jkutner lifecycle currently strips out all of those variables. See https://github.com/buildpacks/lifecycle/blob/4f52db0a90b5637e718328e7019bf8e592fdeece/env/build.go#L8

text/0000-build-config.md
samj1912: This mostly provides a choice to operators. If they don't want to have a higher precedence than user provided variables, they can choose to use `VAR_NAME.default`. If they do want a higher precedence they can choose to use `VAR_NAME.override`. The reason this needs to be at a higher precedence level than the user is to allow modifying user args to conform to any buildpack security/compliance requirements. For eg. being able to prepend/append args to user provided ones or being able to override user-provided invalid arguments for variables like proxies/mirrors etc. if desired. It always allows them to give higher precedence to user provided args via `.DEFAULT`. Buildpacks are already able to do this via setting the `clear-env` flag in their buildpack.toml or choosing to override/modify input user args.

samj1912: Another alternative may be to introduce a `env.pre` and `env.post` folder to allow for both behaviors if that is important. But for us, being able to sanitize/mutate user input is extremely important. Sort of similar to k8s mutating/validation webhooks.

jromero: > Another alternative may be to introduce a env.pre and env.post folder to allow for both behaviors if that is important. But for us, being able to sanitize/mutate user input is extremely important. Sort of similar to k8s mutating/validation webhooks.

Would these folders be in `/cnb/platform`?

My main concern from this RFC, is "yet another way to configure the same thing". If this RFC is primarily focused on order of precedence and not portability (embedded in build image) then I'd prefer to see if we can adapt `/cnb/platform` to provide the desired behavior.

samj1912: @jromero it is both. The portability aspects of it are fairly important.

> Would these folders be in /cnb/platform? 

These would be folders in the build image.

dmikusa-pivotal:
```suggestion
The environment variables may be able to take precedence over user-provided values, if configured to, which ensures that operators can take full control over their builders if that is required.
```

Rephrases this a bit. I agree the default behavior should be to have user-specified config take precedence, but I also know operators might want or need to have control as well.

text/0000-build-config.md
```diff
@@ -0,0 +1,87 @@
...
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in the Buildpacks largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may be able to take precedence over user-provided values, if configured to, which ensures that operators can take full control over their builders if that is required.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The RFC proposes the introduction of the following directory `/cnb/build-config/env` in build images. The directory follows the same convention as a `CNB environment directory`. The notable difference is that the environment variables sourced from this directory are applied **AFTER** processing the user-provided platform environment variables i.e. they should have the highest precedence. These variables should be available during both `detect` and `build` phases (and the `generate` phase in the future).
```
dmikusa-pivotal, line 43:
```suggestion
The notable difference is that the environment variables sourced from this directory are applied **AFTER** processing the user-provided platform environment variables i.e. they should have the highest precedence. This allows them to potentially override user-specified environment variables, but that is not the default behavior. An operator must mark an environment variable as `override` to indicate it should replace a value set by a user.
```

When I first read this, I thought it would squash all my user-specified env variables, but if I'm understanding correctly the behavior is more nuanced. It won't squash them unless `override` is set. This change tries to clarify so it sounds a little less scary.

text/0000-build-config.md
```diff
@@ -0,0 +1,87 @@
...
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in the Buildpacks largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may be able to take precedence over user-provided values, if configured to, which ensures that operators can take full control over their builders if that is required.
+
+# What it is
+[what-it-is]: #what-it-is
+
+The RFC proposes the introduction of the following directory `/cnb/build-config/env` in build images. The directory follows the same convention as a `CNB environment directory`. The notable difference is that the environment variables sourced from this directory are applied **AFTER** processing the user-provided platform environment variables i.e. they should have the highest precedence. This allows them to potentially override user-specified environment variables, but that is not the default behavior. An operator must mark an environment variable as `override` to indicate it should replace a value set by a user. These variables should be available during both `detect` and `build` phases (and the `generate` phase in the future).
```
natalieparellano, line 43: > This allows them to potentially override user-specified environment variables, but that is not the default behavior. An operator must mark an environment variable as `override` to indicate it should replace a value set by a user.

Could you give some examples? That might help to grok it better.

text/0000-build-config.md
natalieparellano: From 9/1 working group:
* a buildpack that always runs first: doesn't set env vars for detect
* `/platform/env`: overridable by the buildpack (via `clear-env`), not portable, ???

text/0000-build-config.md
natalieparellano, line 43: > `/cnb/build-config/env`
What about making this `/cnb/env.build`? (to be consistent with the buildpack API)

jkutner (approved):

text/0000-build-config.md
samj1912: @natalieparellano do we want to be able to provide a way to change the location of this directory by default? If so should we provide something specific to the env directory or `CNB_BUILD_CONFIG_DIR` as it is currently suggested. If so, should that value default to `/cnb` instead?

jabrown85 (approved):

jromero (approved):

hone (approved):

text/0000-build-config.md
```diff
@@ -0,0 +1,108 @@
...
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO.prepend=another-value, FOO.delim=:`
+Final value: `FOO=another-value:value`
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO.append=another-value, FOO.delim=:`
+Final value: `FOO=value:another-value`
+
+In case a buildpack uses `clear-env=true` then it is up to the buildpack to resolve the user provided platform values against the values set by previous buildpacks and the build config.
```
natalieparellano, line 71: This sentence is a bit unclear to me - if `clear-env=true` the buildpack will not see user provided platform values unless it looks in the platform directory - is this what you mean?

text/0000-build-config.md
```diff
@@ -0,0 +1,108 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+Reference implementation available at https://github.com/buildpacks/lifecycle/pull/899/files
+
+Examples - 
+
+Platform Enviroment varaible: `FOO=value`
+Build config: `FOO=another-value`
+Final value: `FOO=value`
```
natalieparellano, line 57: Is it possible to include a buildpack-provided value in this example? If I'm not mistaken, the order of application is:
buildpack -> platform (behavior: override) -> builder (behavior: configurable, with default `default`)

I recall that changing the order of application would be a larger breaking change - but what about changing the platform behavior so that it is:
buildpack -> platform (behavior: configurable, with default `default`) -> builder (behavior: configurable, with default `default`)

Platforms such as `pack` could add `.override` to env var files, to keep the same behavior.

natalieparellano (approved):
I would be really happy if we could make the behavior of the platform env directory more consistent with the builder env directory. That said, if that is too big of a change to tackle right now, I'm also fine to wait until some later date to improve the env API.

text/0000-build-config.md
samj1912: Yup, exactly that.

samj1912:
@natalieparellano the RFC should be up to date now. Given the number of approvals already, I will move the RFC to the voting period.

samj1912:
Moving to voting period with voting ending on 10th November 2022.

samj1912:
@natalieparellano the platform directory change seems like something that could be done independently. I am wary of scope creep on this RFC and gathering consensus again on the new set of changes. Would it be okay if we merge this RFC in first (given that we already have a draft PR implemention in lifecycle) and then have another RFC to make platform behaviour more consistent with the rest of the env API?

natalieparellano:
@samj1912 I am good with that 🙂 ...if there's general consensus that the platform change IS something we'd like to do, I'd be happy to draft the RFC.

natalieparellano:
Should we move this to "voting" with a close date?

samj1912:
/queue-issue buildpacks/lifecycle "Allow setting environment variables using the build image"
/queue-issue buildpacks/spec "Allow setting environment variables using the build image" 
/queue-issue buildpacks/docs "Allow setting environment variables using the build image"

Issue #237, opened 2022-11-09
[RFC #0119] - Export to OCI format


jjbustamante:
[RFC #203](https://github.com/buildpacks/rfcs/blob/main/text/0119-export-to-oci.md) - Export to OCI format

# Phase 1

The goal for this phase is to deliver an experimental version of the feature, wait for the community to use it and gives feedback to keep improving it

### Spec
- [x] https://github.com/buildpacks/spec/issues/327
- [x] Released in platform API version `0.12`

### Imgutil
- [x] https://github.com/buildpacks/imgutil/issues/162
- [x] https://github.com/buildpacks/imgutil/issues/168
- [x] https://github.com/buildpacks/imgutil/issues/174
- [x] https://github.com/buildpacks/imgutil/issues/179

### Lifecycle
- [x] https://github.com/buildpacks/lifecycle/issues/953
- [x] Released in lifecycle version `0.17.0`
- [x] https://github.com/buildpacks/lifecycle/issues/1194
- [x] Released in lifecycle version `0.17.1`

### Pack:
- [x] https://github.com/buildpacks/pack/issues/1548
- [x] Released in pack version `0.30.0`
- [x] https://github.com/buildpacks/pack/issues/1881
- [x] Released in pack version `0.30.1`

### Documentation:
- [x] https://github.com/buildpacks/docs/issues/529

# Phase 2

The initial idea of this phase was to work in the `rebase` capability in pack for an Image exported on disk in OCI layout format, we are not sure if this is useful and if it worth it. We would like to hear from the community if this is something valuable to work on

### Imgutil
- [ ] https://github.com/buildpacks/imgutil/issues/170

### Lifecycle
- [x] https://github.com/buildpacks/lifecycle/issues/957
- [ ] Released in lifecycle version `TBD`

jjbustamante:
Yes! we can close it. We decided not to move forward with phase 2

PR #235, opened 2022-10-12
Proposal for donating kpack to CNB

jjbustamante:
[Readable](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/kpack-donation/text/0000-kpack-donation-to-cnb.md)

Signed-off-by: Juan Bustamante <jbustamante@vmware.com>

PR #238, opened 2022-11-18
This follows up on #202

david-caro:
[Readable](https://github.com/buildpacks/rfcs/blob/rfc/prepare-operation/text/0000-prepare-operation.md)
#### 2022-11-18 ([3bb05f8](https://github.com/buildpacks/rfcs/commit/3bb05f8a915e308ea7612173f3fa81bb4b382b96))
    * [Fix some typos](https://github.com/buildpacks/rfcs/pull/237/commits/94e318065db8df8d76638473e241c64ee7c82156)

#### 2022-02-22 ([94e3180](https://github.com/buildpacks/rfcs/commit/94e318065db8df8d76638473e241c64ee7c82156), [d02754f](https://github.com/buildpacks/rfcs/commit/d02754ffd2b13b0ee42118f502199f30c8407ef8))

    * [Rename io.buildpacks.defaults back to io.buildpacks](https://github.com/buildpacks/rfcs/pull/237/commits/94e318065db8df8d76638473e241c64ee7c82156)

    * [Answer arbitrary properties question](https://github.com/buildpacks/rfcs/pull/237/commits/d02754ffd2b13b0ee42118f502199f30c8407ef8)

#### 2022-02-10 ([8fc7a70](https://github.com/buildpacks/rfcs/commit/8fc7a70b111e2d8d408071d2565a2bf2dae3e15f))

    * Add a detailed summary

    * Give the `io.buildpacks` namespace an independent schema version

#### 2022-02-07 ([99fc145](https://github.com/buildpacks/rfcs/commit/99fc145c93f07e045e62b9c5d42405ecffc53afc))

    * Initial RFC

### DIFF
--- a/text/0000-prepare-operation.md
+++ b/text/0000-prepare-operation.md
@@ -35,7 +35,7 @@ The motiviation for this change was to enable the following goals from being ach
   - **Goal 3: Platform recognition of project.toml**
     > As a user, I would like to ensure that my configuration in `project.toml` is being used.
 
-NOTE: This RFC directly solves for Goal 3, while enabling the solutions for Goals 1 and 2 via a supplimental RFC ([Support for pack.toml][rfc-pr-189]).
+**This RFC directly solves for Goal 3**, while enabling the solutions for Goals 1 and 2 via a supplimental RFC ([Support for pack.toml][rfc-pr-189]).
 
 [rfc-pr-189]: https://github.com/buildpacks/rfcs/pull/189
 
@@ -44,13 +44,15 @@ NOTE: This RFC directly solves for Goal 3, while enabling the solutions for Goal
 
 The proposal is composed of following changes:
 
-1. [Moving `io.buildpacks` properties to `io.buildpacks.defaults`.](#Namespace-iobuildpacksdefaults)
-2. [A replaceable new phase `prepare` that applies platform configuration.](#Preparer)
-3. [Supporting Cloud Native Buildpack utilities.](#Cloud-Native-Buildpacks-utilities)
+- [Changes to `io.buildpacks` namespace.](#changes-to-iobuildpacks-namespace)
+  - [Define criteria for properties in namespace.](#criteria-for-properties)
+  - [Version `io.buildpacks` namespace.](#version-iobuildpacks-namespace)
+- [A replaceable new phase `prepare` that applies platform configuration.](#prepare-phase)
+- [Supporting Cloud Native Buildpack utilities.](#cloud-native-buildpacks-utilities)
 
 ## Changes to `io.buildpacks` namespace
 
-### Namespace `io.buildpacks.defaults`
+### Criteria for properties
 
 This namespace declares default values that platforms should acknowledge. A majority of properties should be spec'd but we would allow for unspec'd additional properties.
 
@@ -58,7 +60,7 @@ The criteria for spec'd properties in this namespace would be based on whether t
 
 Therefore the following changes have been made from the [latest schema][pd-02]:
 
- - Remove `builder`
+ - **Remove `builder` from `io.buildpacks` namespace**
 
 [pd-02]: https://github.com/buildpacks/spec/blob/extensions/project-descriptor/0.2/extensions/project-descriptor.md#iobuildpacks-optional
 
@@ -74,7 +76,7 @@ See [`io.buildpacks` Namespace Schema](#io-buildpacks-Namespace-Schema) section 
 
 ###### Multi-Platform (pack + kpack)
 
-In the following example we show how various platform namespaces can have their own configuration but properties can be promoted to `io.buildpacks.defaults` if the user's intent is to apply it to all platforms.
+In the following example we show how various platform namespaces can have their own configuration but properties can be promoted to `io.buildpacks` if the user's intent is to apply it to all platforms.
 
 ```toml
 [io.buildpacks]
@@ -84,7 +86,7 @@ schema-version = "0.2"
 # common buildpacks config
 ##
 
-[io.buildpacks.defaults]
+[io.buildpacks]
 exclude = ["some-files/**"]
 
 ##
@@ -94,8 +96,7 @@ exclude = ["some-files/**"]
 [io.buildpacks.pack]
 schema-version = "0.1"
 builder = "cnbs/sample-builder:bionic"
-image = "my-app"                        # (example only)
-run-image = "my-run-image:latest"       # (example only)
+cache-volume = "my-app-cache"           # (example only)
 
 ##
 # kpack config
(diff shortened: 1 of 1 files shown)


samj1912:
In general one alternative I would to see reflected in this proposal and discussed is the the following interface -


preparer </path/to/lifecycle-config.toml>

where the lifecycle config is a viper config file that maps to the various lifecycle binary flags + env vars. (See https://github.com/spf13/viper#reading-config-files). 

The preparer would take the above file and modify any of the inputs in the file and pass it back to the platform to be supplied to the lifecycle. 

The above file would also need to include the platform API. This way the preparer becomes a part of the platform API and has a way of modifying all the inputs needed. 

This can be used to implement all of the possible functionality we could ever want from a prepare stage and this interface would be forward-compatible as long as the preparer is compatible with a certain platform API.

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The motivation for this change was to enable the following goals from being achieved:
+
+  - **Goal 1: Serializing CLI configuration**
+    > As a user, I would like to be able to serialize and share the parameters I use with certain platforms such as `pack`.
+  - **Goal 2: A recognizable file in repositories**
+    > As a user, I would like to be able to recognize, based on the file system, if a project is using Cloud Native Buildpacks.
```
jkutner, line 34: I'm not sure I understand how this goal fits in

jkutner: Is it because I can not depend on `project.toml` to work, regardless of platform?

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Criteria for properties
+
+This namespace declares default values that platforms should acknowledge. A majority of properties should be spec'd but we would allow for unspec'd additional properties.
+
+The criteria for spec'd properties in this namespace would be based on whether they affect inputs to the lifecycle phases. For example, the order and environment variables are inputs to the lifecycle. The `builder`, in contrast, is not an input to the lifecycle so it wouldn't meet the criteria.
+
+Therefore the following changes have been made from the [latest schema][pd-02]:
+
+ - **Remove `builder` from `io.buildpacks` namespace**
```
jkutner, line 63: Do we have a plan for backwards compat?

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+[io.buildpacks.pack]
+schema-version = "0.1"
+builder = "cnbs/sample-builder:bionic"
+cache-volume = "my-app-cache"           # (example only)
+
+##
+# kpack config
+##
+
+[com.vmware.kpack]
```
jkutner, line 105: the `com.vmware` is a bit problematic. Does kpack have a vendor neutral namespace?

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+The default implementation COULD take care of applying the following configuration:
+
+- `io.buildpacks.group` →
+    - Download buildpacks
+    - Update `order.toml`
+- `io.buildpacks.build.env` →
+    - Set build env vars in `<platform>/env`
+- Notify users of any other properties in `io.buildpacks`
+
+# Drawbacks
```
jkutner, line 163: Are there any drawbacks around efficiency? Like this change to `excludes` could impact performance, because it means that very large files that are excluded might still have to be sent to some remote platform that runs prepare and then removes the exclude files later

text/0000-prepare-operation.md
```diff
@@ -0,0 +1,387 @@
...
+[drawbacks]: #drawbacks
+
+1. Parsing and applying the `io.buildpacks` namespace becomes responsibility of the platform.
+    - This is mitigated by providing [utilities](#Cloud-Native-Buildpacks-utilities) and the fact that the prepare phase is independent and swappable.
+2. Executing the Prepare operation may require an additional container to be spun up in some platforms; this would effectively lengthen the overall build process.
```
jkutner, line 168: Do we expect this will work with `creator`?

text/0000-prepare-operation.md
david-caro: I'm guessing that will depend on #235, but I'll look into it

david-caro:
> In general one alternative I would to see reflected in this proposal and discussed is the the following interface -
> 
> preparer </path/to/lifecycle-config.toml>
> 
> where the lifecycle config is a viper config file that maps to the various lifecycle binary flags + env vars. (See https://github.com/spf13/viper#reading-config-files).
> 
> The preparer would take the above file and modify any of the inputs in the file and pass it back to the platform to be supplied to the lifecycle.
> 
> The above file would also need to include the platform API. This way the preparer becomes a part of the platform API and has a way of modifying all the inputs needed.
> 
> This can be used to implement all of the possible functionality we could ever want from a prepare stage and this interface would be forward-compatible as long as the preparer is compatible with a certain platform API.
> 
> Related #128
> 
> Maybe it's going too much into implementation details, but standardizing lifecycle input parsing on viper would allow us to expose a consistent format to map lifecycle binary CLI args <> config file <> lifecycle env vars and define an order of precedence for them.

As I understand, this would introduce a new configuration file (the lifecycle-config.toml) and an new configuration file parsing strategy.

* About using the new `lifecycle-config.toml` file, wouldn't that be duplicating what the project descriptor is for? (according to https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)

* About using viper for config parsing and such, I think it would be great to consolidate on it yes :), but out of scope of this RFC? I would not mind working on it though, aside from this maybe. Do you think that just an issue would be enough or a new RFC is required? (I'm not very familiar with the processes of the group yet :} )


(btw. I can absolutely add the option you requested to the doc, just let me know, I'm trying to discuss/understand the details)

text/0000-prepare-operation.md
david-caro: I think so yes. The project descriptor file currently can be ignored by the platform in any case right?
This adds the restriction that now they `MUST` apply at least the `io.buildpacks` section, making the combo (`project.toml` + `io.buildpacks`) a way to identify if the project uses buildpacks, regardless of the platform.

So summarizing my own words, this RFC ensures that if there's a `project.toml`, and it has a `io.buildpacks' section, then you can rely on it using those buildpacks regardless of platform.

Does that make sense? (/me is still trying to understand both the original motivation and current flows, so please be extra verbose with the replies :pray: )

text/0000-prepare-operation.md
david-caro: Not that I see, I think it's relying on the `schema-version` (see below) to do that breaking change.

Do we need one? (I can add bot a backwards compatible/deprecation period and a removal procedure to the rfc)

jabrown85, line 63: Does `pack` support this key today? I think my vote would be for a platform to fail building (or warn, I guess) if the builder key isn't supported. But if we remove it entirely I suppose I won't be too upset. Platforms can use their own namespace to achieve the same thing.

natalieparellano:
@david-caro were you still interested in moving this one forward?

david-caro:
I am yes, though I'm quite busy lately, feel free to take over if you want to push it

natalieparellano:
@david-caro sounds good, I'll move this into draft for now. We can move it out of draft when someone has the bandwidth to pick it up :)

Issue #258, opened 2022-12-02
Idea: provide in-container env var interpolation for direct processes


natalieparellano:
From RFC 0093:

## Using Environment Variables in a Process

One upside to our previous execution strategy was that it enable users to include environment variable references in arguments that were later evaluated in the container. To preserve this feature we can instead adopt the Kubernetes strategy for [environment variables interpolation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config). If a buildpack or user includes `$(<env>)` in the `command` or `args` and `<env>` is the name of an environment variable set in the launch environment, the launcher will replace this string with the value of the environment variable after apply buildpack-provided env modifications and before launching the process.

# How it Works
[how-it-works]: #how-it-works

### Example 1 - A Shell Process
The Paketo .Net Execute Buildpack may generates shell processes similar to the following:
```
[[processes]]
type = "web"
command = "dotnet my-app.dll --urls http://0.0.0.0:${PORT:-8080}"
direct = false
```
NOTE: the buildpack API used by this buildpack (`0.5`) predates the introduction of `default`.

Using the new API this process could look like:
```
[[processes]]
type = "web"
command = ["dotnet", "my-app.dll", "--urls", "http://0.0.0.0:$(PORT)"] # the default value of PORT would need to be provided in a layer
default = true
```
Things to note:
* In the above example I have eliminated the dependency on Bash instead of explicitly adding it to the command, because it is likely unnecessary.
* If the buildpack authors believed that `--urls` should be overridable they could set move the last two arguments from `command` to `args`.

### Example 2 - A Script Process
When a buildpack upgrade to the new buildpack API, it must convert any existing shell processes into direct processes.

The [Paketo Yarn Start Buildpack](https://github.com/paketo-buildpacks/yarn-start) currently may generate a script processes similar to the following:

```
[[processes]]
type = "web"
command = "pre-start.sh && nodejs server.js && post-start.sh"
direct = false
default = false
```

Using the new API this process look like:
```
[[processes]]
type = "web"
command = ["bash", "-c", "pre-start.sh && nodejs server.js && post-start.sh"]
default = false
```

## User Provided Processes

[...]

PR #259, opened 2022-12-02
Remove references to custom env templating

natalieparellano:
As this is a breaking change, we decided to do this in a separate (yet to be created) RFC. Created issue: https://github.com/buildpacks/rfcs/issues/258 In addition to the changes described originally in 0093 we'd like some way of versioning the launcher interface,
  to avoid surprising end-users.

Signed-off-by: Natalie Arellano <narellano@vmware.com>

### DIFF
--- a/text/0093-remove-shell-processes.md
+++ b/text/0093-remove-shell-processes.md
@@ -376,3 +376,28 @@ command = ["<command>"]
 args = ["<arguments>"]
 default = false
 ```
+
+# History
+[history]: #history
+
+## Amended
+### Meta
+[meta-1]: #meta-1
+- Name: Removed references to custom env templating
+- Start Date: 2022-12-02
+- Author(s): natalieparellano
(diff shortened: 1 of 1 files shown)


text/0093-remove-shell-processes.md
```diff
@@ -295,7 +232,6 @@ we would convert it to the following in `metadata.toml`
 type = "hi"
 command = ["echo", "hello", "world"]
 args = []
-api = "0.5"
```
natalieparellano, line 298: This ended up not being necessary, as the buildpack-id is actually a part of this table.

edmorley:
Thank you for updating this RFC - it referring to the env interpolation feature caused me a bit of confusion a few days ago, when I was trying to use the feature (since the rest of this RFC was implemented, making me think env var interpolation was too).

samj1912 (approved):

PR #274, opened 2023-02-01
RFC: Execution Environments

hone:
[Readable](https://github.com/buildpacks/rfcs/blob/execution-environments/text/0000-execution-environments.md)


text/0000-execution-environments.md
```diff
@@ -0,0 +1,185 @@
...
+The specifics of creating development enviroments are out of scope of this RFC, but it's not hard to extrapolate how these kind of changes can assist in creating Buildpacks for development environments.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## `exec-env` key in TOML
+
+In order to support additional execution environments an `exec-env` key will be added to various TOML tables in the project. The value can be any string with `all` having special meaning. `all` will apply to all execution environments and will be the default if not specified. This should make it backwards compatible and optional. When `exec-env` is not set to `all`, the table settings will only be applied to that execution environment.
+
+### Project Descriptor - `project.toml` (App Developers)
```
jabrown85, line 75: Can you add an example of a full project.toml that is used for producing a test image and a production image?

hone: Just for Project Descriptor?

samj1912:
I'm having a difficulty trying to grasp how environment is being used by different toml files. @hone could you please provide examples of places where you imagine it being used and how that information will be leveraged by buildpacks/platform?

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+## `CNB_EXEC_ENV` Environment Variable
+
+This env var will reserve the following values:
+
+* production
+* test
+* development
+
+### Buildpack API
```
jabrown85, line 140: Can we perhaps get a conceptual example of what a buildpack author might do with this new execution environment information?

I have in my head a language buildpack author may...
- install test group dependencies
- create `ruby-tests` process / `ruby-tests-verbose` process
- Skip cleaning up things they might clean up otherwise for production images
- Set env vars to test mode operation (RAILS_ENV or similar)

loewenstein:
Frankly, I don't get this proposal aligned with my mental model of what buildpacks do and how they help.

For me, container images are just another software artefact nowadays produced to ship software. Like a while back one was to produce JAR and/or WAR files that got delivered and finally deployed into some execution environment.
Buildpacks are a great tool to create those, but I don't see how or why I would create a test container that was separate from the production ones. I would want my tests to validate exactly the artefact that I am about to deliver. Otherwise, how do I know I checked the right thing.

From my POV buildpacks should focus on that task, creating the container image, and leave the other CI tasks to specialized tools.

Please help me to adjust my mental model, why's this added complexity in the buildpack spec worth it?

jabrown85:
> Please help me to adjust my mental model, why's this added complexity in the buildpack spec worth it?

I'm not the author but I can speak for my own mental model.

One use case is from the app developer side. Today they can `pack build ` to get a production artifact.

If we imagine the resulting image is ruby + nodejs, the developer has to setup both of those environments locally to develop and AGAIN on say GitHub Actions. The way those installations happen can differ. The GitHub action may install a different minor version of node or ruby to run the tests for instance. 

Setting up CI with buildpacks could make this experience easier. `pack test` would use the same buildpacks and therefore test a _more_ prod-like experience as the buildpack would install the same versions. This is especially important if your buildpack has options like `DO_THIS_GARBAGE_COLLECTION_SETTING` that you would have to know and replicate in your CI environment.

Another future-proofing thing here may be around future ARM support. Building and running an ARM test image via `pack test` seems tractable.

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+These are specific to the Google Cloud Buildpacks for setting a development code path to work with skaffold.
+
+## [Develop API](https://github.com/buildpacks/spec/pull/71)
+
+The original Cloud Native Buildpacks spec included a Develop API, but it was never implemented.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- "env" is overloaded as a word since we also use it for environment variables. Is there a better word here?
```
natalieparellano, line 193: How about "mode"?

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+[[io.buildpacks.build.env]]
+name = "PARALLEL_WORKERS"
+value = "4"
+exec-env = "test"
+```
+
+### `builder.toml` (Builder Authors)
+
+The only table that `exec-env` will be added to is `[[order.group.env]]`.
```
natalieparellano, line 126: Do you mean `[[order.group]]`?

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+[[io.buildpacks.group]]
+id = "buildpacks/headless-chrome"
+version = "latest"
+exec-env = "test"
+
+[[io.buildpacks.post.group]]
+id = "buildpacks/procfile"
+version = "latest"
+
+[[io.buildpacks.buildenv]]
```
natalieparellano, line 108:
```suggestion
[[io.buildpacks.build.env]]
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+App Source -> OCI Image (Test Environment) -> Execute (Test Process)
+```
+
+### Division of Responsibility
+With the test OCI Image, a platform can execute the tests in the pipeline as they see fit. This means a bulk of the responsibilities are platform concerns:
+
+- Set which environment to build for
+- Decide which buildpacks to execute
```
natalieparellano, line 48: Isn't this a lifecycle concern? i.e., when processing a group within an order, should it skip buildpacks that do not declare `exec-env` matching the desired env?

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+- How to execute the tests
+- What is the test result format like [TAP](https://en.wikipedia.org/wiki/Test_Anything_Protocol)?
+- How to process the test results
+- What to do with the results
```
natalieparellano, line 52: Is this ever something we would want to spec? Having a consistent way for buildpacks to e.g., dump test output could help ensure portability across platforms.

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- "env" is overloaded as a word since we also use it for environment variables. Is there a better word here?
+- Should there be builders that are specific to an execution environment? What about `include` or `exclude`?
+- Should the execution environments be an enum or flexible as a string?
+  - enums will help encourage standardization across buildpacks and platforms.
+  - strings can help account for use cases we haven't thought of yet.
+- Should buildpacks be allowed specify allowlist execution environments?
```
natalieparellano, line 198: Are there any downsides to doing this? It would be more flexible and possibly avoid some duplication within orders

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Setting the Execution Enviroment for Build + Buildpack
+A platform will set the `CNB_EXEC_ENV` env var to the execution environment desired. Buildpacks can than read this env var to branch or switch on logic needed based on the execution environment.
+
+In addition, Builder Authors, Buildpack Authors, and App Developers will be able to specify various options to a specific execution enviroment using the `exec-env` key.
+
+## Development Environments
+The specifics of creating development enviroments are out of scope of this RFC, but it's not hard to extrapolate how these kind of changes can assist in creating Buildpacks for development environments.
+
+# How it Works
```
natalieparellano, line 68: It would be nice to see the pack flow as well

natalieparellano: Should we add anything to the app image to designate it as being built for a particular environment? To avoid users accidentally deploying a test image in production...

I could see folks wanting to use the same tag when re-building a test image for production, in order to use previously cached dependencies.

hone: @natalieparellano that's a good question. Do you think cached dependencies should be shared b/t different execution environments?

text/0000-execution-environments.md
hone: When talking to @jabrown85, I thought the Platform/Builder provide the `order.toml` to `lifecycle`?

text/0000-execution-environments.md
hone: I could see us wanting to do that, but I wasn't sure how much we wanted to impose standards there.

jama22:
Really excited for this proposal, and I think there's a lot of interesting ways to make use of it.

**In the context of a self-hosted PaaS** 
I've seen a common dynamic where the "operator" needs to establish fine grained controls over what the final containers look like. These configurations are also prone to change as the artifact itself moves into different environments (service connectors, certificates, environment variables, access to secrets managers, etc.) 

On the developer side, most folks don't ever see production-like environments, so being able to debug their applications in production-like environments becomes incredibly difficult. In the past, I've seen some of these per-environment configurations being managed through GitOps and config files. This gives the user some control over managing the movement of the workload through the system, but even still they may not be able to build that container and interact with it directly

Moving the environment configuration opens up a lot of possibilities in this use case. I could image an operator encoding their per-environment configurations into the builder, and giving the developer the ability to simulate how their application will change in behavior across configurations. 

**For CI/CD**
I think there have been some interesting ideas already discussed above  and in the proposal. But I think a common use case that i can come up with is if you're building a library with buildpacks, you'll want to test on multiple OS-distros and architecture combinations (e.g. x86 vs win vs arm, ubuntu vs debian vs rhel). Those env configurations are commonly configured in CI itself or possibly externalized to some other config. Moving them back into the buildpack may provide for an improved UX

**Changing what "prod-like" means**
I think there's a pretty reasonable use case here where operators may want to give developers the ability to change what "prod-like" means. For example, the GCP buildpacks has [GOOGLE_DEVMODE](https://cloud.google.com/docs/buildpacks/service-specific-configs#google_devmode) that can be enabled for faster changes. I can see it being used as a developer-oriented configuration while maintaining guardrails in prod-like envs

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+## [Heroku Testpack API](https://devcenter.heroku.com/articles/testpack-api)
+
+As part of the classic buildpack design, Heroku extended the API to include support for tests. It added a `bin/test-compile` phase, which is like the normal `bin/compile` phase but specific for setting things up for test. `bin/test` was called for executing tests. This design let a buildpack have a specific codepath for tests and also do intelligent logic for executing tests based on the source code. For example [in Ruby](https://github.com/heroku/heroku-buildpack-ruby/blob/main/bin/support/ruby_test#L62-L76), the buildpack author can run `bin/rspec` if it detects `rspec`, `bin/rails test` when using rails, or default to `rake test`.
+
+There were some flaws in this design. Though it's clean to separate production and test code paths, they end up sharing a lot of code. Many of the bash based Heroku buildpacks would just [call `bin/compile`](https://github.com/heroku/heroku-buildpack-nodejs/blob/main/bin/test-compile#L24) with different parameters/env vars.
+
+## [GOOGLE_DEVMODE](https://cloud.google.com/docs/buildpacks/service-specific-configs#google_devmode)
```
jama22, line 182: Similar to devmode, I just ran across live reloading in the Paketo python buildpacks https://paketo.io/docs/howto/python/#using-bp_live_reload_enabled

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+- Execution Environment - The target environment the OCI image is expected to be run in, i.e. production, test, development.
+
+# Motivation
+[motivation]: #motivation
+
+The main way Buildpacks are being used is building production images, but this is only one piece of the CI/CD pipeline. Without a solid buildpack test environment story, users will be required alternatives for building their environment. A testing environment, while different, shares many of the same broad strokes ultimately producing an execution environment. This is something Buildpacks are well suited to solve.
```
joshwlewis, line 25:
```suggestion
The main way Buildpacks are being used is building production images, but this is only one piece of the software development process. Without a solid buildpack test environment story, users will be required alternatives for building their environment. A testing environment, while different, shares many of the same broad strokes ultimately producing an execution environment. This is something Buildpacks are well suited to solve.
```

I'm thinking this RFC could be useful outside of CI/CD pipelines, most notably for testing changes locally. However, I could also imagine building development environments locally with features like hot reload, auto recompilation, web servers that respond with stack traces, etc.

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Development Environments
+The specifics of creating development enviroments are out of scope of this RFC, but it's not hard to extrapolate how these kind of changes can assist in creating Buildpacks for development environments.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## `exec-env` key in TOML
+
+In order to support additional execution environments an `exec-env` key will be added to various TOML tables in the project. The value can be any string with `all` having special meaning. `all` will apply to all execution environments and will be the default if not specified. This should make it backwards compatible and optional. When `exec-env` is not set to `all`, the table settings will only be applied to that execution environment.
```
joshwlewis, line 73: What would consume this new key? If a platform or user ran `pack build -e CNB_EXEC_ENV=test my-app:test`, a buildpack might contribute new launch entries flagged for test, but how would a platform or user determine which of the launch entries were `all` or `test`? My guess is at the very least, we'd want this key to show up in the `io.buildpacks.build.metadata` `process-types` data?

text/0000-execution-environments.md
joshwlewis: I like "context", but that's just as overloaded. How about "purpose = test" or "intent = test"?

text/0000-execution-environments.md
hone: Is a label the best place to designate the execution environment?

cz4rny:
Two questions:

1. How would one extract artifacts from the test env? Running unit tests, or any other static-analysis tools produces outputs separate from the end OCI image. Currently `pack` performs the build in a tmp directory, which is cleaned out at the end. How would one access the results of all the checks done in the dev env?

2. How would I ensure, that the code produced in *another* pack built, to the `production` environment is tested? Having the `CNB_EXEC_ENV` set could introduce a completely different set of layers and hence a different output. So I would test the source code and the binary produced from the `test` environment, but that doesn't mean I've tested the exact same binary, source code, and artifacts produced in the `production` env.

jabrown85:
> Two questions:
> 
> 1. How would one extract artifacts from the test env? Running unit tests, or any other static-analysis tools produces outputs separate from the end OCI image. Currently `pack` performs the build in a tmp directory, which is cleaned out at the end. How would one access the results of all the checks done in the dev env?

This RFC aims to only _produce_ a test image artifact. A test platform would then execute the image and collect the results (TAP/json/etc).

> 2. How would I ensure, that the code produced in _another_ pack built, to the `production` environment is tested? Having the `CNB_EXEC_ENV` set could introduce a completely different set of layers and hence a different output. So I would test the source code and the binary produced from the `test` environment, but that doesn't mean I've tested the exact same binary, source code, and artifacts produced in the `production` env.

There would be no such guarantee. Production images vary quite a bit from stack to stack and there is no one size fits all solution for testing. For a practical example, a ruby or node app will often have test environment specific dependencies that should _not_ make their way into the final production images. Another example is a simple go app on a scratch/tiny stack. The go tooling itself, `go test`, should not be packaged into the final production image.

Thinking out loud, a test pipeline could grab the SBOM/digests/shasum of the things it cares about and compare them to the production image that was built.

jabrown85:
The way I understand this is the produced image would have `test` processes contributed by buildpacks. 

A golang buildpack would create a layer that contains go testing tools and maybe even run a `go build ./...` during the build phase so you can fail early. The go buildpack could contribute `test` process to `launch.toml` that runs `go test ./...`. That would be the default process launched when the image is executed in a testing pipeline. As you said, the pipeline could also choose to run static analysis as well as any other process on the image (maybe `lint` in this example) and capture the results.

keskad:
Hi,

I wanted to comment the concept. So, I don't see the point in creating a production images in environment X, while testing in environment Y. That's not consistent way of doing CD pipelines :slightly_smiling_face: 

I give buildpacks to teams, they set e.g. BP_NODE_VERSION 16.1 and it will execute with Node 16.1, while in tests it will still execute with hardcoded Node 16.0 because I have to maintain additional images. Using buildpacks in this scenario is loosing sense for me :slightly_smiling_face:

Due to nature of buildpacks - preparing the environment (by installing required tooling, setting up those tools) - I think the test phase is even a neccesary scenario to implement a valid CD pipeline in order to gain reproducible environment everytime. The tests needs to be running on something that is CLOSE TO PRODUCTION :slightly_smiling_face: 

So I see that `build phase` even as neccessary to implement a valid CD pipeline. 

It is similar to the principle, where we should not build artifact twice if promoting to next environment e.g. dev -> prod, but just reuse the artifact. There we should use the same build parameters, same tooling. 

Its difficult to maitain e.g. Node 14, Node 15, Java 8, Java 11, and 15 other base images just for tests, while giving teams a possibility to set a Node or Java version in buildpacks.

text/0000-execution-environments.md
```diff
@@ -0,0 +1,225 @@
...
```
jabrown85, line 73: We should consider wanting multiple values. This might make `[]` mean `all`, which is kind of odd.

```
exec-env: ["staging", "production"] # but not uat
```

text/0000-execution-environments.md
hone: I think the only downside is what if a user wants to override in a way it wasn't intended? Is there a reason to lock stuff out?

text/0000-execution-environments.md
hone: I ended up going with the `*` any stack magic value.

text/0000-execution-environments.md
hone: yep, good call out.

text/0000-execution-environments.md
```diff
@@ -0,0 +1,245 @@
...
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- "env" is overloaded as a word since we also use it for environment variables. Is there a better word here?
+- Should there be builders that are specific to an execution environment? What about `include` or `exclude`?
+- Should the execution environments be an enum or flexible as a string?
+  - enums will help encourage standardization across buildpacks and platforms.
+  - strings can help account for use cases we haven't thought of yet.
+- Should buildpacks be allowed specify allowlist execution environments?
+- What changes are needed in the buildpack registry?
```
dmikusa, line 216: Users will probably want a way to tell if a buildpack supports execution environments and if so, which ones. Maybe searching by that too. i.e. filter by buildpacks that support execution environments.

text/0000-execution-environments.md
hone: For sharing cache, is it a problem if I run a test, get dev dependencies but my production build doesn't want them and prunes that cache. Next time I run a test I would need to install them. Do you just want to share cache for bootstrapping for every build?

dmikusa: I don't feel like that will be an issue for things like the JVM or Node.js runtime, but it sounds like it would be for other layers, like node modules or other things installed via package manager that would change between environments. That would certainly be annoying and not expected as a user.

Caching is per layer, what if the buildpack were to create different layers for cases like this? Maybe incorporate the exec env key into the layer name or something to make it unique? It wouldn't need to do this everywhere, but for things that are env dependent, it could have `npm_modules-test` and `npm_modules-production` layers.

If you ran a build for test and then a build for production, would that result in keeping both of those cached? or would it evict the test layer when you run the production build because the production image didn't use that layer?

text/0000-execution-environments.md
```diff
@@ -0,0 +1,245 @@
...
+This narrows the scope of the Buildpack responsibilities to what it's already good at:
+
+- Produce Test Environment
+- How to launch the process
+  - launch.toml with test process marked as default (recommendation)
+
+## Setting the Execution Enviroment for Build + Buildpack
+A platform will set the `CNB_EXEC_ENV` env var to the execution environment desired. Buildpacks can than read this env var to branch or switch on logic needed based on the execution environment.
+
+In addition, Builder Authors, Buildpack Authors, and App Developers will be able to specify various options to a specific execution enviroment using the `exec-env` key.
```
jkutner, line 63:
```suggestion
In addition, Builder Authors, Buildpack Authors, and App Developers will be able to specify various options to a specific execution enviroment using the `exec-env` key in a `project.toml` file.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,245 @@
...
+
+The spec will reserve the following values to help standardize execution environments:
+
+* production
+* test
+* development
+
+### Buildpack API
+
+A buildpack author will be able to determine the execution environment their buildpack is expected to build for by reading the `CNB_EXEC_ENV` environment variable. If this value is not set, a Buildpack Author can assume it's set to `production`. This will be provided for both `bin/detect` and `bin/build`.
```
jkutner, line 155: Instead of defaulting to `production`, should we have a concept of `all`? Maybe this is similar to the `exec-env` having `*`?

text/0000-execution-environments.md
jkutner: I think "env" was overloaded before we got here, and it's ok to use.

text/0000-execution-environments.md
hone: There are now a few other files like `builder.toml` and `launch.toml`.

text/0000-execution-environments.md
hone: @jkutner do we need a concept of "all" for an individual lifecycle execution?

text/0000-execution-environments.md
hone: @jkutner should we make changes to the registry index to account for this somehow? Add it into `buildpack.toml`?

text/0000-execution-environments.md
hone: Talking with @jabrown85 , I think compiled languages like Go might want to remove the source tree from `/workspace` for production but not test which would extra complicate things.

hone:
Moving this to voting status with a close of next Thursday the 30th of Jan.

jkutner (approved):

jabrown85:
@hone the voting window is closed. There were no 👎  - feel free to approve and merge

jjbustamante (approved):

jabrown85:
@hone can you work on merging this? I think the scripts are broken..

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+Add support for different execution environments for buildpacks, with testing as the first use case.
```
runesoerensen:
```suggestion
Add buildpack support for multiple execution environments, with better support for testing as the first use case.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+- Execution Environment - The target environment the OCI image is expected to be run in, i.e. production, test, development.
```
runesoerensen:
```suggestion
- Execution Environment - The target environment the OCI image is expected to be executed in, i.e. production, test, development.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The main way Buildpacks are being used is building production images, but this is only one piece of the software development process. Without a solid buildpack test environment story, users will be required use alternatives for building their environment. A testing environment, while different, shares many of the same broad strokes ultimately producing an execution environment. This is something Buildpacks are well positioned to solve.
```
runesoerensen:
```suggestion
Buildpacks are mainly being used to build images for production environments, but this is only one piece of the software development process. Without test environment support for buildpacks, users currently have to implement alternatives for testing their source code. Buildpacks are well positioned to help bridge that gap, as building production and test environments often share many of the same broad strokes.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+- What to do with the results
+
+This narrows the scope of the Buildpack responsibilities to what it's already good at:
+
+- Produce Test Environment
+- How to launch the process
+  - launch.toml with test process marked as default (recommendation)
+
+## Setting the Execution Enviroment for Build + Buildpack
+A platform will set the `CNB_EXEC_ENV` env var to the execution environment desired. Buildpacks can than read this env var to branch or switch on logic needed based on the execution environment.
```
runesoerensen:
```suggestion
A platform will set the `CNB_EXEC_ENV` env var to the desired execution environment. Buildpacks can then read this env var to branch or switch on logic needed for the execution environment.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+In addition, Builder Authors, Buildpack Authors, and App Developers will be able to specify various options to a specific execution enviroment using the `exec-env` key.
```
runesoerensen:
```suggestion
In addition, Builder Authors, Buildpack Authors, and App Developers will be able to configure various options for specific execution environments using the `exec-env` key.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Development Environments
+The specifics of creating development enviroments are out of scope of this RFC, but it's not hard to extrapolate how these kind of changes can assist in creating Buildpacks for development environments.
```
runesoerensen:
```suggestion
The specifics of creating development environments are out of scope of this RFC, but it's not hard to extrapolate how the proposed changes can assist in creating Buildpacks for development environments.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## `exec-env` key in TOML
+
+In order to support additional execution environments an `exec-env` key will be added to various TOML tables in the project. This will be an array that takes string values. An individual element can be any string with `*` having special meaning. Similar to the ["any stack RFC"](https://github.com/buildpacks/rfcs/blob/main/text/0056-any-stack-buildpacks.md) `*` will apply to all execution environments. `["*"]` will be the default if not specified. This should make the key backwards compatible and optional. When `exec-env` is not empty and does not include `*`, the table settings will only be applied to the specificed execution environments.
```
runesoerensen:
```suggestion
In order to support additional execution environments, an optional `exec-env` key will be added to various TOML tables in the project. This will be an array that takes string values. An individual element can be any string with `*` having special meaning. Similar to the ["any stack RFC"](https://github.com/buildpacks/rfcs/blob/main/text/0056-any-stack-buildpacks.md), `*` will apply to all execution environments. `["*"]` will be the default if not specified. This should make the key backwards compatible. When `exec-env` is not empty and does not include `*`, the table settings will only be applied to the specified execution environments.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Project Descriptor - `project.toml` (App Developers)
+
+An app developer may have execution environment configuration like only using a metrics agent in production or headless user agent needs in test. In order to facilitate these needs, the project descriptor will be extended by adding `exec-env` to the following tables:
```
runesoerensen:
```suggestion
An app developer may have execution environment specific configuration, like only using a metrics agent in production, or a headless user agent in test. In order to facilitate these needs, the project descriptor will be extended by adding `exec-env` to the following tables:
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+[[io.buildpacks.build.env]]
+name = "PARALLEL_WORKERS"
+value = "4"
+exec-env = ["production"]
+```
+
+### `builder.toml` (Builder Authors)
+
+The the only table `exec-env` will be added to is `[[order.group]]` and `[[build.env]]`.
```
runesoerensen:
```suggestion
The `exec-env` key will be added to the `[[order.group]]` and `[[build.env]]` arrays of tables.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+* test
+* development
+
+In addition, the `/` character is reserved in case we need to introduce namespacing in the future.
+
+### Buildpack API
+
+A buildpack author will be able to determine the execution environment their buildpack is expected to build for by reading the `CNB_EXEC_ENV` environment variable. If this value is not set, a Buildpack Author can assume it's set to `production`. This will be provided for both `bin/detect` and `bin/build`.
+
+This would let a buildpack author do different things based on the execution environment. For example, it's common for a production build to not include test dependencies or files, while a test environment would include those things. For compiled languages, a production might even remove the source code and just leave the compiled binary with optimizations. In a test environment, it may include debug symbols, not run with optimizations, not prune the source tree.
```
runesoerensen:
```suggestion
This would let a buildpack author do different things based on the execution environment. For example, it's common for a production build to exclude test dependencies or files that a test environment would include. For compiled languages, a production build might even remove the source code and just leave the compiled binary with optimizations. In a test environment, a build may include debug symbols, be compiled without optimizations, and leave the source code untouched.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+## [Heroku Testpack API](https://devcenter.heroku.com/articles/testpack-api)
+
+As part of the classic buildpack design, Heroku extended the API to include support for tests. It added a `bin/test-compile` phase, which is like the normal `bin/compile` phase but specific for setting things up for test. `bin/test` was called for executing tests. This design let a buildpack have a specific codepath for tests and also do intelligent logic for executing tests based on the source code. For example [in Ruby](https://github.com/heroku/heroku-buildpack-ruby/blob/main/bin/support/ruby_test#L62-L76), the buildpack author can run `bin/rspec` if it detects `rspec`, `bin/rails test` when using rails, or default to `rake test`.
```
runesoerensen:
```suggestion
As part of the classic buildpack design, Heroku extended the API to include support for tests. It added a `bin/test-compile` phase, which is similar to the `bin/compile` phase, but specific for setting things up for test. `bin/test` was called for executing tests. This design lets buildpack authors write codepaths and logic optimized specifically for executing tests based on the source code. For example, [this Ruby buildpack](https://github.com/heroku/heroku-buildpack-ruby/blob/fb393cb46e23ab809e21daeef2a97cbd5f04a370/bin/support/ruby_test#L62-L76) will run `bin/rspec` if it detects `rspec`, `bin/rails test` when using rails, or default to `rake test`.
```

text/0000-execution-environments.md
```diff
@@ -0,0 +1,260 @@
...
+
+For composite buildpacks, `exec-env` will be added to is `[[buildpack.order.group]]`.
+
+### `launch.toml` (Buildpack Authors)
+
+Not all process types make sense for every execution environment. In order to help hint to the platform the intention how a process should be used, the `exec-env` key will be added to the `[[processes]]` table.
+
+### `metadata.toml` (Platform Operators)
+
+On the platform side, the `exec-env` key will be added to `metadata.toml` in the `[[processes]]` table to mirror `launch.toml`. This will help platforms to make decisions on which processes are desired for each execution environment. `lifecycle` will list all processes and will not exclude any based on the execution environment Instead, a platform can use the available information to make a decision.
```
runesoerensen:
```suggestion
On the platform side, the `exec-env` key will be added to `metadata.toml` in the `[[processes]]` table to mirror `launch.toml`. This will help platforms decide which processes are desired for each execution environment. `lifecycle` will list all processes and will not exclude any based on the execution environment Instead, a platform can use the available information to make a decision.
```

runesoerensen (approved):
Happy to see this was approved and that the [POC implementation](https://github.com/buildpacks/pack/pull/2324) is already well under way!

Just adding a few (somewhat pedantic) wording/spelling/phrasing suggestions for you to consider before merging :)

hone:
Merged as part of [6ce039b](https://github.com/buildpacks/rfcs/commit/6ce039b680dc7b4d2ab5f2fa50ef6523e9b3f7f1)

PR #278, opened 2023-02-24
SBOM layer support

RealHarshThakur:


samj1912:
@RealHarshThakur thanks for the RFC. Have you had a chance to look at https://github.com/buildpacks/rfcs/pull/195 which covers the cosign sbom format as well as storing sboms as attestations.

RealHarshThakur:
@samj1912  thanks for pointing it out. Just skimmed through it.  I think we're after the same goal. OCI v1.1 draft spec is helping with interoperability. 

Re: attestations
Feels like it's easier to support SBOMs as dedicated layers in OCI manifest first. This will set the ground for attestations too but I didn't want to include it here as I haven't seen the popular builder images emitting attestations and there isn't a `pack attest` subcommand yet. Attestation support might deserve its own RFC IMO. 


Re: using cosign as a library
I noticed cosign also relies on go-containerregistry library that Buildpack does. It felt easier to use the underlying library directly than to introduce a new dependency.

Re: similarity to RFC
Is there a notion of sub-RFCs or similar since it feels like this RFC can be a sub-RFC to the one you've proposed? If not, I'm happy to close this one if we can cover the OCI spec progress in the other one

text/0119-sbom-oci-format.md
```diff
@@ -0,0 +1,106 @@
...
+* Layers directory (downloaded through  `pack sbom download`)
+
+
+From implementation perspective, this will have to be split up into [imgutil](https://github.com/buildpacks/imgutil/blob/main/image.go#L36) package which lifecycle uses in Export phase. 
+
+
+# Migration
+[migration]: #migration
+
+N/a
```
AidanDelaney, line 80: How would `pack sbom download mytag` work if we adopt this RFC?

loewenstein:
Should we consider https://github.com/buildpacks/rfcs/pull/186 a dependency? Feels like a prerequisite to have complete sbom information before we condense it into an "image sbom"...

text/0119-sbom-oci-format.md
RealHarshThakur: OCI artifact will have the SBOM embedded during build time. CLI can query the container registry for it. 
To remain backward-compaitable, I guess CLI will have to fallback to extracting the layers if it doesn't find any.

RealHarshThakur:
@loewenstein Agreed! OCI spec allows annotations to attach metadata like that.

loewenstein:
@RealHarshThakur well, I primarily meant this to say, we - as the CNB project - should make sure to allow complete application image boms before exposing them via standards. Or was your comment to say there are annotations for "warning: this sbom is incomplete"?

RealHarshThakur:
I meant we should provide complete image sboms too but adhere to OCI 1.1 spec on how it suggests on doing it. In future, when we do runtime SBOMs, etc- annotations will help us differentiate it.

PR #279, opened 2023-02-24
draft for LTS releases

joe-kimmel-vmw:


text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+It is risk mitigation for consumer and enterprise customers who want to apply critical patches without re-testing their entire platform for regressions which could occur during a full upgrade cycle.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Lifecycle will maintain a publicly viewable table of LTS versions with anticipated sunset dates for support.
```
natalieparellano, line 43: We currently have this table: https://github.com/buildpacks/lifecycle#supported-apis

Which should probably be updated to not even mention lifecycle versions that are more than a year old. It could be modified to (**) the ones that are LTS versions

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+[what-it-is]: #what-it-is
+
+It is risk mitigation for consumer and enterprise customers who want to apply critical patches without re-testing their entire platform for regressions which could occur during a full upgrade cycle.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Lifecycle will maintain a publicly viewable table of LTS versions with anticipated sunset dates for support. 
+Not all Lifecycle releases will need to be LTS, but 3-5 each year probably will be marked as LTS. Maintainers will be responsible for marking releases as LTS.
+Some additional release automation will be built to support patching and releasing LTS releases.
```
natalieparellano, line 45: Potentially:
* Automation around release notes
* Some system where tagging a PR as applicable to an LTS release creates new PRs against the LTS tags

It's probably not necessary to put this in the RFC, I'm just thinking out loud here

natalieparellano (approved):

text/0000-lts-releases.md
joe-kimmel-vmw: the argument(s) I can think of for keeping a long historical tail:
- shows the project has been around for a while
- gives you something to point at if/when somebody _is_ using an ancient version ("that version has been out of support for 18 months!" is more convincing with a link)
- helps someone see patterns in support/sunset windows

though i agree generally people are only going to care about the top couple rows of any such table

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Like all efforts this does obviously come with an opportunity cost of alternative lines of work. Additionally, it's especially clear that the cost of this commitment will be not just one-time, but ongoing, until the community sunsets this LTS initiative itself.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Just upgrade to the latest: This is our current approach, and works pretty well until some large enterprise partners don't want to do it.
```
natalieparellano, line 62:
```suggestion
  - Just upgrade to the latest: This is our current approach, and works pretty well until some large enterprise partners don't want to do it.
```

(Thinking about this more) It might be worth explicitly calling out how this works - the lifecycle accepts a `CNB_PLATFORM_API` environment variable which controls its behavior with respect to the Platform API, allowing older platforms to keep older behavior even when using the latest binary. This RFC doesn't contemplate changing the way this works at all - older platforms can keep updating to the latest lifecycle if they trust the implementation team :)

natalieparellano:
I'd be happy to steward this one

text/0000-lts-releases.md
joe-kimmel-vmw: good point -- i'll modify

natalieparellano:
@jabrown85 any thoughts on this one?

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Lifecycle to define approximately 4 releases per year as LTS.
```
samj1912, line 15: This seems fairly high. I don't think we actually do enough minor releases an year that picking 4 would help. This would roughly result in every lifecycle minor being an LTS. I would be more comfortable if we reduce this cadence.

I also am not sure how much benefit this provides us given lifecycle is supposed to provide strong backwards compatibility guarantees.

samj1912 (changes requested):
I am not convinced of this RFC yet. I would like the lifecycle maintainers to pitch in here and consider the maintanance costs and also appropriate recommendations for the end user. Given lifecycle's strong backwards compatibility guarantees, I'm not sure how much it's worth porting patch releases to older lifecycle versions unless we decide to introduce deprecations in a newer version. Even there I don't believe we have ever deprecated any buildpack or platform api versions that are less than an year old.

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+
+**CVE**: While some scholars claim this acronym has origins in the Mayan codices[1], others seem to support a Norse origin[2]. Regardless of its hagiography, CVEs come in varying severities such as Critical, High, and "you probably don't need to patch it but we still wanted to let you know."
+These severity levels are issued by the shadowy council of security elders and cannot be modified by mere practitioners such as ourselves.
+However there are multiple councils[3] and they will occasionally grade the same vulnerability differently.
+This proposal only relates to CVEs graded Critical by at least one council.
+
+
+# Motivation
+[motivation]: #motivation
+
```
samj1912, line 30: I would like this section to be more fleshed out and list scenarios where we would recommend patches of older lifecycle versions rather than the latest.

text/0000-lts-releases.md
samj1912: If the argument were instead made for one lifecycle release every 1 year (or possibly every 2 years) being marked as LTS, this would make more sense to me.

jabrown85: I agree with Sam. I think 1-2 LTS releases a year seem like plenty.

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+- Why should we do this? Some organizations that bundle lifecycle into other products are sufficiently risk-averse to be opposed to upgrading to the latest version, but still desire critical CVE patches. This LTS patch-backporting strategy allows us to support these slower moving vendors with minimal effort.
+- What use cases does it support? See above.
+- What is the expected outcome? Lifecycle will have to maintain a trailing list of supported LTS versions with approximate sunset dates for each. Then for certain CVEs there will be some additional "release engineering" effort to ease backporting and then releasing patch versions.
```
jabrown85, line 33: Does having LTS give lifecycle maintainers any advantages?

text/0000-lts-releases.md
```diff
@@ -0,0 +1,109 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Like all efforts this does obviously come with an opportunity cost of alternative lines of work. Additionally, it's especially clear that the cost of this commitment will be not just one-time, but ongoing, until the community sunsets this LTS initiative itself.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - Just upgrade to the latest: Our current approach encourages all users to upgrade to the latest binary but allows them to specify the version of the API to use. This works pretty well until some large enterprise partners don't want to do it because they're so risk averse that they want the reassurance that the changeset is the minimal one necessary to addrress the CVEs.
```
jabrown85, line 62: With the introduction of LTS - do we see a future where lifecycle _stops_ taking in Platform API all together?

text/0000-lts-releases.md
joe-kimmel-vmw: Thanks - that makes sense

text/0000-lts-releases.md
joe-kimmel-vmw: Not that I'm aware of

joe-kimmel-vmw:
Thanks all for the feedback. I think we are closing this in favor of a proposal with substantially less commitment, and hopefully also a better motivation / addressing some of the comments above: https://github.com/buildpacks/rfcs/pull/281

PR #280, opened 2023-03-06
Fix 0096-remove-stacks-mixins : architecture names

joe-kimmel-vmw:
previously this RFC used x86_64 which was a misspelling of amd64.


hone (approved):

AidanDelaney (approved):
LGTM

PR #282, opened 2023-04-03
decorrelate source and target directories

xfrancois:
Proposition to decorrelate source and target directories
The idea is to introduce a new variable named `CNB_SOURCE_APP_DIR`, that could be specified to tell where the source code is. But the final image directory structure will still be created following the `CNB_APP_DIR` variable.
That way, source and target directories structure could be decorrelate. 
[This will fix this issue](https://github.com/buildpacks/lifecycle/issues/1034).

### DIFF
--- a/text/0000-decorrelate-source-targer-dirs.md
+++ b/text/0000-decorrelate-source-targer-dirs.md
@@ -1,0 +1,118 @@
+# Meta
+[meta]: #meta
+- Name: Decorrelate source dans target directories
+- Start Date: 2023-04-03
+- Author(s): @xfrancois
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: [Can't make final image directory different from the CNB_APP_DIR](https://github.com/buildpacks/lifecycle/issues/1034)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Allow CNB users to build image from one directory and have the final image directory structure to be something else. Currently it's possible to change the default target directory (/workspace) by setting the `CNB_APP_DIR` environement variable, but the source directory must also be the one specified in `CNB_APP_DIR`. This behavior is not well suited for constrained environements, like CI.
+
+# Definitions
+[definitions]: #definitions
+
+N/A
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+
+We should do this to provide more flexibility for end-users.
+
+- What use cases does it support?
+
+Building from a CI tool, for example Gitlab-CI. We can't chose the source directory from which we are building. On Gitlab-CI the Git folder is checkout on `/builds/<group>>/<project>`. By having a way to decorrelate source and target directories, we could chose the finaly directory structure we want.
+
+- What is the expected outcome?
+
+Support of a new environment variable / parameter that allows to set the source directory. We probably should document that  `CNB_APP_DIR` only handles the target directory.
+
+| Variable           | Default value | Description                |
+|--------------------|---------------|----------------------------|
+| `CNB_APP_DIR`       | `/workspace`    | Target directory structure |
+| `CNB_SOURCE_APP_DIR` | `$CNB_APP_DIR`  | Source directory structure |
+
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+
+# How it Works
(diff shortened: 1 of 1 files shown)


natalieparellano (commented):
This seems reasonable to me. I'd be happy to steward this one. If implemented I could see this being a new input to the exporter.

text/0000-decorrelate-source-targer-dirs.md
```diff
@@ -0,0 +1,118 @@
...
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+In the Gitlab-CI example, I can't chose the source directory but I want to be able to chose the target directory. I will set the variables like this :
+
+`CNB_SOURCE_APP_DIR`: /builds/group/project
```
joe-kimmel-vmw, line 58: given the way the [appPath is currently used in pack](https://github.com/buildpacks/pack/blob/main/internal/commands/build.go#L357) it may be more backwards compatible to keep the "APP_DIR" as the variable which means "both source and destination" 
OR as the variable which means "source only" rather than as the variable that means "destination only". 

Looking in lifecycle it also looks like it's pretty thoroughly "the source and destination" directory, so i wonder if it would be better to add 2 entirely new env vars and then do some validation that they're used exclusively with the existing env vars and flags?

text/0000-decorrelate-source-targer-dirs.md
```diff
@@ -0,0 +1,118 @@
...
+The final image will be available
+
+# Migration
+[migration]: #migration
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
```
joe-kimmel-vmw, line 71: it's a little disingenuous to claim there's no drawbacks --  at the very least, any new variable is something that comes with a small maintenance and complexity cost.

text/0000-decorrelate-source-targer-dirs.md
```diff
@@ -0,0 +1,118 @@
...
+N/A
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What is the impact of not doing this?
+
+The impact is that people wanting to use CNB on CI environment will have to make some workaround (like copying source folders in another location) in order to have a proper directory structure in the final image.
```
joe-kimmel-vmw, line 79: what is the impact of having an "improper" directory structure?  What gets harder or worse other than an abstract sense of "proper-ness" ?

joe-kimmel-vmw (changes requested):
Thanks @xfrancois for contributing this!  

I've given you some critiques in this review but I want to be clear that I'm not opposed to moving forward with this idea,  just that I think it's better for all of us to explain the justification for doing so more clearly,  and to consider whether it's possible to re-use the existing control points (env vars and flags for respective commands) at all or whether we'd be better off introducing two new ones that are mutually exclusive with the existing one.

text/0000-decorrelate-source-targer-dirs.md
xfrancois: It's a bit complicated to explain but let me try : 
On Gitlab-CI I'm building an application A that uses an app B as a "service" (equivalent to docker-compose in gitlab-ci).
This app B in my context has also been built with buildpack on Gitlab-CI. As it's been build on Gitlab-CI, by default the working directory is /workspaces, so we need to set up the CNB_APP_DIR to /worspaces. Then the final image of the application B would have a directory of /workspaces.

Back to the application A build : when launching the app B as a service, the directory structure of the B app will clash with Gitlab-CI default, and then the app B won't be able to run.

To sum up, the inability to chose the "proper" directory structure prevents us from running my app as a service dependency in Gitlab-CI context.

The only workaround that works is when building the "B" app, copy the content from /worskpaces (default Gitlab CI folder) to another location, set the CNB_APP_DIR to this other location. That way the final directory structure will not be "/workspaces" and it won't clash when loading this app as a service when building app A.
Hope the need is clearer

joe-kimmel-vmw: Thanks @xfrancois  - i think it's coming into focus but it's worth really understanding. 

This situation arises when:

- you are invoking the lifecycle directly,  so you're managing your own "unit of hermetic building" via CI.  (If you were using pack,  the 'hermetic unit' would be the container created by pack, which is fresh for each invocation)

- you want to build 2 (coupled) applications in the same "hermetic unit";  i.e. the build and run for apps A and B share the same filesystem. 

- you want two different images / or "runnable units"

- you are constrained by your "provider of hermetic units" (your CI) to have a hardcoded directory name


- and the problem is then that you wind up with two applications in a single directory?  

- and the current workaround is to manually copy a directory? 


To be honest i'm still a little confused.   Maybe you can draw a more detailed sketch.  I'm not finding the word "proper" helpful as I think you're referring to a convention that I don't know,  so it might help to spell out that convention? 

Are you trying to wind up with one image that can run either app A or B?   

Why wouldn't it work to do something like 
```
cp -r path/to/appA/src /workspaceA
lifecycle ... -app /workspaceA
cp -r path/to/appB/src /workspaceB
lifecycle ... -app /workspaceB
```
?

PR #283, opened 2023-04-19
Proposal for adding commands to handle manifest list in pack

jjbustamante:
[Readable](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/pack-manifest-list-commands/text/0000-pack-manifest-list-commands.md)

### DIFF
--- a/text/0000-pack-manifest-list-commands.md
+++ b/text/0000-pack-manifest-list-commands.md
@@ -167,10 +167,19 @@ The proposal is to implement an abstraction of an OCI *Image Index* and expose i
 
 ## Image Index Abstraction
 
+A new high level abstraction to represent an OCI Image Index is proposed, similar to the [Image](https://github.com/buildpacks/imgutil/blob/main/image.go) interface exposed in *imgutil* repository, 
+we proposed a new *ManifestList* interface to expose the behavior of an OCI Image Index.
+
 ```mermaid
 classDiagram
-    ManifestList <|-- remote_ManifestList
-    ManifestList <|-- local_ManifestList 
+    class ManifestList {
+       <<interface>>
+        +Add(repoName string) error
+        +Remove(repoName string) error
+        +Delete(additionalNames []string) error
+        +AnnotateManifest(manifestName string, opts AnnotateFields) error
+        +Save() error
+    }
 
     class remote_ManifestList {
         +NewManifestList(repoName string, keychain authn.Keychain) ManifestList 
@@ -179,16 +188,18 @@ classDiagram
     class local_ManifestList {
          +NewManifestList(repoName string, path string) ManifestList
     }
-
-    class ManifestList {
-       <<interface>>
-        +Add(repoName string) error
-        +Remove(repoName string) error
-        +Save() error
+    
+    class AnnotateFields {
+         +String Architecture
+         +String OS
+         +String Variant
     }
+    
+    ManifestList <|-- remote_ManifestList
+    ManifestList <|-- local_ManifestList 
 ```
 
-Two implementations: *remote* and *local* are propose, *remote* will take care of implementing the interaction with an OCI registry and *local* will deal with the local storage operations.
+Two implementations: *remote* and *local* are proposed, *remote* will take care of implementing the interaction with an OCI registry and *local* will deal with the local storage operations.
 
 ### Component Diagram
 
@@ -201,6 +212,15 @@ Using a [C4 component diagram](https://c4model.com/#ComponentDiagram), we can de
   - As we can see, depending on the implementation it will interact with the file system or with a remote registry
 
 ### Considerations
+
+#### When a user wants to create a manifest list using a manifest outside the user's repo. 
+
+Let's suppose the following user case: A user wants to create a manifest list `foo/my-manifest:my-tag` using a manifest outside his repository `foo` for example `other/external-manifest:latest`. 
+
+In this case, pack will need to *copy* the external manifest `other/external-manifest:latest` into `foo` repository `foo/external-manifest:latest` and then uses this reference into the manifest list created. Pack should at least *warn* the user about this operation.
+
+#### When a user wants to create a manifest list referencing a manifest list
+
 
 # Migration
 [migration]: #migration
--- a/text/0000-pack-manifest-list-commands.md
+++ b/text/0000-pack-manifest-list-commands.md
@@ -1,0 +1,148 @@
+# Meta
+[meta]: #meta
+- Name: Manifest List Commands for Pack
+- Start Date: 2023-04-19
+- Author(s): [Juan Bustamante](Juan Bustamante)
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The problem for adding support for multi-arch buildpacks can be divided into two parts:
+- Support buildpack authors to **migrate their existing buildpacks** to support multi-arch .
+- Support buildpack authors to **create new buildpacks and builders** that handle multi-arch from the beginning.
+
+This RFC proposes to create a new set of CRUD commands in pack to handle manifest lists, which will be used to support the first part of the problem.
+
+# Definitions
+[definitions]: #definitions
+
+- Image Manifest: The image manifest provides a configuration and set of layers for a single container image for a specific architecture and operating system. See [spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
+- Image Index: The image index is a higher-level manifest which points to specific image manifests, ideal for one or more platforms. See [spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md) 
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+
+The uses of ARM architecture in the cloud and edge computing has been growing rapidly. The CNCF community has been also growing in the last years, and there is a need to support multi-arch for all the projects. The buildpacks community is not an exception, issues like:
+- [It would be nice to support easily creating a manifest list packages and builders](https://github.com/buildpacks/pack/issues/1460)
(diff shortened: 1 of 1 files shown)


text/0000-pack-manifest-list-commands.md
```diff
@@ -0,0 +1,148 @@
...
+
+The expected outcome is to have a set of commands in pack to handle manifest lists, for example:
+- A command to create a manifest list from a set of images
+- A command to push the manifest list to a registry
+- A command to update or delete a manifest list
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal is to add a new command `pack manifest` and different subcommands.
```
AidanDelaney, line 56:
```suggestion
The proposal is to add a new _experimental_ command `pack manifest` and different subcommands.   The `pack manifest` commands will initially be gated behind `pack config experimental`.  The `pack manifest` commands will move from experimental status to supported status when maintainers deem it appropriate.
```

text/0000-pack-manifest-list-commands.md
```diff
@@ -0,0 +1,234 @@
...
+We also have in mind to improve existing commands like `pack builder create` and `pack buildpack package` to support the creation of a manifest list without the need of a new command. However, this approach could not be suitable for buildpack authors who are maintaining existing buildpacks and they will need to change their current process to generate multi-arch images.
+
+- Why is this proposal the best?
+
+Because we will provide the tool to our end users to solve their problems without the need of using other tools.
+
+- What is the impact of not doing this?
+
+The impact of not doing this is that users will need to use other tools to create and handle their manifest list, which is a poor user experience.
+
```
dmikusa, line 194: My $0.02 only, but I would really rather have `pack builder create` and `pack buildpack package` to support multi-arch natively. If this is a required step to get there, then ignore me. 

I have little desire to use these commands directly and would be totally fine using `docker` to manage manifest lists if I absolutely had to in the interim.

text/0000-pack-manifest-list-commands.md
```diff
@@ -0,0 +1,234 @@
...
+[migration]: #migration
+
+This section should document breaks to public API and breaks in compatibility due to this RFC's proposed changes. In addition, it should document the proposed steps that one would need to take to work through these changes. Care should be give to include all applicable personas, such as platform developers, buildpack developers, buildpack users and consumers of buildpack images.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+We should decide to do not add this feature and users could use tools like `docker` or `podman` to create and handle their manifest list, however this is a poor user experience forcing users to use different tools to fulfill scenarios that are part of the business domain of pack.
```
dmikusa, line 178: I would agree with this if it required additional tools to be installed. If you're using `pack`, you have `docker` or `podman` already though, so I'm not sure it's really that big of a deal. Just my $0.02.

dmikusa:
One thing that has been a little weird working with multi-arch support and image indexes is that the different images do not have to be the same for each arch. My amd64 image can be quite different from the arm64 image. I'm not entirely sure how I feel about that.

In terms of buildpacks, I feel like it seems logical that the two could be different. Maybe a particular buildpack doesn't function on arm64 (maybe there are just no binaries that it needs)? So you don't include that one. 

The problem is that I think users will expect them to be the same, and I don't think there is presently a good way for users to interrogate and see what's in each architecture. If you `pack buildpack inspect` an image, it's not clear what it shows exactly (I'm guessing my native architecture but maybe just the first one in the image?). Anyway, I think we need to make it clear what's being show in the inspect related commands. I also think we'll probably need an arch flag to specify a non-native arch if you want to see what's included in a different arch.

text/0000-pack-manifest-list-commands.md
jjbustamante: Hi @dmikusa. We are working on these commands as part of a mentorship program for LFX and initially, we wanted to work on `pack builder create` and `pack buildpack package` but we are also working in parallel removing the support for `stacks`, redefining the toml files and there was some interception between initiatives that make us work on this direction first, without having to wait for `stack removal` but still adding value for end users

jkutner:
Is this ready for review?

text/0000-pack-manifest-list-commands.md
```diff
@@ -0,0 +1,301 @@
...
+- A command to update or delete a manifest list
+
+# What it is
+[what-it-is]: #what-it-is
+
+The proposal is to add a new _experimental_ command `pack manifest` and different subcommands.   The `pack manifest` commands will initially be gated behind `pack config experimental`.  The `pack manifest` commands will move from experimental status to supported status when maintainers deem it appropriate.
+- `pack manifest create` will create a local manifest list for annotating and pushing to a registry
+- `pack manifest annotate` will add additional information like os, arch or variant to an existing local manifest list
+- `pack manifest add` will add an image to an existing manifest list
+- `pack manifest delete` will delete a manifest list from local storage
```
drac98, line 60: We do not have a `delete` command. The `remove` command implements the functionality you have mentioned here. Podman have this same functionality in the `rm` command and it also have a `remove` command which is used to remove an image manifest from a local manifest list. So if we want to make the user experience same, we can rename our commands.

jjbustamante:
> Is this ready for review?

Hi @jkutner ! 

Yes, I think it is ready for a review, it is not perfect but I put most of the stuff I had in mind for these new set of commands. I will be taking some days off (3 weeks) but Jerico or Husni will follow up on this, the idea is to get all the feedback as possible. 

cc @drac98 @jericop @AidanDelaney

jjbustamante:
@chenbh Thank you very much for this awesome feedback!!! 

Definitly it is a very interesting user experience the one you are proposing! I think it worth to be discuss. 

@husni-faiz , @jericop , @AidanDelaney any thoughts about it? I think we will end up with the same outcome for users but with a very different flow.

jjbustamante:
A quick note from our conversation during the WG today

- We like the idea of the [abstraction](https://github.com/buildpacks/rfcs/pull/283#issuecomment-1684073529) suggested by Daniel. It is something we see possible to develop in the future.

hone:
@jjbustamante once we have said future abstractions, do you see us removing this lower level primitive commands since many other tool support manipulating the image index/manifests like @chenbh / @dmikusa has suggested.

jjbustamante:
> @jjbustamante once we have said future abstractions, do you see us removing this lower level primitive commands since many other tool support manipulating the image index/manifests like @chenbh / @dmikusa has suggested.

If we can hide the complexity of the primitives command behind the scene with a better user experience as Daniel suggest. I think in the future we can remove those commands, but probably the code will still be necessary to implement the abstraction. That's something the community can guide us, right? if the abstraction is so good that it fills most the scenarios, maybe the primitives will not be required anymore.

jjbustamante:
@hone, @chenbh 

Thinking a little bit about this, right now pack creates and push OCI artifacts with 3 commands. `pack build`, `pack builder create` and `pack buildpack package`. We know the primitives suggested in this RFC are just a first step in the multi-arch journey.

Based on the scenario proposed by Daniel, what about this idea:

### Propose adding flags like `--platform` and `--append` to our current 3 commands mentioned above. 

For example:

Let's take the Paketo Java Buildpack `gcr.io/paketo-buildpacks/java`, it is distributed without an Image Index and with `os=linux` and `architecture=""`

Let's suppose we already defined on how a multi-arch buildpack must be structured and the buildpacks authors execute:

```bash
pack buildpack package  --append --platform linux/amd64,linux/arm64 --publish gcr.io/paketo-buildpacks/java
```
We should expect:
- Pack will create two OCI artifacts, one for each platform specified.
- Pack will create an Image Index and adds the two OCI artifacts to that index
- Pack should **append** the current manifest and replace it with the Image Index? 

Similar to the buildpack package, we can add similar behavior to the other two commands

```bash
pack build --platform linux/amd64,linux/arm64 --publish
pack builder create --platform linux/amd64,linux/arm64 --publish
```

In the reference doesn't exists in the registry, we can create the Image Index by default or maybe also add a flag? I think an idea in this direction could be next steps after these primitives

hone:
@jjbustamante are you still planning on moving away from `manifest` to another name?

jjbustamante:
> @jjbustamante are you still planning on moving away from `manifest` to another name?

I haven't thought of a different name honestly. I am okay with keeping it. Again, the idea is to keep working on this and open a new RFC to start discussing the idea of: 

```bash
pack build --platform linux/amd64,linux/arm64 --publish
pack builder create --platform linux/amd64,linux/arm64 --publish
```

hone (approved):

hone:
> @hone, @chenbh
> 
> Thinking a little bit about this, right now pack creates and push OCI artifacts with 3 commands. `pack build`, `pack builder create` and `pack buildpack package`. We know the primitives suggested in this RFC are just a first step in the multi-arch journey.
> 
> Based on the scenario proposed by Daniel, what about this idea:
> ### Propose adding flags like `--platform` and `--append` to our current 3 commands mentioned above.
> 
> For example:
> 
> Let's take the Paketo Java Buildpack `gcr.io/paketo-buildpacks/java`, it is distributed without an Image Index and with `os=linux` and `architecture=""`
> 
> Let's suppose we already defined on how a multi-arch buildpack must be structured and the buildpacks authors execute:
> 
> ```shell
> pack buildpack package  --append --platform linux/amd64,linux/arm64 --publish gcr.io/paketo-buildpacks/java
> ```
> 
> We should expect:
> 
>     * Pack will create two OCI artifacts, one for each platform specified.
> 
>     * Pack will create an Image Index and adds the two OCI artifacts to that index
> 
>     * Pack should **append** the current manifest and replace it with the Image Index?
> 
> 
> Similar to the buildpack package, we can add similar behavior to the other two commands
> 
> ```shell
> pack build --platform linux/amd64,linux/arm64 --publish
> pack builder create --platform linux/amd64,linux/arm64 --publish
> ```
> 
> In the reference doesn't exists in the registry, we can create the Image Index by default or maybe also add a flag? I think an idea in this direction could be next steps after these primitives

This looks great and fits more inline with what I would expect most people to use. Definitely a lot of things to work through. Just some of the questions that are coming to my head:

```
pack build --platform linux/amd64,linux/arm64 --publish
```
does this execute two different builds (runs through lifecycle)?

```
pack buildpack package  --append --platform linux/amd64,linux/arm64 --publish gcr.io/paketo-buildpacks/java
```
What are the intermediate images for each platform named/called? What happens if you don't put `--append` here?

PR #285, opened 2023-05-24
RFC: Private Registry Mirrors

jabrown85:
[Readable](https://github.com/jabrown85/rfcs/blob/jab/private-mirrors/text/0000-private-mirrors.md)


loewenstein:
What images are requested inside of the lifecycle?

hone:
> What images are requested inside of the lifecycle?

If the mirrors are available, I believe it would use the image from the mirror while the image will be written with the original source.

loewenstein:
Understood. Thanks. I was wondering if there is anything else but run images that are pulled by the lifecycle, because we just recently contributed https://github.com/buildpacks/spec/pull/357 and https://github.com/buildpacks/lifecycle/pull/1024 to improve the handling of stack provided run image mirrors.

Should we add a similar provision here and state that 
1. Multiple mirrors could be defined
2. The first site that is both reachable and readable is taken
3. The original site is considered last

cc @natalieparellano @pbusko @c0d1ngm0nk3y

text/0000-private-mirrors.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How will will teach kaniko-style extensions to use the mirrors?
```
natalieparellano, line 101: With luck I think the lifecycle could handle all of the mirror translations. We don't pull any images when running kaniko...

text/0000-private-mirrors.md
```diff
@@ -0,0 +1,130 @@
...
+
+The `CNB_REGISTRY_MIRRORS` environment variable will be a list of mirror configurations. Each mirror configuration will be a key/value pair, where the key is the registry URL and the value is the mirror URL. The key/value pairs will be separated by a semicolon (`;`).
+
+For example, if we wanted to configure a mirror for Docker Hub and Quay.io, we could set the `CNB_REGISTRY_MIRRORS` environment variable to the following:
+
+```
+docker.io=https://docker.mirror.example.com;quay.io=https://quay.mirror.example.com
+```
+
+When `lifecycle` requests an image during any phase (`analyze`, `export`, `extend`, etc.), it will first check the `CNB_REGISTRY_MIRRORS` environment variable. If the requested image's registry is configured in the `CNB_REGISTRY_MIRRORS` environment variable, it will use the mirror URL instead of the original registry URL.
```
natalieparellano, line 60: Could we outline how this intersects with run.toml/stack.toml and "run image mirrors"?

natalieparellano, line 60:
```suggestion
When `lifecycle` requests an image during any phase (`analyze`, `restore`, `export`, `rebase`), it will first check the `CNB_REGISTRY_MIRRORS` environment variable. If the requested image's registry is configured in the `CNB_REGISTRY_MIRRORS` environment variable, it will use the mirror URL instead of the original registry URL.
```

text/0000-private-mirrors.md
```diff
@@ -0,0 +1,146 @@
...
+ mirrors = ["quay.io/public/stack:run-image"]
+```
+
+If `CNB_REGISTRY_MIRRORS` has the value of:
+
+```
+docker.io=https://docker.mirror.example.com;quay.io=https://quay.mirror.example.com
+```
+
+When `lifecycle` attempts to resolve the `public/stack:run-image` image, `lifecycle` will attempt to fetch the image from `docker.mirror.example.com/public/stack:run-image`. If the [Run Image Resolution](https://github.com/buildpacks/spec/blob/main/platform.md#run-image-resolution) resulted in `quay.io/public/stack:run-image` being chosen, `lifecycle` will attempt to fetch the image from `quay.mirror.example.com/public/stack:run-image` instead. This new processing happens AFTER the [Run Image Resolution](https://github.com/buildpacks/spec/blob/main/platform.md#run-image-resolution) has executed.  The run image selection will NOT take `CNB_REGISTRY_MIRRORS` into account, but the final image resolution will. Think of `CNB_REGISTRY_MIRRORS` as a just-in-time override of the final image resolution.
```
natalieparellano, line 78: Just thinking out loud here, I'm not sure if this is relevant or not...

There exists today a `pack config run-image-mirrors` that will
* [transparently](https://github.com/buildpacks/pack/blob/3364a6b4b8b8aab03e90722f479a918f9dd3a8a9/pkg/client/build.go#L324) translate the run image reference

There is also a `pack config registry-mirrors` that will 
* [quietly](https://github.com/buildpacks/pack/blob/3364a6b4b8b8aab03e90722f479a918f9dd3a8a9/pkg/client/client.go#L218) translate the registry before fetching ANY image for ANY operation
* [transparently](https://github.com/buildpacks/pack/blob/3364a6b4b8b8aab03e90722f479a918f9dd3a8a9/pkg/client/build.go#L482) translate the registry for the run image reference AFTER `run-image-mirrors` have been applied

It's worth noting that while these translations update the `-run-image` provided to the lifecycle, they do NOT update stack.toml/run.toml i.e., these mirrors are "hidden" today in the sense that they are not persisted as metadata on the app image.

natalieparellano: Now I remember - it would be helpful to describe how `CNB_REGISTRY_MIRRORS` env var for the lifecycle would be exposed in pack given the above stuff that exists today. I would really hope that we could avoid adding yet another toggle here.

natalieparellano: @buildpacks/platform-maintainers @jjbustamante do you have any thoughts here?

jjbustamante: I tried to search in past issues and pull request to understand the reason for `pack config mirrors` commands. I found the following:

The order of preference applied for for run image when `pack build` is executed:

1. --run-image flag
2. default run image mirror, as discussed in the --publish/daemon cases above
    2.1 when `publish`, default to using the run-image mirror in the registry you are publishing the image to  
    2.2 when `daemon`, defaults to using the run-image mirror in the same registry (colocated) as the builder 
3. mirrors set using `pack config run-image-mirrors` 
4. mirrors defined on the builder

Based on this, the new `CNB_REGISTRY_MIRRORS` seems to be like a new option to override point **4**, right? something like:

4. expose some way to set `CNB_REGISTRY_MIRRORS` and pass it through the lifecycle
5. mirrors defined on the builder

text/0000-private-mirrors.md
```diff
@@ -0,0 +1,146 @@
...
+
+Importantly, the presence of a mirror should be invisible outside of the platform. The resulting image should not contain any references to the mirror. This will allow the image to be used in any environment, regardless of whether the mirror is available. For example, the metadata set on the resulting image would reference the original registry URL, not the mirror URL. This will allow the resulting image to be used as if there was no mirror configured. This is important for future actions against the resulting image, such as `pack rebase` or `pack inspect`.
+
+- What use cases does it support?
+If a public registry was having service interruptions, an operator that had a previously configured mirror would be able to continue to build images without interruption.
+
+If a public registry introduced lower rate limits, an operator that had a previously configured mirror would be able to continue to build images without interruption or fear of hitting the rate limit.
+
+- What is the expected outcome?
+Operators concerned with the reliability of their builds that use public images will be able to configure private registry mirrors for the most popular registries.
```
edmorley, line 38: (Fixes the spacing/indentation)

```suggestion
- What use cases does it support?

  If a public registry was having service interruptions, an operator that had a previously configured mirror would be able to continue to build images without interruption.

  If a public registry introduced lower rate limits, an operator that had a previously configured mirror would be able to continue to build images without interruption or fear of hitting the rate limit.

- What is the expected outcome?

  Operators concerned with the reliability of their builds that use public images will be able to configure private registry mirrors for the most popular registries.
```

text/0000-private-mirrors.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How will will teach kaniko-style extensions to use the mirrors?
```
edmorley, line 117:
```suggestion
- How will we teach kaniko-style extensions to use the mirrors?
```

natalieparellano (approved):

natalieparellano:
Whoops, I mean to review this, not approve it (yet)

PR #288, opened 2023-06-16
first draft of pack direct podman call rfc

dvaumoron:
This PR follows my message in https://github.com/buildpacks/pack/issues/413

### DIFF
--- a/.github/workflows/issues-generation.yml
+++ b/.github/workflows/issues-generation.yml
@@ -15,7 +15,7 @@ jobs:
     if: ${{ github.event.issue.pull_request || github.event.pull_request }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Run Issue Generation
         uses: jromero/issue-generation-action@v1.0.0-beta.4
         id: issues-generation


text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+A flag `--container-engine` will be added to the `pack` CLI, the allowed flag value will be `docker` (the current functioning) and `podman` to call podman directly (without socket call)
```
AidanDelaney, line 15: Can we add a little motivation on why we want to avoid the socket call?

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Adding `--container-engine podman` to the `pack` command which currently need docker call would allow them to work without docker installed.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The flag will change the initialization of the CommonAPIClient passed to call docker (interface defined in "github.com/docker/docker/client") by an adapter to call podman as a library.
```
AidanDelaney, line 35: Could we sketch out how the adapter would work?

text/0000-pack-direct-podman-call.md
dvaumoron: This is a way to make pack standalone, developer would be able to use pack without docker or podman, however they will still need a tool like skopeo to move OCI images.

text/0000-pack-direct-podman-call.md
dvaumoron: the goal is a to transmit/translate call to an initialized Runtime https://github.com/containers/podman/blob/v4.5.1/libpod/runtime.go#L63

dvaumoron: I can limit the adapter to the sub interface DockerClient https://github.com/buildpacks/pack/blob/v0.29.0/pkg/client/docker.go#L14, because the other method of the CommonAPIClient are not used within pack

dvaumoron:
what do you think of an option in the configuration file in addition to the flag ?

text/0000-pack-direct-podman-call.md
abitrolly: > Can we add a little motivation on why we want to avoid the socket call?

`podman` doesn't react the socket by default and it needs to be started by calling `systemd` manually. https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md

Probably the socket concern is the security. I don't know if another process can interfere with existing socket communication, but if it can call podman, I guess there is no difference in security.

dvaumoron: docker act as root, podman doesn't need that, so you can use it on a computer without root access (like a company one), in that case why bother to launch a socket in user space ? Moreover as i have already said, using podman as a library, we can make the pack utility standalone, that could be usefull for some case where you don't run container locally, because you will not need docker or podman at all.

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Summary
+[summary]: #summary
+
+A flag `--container-engine` will be added to the `pack` CLI, the allowed flag value will be `docker` (the current functioning) and `podman` to call podman directly (without socket call)
+
+# Definitions
+[definitions]: #definitions
+
+N/A
```
AidanDelaney, line 20: The underlying container runtime is `containerd` or `runc` or similar.  I think here we're discussing an alterative to the Docker-API-via-socket with the Podman API.  Is "Orchestration Client API" a good collective term for these?  Is there a more commonly used term?

```suggestion
Orchestration Client API: an API such as Docker's `DockerClient` or Podman's `Runtime` that 
```

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This will allow to use pack as a standalone on the OS supported by podman (currently Linux, as Windows and macOS need podman machine), making possible to use pack without installing docker or podman (a tool like skopeo will still be necessary to send image in registries)
```
AidanDelaney, line 25: Can we expand on the motivation to explain what the difference is between the proposed mode of operation and the existing mode of operation?

```suggestion
`pack` currently requires an orchestration client API available via a socket.  Both `docker` and `podman` are compatible with the [socket API](https://docs.docker.com/engine/api/v1.24/).  This RFC proposes that `pack` is extended to support a "daemonless" mode.  In a "daemonless" mode `pack` will directly use an orchestration client API.  This avoids running a `docker` or `podman` daemon and avoids opening up a local socket.  This will allow to use pack as a full-userspace, standalone application on the OS supported by podman (currently Linux, as Windows and macOS need podman machine).
```

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+Adding `--container-engine podman` to the `pack` command which currently need docker call would allow them to work without docker installed.
```
AidanDelaney, line 30: I wonder if `--daemonless` is a better flag name?  Does this imply `--publish` which publishes directly to an OCI registry, rather than using the local daemon for storage.

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
```
AidanDelaney, line 34: Do we need to expand on this section?  I'd like a 

1. high-level overview of how-it-works
2. details on the adapter interface
3. discussion on how layers are cached
4. discussion on where the output image ends up

I expect that we're proposing here to use an OCI registry for the layer cache and for output images.  But I think we need to make all this explict to RFC readers.

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+The flag will change the initialization of the CommonAPIClient passed to call docker (interface defined in "github.com/docker/docker/client") by an adapter conforming to the used subset [DockerClient](https://github.com/buildpacks/pack/blob/v0.29.0/pkg/client/docker.go#L14) to call podman as a library (forwarding calls to an initialized instance of [Runtime](https://github.com/containers/podman/blob/v4.5.1/libpod/runtime.go#L63)).
+
+# Migration
+[migration]: #migration
+
+N/A
```
AidanDelaney, line 40: Do we need migration documentation?

```suggestion
The current mode of `pack` operation is the default mode of operation.  The proposed `--daemonless` mode of operation is an optional mode of execution.  We will need to document how to switch between operation modes and the motivations for choosing an appropriate mode of operation.
```

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+pack contributor could need to update the podman dependencies in go.mod for bugfixes
```
AidanDelaney, line 45: Does the `podman` API support running `arm64` containers on `amd64`?  If not, then one drawback is that this proposed RFC will not benefit from changes in our multi-architecture RFC.

AidanDelaney (commented):
I like the direction that this RFC is taking.  I've suggested expanding some explanations.  In general, I'd like an RFC to contain enough information that we could hand-off the implementation to any developer.

text/0000-pack-direct-podman-call.md
dvaumoron: I am not sure that  "Orchestration Client API" would be a good choice, "Container Orchestration" refer to tools like Kubernetes or Docker Swarm

dvaumoron: Maybe "Container Manager API" to distinguish from "Container Runtime API" which are lower level API

text/0000-pack-direct-podman-call.md
dvaumoron: I think a `--daemonless` flag without argument is a better solution, this mode could be used without `--publish` as podman store OCI images in user space (generally in `$HOME/. local/share/containers/storage/`), it depends on the will of the user to run image locally

text/0000-pack-direct-podman-call.md
dvaumoron: In order to store locally OCI images with the new mode, change in the lifecycle tool will be needed too

text/0000-pack-direct-podman-call.md
dvaumoron: This seems to be ok : https://github.com/containers/podman/discussions/15741
podman use the --arch flag which is transmited to the interface ContainerEngine as an option (i changed the how it works part to change Runtime by ContainerEngine (the Runtime struct is called by one of the ContainerEngine implementation))

text/0000-pack-direct-podman-call.md
dvaumoron: I have added a short code sample to show how an adapter could be done to answer 2

dvaumoron: I have digged deeper into the caching part of pack and lifecycle, it don't seem needed to modify those part (volume caching store in local host by mounting it in the container using the interface DockerClient, image caching use registries (licefycle use github.com/google/go-containerregistry via imgutil))

AidanDelaney:
For reference, a previous discussion on replacing the daemon is at https://github.com/buildpacks/rfcs/pull/201

jjbustamante:
Hi @dvaumoron.

As my understanding the proposal is to offer a new implementation of the our  [DockerClient](https://github.com/buildpacks/pack/blob/main/pkg/client/docker.go) interface through the Adapter. The following diagram shows the components that depends on this docker client implementation, maybe it could help us on thinking potential edges cases we need to take care of.

```mermaid 
classDiagram

    
    DockerClient <|-- docker_client
    DockerClient <|-- podmanAdapter
    ImageFactory <|-- pack_imageFactory
    ImageFetcher <|-- image_Fetcher
    LifecycleExecutor <|-- build_LifecycleExecutor
    BuildpackDownloader <|-- buildpack_buildpackDownloader

    class ImageFactory {
        <<interface>>
    }

    FetchOptions <.. ImageFetcher

    class ImageFetcher {
        <<interface>>
    }

    class BlobDownloader {
        <<interface>>
    }

    class LifecycleExecutor {
        <<interface>>
    }

    build_LifecycleExecutor o-- DockerClient

    class build_LifecycleExecutor {

    }

    class BuildpackDownloader {
        <<interface>>
    }

    class buildpack_buildpackDownloader {

    }

    class pack_Client {

    }

    class FetchOptions {
        +Bool Daemon
        +String Platform
        +PullPolicy PullPolicy
        +LayoutOption LayoutOption
    }

    pack_Client o-- ImageFactory
    pack_Client o-- ImageFetcher
    pack_Client o-- DockerClient
    pack_Client o-- BlobDownloader
    pack_Client o-- LifecycleExecutor
    pack_Client o-- BuildpackDownloader
    buildpack_buildpackDownloader o-- ImageFetcher

    class DockerClient {
        <<interface>>
        ImageHistory()
        ImageInspectWithRaw()
        ImageTag() 
        ImageLoad()
        ImageSave()
        ImageRemove()
        ImagePull()
        Info()
        VolumeRemove()
        ContainerCreate()
        CopyFromContainer()
        ContainerInspect()
        ContainerRemove()
        CopyToContainer()
        ContainerWait()
        ContainerAttach()
        ContainerStart()
    }

    pack_imageFactory o-- DockerClient
    image_Fetcher o-- DockerClient

    class pack_imageFactory {

    }

    class image_Fetcher {

    }
```

From the diagram, I noticed:
- [ImageFactory](https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L262): This is just an implementation of  the [Factory Method ](https://refactoring.guru/design-patterns/factory-method) pattern, requires the docker client for those cases where we need to instantiate a imgutil.local Image. 
  -  Lifecycle and Pack uses the [Image](https://github.com/buildpacks/imgutil/blob/main/image.go) interface defined in imgutil, currently we have 3 implementations: local (for images saves in the daemon), remote (for images saves in a regristry), layout (for images saves in disk in OCI layout format). **Are you planning to reuse one of these implementations?** or **Do you need a different one that uses this podman adapter? like the local one uses the docker client**
- [ImageFetcher](https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L44): As the documentation said `is an interface representing the ability to fetch local and remote images`, if you see in the diagram above the method on this interfaces uses the *FetchOptions* class to determine from where we need to fetch the requested image, I think based on the my previous questions you will need to define new Options here.
- [LifecycleExecutor](https://github.com/buildpacks/pack/blob/main/pkg/client/build.go#L67): from the documentation `executes the lifecycle which satisfies the Cloud Native Buildpacks Lifecycle specification`, I think here you will not have a lot  of problem, maybe here @natalieparellano can help me on thinking possible side effects or things we need to take into consideration.

I hope this can be helpful, I just tried to put my thoughts on paper and share them 😄 . Also can you fix the DCO [error](https://github.com/buildpacks/rfcs/pull/288/checks?check_run_id=16760213230) ?

dvaumoron:
Hi @jjbustamante,

The methods in DockerClient interface return struct, the goal is to populate them with the data returned by the ContainerEngine implementation, the Image implementation returned by the method NewImage from "github.com/buildpacks/imgutil/local" (called at https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L271) wich use the DockerClient interface doesn't need to change (see https://github.com/buildpacks/imgutil/blob/main/local/local.go#L280 or https://github.com/buildpacks/imgutil/blob/main/local/save.go#L76 among other examples).

As earlier the Fetcher struct which implements the ImageFetcher interface use the DockerClient interface (see https://github.com/buildpacks/pack/blob/main/pkg/image/fetcher.go#L56)

The LifecycleExecutor struct use the DockerClient interface too (see https://github.com/buildpacks/pack/blob/main/internal/build/lifecycle_executor.go#L49), the lifecycle part was unclear so i have checked (reported at https://github.com/buildpacks/rfcs/pull/288#discussion_r1247989303)

All my verifications convey the same conclusion, by (a good) design the DockerClient interface is the single point of contact with the Docker API, a correct replacement with an adapter which call the Podman API instead should be enough to make the feature work.

dvaumoron:
I didn't get the point with the DCO check, it doesn't accept my signed-of-by, but looking at https://github.com/buildpacks/rfcs/commit/714a7f0637d75968d1bf3938aedb3e493f39a10b, it seems possible to use your true name with an email with doesn't match, and my name associated with my github account is correct, i need to dive deeper...

text/0000-pack-direct-podman-call.md
natalieparellano: > In order to store locally OCI images with the new mode, change in the lifecycle tool will be needed too

Could you talk about what these changes would look like?

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+Container Manager Client API: an API such as Docker's `CommonAPIClient` or Podman's `ContainerEngine` allowing interaction with container and OCI image
+
+# Motivation
+[motivation]: #motivation
+
+`pack` currently requires a container manager client API available via a socket.  Both `docker` and `podman` are compatible with the [socket API](https://docs.docker.com/engine/api/v1.24/).  This RFC proposes that `pack` is extended to support a "daemonless" mode.  In a "daemonless" mode `pack` will directly use a container manager client API.  This avoids running a `docker` or `podman` daemon and avoids opening up a local socket.  This will allow to use pack as a full-userspace, standalone application on the OS supported by podman (currently Linux, as Windows and macOS need podman machine).
```
natalieparellano, line 25: `pack` uses the docker daemon both to spin up containers in which the lifecycle executes, as well as an export target in which to store application images. Does this RFC support both use cases? I assume so, but thought it best to clarify.

text/0000-pack-direct-podman-call.md
natalieparellano: I'd love to know more about @AidanDelaney's points 1-4. Not being super familiar with podman myself, a little background information would be helpful.

dvaumoron: > > In order to store locally OCI images with the new mode, change in the lifecycle tool will be needed too
> 
> Could you talk about what these changes would look like?

As stated in the following comment, after a deeper check, that part doesn't need change, it relies on the DockerClient interface or an independent library no tied to the Docker socket

dvaumoron: > I'd love to know more about @AidanDelaney's points 1-4. Not being super familiar with podman myself, a little background information would be helpful.

1 & 2 have been detailed in the RFC document
3 layer are cached with a "VolumeCache" mecanism, which rely on the DockerClient interface
4 podman store local image in user space (/home/dvaumoron/.local/share/containers on my computer), and docker store them in common space using root right (/var/lib/containers/storage on my computer)

text/0000-pack-direct-podman-call.md
dvaumoron: It depend of what kind of export we are talking about, if you mean export to a registry, pack use the github.com/google/go-containerregistry library to do it, and do not depend on Docker for that.

If you talk about local storage, pack call the Docker socket via the DockerClient interface, changing this implementation will change the place where image will be stored, however i don't think this is a problem.

Users using the new flag to avoid installing Docker will have podman if they need to run locally containers (without having to use the podman compatibility socket), or no tools at all because pack will be usable as a standalone.

User with several tool including Docker, would be able to use them to "send" images to Docker if they prefer to use the flag (podman or buildah (which use the same image storage)  allow to do that).

User with only Docker have no reason to use the flag, if they use it and want to run the generated images locally without installing more tools, they will need to export them to a registry and retrieve them from it. Maybe adding a warning when the flag is used will avoid this kind of mistake.

text/0000-pack-direct-podman-call.md
natalieparellano: > it relies on the DockerClient interface or an independent library no tied to the Docker socket

Here's where we initialize the docker client in the lifecycle: https://github.com/buildpacks/lifecycle/blob/367b451eb708dff1aa6e5ba179e1c88c4b9917e2/priv/docker.go#L16

I'm assuming `shouldConnectSock` will be false in a daemonless scenario as pack will not mount the socket. But don't we need to specify other options to enable the podman wrapper? Or are you expecting to use the lifecycle in "OCI layout mode" and handle the processing in pack?

dvaumoron: After some more check, depending on which lifecycle command is launched the fallback when the flag use daemon is false, will be the layout version or the remote version, so the options passed to lifecycle should be adapted to ensure correctness.

dvaumoron: I added precisions in the RFC corresponding to my previous comment

text/0000-pack-direct-podman-call.md
```diff
@@ -0,0 +1,110 @@
...
```
abitrolly, line 25:
```suggestion
`pack` currently interacts with `podman` through API service.  Both `docker` and `podman` can run the service, which provides compatible [socket API](https://docs.docker.com/engine/api/v1.24/).  This RFC proposes alternative "daemonless" mode with direct calls to `podman` without opening up a local socket or running API service. This makes `pack` userspace, standalone application (currently Linux, as Windows and macOS need `podman` machine).
```

abitrolly:
A [callgraph](https://github.com/ondrajz/go-callvis) of functions that are calling socket API while building the sample project, would greatly help to estimate the critical path that needs to be patched.

dvaumoron:
[go-callvis](https://github.com/ondrajz/go-callvis) doesn't seem to help for that, it does not keep track of object passed as argument.

The diagram from @jjbustamante is more helpful than those I obtain with go-calvis

dvaumoron:
To be more accurate, you are limited to path calling some part of a package (pkg/client) from another (cmd/pack), the resulting graph is not very readable

PR #290, opened 2023-07-20
RFC Flatten builders

dlion:
[Readable](https://github.com/dlion/rfcs/blob/rfc-flatten-feature/text/0000-flatten-feature.md)


hone:
Thanks for writing this up! For a small nit can you rename your file to `text/XXX-flatten-feature.md`.

As mentioned in WG today:

* Is there a need for flattening buildpacks or just builders? My concern about flattening buildpacks is if it's distributed this way, then it removes the flexibility for the builder author (if they don't have the layer limit issue for their composition).
* Can you provide an example for exclude and include being used together?
* I'd like to see any of the user research you all are doing being added as context into this RFC!

text/XXX-flatten-feature.md
```diff
@@ -0,0 +1,325 @@
...
+# Summary
+[summary]: #summary
+
+We propose to add new capabilities to the Pack tool that allow end users to reduce the number of Buildpack's layers in an OCI image by flattening according to their requirements.
+
+# Definitions
+[definitions]: #definitions
+
+- Buildpack: A buildpack is a set of executables that inspects your app source code and creates a plan to build and run your application.
+- Builder: A builder is an image that contains all the components necessary to execute a build. A builder image is created by taking a build image and adding a lifecycle, buildpacks, and files that configure aspects of the build including the buildpack detection order and the location(s) of the run image
```
natalieparellano, line 21: We should also define "buildpack package"

text/XXX-flatten-feature.md
```diff
@@ -0,0 +1,325 @@
...
+
+- `pack builder create`
+- `pack buildpack package`
+
+The new flags will move from experimental status to supported status when maintainers deem it appropriate.
+
+The new flags to be included are:
+
+- `--flatten` will flatten all the Buildpacks in one layer
+- `--flatten --depth=<int>` If there are composite buildpack to be flatten, `depth` is the high in the buildpack dependency tree to start flattening the buildpacks
```
natalieparellano, line 61: I would love to know if this is useful in practice. If it is not, could we remove it? As it adds complexity to the code...

text/XXX-flatten-feature.md
```diff
@@ -0,0 +1,325 @@
...
+
+
+$$ 
+ total = \text{3 layers}
+$$
+
+
+---
+
+* `--flatten --flatten-include=<buildpacks>` i.e. `--flatten --flatten-include=<BP1,BP2> --flatten-include=<BP3,BP4>`
```
natalieparellano, line 199: I like this for its simplicity and ease of understanding how buildpacks will be flattened together.

text/XXX-flatten-feature.md
```diff
@@ -0,0 +1,325 @@
...
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to be resolved before this gets merged?
+
+We think the `--depth` command MAY not be helpful, after implementing a PoC. Should we get rid of this flag?
```
natalieparellano, line 285: Actually, same question for `--flatten-exclude`. Do we need it?

text/XXX-flatten-feature.md
natalieparellano: We could potentially get rid of `--flatten-exclude` and even `--flatten` if the above is sufficient. We could perhaps introduce the special case of `--flatten-include *` if we want to flatten "all".

dlion:
> Thanks for writing this up! For a small nit can you rename your file to text/XXX-flatten-feature.md

The file is already called `https://github.com/dlion/rfcs/blob/rfc-flatten-feature/text/XXX-flatten-feature.md` did I miss anything? 🤔 

> Is there a need for flattening buildpacks or just builders? My concern about flattening buildpacks is if it's distributed this way, then it removes the flexibility for the builder author (if they don't have the layer limit issue for their composition).

Uhm, I don't know about the specific need around one or another choice since I wasn't there (I remember that @jjbustamante told me that he has been asked for this implementation) , maybe @natalieparellano has some context about it.
 
 > Can you provide an example for exclude and include being used together?

Uhm, let’s say we have something like:
BP 1, BP 2, BP 3, BP 4, BP 5, BP 6
 
```mermaid
flowchart TD
    A[CB1]
    A --> BP1[BP1]
    A --> BP2[BP2]
    A  --> BP3[BP3]
    A --> BP4[BP4]
    A --> BP5[BP5]
    A --> BP6[BP6]
```

and I want just to specifically merge BP 1 and BP2, but exclude BP 5 and leave the rest flatten to obtain something like: BP1-BP2 BP5 BP3-BP4-BP6

```mermaid
flowchart TD
A[CB1]
A --> BP1BP2[BP1-BP2]
A --> BP3BP4[BP3-BP4-BP6]
A --> BP5
```
Does it make sense?

> I'd like to see any of the user research you all are doing being added as context into this RFC!
We are planning to have the session this Thursday, I'll keep it updated @hone  😄

text/XXX-flatten-feature.md
dlion: We are going to talk about it during the Thursday session, let's see how it goes

text/XXX-flatten-feature.md
```diff
@@ -0,0 +1,325 @@
...
+- `pack buildpack package`
+
+The new flags will move from experimental status to supported status when maintainers deem it appropriate.
+
+The new flags to be included are:
+
+- `--flatten` will flatten all the Buildpacks in one layer
+- `--flatten --depth=<int>` If there are composite buildpack to be flatten, `depth` is the high in the buildpack dependency tree to start flattening the buildpacks
+- `--flatten --flatten-exclude=<buildpacks>` will flatten all the Buildpacks in one layer excepts the ones specified under the `flatten-exclude` flag 
+- `--flatten --flatten-include=<buildpacks>` will flatten the Buildpacks specified after the `flatten-include` flag into a single layer. Can be used more than once, with each use resulting in a single layer.
```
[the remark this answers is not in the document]
dlion: I guess it should stay untouched.

text/XXX-flatten-feature.md
dlion: We should validate it with our users just to be sure

hone:
@dlion 
> The file is already called https://github.com/dlion/rfcs/blob/rfc-flatten-feature/text/XXX-flatten-feature.md did I miss anything? thinking
Sorry, i meant it should be rename from that with `0000` instead of `XXX` as talked about in the [README](https://github.com/buildpacks/rfcs#proposal).

dlion:
> Sorry, i meant it should be rename from that with 0000 instead of XXX as talked about in the [README](https://github.com/buildpacks/rfcs#proposal).
Ops my bad, I misinterpreted the sentence in the readme `where 'my-feature' is descriptive. don't assign an RFC number yet` as `Don't even put 000`.
I'll fix it asap, thanks @hone

dlion:
I've just updated this RFC with the feedback we've got from some of our stakeholders:

* Remove `depth` since it's not used
* Remove `--flatten --flatten-exclude` since it's too complicate and it requires some mental gymnastic to be understood and used
* Rename `--flatten --flatten-include`  into just `--flatten`, it will make the feature simpler and the flatten `all` it's not the default mode anymore
* It doesn't really matter if this feature is behind the `experimental` flag.

Here an image of the MIRO board we used to ran the exercise with our users:
image


Thank you

text/XXX-flatten-feature.md
simonjjones: Given that Paketo has started to use this argument in a pre-release version of pack, could I ask that if you intend to remove exclude, that you phase it out over a release or two, to give us time to migrate to `--flatten-include`

text/XXX-flatten-feature.md
jjbustamante: @dlion can we add those definitions?

A **component buildpack** is a buildpack containing `/bin/detect` and `/bin/build` executables. Component buildpacks implement the [Buildpack Interface](#buildpack-interface).

A **composite buildpack** is a buildpack containing an order definition in `buildpack.toml`. Composite buildpacks do not contain `/bin/detect` or `/bin/build` executables. They MUST be [resolvable](#order-resolution) into a collection of component buildpacks.

A **A buildpackage** is a buildpack packaged for [distribution](https://github.com/buildpacks/spec/blob/main/distribution.md#buildpackage).

text/0000-flatten-feature.md
```diff
@@ -0,0 +1,222 @@
...
+
+The current [distribution spec](https://github.com/buildpacks/spec/blob/main/distribution.md#buildpackage) defines:
+
+```
+Each buildpack layer blob MUST contain a single buildpack at the following file path:
+
+/cnb/buildpacks/<buildpack ID>/<buildpack version>/
+```
+
+A Buildpackage flattened with this new feature would not be consumable by older platform implementations because they are not expecting to find more than one buildpack on a blob layer.
```
jjbustamante, line 144: We should also include a Builder

dlion:
Hi @hone, is this good to be voted?
/cc @jjbustamante @natalieparellano

jkutner (changes requested):
Could we limit this RFC to `pack builder create` with flattened layers?

It seems like that's really the long pole, since that's where most people would accumulate more than 127 layers (are there really examples of individual buildpacks that have more 127 child buildpacks?)

jjbustamante:
> Could we limit this RFC to `pack builder create` with flattened layers?
> 
> It seems like that's really the long pole, since that's where most people would accumulate more than 127 layers (are there really examples of individual buildpacks that have more 127 child buildpacks?)

I think it makes sense, just want to confirm @ForestEckhardt Paketo is flattening only the builders, right?

As an example, I took the Java buildpack from Paketo and count the number of layers there:

```bash
> crane manifest gcr.io/paketo-buildpacks/java | jq '.layers | length'
27
```
I got **27**, so I think ti is fine.

dlion:
> Could we limit this RFC to pack builder create with flattened layers?

I think it's a great idea, as I said during our sync would be better to just focus on the builder part to be sure we focus on deliverying value to our users.
From that we can start collecting feedback and if this functionality it is needed for buildpacks as well we can considerate it right away. 🚀 

I'll modify the RFC accordingly and start creating the first issue about it. Thanks for the feedback everyone!

jjbustamante:
From our WG meeting today:

- Let's remove from the RFC the idea of flatening a buildpack package, not because it is not possible or valuable, because it requires another conversation for the impact of doing that.
- We want to move forward with this RFC, so once the change are done, @jkutner  will take a look and if it is ok, then he can moves it to next state in the RFC process.
- We can open another RFC for flattening a **buildpack package** and keep working in that without blocking flattening a builder.

dlion:
Hi all, I updated the RFC removing any reference that might not be related to a Builder.
Let me know if it is fine 😄 
In the meantime I also created the issue related to this RFC: https://github.com/buildpacks/pack/issues/1880 🚀

text/0000-flatten-feature.md
```diff
@@ -0,0 +1,222 @@
...
+
+<!--
+This section should document breaks to public API and breaks in compatibility due to this RFC's proposed changes. In addition, it should document the proposed steps that one would need to take to work through these changes. Care should be give to include all applicable personas, such as platform developers, buildpack developers, buildpack users and consumers of buildpack images.
+-->
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+It could create artifacts that are not consumable by older platforms.
```
hone, line 159: This use case, only happens for platforms that create ephemeral builders (like `pack` and kpack) right? I was looking at how we consume builders at Heroku and I think this change would "just work".

hone (approved):
Scoping this to builders minimizes the impact, which I think is a good first step. How many layers is the paketo builder at this point?

jkutner (approved):

hone:
I'm moving this to voting with a close of Sept 13th. @jkutner I'll be out for half of next week, so can you handle finalizing this?

dlion:
Hey @jkutner, is there anything else I need to do to have this PR merged? 😄

PR #291, opened 2023-08-29
RFC for exporting images in parallel

ESWZY:
We already have some discussion here: [lifecycle#1167](https://github.com/buildpacks/lifecycle/pull/1167).

[Readable](https://github.com/ESWZY/buildpacks-rfcs/blob/main/text/0000-lifecycle-parallel-export.md)

### DIFF
--- a/text/0000-lifecycle-parallel-export.md
+++ b/text/0000-lifecycle-parallel-export.md
@@ -51,6 +51,33 @@ The target personas affected by this change are:
 
 This proposal, in addition to acceleration, has the greatest impact on the export time of app image. This will lead to resource competition, that is, when the app image and cache image are exported at the same time, the app image export time will become longer. For some users, they only care about the export time of the app image, but not the overall optimization effect, and enabling this capability will affect their performance. Therefore, this ability is optional.
 
+The flag name, refer to other boolean flag like `daemon`, `layout`, I think it can be named `parallel`. And then the usage is just like this:
+```
+/cnb/lifecycle/exporter \
+  [-analyzed <analyzed>] \
+  [-app <app>] \
+  [-cache-dir <cache-dir>] \
+  [-cache-image <cache-image>] \
+  [-daemon] \ # sets <daemon>
+  [-extended <extended>] \
+  [-gid <gid>] \
+  [-group <group>] \
+  [-launch-cache <launch-cache> ] \
+  [-launcher <launcher> ] \
+  [-launcher-sbom <launcher-sbom> ] \
+  [-layers <layers>] \
+  [-layout] \ # sets <layout>
+  [-layout-dir] \ # sets <layout-dir>
+  [-log-level <log-level>] \
+  [-parallel] \ # sets <parallel>
+  [-process-type <process-type> ] \
+  [-project-metadata <project-metadata> ] \
+  [-report <report> ] \
+  [-run <run>] \
+  [-uid <uid> ] \
+  <image> [<image>...]
+```
+
 # How it Works
 [how-it-works]: #how-it-works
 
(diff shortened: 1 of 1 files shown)


jjbustamante:
Hi @ESWZY!

Thank you so much for working on this proposal and the RFC it is really awesome!

My suggestions:

- Add to the *what is this* section the propose flag to the [exporter](https://github.com/buildpacks/spec/blob/main/platform.md#exporter) that will enable the feature.  Usually we just add a table, similar to the one in the spec, with the new flag we want to incorporate.
- Once you have the flag, in the *how it works* section, you can just add a few examples of invocations of the exporter with the new flag and you can explain the expected outcome and behavior. 

The new flags, at the end, will require a new API spec, so you can also put them under the *Spec. Changes * section, but that is something can be done later once you get consensus from the community around the feature.

ESWZY:
> Hi @ESWZY!
> 
> Thank you so much for working on this proposal and the RFC it is really awesome!
> 
> My suggestions:
> 
> * Add to the **what is this** section the propose flag to the [exporter](https://github.com/buildpacks/spec/blob/main/platform.md#exporter) that will enable the feature.  Usually we just add a table, similar to the one in the spec, with the new flag we want to incorporate.
> * Once you have the flag, in the **how it works** section, you can just add a few examples of invocations of the exporter with the new flag and you can explain the expected outcome and behavior.
> 
> The new flags, at the end, will require a new API spec, so you can also put them under the **Spec. Changes** section, but that is something can be done later once you get consensus from the community around the feature.

Thanks for your suggestion!

I added a proposal flag to this PRC and made a preliminary attempt in [lifecycle#1167](https://github.com/buildpacks/lifecycle/pull/1167).

text/0000-lifecycle-parallel-export.md
```diff
@@ -0,0 +1,182 @@
...
+If export one image, the effect of this function is not very obvious.
+
+```shell
+> export CNB_PARALLEL_EXPORT=true
+> /cnb/lifecycle/exporter -app cr1.example.com/foo:app
+
+# OR
+
+> /cnb/lifecycle/exporter -app cr1.example.com/foo:app -parallel
+
```
jjbustamante, line 118: I would suggest adding a `WARN` message to the end-user to indicate that using the `parallel` flag without exporting a cache image has no effect.

text/0000-lifecycle-parallel-export.md
```diff
@@ -0,0 +1,182 @@
...
+N/A.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- How to allow users to choose the export method? environment variable? Or a parameter of the creator? 
+- Does it also need to be specified in the pack tool?
+- Should this feature be enabled by default?
+
```
jjbustamante, line 150: From my point of view:
- How to allow users to choose the export method? environment variable? Or a parameter of the creator? 
I like how you are proposing it, with the flag the env variable

- Does it also need to be specified in the pack tool?
Probably, once we get the RFC approved, we can create the issue on `pack` to expose the behavior for end-users

- Should this feature be enabled by default?
I'd say **no**  to avoid unexpected behaviors to people consuming the *lifecycle*.

jjbustamante:
Awesome work @ESWZY! I really like this RFC

text/0000-lifecycle-parallel-export.md
ESWZY: This makes sense, I agree with you.

natalieparellano (approved):

jabrown85 (approved):

jjbustamante (approved):

text/0000-lifecycle-parallel-export.md
```diff
@@ -0,0 +1,213 @@
...
+  [-gid <gid>] \
+  [-group <group>] \
+  [-launch-cache <launch-cache> ] \
+  [-launcher <launcher> ] \
+  [-launcher-sbom <launcher-sbom> ] \
+  [-layers <layers>] \
+  [-layout] \ # sets <layout>
+  [-layout-dir] \ # sets <layout-dir>
+  [-log-level <log-level>] \
+  [-parallel] \ # sets <parallel>
```
natalieparellano, line 72: Suggestion from working group - when these flags are combined in the `creator` `-parallel-export` might make more sense. However `-parallel-export` might look funny as a flag to the exporter.

Perhaps `-parallel-export` as a flag to the creator and `-parallel` as a flag to the exporter? I don't have a strong opinion either way, but thought I'd surface this idea.

natalieparellano: @ESWZY @jabrown85 perhaps you could opine here? Then I'll move this one back into voting. We already have enough approvals.

AidanDelaney (approved):

natalieparellano:
Moved to voting, as this is a sub-team RFC and has approvals from both implementation maintainers I think we can merge it early cc @jabrown85

PR #295, opened 2023-09-14
Support for multi-platform artifacts when creating builders and buildpack package

jjbustamante:
[readable](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/multi-arch-phase-2/text/0000-multiarch-builders-and-package.md)

Fixes pack [#1459](https://github.com/buildpacks/pack/issues/1459)

### DIFF
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -346,7 +346,7 @@ folder structure in the OCI image for each buildpack package will be:
                 └── package.toml
 ```
 
-On the other hand, When target is `linux/arm64, the **platform root folder** determined is `<buildpack root folder>/linux/arm64`
+On the other hand, When target is `linux/arm64`, the **platform root folder** determined is `<buildpack root folder>/linux/arm64`
 and the output OCI image folder structure looks like:
 ```bash
 .
@@ -384,12 +384,12 @@ let end users know things are changing.
 
 ```bash
 pack buildpack package <buildpack> --config ./package.toml --publish 
-Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux/amd64' as default
+Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux' as default
 Successfully published package <buildpack> and saved to registry
 
 # Or
 pack buildpack package  <buildpack> --config ./package.toml --format file 
-Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux/amd64' as default
+Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux' as default
 Successfully created package <buildpack> and saved to file
  ```
 **Output**: pack will create a buildpack package image (as it is doing it today) with the provided binaries and a 
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -679,7 +679,7 @@ with a content similar to:
 If the Buildpack Author wants to create a single buildpack package they will use the `target` flag, similar to our previous 
 examples.
 
-### Important considerations for share files in the new folder structure
+### Important considerations for sharing files with the new folder structure
 
 Let's suppose we have a multi-arch buildpack **A** with local folder structure like this:
 
@@ -704,22 +704,26 @@ Let's suppose we have a multi-arch buildpack **A** with local folder structure l
           └── bar
 ```
 
-As we already discussed in this RFC, we are expecting to create two OCI images artifacts, combined with an  [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). 
+As we already discussed in this RFC, we are expecting to create two OCI images artifacts and combine with an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). 
 But what happen with the layers on those OCI images and the files that are shared? 
 
-If we update the buildpack package creation process as follows:
+Let's suppose we have the following:
+
+Given
 
 - `A` is a multi-arch buildpack  
-- `Platform` is a target defined for the buildpack A that follows a folder structure like `{os}/{arch}[/{variant}/{name@version-1}]/bin`
+- `Platform` is a target defined for `A` that follows a structure like `{os}/{arch}[/{variant}/{name@version-1}]/bin`
+
+We define the buildpack package layers as:
 
 $$BuildpackPackageLayers(A, Platform) = ShareLayers(A) + PlatformLayers(A, Platform)$$
 
 Where
 
-- **ShareLayers**: Will be a layer created with the content of the root folder filtering out any `{os}` folder
-- **PlatformLayers**: Will be a layer created with the content of the `Platform` folder given
+- **ShareLayers**: Will be a layer created with the content of the root folder filtering out all the `{os}` folders
+- **PlatformLayers**: Will be a layer created with the `bin` folder content of the given `Platform` 
 
-With our previous example, it means
+Using our previous example:
 
 $$ ShareLayers(A) = \{\text{LICENSE},\text{NOTICE},\text{README.md},\text{buildpack.toml},\text{resources/config.properties}\} $$
 
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -401,7 +401,7 @@ Successfully published package <buildpack> and saved to registry
 
 # Or
 pack buildpack package  <buildpack> --config ./package.toml --format file 
-Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux' as default
+Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux/amd64' as default
 Successfully created package <buildpack> and saved to file
  ```
 **Output**: pack will create a buildpack package image (as it is doing it today) with the provided binaries and a 
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -709,8 +709,8 @@ But what happen with the layers on those OCI images and the files that are share
 
 If we update the buildpack package creation process as follows:
 
-- If `A` is a multi-arch buildpack  
-- If `Platform` is a target defined for the buildpack A that follows a folder structure like `{os}/{arch}[/{variant}/{name@version-1}]/bin`
+- `A` is a multi-arch buildpack  
+- `Platform` is a target defined for the buildpack A that follows a folder structure like `{os}/{arch}[/{variant}/{name@version-1}]/bin`
 
 $$BuildpackPackageLayers(A, Platform) = ShareLayers(A) + PlatformLayers(A, Platform)$$
 
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -32,6 +32,11 @@ that combines them into one single consumable tag for end-users.
 - Builder: A builder is an image that contains all the components necessary to execute a build. A builder image is created by taking a build image and adding a lifecycle, buildpacks, and files that configure aspects of the build including the buildpack detection order and the location(s) of the run image
 - Image Manifest: The image manifest provides a configuration and set of layers for a single container image for a specific architecture and operating system. See [spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
 - Image Index: The image index is a higher-level manifest which points to specific image manifests, ideal for one or more platforms. See [spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md)
+- Buildpack root folder: is the top-most directory where a `buildpack.toml` can be found
+- Platform root folder: Based on our new folder structure, the **platform root folder** is the top-most directory that identifies a target in **buildpack root folder**
+  For example:
+    - given a target `linux/amd64` the **platform root folder** will be `<buildpack root folder>/linux/amd64`
+    - given a target `windows/amd64:windows@10.0.20348.1970` the **platform root folder** will be `<buildpack root folder>/windows/amd64/windows@10.0.20348.1970`
 
 # Motivation
 [motivation]: #motivation
@@ -239,8 +244,9 @@ this will be the way for end-users to specify the platform for which they want t
 # Option 1 - no variant is required
 .
 ├── buildpack.toml                 // mandatory
+├── package.toml                   // mandatory
 └── {os}                           // optional
-    └── {arch}                     // optional
+    └── {arch}                     // optional (becomes the platform root folder)
         └── bin
             ├── build              // platform dependant binary (mandatory)
             └── detect             // platform dependant binary (mandatory)
@@ -248,19 +254,24 @@ this will be the way for end-users to specify the platform for which they want t
 # Option 2 - variant is required
 .
 ├── buildpack.toml                  // mandatory
+├── package.toml                    // mandatory
 └── {os}                            // optional
     └── {arch}                      // optional
         └── {variant}               // optional
-            ├── {name@version-1}    // optional
-            │ └── bin
-            │     ├── build         // platform dependant binary (mandatory)
-            │     └── detect        // platform dependant binary (mandatory)
-            └── {name@version-2}    // optional
+            ├── {name@version-1}    // optional  (becomes the platform root folder)
+            │   └── bin
+            │       ├── build       // platform dependant binary (mandatory)
+            │       └── detect      // platform dependant binary (mandatory)
+            └── {name@version-2}    // optional (becomes the platform root folder)
                 └── bin
                     ├── build       // platform dependant binary (mandatory)
                     └── detect      // platform dependant binary (mandatory)
 ```
-> **Note** 
+- `buildpack.toml` file MUST be present at the **buildpack root folder**
+- `package.toml` file MUST be present at the **buildpack root folder**
+- For each platform, Buildpack Authors are responsible for copying or creating symlink or hard link for files into each **platform root folder** 
+
+> **Note**
 > For cross-compile buildpacks like Paketo, it looks easy to add a step to their Makefile to compile and separate the binaries following this structure. It is important to mention
 > that the final buildpack image will not change, this will only change the buildpack structure from `pack` perspective
 
@@ -285,9 +296,8 @@ versions = ["<distribution version>"]
 - When `more than 1 target is defined`
   - When `--publish` or `--format file` is specified
     - For each `target` an OCI image will be created, following these rules
-      - `pack` will copy everything in the root folder excluding all `{os}/**` folder 
-      - `pack` will determine the **target root binary folder**, this is the specific root folder for a given `target` (based on the `targets.path` in the buildpack.toml or inferring it from a folder structure similar to the one show above)
-        - `pack` will copy everything from the **target root binary folder** into the base buildpack package folder, in case a conflict file name is found, `pack` will **override** the existing file with the latest one
+      - `pack` will determine the **platform root folder**, this is the specific root folder for a given `target` (based on the `targets.path` in the buildpack.toml or inferring it from a folder structure similar to the one show above)
+      - `pack` will execute the current process to create a buildpack package using the **platform root folder** and the `target` values  
     - If more than 1 OCI image was created, an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md) will be created to combine them
   - When `--daemon` is specified
     - `pack` can keep using `docker.OSType` to determine the target `os` and probably can do some validations it the `os` is valid target
@@ -296,15 +306,14 @@ versions = ["<distribution version>"]
 
 Let's use some examples to explain the expected behavior in different use cases
 
-#### Overriding files rules
+#### How to determine the platform root folder
 
 Let's suppose the Buildpack Author creates a multi-arch folder structure and wants to create multiple buildpack packages
 
 ```bash
 .
 ├── buildpack.toml
-├── package.toml
-├── foo             // some common file for all the platforms
+├── package.toml            
 └── linux
    ├── amd64
    │   └── bin
@@ -312,33 +321,15 @@ Let's suppose the Buildpack Author creates a multi-arch folder structure and wan
    │      ├── detect
    │      └── foo
    └── arm64
-       ├── foo     // for some reason we need a different binary for arm64
+       ├── foo
        └── bin
           ├── build
           ├── detect
           └── bar
 ```
 
-In this case:  
- - The targets `os` provided are `["linux"]`, then 
- - For each `target` pack will exclude everything from `linux/**` to be copied into the initial buildpack package
-
-Resulting into a base structure like:
-
-```bash
-.
-└── cnb
-    └── buildpacks
-        └── {ID}
-            └── {version}
-                ├── buildpack.toml
-                ├── foo             // this is our common file
-                └── package.toml
-```
-
-When target is `linux/amd64`
-  - `pack` will determined **target root binary folder** is `./linux/amd64` 
-    - `pack` will copy everything from the `./linux/arm64` into the base buildpack package folder, because there are no conflicts the expected output will be:
+- When `linux/amd64` the **platform root folder** determined is `<buildpack root folder>/linux/amd64`, and the expected
+folder structure in the OCI image for each buildpack package will be:
 
 ```bash
 
@@ -352,14 +343,11 @@ When target is `linux/amd64`
                 │    ├── detect
                 │    └── foo         // specific platform binary
                 ├── buildpack.toml
-                ├── foo              // common file
                 └── package.toml
 ```
 
-On the other hand, When target is `linux/arm64`
-- `pack` will determined **target root binary folder** is `./linux/arm64`
-  - `pack` will copy everything from the `./linux/arm64` into the base buildpack package folder, because there is a conflict trying to copy `./linux/arm64/foo` into `cnb/buildpacks/{ID}/{version}/foo`, `pack` will override it and moves forward, resulting into:
-
+On the other hand, When target is `linux/arm64, the **platform root folder** determined is `<buildpack root folder>/linux/arm64`
+and the output OCI image folder structure looks like:
 ```bash
 .
 └── cnb
@@ -371,7 +359,7 @@ On the other hand, When target is `linux/arm64`
                 │    ├── build
                 │    └── detect
                 ├── buildpack.toml
-                ├── foo             // override files with the content of the arm64 folder
+                ├── foo
                 └── package.toml
 ```
 
@@ -759,63 +747,6 @@ with a content similar to:
 If the Buildpack Author wants to create a single buildpack package they will use the `target` flag, similar to our previous 
 examples.
 
-### Important considerations for sharing files with the new folder structure
-
-Let's suppose we have a multi-arch buildpack **A** with local folder structure like this:
-
-```bash
-.
-├── LICENSE
-├── NOTICE
-├── README.md
-├── buildpack.toml
-├── package.toml
-├── some-file             // a file with a big size for example
-├── resources
-│ ├── config.properties
-└── linux
-   ├── amd64
-   │   └── bin
-   │      ├── build
-   │      ├── detect
-   │      └── foo
-   └── arm64
-       └── bin
-          ├── build
-          ├── detect
-          └── bar
-```
-
-As we already discussed in this RFC, we are expecting to create two OCI images artifacts and combine with an 
-[image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md). The current [distribution spec](https://github.com/buildpacks/spec/blob/main/distribution.md#buildpackage) 
-defines:
-
-```
-Each buildpack layer blob MUST contain a single buildpack at the following file path:
-
-/cnb/buildpacks/<buildpack ID>/<buildpack version>/
-```
-
-This means, we will end up duplicating the big file in each OCI images, for example our `linux/amd64` buildpack package 
-will have 1 layer with the following content under `/cnb/buildpacks/<buildpack ID>/<buildpack version>/` 
-
-```bash
-├── LICENSE
-├── NOTICE
-├── README.md
-├── buildpack.toml
-├── some-file             // a file with a big size for example
-├── resources
-│ └── config.properties
-└── bin
-  ├── build
-  ├── detect
-  └── foo
-```
-
-And the `linux/arm64` version will be similar but with the `bar` file inside the `bin/` folder, as a consequence, the big file
-will be duplicated on two different layers increasing the space requirements in the registry
-
 ## Builder
 
 Similar to how we did it for the `buildpack package`, lets summaries, our current process to create a **Builder**:
@@ -1005,10 +936,6 @@ Tip: Run pack build <image-name> --builder <builder> to use this builder
 **Output**: We keep our current behavior and detect the `os` and `architecture` from the daemon. Because there is `target` 
 that matches the daemon `os/arch` the builder is being built.
 
-Using `--target` flag against the daemon with a different platform
-
-TODO check if this is possible!
-
 <!--  
 ```bash
 pack builder create <builder> --config ./builder.toml --target linux/arm64
@@ -1091,6 +1018,8 @@ create [image indexes](https://github.com/opencontainers/image-spec/blob/master/
 - What are the intermediate images for each target named/called?
 - What happen if I want to exclude some buildpack for a particular target?
 - What happen if I want to include the same file or folder for every image, do I have to copy then inside the {os}-{arch} folder?
+- Initially we proposed a shared file strategy but, we decided to leave that complexity out of the scope of this RFC and can be 
+revisited later if it required
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
--- a/text/0000-multiarch-builders-and-package.md
+++ b/text/0000-multiarch-builders-and-package.md
@@ -689,6 +689,8 @@ Let's suppose we have a multi-arch buildpack **A** with local folder structure l
 ├── NOTICE
 ├── README.md
 ├── buildpack.toml
+├── package.toml
+├── some-file             // a file with a big size for example
 ├── resources
 │ ├── config.properties
 └── linux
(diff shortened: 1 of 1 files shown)


text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+2. Support buildpack authors to **create new buildpacks and builders** that handle multi-arch from the beginning.
+3. Support application developers to **create application images** using multi-arch buildpacks and builders.
+
+The purpose of this RFC is to solve the statement 2, adding the capability to the commands:
+
+- `pack buildpack package`
+- `pack builder create`
+
+to create individuals OCI images artifacts per each os and arch (builders and buildpack packages) and handle the creation for the  [image index,](https://github.com/opencontainers/image-spec/blob/master/image-index.md)
+that combines them into one single consumable tag for end-users.
```
dmikusa, line 26: To confirm, this will support `--publish` of the generated OCI images also? We'd been creating OCI archives in some of our pipelines and when we started generating multi-arch images in OCI archive files, it broke the tools we were using to move things around (`crane` and `skopeo`). It'd be really helpful to have that support in the tool.

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+```
+
+But in this case, we actually **remove** the [platform](https://buildpacks.io/docs/reference/config/package-config/) section because it will be taken from the `buildpack.toml`.
+
+Packaging a multi-arch buildpack will require the output to be **publish** to a registry or **saved on disk** in OCI layout format.
+
+```bash
+pack buildpack package my-buildpack --config ./package.toml --publish --multi-arch
+# Or
+pack buildpack package my-buildpack.cnb --config ./package.toml --format file --multi-arch
```
dmikusa, line 137: Could the `--multi-arch` flag be inferred? For example, if `buildpack.toml` contains multiple target/arch's then this flag is automatic. Or are we worried someone might have that but still only want to build a buildpack for a single arch?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+```
+
+In these cases each `target` entry corresponds to a different buildpack image that is exported into an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md)
+
+### Package a multi-arch Builder
+
+In case of packing a **Builder**, we assume the following premises: 
+
+1. Buildpack authors migrated their `builder.toml` and [remove the stack](https://github.com/buildpacks/pack/issues/1303) concept.
```
dmikusa, line 146: I'm concerned about this. If we make the change in #1303, will we still be able to support stacks? We have a lot of users in production. We need to support both for a while so we can give folks plenty of time to migrate.

If this requires removing and stopping support for stacks then I suspect that this will hold us up for a while on being able to use this functionality.

Is there any way we could change this to something like `Buildpack authors migrated `builder.toml` and have added the entries to `builder.toml` for the new run/build image format.`. So that we could have both for some period of time as we migrate?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+We propose:
+- Add a new `--platform` flag with format `[os][/arch][/variant]:[name@version]` to build for a particular target, once the `platform.os` field is removed, 
+this will be the way for end-users to specify the platform for which they want to create single OCI artifact.
+> **Note**
+> We want to avoid duplicated information, we expect to have `targets` with all the platform supported by a buildpack
+
+- Add a new boolean `--multi-arch` flag to indicate pack it must create multiples OCI artifacts and combine them with an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md), 
+this flag must be used in conjunction with `--publish` or `--format file` flags and error out when `daemon` is selected.
+> **Note**
+> I got feedback to avoid creating an Image Index if users are not expecting that, the flag is an acknowledgment for creating the index
```
dmikusa, line 220: Similar thoughts here. Could this be inferred based on the presence of multiple targets? I get that we don't want to just suddenly start making image indexes as that changes things. But couldn't we only do this if multiple targets with different arch's are specified? Or is there a case where you'd have multiple target/arch's and still not want that behavior?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+        │   └── bin
+        │       ├── build
+        │       └── detect
+        └── {version-2}     // optional
+            └── bin
+                ├── build
+                └── detect
+```
+> **Note** 
+> For cross-compile buildpacks like Paketo, it looks easy to add a step to their Makefile to compile and separate the binaries following this structure.
```
dmikusa, line 238: Yes, that shouldn't be an issue.

Would this only change the buildpacks structure presented to `pack` or would it also change the buildpack structure in the actual image that's generated? The latter would require buildpack API/spec changes, right?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+```bash
+pack buildpack package <buildpack> --config ./package.toml --publish --platform linux/arm64 --platform linux/amd64 --multi-arch
+A multi-arch buildpack package will be created for platforms: 'linux/amd64', 'linux/arm64'
+Successfully published package <buildpack> and saved to registry
+
+# Or
+pack buildpack package  <buildpack> --config ./package.toml --format file --platform linux/arm64 --platform linux/amd64  --multi-arch
+A multi-arch buildpack package will be created for platforms: 'linux/amd64', 'linux/arm64'
+Successfully created package <buildpack> and saved to file
+```
```
dmikusa, line 342: So this command could be used if one hasn't updated to use targets yet? i.e. still using stacks?

What about buildpacks that are migrating from stacks to no-stacks and have both present? This use case concerns me the most, cause I know we (Paketo) will be in this state for a while.

dmikusa: I'd be interested to hear more in general about this state of migrating from stacks to no-stacks. What commands would use the old stack metadata? what commands would use the new target info? I'm assuming we'd keep the two synchronized, but just curious more about how this would all work.

Or am I overthinking this, and the net result (i.e. the images generated) will be the same regardless of stacks or no stacks?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+In this case, because `targets` are specified and `--multi-arch` is being used `pack` will follow the builder creation process for **each provided platform**, 
+pulling the correct (based on os/arch) buildpacks, build and run images and creating different builders images that are 
+exported and combined into an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md)
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Buildpack Package
+
+As a quick summary, our current process to create a buildpack package involves:
```
dmikusa, line 193: Would anything change for a composite/meta buildpack? I'd assume that you would need to have multi-arch images for every buildpack in the composite buildpack. Otherwise, would it work the same as a regular component buildpack?

text/0000-multiarch-builders-and-package.md
jjbustamante: Yes, the idea is to push the generated OCI images and create the Image Index

text/0000-multiarch-builders-and-package.md
jjbustamante: It is exactly this: 
> Or are we worried someone might have that but still only want to build a buildpack for a single arch?
At first, I was thinking on doing it without the `--multi-arch` flag, but then I decided to force the user to give the acknowledgment

text/0000-multiarch-builders-and-package.md
jjbustamante: Let me double check this, I need the targets specify in the `builder.toml`. If the users still specify stacks for compatibility I think we can workaround it.

text/0000-multiarch-builders-and-package.md
jjbustamante: This will only change the buildpack structure presented to `pack`, the final image should keep the structure as we defined in the spec.

text/0000-multiarch-builders-and-package.md
jjbustamante: That's correct, I will add a new example with the composite builpack, but we are assuming you'd need to have multi-arch images for every buildpack in the composite buildpack

text/0000-multiarch-builders-and-package.md
jjbustamante: Good point, regarding the stack removal [plan](https://docs.google.com/document/d/1bExpr31U5R5yQ6fncpl5YcdosWVwYcXgkt12vE-lpvU/edit#heading=h.jpmej4uevvh8) we are planning to support both scenarios (stacks and run, build images), based on your feedback the idea will be to clarify how a world with:
- stacks
- targets
- multi-arch support 
will work

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+name = "windows"
+versions = ["10.0.20348.1970"]
+```
+
+And organizes the binaries according to their os/arch with a structure similar to this one:
+
+```bash
+my-multiarch-buildpack
+├── buildpack.toml
+├── linux-amd64
```
natalieparellano, line 103: Is there any case for doing something like

```
└── linux
    └── amd64
        └── bin
            ├── build
            └── detect
```

or (in the case that the buildpack runs on any arch)

```
└── linux
    └── bin
        ├── build
        └── detect
```

with both being accepted? Asking because that pairs nicely with the command line input `[os][/arch][/variant]:[name@version]`. Having to type linux-amd64 sometimes and linux/amd64 other times might be confusing.

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+- The end-users defined `os` for the OCI image using the [package.toml](https://buildpacks.io/docs/reference/config/package-config/) when they want to export a buildpack package.
+- The only values allowed are `linux` and `windows` and by default when is not present, `linux` is being used.
+- When exporting to daemon, the `docker.OSType` must be equal to `platform.os`
+
+### To keep compatibility 
+
+We propose:
+- Deprecate the `platform.os` field from [package.toml](https://buildpacks.io/docs/reference/config/package-config/). It will be removed after two pack releases with the new feature
+- When `platform.os` is present in [package.toml](https://buildpacks.io/docs/reference/config/package-config/), throw a warning messages indicating the field will be removed 
+and `--platform` flag must be used
```
natalieparellano, line 204: This was brought up elsewhere, but `--platform` might be confusing when the corresponding word in buildpack.toml and builder.toml is `--target`.

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,944 @@
...
+api = "0.12"
+
+# Buildpack ID and metadata
+[buildpack]
+  id = "examples/my-multiarch-buildpack"
+  version = "0.0.1"
+
+# List of targets operating systems, architectures and versions
+
+[[targets]]
```
natalieparellano, line 81: I am sorry for opening this can of worms, but technically "target" when used in the context of say `CNB_TARGET_OS` or `CNB_TARGET_ARCH` refers exclusively to the run image. Right now, all target data for the build image and the run image must match - but perhaps that won't always be the case? What will we do then?

text/0000-multiarch-builders-and-package.md
natalieparellano: I think for the purposes of this RFC any `[[stacks]]` in builder.toml can just be ignored.

text/0000-multiarch-builders-and-package.md
natalieparellano: I agree that having to type `--multi-arch` everywhere is cumbersome. I think if the buildpack author took the trouble to arrange their binaries according to the nested directory structure then we should just infer that they want a multi-arch image. Or maybe there could be a `pack config` option that sets multi-arch (or not) as the default.

text/0000-multiarch-builders-and-package.md
natalieparellano: I am trying to think of who would be specifying multiple targets today and I'm guessing it would just be bash buildpacks. Maybe we could print a warning if multiple targets are defined but there is only one ./bin/detect or ./bin/build.

text/0000-multiarch-builders-and-package.md
jjbustamante: I agreed, I created the structure first and then I thought on the `flag`. I will change it

text/0000-multiarch-builders-and-package.md
jjbustamante: I removed the `--multi-arch` flag

text/0000-multiarch-builders-and-package.md
jjbustamante: Yeah, I think stacks can be ignored, and we will follow the steps described in the [plan](https://docs.google.com/document/d/1bExpr31U5R5yQ6fncpl5YcdosWVwYcXgkt12vE-lpvU/edit#heading=h.jpmej4uevvh8) to keep compatibility

text/0000-multiarch-builders-and-package.md
jjbustamante: The scenarios I was thinking were these:
1. I do not have `targets` on my `buildpack.toml` 
   - In this case, the idea is to `pack buildpack package` do the same thing as it is doing today. 
1. I do have `targets` on my `buildpack.toml` 
  - **Daemon** case: `pack` should build my **buildpack package** with the platform defined by the daemon (as it is doing today) without requiring any new flag. 
  - **Registry or File** case: 
     -  If more than 1 target is defined: `pack` will create the `Image Index`, If I just want to create 1 single package, I will need to use `--target` flag to defined which one.
     - If only 1 target is available, nothing changes

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+[[targets.distributions]]
+name = "<distribution ID>"
+versions = ["<distribution version>"]
+```
+- When `more than 1 target is defined`
+  - When `--publish` or `--format file` is specified
+    - For each `target` an OCI image will be created, following these rules
+      - `pack` will copy everything in the root folder excluding all `{os}/**` folder 
+      - `pack` will determine the **target root binary folder**, this is the specific root folder for a given `target` (based on the `targets.path` in the buildpack.toml or inferring it from a folder structure similar to the one show above)
+        - `pack` will copy everything from the **target root binary folder** into the base buildpack package folder, in case a conflict file name is found, `pack` will **override** the existing file with the latest one
```
joshwlewis, line 290: Will these two copy steps outlined happen in different image layers? It seems like the idea here is to save disk space for files that are common across architectures. To achieve that, I think you'd need an image layer with the common files, that images for each architecture could include in their layer hierarchy. Maybe this was the idea and I missed it. If so, maybe worth mentioning somewhere.

Otherwise, if common files are in the same image layer as the architecture-specific files, clients would essentially be downloading and storing the common files once for each architecture.

joshwlewis: Ok, I did find the answer a bit further on: https://github.com/buildpacks/rfcs/pull/295/files#diff-cf9fc6f40b9a670966257d2475741958b49fb15684dd37cf8019514fe779885bR799

joshwlewis: If the common files are duplicated across architectures anyway, does the common file inheritance have much value? End users of buildpacks and builders won't really get much benefit. And the CNBs I work on (Heroku's), they won't have many (maybe 0) files shared across architectures.

Just thinking common file inheritance is a heavy feature with some potential gotchas, and buildpack authors who want/need to dedupe files could probably get pretty far just by symlinking or hardlinking files into multiple `{os}/{arch}` folders.

jjbustamante: Hi @joshwlewis 

Thank you very much for your feedback! I think that at least the `buildpack.toml` and the `package.toml` will be two common files shared across architectures, that's actually a win for a Buildpack Authors, right? because they do not need to keep separate files for each os/arch they want to support.

natalieparellano: > common file inheritance is a heavy feature with some potential gotchas

I agree. Maybe it makes sense to take it out of this RFC and add it later if buildpack authors want it

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+and `--target` flag must be used
+- When `platform.os` is not present in [package.toml](https://buildpacks.io/docs/reference/config/package-config/) and `--target` flag is not used, throw a warning messages indicating 
+a new `--target` flag is available
+- Keep doing our current process to package a buildpack
+
+### To improve user experience
+
+We propose:
+- Add a new `--target` flag with the follwing format `[os][/arch][/variant]:[name@version]` to build for a particular target, once the `platform.os` field is removed, 
+this will be the way for end-users to specify the platform for which they want to create single OCI artifact.
```
natalieparellano, line 233: I like to think of `--target` as more of a helpful way to narrow down the os/arch when buildpack.toml has multiple definitions, or to avoid creating a manifest list. Is that correct?

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+                └── bin
+                    ├── build       // platform dependant binary (mandatory)
+                    └── detect      // platform dependant binary (mandatory)
+```
+> **Note** 
+> For cross-compile buildpacks like Paketo, it looks easy to add a step to their Makefile to compile and separate the binaries following this structure. It is important to mention
+> that the final buildpack image will not change, this will only change the buildpack structure from `pack` perspective
+
+In case, this folder structure is not suitable for Buildpack Authors, **we propose** a new `path` attribute to be included 
+into the `target` section of the `buildpack.toml` to specify where the binaries are located in the filesystem.
```
natalieparellano, line 268: Would that be the location of the `bin` directory, or the buildpack root directory? I think the latter makes more sense, but I could be wrong. Do we need to add this now, or is this something that could be added later if buildpack authors request it?

text/0000-multiarch-builders-and-package.md
natalieparellano: > I think that at least the buildpack.toml and the package.toml will be two common files shared across architectures

I feel like `buildpack.toml` could be the one and only special case. IDK if `package.toml` is something that we even expect to be inside a buildpack root directory at build time. It's just easier to comprehend the packaging step if we don't move too much stuff around inside the directory tree. Perhaps we could spec out that we'd resolve symlinks when creating the layer if folks want to take advantage of that.

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+├── buildpack.toml
+└── package.toml
+```
+
+In these cases: We expect `pack` to keep doing what is doing today, but with the warning messages we mentioned above to 
+let end users know things are changing. 
+
+```bash
+pack buildpack package <buildpack> --config ./package.toml --publish 
+Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux' as default
```
natalieparellano, line 399:
```suggestion
Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux/amd64' as default
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+Successfully published package <buildpack> and saved to registry
+
+# Or
+pack buildpack package  <buildpack> --config ./package.toml --format file 
+Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux' as default
```
natalieparellano, line 404:
```suggestion
Warning: A new '--target' flag is available to set the platform for the buildpack package, using 'linux/amd64' as default
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+Successfully created package <buildpack> and saved to file
+ ```
+**Output**: pack will create a buildpack package image (as it is doing it today) with the provided binaries and a 
+[configuration](https://github.com/opencontainers/image-spec/blob/main/config.md#properties) with the following target 
+platform:
+
+```json
+{
+  "architecture": "",
```
natalieparellano, line 413: This is how it works today, but this feels like a bug. The OCI image spec says this field is required.

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1119 @@
...
+[[run.images]]
+image = "cnbs/sample-base-run:jammy"
+```
+
+In these cases, the expected output will be similar to:
+
+```bash
+pack builder create <builder> --config ./builder.toml 
+Warning: "stack" has been deprecated, prefer "targets" instead: https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md
+Warning: A new '--target' flag is available to set the target platform for the builder, using 'linux/amd64' as default
```
natalieparellano, line 922: This assumes that the build image is a manifest list, no? Because if the build image is linux/arm64 (only) then we create an arm64 image, no?

natalieparellano:
Awesome work @jjbustamante. I would really prefer if we kept "shared" files out of this RFC, it would make it a lot simpler to understand and implement. But otherwise this looks great.

text/0000-multiarch-builders-and-package.md
jjbustamante: I was thinking on the **buildpack root directory**, well it came up in a WG meeting suggested by @hone. on the implementation side, I'd propose we could create the ticket and do it later that will simplify the first implementation.

text/0000-multiarch-builders-and-package.md
jjbustamante: Maybe it is a bug, but apparently nobody is complaining about it! I didn't want to start filling this attribute and break the current behavior for end-users.

text/0000-multiarch-builders-and-package.md
jjbustamante: Yeah, the `pack builder create` command assumes the build, run and buildpack images are multi-arch, and when building for a particular platform, `pack` should validate it

text/0000-multiarch-builders-and-package.md
jjbustamante: Well, if I take a look at some buildpacks for Paketo for example, they have an structure like this one in the root folder:

```bash
├── LICENSE
├── NOTICE
├── README.md
├── buildpack.toml
├── package.toml
├── resources
│ ├── config.properties
```

In these cases, I think they want to shared those files on each platform images right?

text/0000-multiarch-builders-and-package.md
jjbustamante: Yes, **exactly** when you actually want to avoid creating multiple images, then `target` will help you

text/0000-multiarch-builders-and-package.md
joshwlewis: > In these cases, I think they want to shared those files on each platform images right?

They could copy, symlink, or hard link files into each platform folder. I don't think that's a big challenge for buildpack authors (they probably script this in some way anyway). Honestly, I think it might be a bit more straightforward for them to understand, too. Copy, symlink, and hardlink are established concepts; they won't need to read this RFC or the buildpack spec to know how to do that.

I still think there's convenience in this shared file logic. I'm mostly just thinking you could simplify the initial implementation, long term maintenance, and probably this RFC without it.

text/0000-multiarch-builders-and-package.md
natalieparellano: Actually this did come up recently ([here](https://github.com/buildpacks/pack/issues/1968) and [here](https://cloud-native.slack.com/archives/C0331B61A1Y/p1706546411819879)) I __think__ we said this RFC would fix those cases but maybe we assumed first the buildpack author would upgrade to use `targets`. In any case, it probably doesn't hurt... I'm fine to leave it out of this RFC though

jjbustamante:
@natalieparellano , @joshwlewis 

Thank you very much for your feedback. Just to quick summarize the changes that I am going to do:

1. From the **root folder** `pack` will read: **buildpack.toml** and **package.toml** 
2. Based on that information, `pack` will determine the **target root folder**, which in theory must contain all the files the Buildpack Author wants to put into the OCI image. 
3. Buildpack Authors will be responsible for copying or creating symlink or hard link for each files into each platform folder

In this way, we will eliminate the complexity of copying partial files and recreating the platform content with the shared files strategy.

did I get it right?

jjbustamante:
@natalieparellano @joshwlewis 

If you can take a look at [these]([50eff2e](https://github.com/buildpacks/rfcs/pull/295/commits/50eff2e8080595ea9fa377035915fa032514492d)) changes will be awesome

text/0000-multiarch-builders-and-package.md
hone: @natalieparellano what do you think we should use here?

jjbustamante:
Also, should I put something on how it is supposed to work for composite buildpacks? based on the discussions on [this](https://cloud-native.slack.com/archives/C033DV9EBDF/p1707401225376109) thread

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1048 @@
...
+this will be the way for end-users to specify the platform for which they want to create single OCI artifact.
+
+- Add `targets` section to `buildpack.toml` to help Buildpack Authors to include support for new platforms without having to update their `pack buildpack package` command in their CI/CD pipelines
+
+- A new folder structure to organize the buildpacks binaries for multi-arch images, similar to this one:
+```bash
+# Option 1 - no variant is required
+.
+├── buildpack.toml                 // mandatory
+├── package.toml                   // mandatory
```
natalieparellano, line 247:
```suggestion
├── package.toml                   
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1048 @@
...
+└── {os}                           // optional
+    └── {arch}                     // optional (becomes the platform root folder)
+        └── bin
+            ├── build              // platform dependant binary (mandatory)
+            └── detect             // platform dependant binary (mandatory)
+
+# Option 2 - variant is required
+.
+├── buildpack.toml                  // mandatory
+├── package.toml                    // mandatory
```
natalieparellano, line 257:
```suggestion
├── package.toml                    
```

text/0000-multiarch-builders-and-package.md
natalieparellano: I don't think package.toml is required today - or at least, it isn't required to live in the buildpack root directory

Looking at this old pack [issue](https://github.com/buildpacks/pack/issues/1082) we might not even need package.toml for component buildpacks if `pack buildpack package` is run inside the buildpack directory containing buildpack.toml.

Taking the above into account, package.toml is probably most useful for composite buildpacks. Following on the discussion of whether composite buildpacks need targets, perhaps we should introduce `[targets]` in package.toml, with the understanding that this is really for composite buildpacks.

jjbustamante: I am working on implementing a PoC and I actually got a problem when trying to create a **multi-arch composite buildpack**. We need a place where to define `[targets]` and `buildpack.toml` can't be that place because we don't allow `Orders` and `Targets | Stacks` at the same time.

text/0000-multiarch-builders-and-package.md
natalieparellano: I am pretty sure that the current architecture will accommodate this. What would probably end up happening is that buildpack.toml would have a list of `[[targets]]` and the buildpack os/arch/bin directories would be a subset of that list. Then at build time (somehow, but probably in run.toml) you'd have a list of run images to fulfill all the os/arch combinations in `[[targets]]` and by iterating through that list, you have a multi-arch application image. The key is that the mapping between buildpack.toml `[[targets]]` and the buildpack os/arch/bin directories would not necessarily be 1:1, whereas today we assume it is.

text/0000-multiarch-builders-and-package.md
jjbustamante: Yeah, I think is what @jericop is trying to fix also [here](https://github.com/buildpacks/pack/pull/2080)

natalieparellano (approved):

natalieparellano:
Moving to voting, closing next Thursday March 21

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+  "architecture": "arm64",
+  "os": "linux"
+}
+```
+
+> **Important**
+> Pack will assume the binaries are appropriate for the given target platform, what the flag is doing is expose a mechanism
+> to update the metadata present in the OCI config file
+
+what about creating a multi-arch image for several target platforms?
```
runesoerensen, line 431:
```suggestion
what about creating a multi-platform image for several targets?
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+
+```bash
+pack buildpack package <buildpack> --config ./package.toml --publish --target linux/arm64 --target linux/amd64
+A multi-arch buildpack package will be created for target platforms: 'linux/amd64', 'linux/arm64'
```
runesoerensen, line 435:
```suggestion
A multi-platform buildpack package will be created for: 'linux/amd64', 'linux/arm64'
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+Successfully published package <buildpack> and saved to registry
```
runesoerensen, line 436:
```suggestion
Successfully published <buildpack>
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+        "architecture": "amd64",
+        "os": "linux"
+      },
+      "size": 424
+    },
+    {
+      "digest": "sha256:2589fe6bcf90466564741ae0d8309d1323f33b6ec8a5d401a62d0b256bcc3c37",
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "platform": {
+        "architecture": "arm",
```
runesoerensen, line 467:
```suggestion
        "architecture": "arm64",
```

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+[prior-art]: #prior-art
+
+- Stack Removal [RFC #096](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/multi-arch-phase-2/text/0096-remove-stacks-mixins.md)
+- This RFC is a continuation of the work started with the proposal to add commands to handle manifest list in pack, see the [RFC](https://github.com/buildpacks/rfcs/pull/283)
+- Paketo [RFC #288](https://github.com/paketo-buildpacks/rfcs/pull/288) to publish multi-arch buildpacks
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- If I already have an [image indexe](https://github.com/opencontainers/image-spec/blob/master/image-index.md) created and I want to add support for new plaform, how do I do it?
```
runesoerensen, line 1059:
```suggestion
- How would I add support for a new platform to an existing [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md)?
```

runesoerensen (commented):
Very excited that this has progressed to `status/voting`, and personally can't wait to see better multi-platform support in our tooling!

At the risk of coming off a bit pedantic, I think it'd be worthwhile to consider the terminology used in this RFC (and our tooling). While supporting multiple architectures on linux is undoubtedly the most pressing reason most of us want this, it seems a bit inaccurate to refer to the changes proposed here as "multi-arch"?

The title is currently "Support for multi-arch artifacts when creating builders and buildpack package", but it seems something like "Support creating multi-platform builder and buildpack artifacts" would more accurately encapsulate/describe the proposed changes? This would also align better with [Docker's terminology](https://docs.docker.com/build/building/multi-platform/).

I've suggested some changes (non-exhaustively) to the wording in case that's helpful - feel free to use/dismiss as you see fit. Appreciate you all's efforts, and again, I very much hope we'll soon see these changes introduced, regardless of what we call them :)

jjbustamante:
> Very excited that this has progressed to `status/voting`, and personally can't wait to see better multi-platform support in our tooling!
> 
> At the risk of coming off a bit pedantic, I think it'd be worthwhile to consider the terminology used in this RFC (and our tooling). While supporting multiple architectures on linux is undoubtedly the most pressing reason most of us want this, it seems a bit inaccurate to refer to the changes proposed here as "multi-arch"?
> 
> The title is currently "Support for multi-arch artifacts when creating builders and buildpack package", but maybe something like "Support creating multi-platform builder and buildpack artifacts" would more accurately encapsulate/describe the proposed changes? This would also align better with [Docker's terminology](https://docs.docker.com/build/building/multi-platform/).
> 
> I've suggested some changes (non-exhaustively) to the wording in case that's helpful - feel free to use/dismiss as you see fit. Appreciate you all's efforts, and again, I very much hope we'll soon see these changes introduced, regardless of what we call them :)

@runesoerensen Thank you so much for your suggestions! I will take a look at them but I truly appreciate it. I am not an native english speaker 😄  and this kind of feedback is always great for me to keep improving!

jjbustamante (commented):
Applying some grammar suggestions

text/0000-multiarch-builders-and-package.md
jjbustamante: This is actually the current messaged showed in pack

text/0000-multiarch-builders-and-package.md
```diff
@@ -0,0 +1,1090 @@
...
+os = "linux"
+arch = "arm64"
+
+[[targets]]
+os = "windows"
+arch = "amd64"
+
+[[targets.distributions]]
+name = "windows"
+versions = ["10.0.20348.1970"]
```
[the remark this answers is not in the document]
schneems: I'm wondering what an automated checks around non-code PRs such as a `buildpack.toml` linter (for example) might look like.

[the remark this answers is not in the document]
jjbustamante: > @jjbustamante I believe this should be `[[targets.distros]]` plus `versions` should be `version` (singular) and a string not an array?
> 
> (Plus the same change made to the 3-4 similar references in this file)

Yes!! You are right @edmorley! I will ask how could I do the fix now that this was merged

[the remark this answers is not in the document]
jjbustamante: I just saw you opened a PR to fix a previous RFC, is it possible for you to open one for this?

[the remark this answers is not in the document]
edmorley: Sure - I've opened #312.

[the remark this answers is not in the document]
jjbustamante: Thank you!! @edmorley

PR #298, opened 2023-09-29
Draft: SBOM refactor

natalieparellano:


text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,260 @@
...
+
+During `rebase`, the lifecycle should verify attestations for the old application image, and remove any attestations where the scopes are no longer valid
+when signing the new image digest
+(e.g., if the scope `layer=sha256:$DIGEST` references a layer in the old run image that is no longer present, the attestation will be removed).
+
+It should optionally accept configuration for new attestations to describe the new run image (or whatever the platform wants to add).
+
+## Next builds
+
+During `analyze`, the lifecyle should verify attestations for the previous application image, and download the SBOM data as files in `<layers>/sbom/launch/`
```
natalieparellano, line 191: If there is an SBOM in the registry for the run image, we could do interesting things, like provide that to buildpacks (somehow, somewhere) so they could know more about what's available. But that should probably be out of scope for now.

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,260 @@
...
+    [[attestations.annotations]]
+    name = "io.buildpacks.buildpack.id"
+    value = "my-buildpack"
+    
+    [[attestations.annotations]]
+    name = "io.buildpacks.sbom.scope"
+    value = "all"
+```
+
+And the `signer` will turn this into an in-toto statement with `predicateType` `https://cyclonedx.org/bom`:
```
natalieparellano, line 160: We should talk more about how the attestation will be attached to the image (the referrers api with fallback as expected in the v1.1 specs but maybe we could give the operator more configuration knobs here).

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,260 @@
...
+   - a manifest "list" published to tag `sha256-<image digest>` where each list element should contain:
+     - `artifactType` corresponding to the `config.mediaType` in the manifest
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+
+Our current approach has a few drawbacks, namely:
+* It can make application images quite large
```
natalieparellano, line 49: Examples here would be nice

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,260 @@
...
+The lifecycle will ship with a new `signer` binary.
+In the past we've talked about making this totally separate from the lifecycle,
+but given the need to restore SBOM data and thus teach the lifecycle about attestations, there isn't much value there.
+We should make it possible to invoke the `signer` separate from other phases.
+
+Buildpacks should continue to output SBOM files as before. We won't need to bother buildpack authors with this change.
+
+The `exporter` will accept a new optional flag `-omit-sbom` that defaults to `false`.
+* If `-omit-sbom=false` the exporter will behave as today, and include the SBOM layer containing the `<layers>/sbom/launch/` directory in the application image.
+* If `-omit-sbom=true` the exporter will not create the SBOM layer, but `<layers>/sbom/launch/` will remain on the filesystem and be available to the `signer`.
```
natalieparellano, line 83: If `-daemon` is true we might log a warning or error

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,260 @@
...
+    value = "all"
+```
+
+And the `signer` will turn this into an in-toto statement with `predicateType` `https://cyclonedx.org/bom`:
+
+TODO
+
+### Example 3 - platform-provided SBOM file
+
+Platforms could inject TOML to describe the run image, e.g.:
```
natalieparellano, line 166: This could be concatenated to the configuration file output by the `exporter`

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,258 @@
...
+type = "application/vnd.cyclonedx+json"
+uri = "/platform/some-file.cdx.json"
+    
+    [[attestations.annotations]]
+    name = "io.buildpacks.sbom.scope"
+    value = "run" # maybe?
+```
+
+Potentially, the lifecycle could translate the special value `"run"` into a list of layer digests such as: `layer=sha256:$DIGEST-1,layer=sha256:$DIGEST-2`
+to save work for the platform to produce a more meaningful scope.
```
natalieparellano, line 177: Or, the lifecycle could read attestations for the run image directly. Or, perhaps this should also be out of scope.

RealHarshThakur:
Couple thoughts:
There's been recent work in Buildkit and Docker to provide SBOM and provenance as part of the image. I think there's a lot that can be potentially re-used or integrated directly. 

* Attestation storage: They've achieved it via storing all the attestations in a single manifest, which I thought was pretty neat compared to the "image tag of `sha256-hash` " way of doing things- https://docs.docker.com/build/attestations/attestation-storage 

* Other aspect that I think worth looking at is [scanners](https://github.com/docker/buildkit-syft-scanner) . This will change the model of "who generates the SBOM?" . Often times, I've found not all buildpacks provide SBOMs and there's no easy fallback mechanism. I think having such scanners might help to provide a fallback or even merge the SBOMs to enrich them. 

* Buildkit supports attestations natively on the newer versions. So that might be handy. https://github.com/moby/buildkit/blob/master/docs/attestations/slsa-provenance.md

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,496 @@
...
+The exporter will additionally generate configuration for the `signer` irrespective of the value of `-omit-sbom`.
+
+## Sign
+
+The `signer` will accept configuration that will allow it to sign images with `cosign`. 
+An example `cosign.toml` was provided in [PR 195](https://github.com/buildpacks/rfcs/pull/195), however there were concerns about encoding sensitive data (such as passwords) in files.
+We should explore secure ways of allowing the platform to inject this data (TODO).
+It's worth noting that `cosign` as of v2 prefers keyless signing.
+
+The `signer` can both sign the image and attach attestations to it. By default, we should do both - but we may want to make this configurable.
```
AidanDelaney, line 94: `signer` is then a registry operation?  This means that both `signer` and `exporter` now write to the registry?

AidanDelaney (commented):
In general, this looks good to me. I have one query as to whether we expect `signer` to write to a registry, or if it is generating an on-disk signature that `exporter` will export (I'm assuming the latter from the examples provided).

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,496 @@
...
+
+See [How it Works](#how-it-works).
+
+# How it Works
+[how-it-works]: #how-it-works
+
+## Setup
+
+The lifecycle will ship with a new `signer` binary.
+In the past we've talked about making this binary totally separate from the lifecycle,
```
natalieparellano, line 72: From 10/5 Working Group - cosign is nevertheless a significant dependency to take into the lifecycle. If cosign took a configuration file (https://github.com/sigstore/cosign/issues/1288 or similar) maybe we could just set that configuration up according to the [mapping](#mapping-layers-to-attestations) defined below. That would take care of the signing part.

For retrieving SBOMs from the previous image, we could:
* explore making verifying attestations the responsibility of the "preparer" binary, if we ever have one; the preparer could write the predicate data as json for the analyzer to find
* put this in the analyzer and hope the dependency footprint would be smaller if we're only verifying signatures (not creating them)

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,496 @@
...
+
+## Export
+
+The `exporter` will accept a new optional flag `-omit-sbom` that defaults to `false`.
+* If `-omit-sbom=false` the exporter will behave as today, and include the SBOM layer containing the `<layers>/sbom/launch/` directory in the application image.
+* If `-omit-sbom=true` the exporter will not create the SBOM layer, but `<layers>/sbom/launch/` will remain on the filesystem and be available to the `signer`.
+
+The exporter will additionally generate configuration for the `signer` irrespective of the value of `-omit-sbom`.
+
+## Sign
```
natalieparellano, line 87: From 10/5 Working Group - adding signatures could be a big jump for platforms that just want to get the SBOM out of the application image. We could provide an option to generate **unsigned** attestations as OCI artifacts in the registry which might ease the migration path for some folks.

text/0000-refactor-sbom.md
```diff
@@ -0,0 +1,496 @@
...
+## What is the impact of not doing this?
+
+SBOMs for CNB-built images feel clunky and outdated.
+
+# Prior Art
+[prior-art]: #prior-art
+
+TODO
+
+Discuss prior art, both the good and bad.
```
tomkennedy513, line 455: Maybe the current cosign implementation in kpack could be referenced here?

PR #299, opened 2023-10-06
Build Observability via OpenTelemetry Traces

joshwlewis:
This RFC introduces a solution to build observability via OpenTelemetry Traces and the OpenTelemetry File Export format.

[Readable](https://github.com/joshwlewis/rfcs/blob/build-observability/text/0000-build-observability.md)

### DIFF
--- a/text/0000-build-observability.md
+++ b/text/0000-build-observability.md
@@ -1,0 +1,411 @@
+# Meta
+[meta]: #meta
+- Name: Buildpack Observability
+- Start Date: 2022-10-05
+- Author(s): @joshwlewis
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes leveraging [OpenTelemetry](https://opentelemetry.io/) to
+grant platform operators and buildpack operators more insight into buildpack
+performance and behavior. This RFC describes new opt-in functionality
+for pack, the lifecycle, and the buildpack spec such that OpenTelemetry data may be
+exported to the build file system.
+
+# Definitions
+[definitions]: #definitions
+
+- [OpenTelemetry](https://opentelemetry.io/): A collection of APIs, SDKs, and tools that can be used it to instrument, generate, collect, and export telemetry data.
+- [Traces](https://opentelemetry.io/docs/concepts/signals/traces/): Telemetry
+  category that describes the path of software execution.
+
+
+# Motivation
+[motivation]: #motivation
+
+Buildpack authors and platform operators desire insight into usage, error
+scenarios, and performance of builds and buildpacks on their platform. The
+following questions are all important for these folks, but difficult to answer:
+
+- "Which buildpacks commonly fail to compile?"
+- "How often does a particular error scenario occur?"
+- "How long does each buildpack compile phase take?"
+- "How often is a certain buildpack used?"
+- "Which versions of Go are being installed?"
+- "How long does it take to download node_modules?"
+
+Instrumenting lifecycle and buildpacks with opt-in OpenTelemetry tracing will
+allow  platform operators to better understand performance and behavior of their
+builds and buildpacks and as a result, provide better service and build
+experiences.
+
+To protect privacy and prevent unnecessary collection of data, this
+functionality shall be optional and anonymous.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC aims to provide a solution for two types of OpenTelemetry traces:
+
+1) Lifecycle tracing: Buildpack-agnostic trace data like which buildpacks were
+available, which buildpacks were detected, how long the detect, build, or
+export phase took, and so on. This telemetry data may be exported by lifecycle.
+2) Buildpack tracing: Telemetry data specific to a buildpack like how long it
+took to download a language binary, which language version was selected, and so
+on. This telemetry data may be exported by buildpacks.
+
+Though the sources and contents of the telemetry data differ, both types may
+be emitted to the build file system in OpenTelemetry's [File Exporter
+Format](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/).
+
+In this solution, each lifecycle phase would write a `.jsonl` file with
+tracing data for that phase. For example, `lifecycle detector --telemetry`
+would write to `/layers/tracing/lifecycle/detect.jsonl`. Additionally each
+buildpack may also write tracing data to it's own `.jsonl` files (at
+`/layers/tracing/buildpacks/#{id}@#{version}-#{phase}.jsonl`).
+
+These `.jsonl` files may be read by platform operators for consumption,
+transformation, enrichment, and/or export to an OpenTelemetry backend. Given
+that builds may crash or fail at any point, these files must be written to
+often and regularly to prevent data loss.
+
+Platform operators will likely want to view or analyze this data. These
+telemetry files are in OTLP compatible format, so may be exported to one or
+more OpenTelemetry backends like Honeycomb, Prometheus, and [many
+others](https://opentelemetry.io/ecosystem/vendors/).
+
+Additionally, these traces may be correlated with traces in a platform
+operator's system via context propagation. The `CNB_OTEL_TRACEPARENT` may be
+provided by a platform to the build environment, such that generated traces
+inherit `trace-id` and `parent-id` from platform systems.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+### Lifecycle telemetry files
+
+If `lifecycle` is provided the telemetry opt-in flag (such as `--telemetry`),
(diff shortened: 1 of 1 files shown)


text/0000-build-observability.md
```diff
@@ -0,0 +1,358 @@
...
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes leveraging [OpenTelemetry](https://opentelemetry.io/) to
+grant platform operators and buildpack operators more insight into buildpack
+performance and behavior. This RFC describes new opt-in functionality
+for both pack and the buildpack spec such that OpenTelemetry data may be
```
natalieparellano, line 18:
```suggestion
for pack, the lifecycle, and the buildpack spec such that OpenTelemetry data may be
```

text/0000-build-observability.md
```diff
@@ -0,0 +1,358 @@
...
+took to download a language binary, which language version was selected, and so
+on. This telemetry data may be exported by buildpacks.
+
+Though the sources and contents of the telemetry data differ, both types may
+be emitted to the build file system in OpenTelemetry's [File Exporter
+Format](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/).
+
+In this solution, each lifecycle phase would write a `.jsonl` file with
+tracing data for that phase. For example, `lifecycle detector --telemetry`
+would write to `/cnb/telemetry/lifecycle-detect.jsonl`. Additionally each
```
natalieparellano, line 69: Right now we use the layers directory as the de-facto place for "output" (layers/report.toml, layers/sbom, etc.). Would it be easier to find a new place under layers?

That said adding yet another output directory makes me want to look at something like https://github.com/buildpacks/rfcs/pull/145 again. Not that we need to block this RFC on that, but maybe it's worth revisiting in tandem.

joshwlewis: Yeah, I thought about `/layers`. I didn't choose it because I don't think it's available during some of the earlier lifecycle phases (like `detect`)? Maybe I should confirm that - I know at the very least, the `/layers` subdirectories don't exist until later phases.

`/cnb/telemetry` is definitely a strawman - I'm not committed to it at all.

text/0000-build-observability.md
```diff
@@ -0,0 +1,358 @@
...
+are persisted in cases of failures to detect, failures to build, process
+terminations, or crashes. The `jsonl` format allows telemetry libraries to
+safely append additional json objects to the end of a telemetry file, so
+telemetry data can be flushed to the file frequently. Telemetry files should
+not be truncated or deleted so that telemetry processing by a platform can
+happen during or after a build. Telemetry files should not be included in the
+build result, as they are not relevant, and would likely negatively impact
+image size and reproduceability.
+
+### Access
```
jkutner, line 213: I'd like to see a couple statements in this section about _intent_. Mainly:

* Buildpacks should not use the telemetry files to pass materiel to another buildpack, or to implement any kind of contract between buildpacks.
*  A platform may access these files at any time during the build in order to evacuate them for safe keeping in the event of a build failure. 

or something like that?

text/0000-build-observability.md
```diff
@@ -0,0 +1,358 @@
...
+
+The telemetry files should be readable so that they may be analyzed by
+the user and/or platform. However, they should be write protected
```
natalieparellano, line 216: How will we accomplish this if buildpacks all run as the same user?

text/0000-build-observability.md
```diff
@@ -0,0 +1,358 @@
...
+This RFC outlines functionality that could be percieved as user tracking. To
+help remediate those concerns, these are some factors to remember about this
+design:
+
+1) This functionality is opt-in. `lifecycle` and `pack` will not emit telemetry
+  data unless the `--telemetry` flag is used.
+2) This functionality emits telemetry data only to the build file system. For
+  `pack` users, the telemetry files are stored in docker volumes on the local
+  machine. Neither `pack` nor `lifecycle` will "phone home" with telemety data.
+3) Neither `pack` nor `lifecycle` collect user-identifiable data (no emails,
```
natalieparellano, line 265: Should image/registry names be considered sensitive?

text/0000-build-observability.md
joshwlewis: Added an intent section here: https://github.com/buildpacks/rfcs/blob/6438aed9b16f226a6f54454d6a12dd0871f266d1/text/0000-build-observability.md?plain=1#L258. 

I covered the point about access in the consumption section: https://github.com/buildpacks/rfcs/blob/6438aed9b16f226a6f54454d6a12dd0871f266d1/text/0000-build-observability.md?plain=1#L234.

text/0000-build-observability.md
joshwlewis: During build, each buildpack runs synchronously, so we could prep the files/permissions prior to bin/build, so that each buildpack can only write to the expected files. 

During detect, many buildpacks will be running at the same time, and all under the same user. I can't think of a way to do block one buildpack from writing to another buildpack's detect telemetry.

To be fair, buildpacks can write to other buildpacks' `/layers` today. So maybe this section is a little overzealous. 

Maybe instead of blocking writes, we say something about the intent - buildpacks shall write only to their own telemetry files.

text/0000-build-observability.md
joshwlewis: I've updated the file locations to `/layers`.

text/0000-build-observability.md
joshwlewis: Hmm - tough question. Yeah, I suppose some folks would consider the resulting image name/registry sensitive. I'll update.

jkutner (approved):

jkutner:
@joshwlewis can you DCO those last commits (and maybe squash em)

text/0000-build-observability.md
```diff
@@ -0,0 +1,374 @@
...
+
+Though the sources and contents of the telemetry data differ, both types may
+be emitted to the build file system in OpenTelemetry's [File Exporter
+Format](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/).
+
+In this solution, each lifecycle phase would write a `.jsonl` file with
+tracing data for that phase. For example, `lifecycle detector --telemetry`
+would write to `/layers/lifecycle-detect.jsonl`. Additionally each
+buildpack may also write tracing data to it's own `.jsonl` files (at
+`/layers/#{buildpack-id}@#{buildpack-version}.jsonl`).
```
natalieparellano, line 71: Should this include the phase as well? E.g., `/layers/#{buildpack-id}@#{buildpack-version}-detect.jsonl`. And what about extensions?

natalieparellano: Bikeshedding:

```
<layers>/
└── tracing
    ├── buildpacks
    │   ├── other-id@other-version-detect.jsonl.   # this buildpack didn't detect
    │   ├── some-id@some-version-build.jsonl
    │   └── some-id@some-version-detect.jsonl
    └── extensions
        ├── some-id@some-version-detect.jsonl    # extensions and buildpacks can share the same ID
        └── some-id@some-version-generate.jsonl
```

This does add one more reserved buildpack ID, i.e., forbidden directory under `<layers>`. But, we already have `app`, `config`, `extended`, `generated`, and `sbom` ...is there harm in one more? (Really)

Eventually I would love to solve the reserved buildpack ID problem, maybe by moving at least `extended`, `generated`, and `sbom` under something like `<layers>/cnb`. But I don't think we need to do it right here and right now.

natalieparellano (approved):
I very much support this proposal. I have my quibbles about where the files should be output, but I think whatever we eventually settle on will be fine.

text/0000-build-observability.md
joshwlewis: > Should this include the phase as well? E.g., /layers/#{buildpack-id}@#{buildpack-version}-detect.jsonl. And what about extensions?

I think this might be useful for a platform operator -- their OTel collector could possibly enrich traces with attributes like `buildpack-id`, `buildpack-version`, `buildpack-phase`, `buildpack_or_extension`  based on the filename. Though, they would have to build that in to their collector. And in my mind, it's fairly likely that each buildpack would already include attributes like these in the tracing data they emit anyway.

That's about the only use-case I could see for this differentiation. So it's maybe not super useful in my mind, but it's low complexity for buildpack authors and telemetry consumers. It's probably worth the tradeoff.

joshwlewis: > This does add one more reserved buildpack ID, i.e., forbidden directory under <layers>.

I like the `/<layers>/tracing` idea. I can add that. 

Since there are a bunch of `lifecycle-*.jsonl` files, might as well put them in a directory too?

```
<layers>/
└── tracing
    ├── buildpacks
    │   ├── other-id@other-version-detect.jsonl
    │   ├── some-id@some-version-build.jsonl
    │   └── some-id@some-version-detect.jsonl
    ├── extensions
    │   ├── some-id@some-version-detect.jsonl
    │   └── some-id@some-version-generate.jsonl
    └── lifecycle
        ├── analyze.jsonl
        ├── build.jsonl
        ├── detect.jsonl
        ├── export.jsonl
        ├── extend.jsonl
        └── restore.jsonl
```

natalieparellano: Sounds good to me :)

jabrown85:
Ready to merge @hone !

PR #300, opened 2023-10-09
RFC: add opentelemetry support

qiaoleiatms:
Had a discussion here:
https://github.com/buildpacks/lifecycle/issues/1208

### DIFF
--- a/text/0000-opentelemetry-metrics.md
+++ b/text/0000-opentelemetry-metrics.md
@@ -1,0 +1,115 @@
+# Meta
+[meta]: #meta
+- Name: Enable OpenTelemetry Metrics
+- Start Date: 2023-10-09
+- Author(s): qiaoleiatms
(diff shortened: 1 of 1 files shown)


hone:
There seems to overlap between this and #299. Do you want to work with the author @joshwlewis ?

natalieparellano:
@qiaoleiatms  @joshwlewis any further thoughts here?

joshwlewis:
@qiaoleiatms I've updated #300. It does include some reasoning as to why I prefer it over this approach. I'd love to hear your thoughts.

It might be easy to chat in slack by the way. You can find me in the slack linked on [the community page](https://buildpacks.io/community/).

AidanDelaney:
I'm going to close this as we've made a lot of progress with #299.  Please re-open if we need to build on #299 and add additional metrics.

PR #301, opened 2023-10-09
Add RFC for extension layers

c0d1ngm0nk3y:
[readable](https://github.com/sap-contributions/buildpacks-rfcs/blob/extension-distroless/text/0000-extension-layer.md)

### DIFF
--- a/text/0000-extension-layer.md
+++ b/text/0000-extension-layer.md
@@ -13,7 +13,7 @@
 # Summary
 [summary]: #summary
 
-This RFC introduces support for Extension Layers to allow data transfer between the build environment and the Kaniko execution.
+This RFC introduces support for Extension configurable context to allow data transfer between the build environment and the Kaniko execution.
 
 # Motivation
 [motivation]: #motivation
--- a/text/0000-extension-layer.md
+++ b/text/0000-extension-layer.md
@@ -25,7 +25,7 @@ This would allow distroless run images to be extended.
 # What it is
 [what-it-is]: #what-it-is
 
-This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive layer. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific layers is ignored at build and launch time, it serves only the extension phase.
+This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive output directory. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific context is ignored at build and launch time, it serves only the extension phase.
 
 # How it Works
 [how-it-works]: #how-it-works
--- a/text/0000-extension-layer.md
+++ b/text/0000-extension-layer.md
@@ -30,14 +30,26 @@ This follows up on RFC-0105 and proposes that during the execution of the extens
 # How it Works
 [how-it-works]: #how-it-works
 
-- New root directory `/layers-ext` is introduced which contains extension layers.
-- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable directory `/layers-ext/<extension-id>` for each extension which passed detection.
-- The `/layers-ext/<extension-id>` is provided to the `./bin/generate` as `<output>` directory.
-- In addition to the files specified in [RFC#0105](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md), the extension may create the `<output>/context` folder with an arbitrary content.
-- If the folder `<output>/context` is present it will be set as Kaniko build context during the `extend` phase instead of the `<app>` directory.
-- If the folder `<output>/context` is not present, Kaniko build context defaults to the `<app>` folder.
+- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable layer `$CNB_GENERATED_DIR/<extension-id>` for each extension which passed detection.
+- The `$CNB_GENERATED_DIR/<extension-id>` is provided to the `./bin/generate` as `<output>` (`$CNB_OUTPUT_DIR`) directory.
+- In addition to the files specified in [RFC#0105](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md), the extension may create the following folders with an arbitrary content:
+
+    either:
+
+    - `<output>/context`
+
+    or the image-specific folders:
+
+    - `<output>/context.run`
+    - `<output>/context.build`
+  
+  If the `<output>/context` is provided together with any of the image-specific folders the detection phase must fail.
+- If the folder `<output>/context` is present it will be set as Kaniko build context during the `extend` phase of the build and run images.
+- If the folder `<output>/context.run` is present it will be set as Kaniko build context during the `extend` phase of the run image only.
+- If the folder `<output>/context.build` is present it will be set as Kaniko build context during the `extend` phase of the build image only.
+- If none of these folders is not present, Kaniko build context defaults to the `<app>` folder.
  
-The `/layers-ext` will not be included in the final image by the lifecycle.
+The `$CNB_GENERATED_DIR/<extension-id>` folders will not be included in the final image by the lifecycle.
 
 ### Example: Extend distroless run image with Debian packages.
 
@@ -48,10 +60,10 @@ This example extension would allow to install `tar` package on the run image wit
 ```bash
 #!/bin/sh
 
-mkdir -p ${CNB_OUTPUT_DIR}/context
+mkdir -p ${CNB_OUTPUT_DIR}/context.run
 
-cp ${CNB_EXTENSION_DIR}/bin/custom-installer ${CNB_OUTPUT_DIR}/context/
-curl -o ${CNB_OUTPUT_DIR}/context/tar.deb http://security.ubuntu.com/ubuntu/pool/main/t/tar/tar_1.34+dfsg-1ubuntu0.1.22.04.1_amd64.deb
+cp ${CNB_EXTENSION_DIR}/bin/custom-installer ${CNB_OUTPUT_DIR}/context.run/
+curl -o ${CNB_OUTPUT_DIR}/context.run/tar.deb http://security.ubuntu.com/ubuntu/pool/main/t/tar/tar_1.34+dfsg-1ubuntu0.1.22.04.1_amd64.deb
 
 cat >> "${CNB_OUTPUT_DIR}/run.Dockerfile" <<EOL
 ARG base_image
@@ -96,8 +108,9 @@ Discuss prior art, both the good and bad.
 
 This RFC requires changes to the layers metadata and the `extend` phase:
 
-- allow optional folder `<output>/context` with an arbitrary content to be provided by extension.
-- if the `<output>/context` is present, kaniko context should be set to this folder instead of the `<app>`.
+- the `$CNB_OUTPUT_DIR` must point to the `$CNB_GENERATED_DIR/<extension-id>` folder instead of a temporary directory.
+- allow optional folders `$CNB_GENERATED_DIR/<extension-id>/context`, `$CNB_GENERATED_DIR/<extension-id>/context.run` and `$CNB_GENERATED_DIR/<extension-id>/context.build` with an arbitrary content to be provided by extension.
+- if the context folders are present, kaniko context should be set to the corresponding folder instead of the `<app>` (following the rules defined in [#how-it-works](#how-it-works)).
 
 <!--
 ## Amended
--- a/text/0000-extension-layer.md
+++ b/text/0000-extension-layer.md
@@ -18,7 +18,7 @@ This RFC introduces support for Extension configurable context to allow data tra
 # Motivation
 [motivation]: #motivation
 
-This change allows extensions to possess their own layers to utilize during the generation/extend process. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
+This change allows extensions to create their own context for the extend phase during the generation phase. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
 
 This would allow distroless run images to be extended.
 
--- a/text/0000-extension-layer.md
+++ b/text/0000-extension-layer.md
@@ -25,26 +25,56 @@ This would allow distroless run images to be extended.
 # What it is
 [what-it-is]: #what-it-is
 
-This follows up on RFC-0105 and proposes that during the execution of the extension's `bin/generate`, an extension is allowed to write arbitrary data to its exclusive layer. This data then becomes accessible during the execution of the `extend` phase via Kaniko context. The content of these extension-specific layers is ignored at build and launch time, it serves only the extension phase.
+This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive layer. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific layers is ignored at build and launch time, it serves only the extension phase.
 
 # How it Works
 [how-it-works]: #how-it-works
 
-Before execution of the `bin/generate`, the lifecycle will create a distinct writable layer for each extension which passed detection. The extensions can then write to these layers, and the Kaniko context is set to the corresponding layer during the `extend` phase instead of the `<app>` directory. The Extension Layers will not be included in the final image by the lifecycle.
+- New root directory `/layers-ext` is introduced which contains extension layers.
+- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable layer `/layers-ext/<extension-id>` for each extension which passed detection.
+- The `/layers-ext/<extension-id>` is provided to the `./bin/generate` as `<output>` directory.
+- In addition to the files specified in [RFC#0105](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md), the extension may create the `<output>/context` folder with an arbitrary content.
+- If the folder `<output>/context` is present it will be set as Kaniko build context during the `extend` phase instead of the `<app>` directory.
+- If the folder `<output>/context` is not present, Kaniko build context defaults to the `<app>` folder.
+ 
+The `/layers-ext` will not be included in the final image by the lifecycle.
 
-The location of the Extension Layer will be provided to the `bin/generate` via additional environment variable `CNB_EXT_LAYER_DIR`.
+### Example: Extend distroless run image with Debian packages.
+
+This example extension would allow to install `tar` package on the run image without package manager (distroless image). The extension contains `./bin/generate` and `./bin/custom-installer` file, which installs `.deb` files.
+
+##### `./bin/generate`
+
+```bash
+#!/bin/sh
+
+mkdir -p ${CNB_OUTPUT_DIR}/context
+
+cp ${CNB_EXTENSION_DIR}/bin/custom-installer ${CNB_OUTPUT_DIR}/context/
+curl -o ${CNB_OUTPUT_DIR}/context/tar.deb http://security.ubuntu.com/ubuntu/pool/main/t/tar/tar_1.34+dfsg-1ubuntu0.1.22.04.1_amd64.deb
+
+cat >> "${CNB_OUTPUT_DIR}/run.Dockerfile" <<EOL
+ARG base_image
+FROM \${base_image}
+ARG build_id=0
+
+ADD custom-installer .
+ADD tar.deb .
+RUN ./custom-installer -p ./tar.deb
+EOL
+```
 
 # Migration
 [migration]: #migration
 
-Since we would change the Kaniko context, this would be breaking for existing extensions.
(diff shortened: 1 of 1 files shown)


loewenstein:
@natalieparellano how do we manage to get feedback? We don't seem to be able to request a review.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+[summary]: #summary
+
+This RFC introduces support for Extension Layers to allow data transfer between the build environment and the Kaniko execution.
+
+# Motivation
+[motivation]: #motivation
+
+This change allows extensions to possess their own layers to utilize during the generation/extend process. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
+
+This would allow distroless run images to be extended.
```
natalieparellano, line 23: IIUC this would allow an extension to use the package manager from the build image to download packages for the run image (by copying the package manager to the layers dir).

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+[what-it-is]: #what-it-is
+
+This follows up on RFC-0105 and proposes that during the execution of the extension's `bin/generate`, an extension is allowed to write arbitrary data to its exclusive layer. This data then becomes accessible during the execution of the `extend` phase via Kaniko context. The content of these extension-specific layers is ignored at build and launch time, it serves only the extension phase.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Before execution of the `bin/generate`, the lifecycle will create a distinct writable layer for each extension which passed detection. The extensions can then write to these layers, and the Kaniko context is set to the corresponding layer during the `extend` phase instead of the `<app>` directory. The Extension Layers will not be included in the final image by the lifecycle.
+
+The location of the Extension Layer will be provided to the `bin/generate` via additional environment variable `CNB_EXT_LAYER_DIR`.
```
natalieparellano, line 35: Where should the root directory reside within the build container? Thinking of something like `/layers-ext` for the root and `/layers-ext/<extension-id>` for each extension. As extensions and buildpacks share the same id, we would need a new root directory to avoid collisions. We could then simply not mount `/layers-ext` to the build or export containers to avoid polluting the application image.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the `bin/generate` be executed during the `extend` phase instead of the `detect` phase?
```
natalieparellano, line 62: I guess this would complicate sharing data across the build and run image, no? Because you would need `./bin/generate` to run on the build image and `run.Dockerfile` to be applied to the run image in order to support your use case.

text/0000-extension-layer.md
natalieparellano: Worth noting, today we provide `CNB_OUTPUT_DIR` which is where we expect to find build.Dockerfiles and run.Dockerfiles (currently, just a temporary directory). Is there any value in combining this directory with the "layers" directory?

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+- The Kaniko context might consist of two folders: `<app>` and the Extension Layer for better compatibility.
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+This RFC requires changes to the layers metadata and the `extend` phase:
+
+- layer metadata needs ti ubducate if a layer is a "extension layer"
```
natalieparellano, line 70: I think having a separate root directory for extension layers would mitigate this concern.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+[migration]: #migration
+
+Since we would change the Kaniko context, this would be breaking for existing extensions.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+The workspace would no longer be the Kaniko context. But the benefit of having it in the first place seems quite limited anyway.
```
natalieparellano, line 47: Perhaps we could denote the location of the app directory with a `CNB_APP_DIR` env var

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Migration
+[migration]: #migration
+
+Since we would change the Kaniko context, this would be breaking for existing extensions.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
```
natalieparellano, line 45: This might make build.Dockerfiles and run.Dockerfiles from extensions less portable - for example at the moment, it doesn't matter whether the build engine is `docker build` or `kaniko` but having Dockerfiles expect volumes to be present might change that.

I could be wrong, but there doesn't seem to be a transparent way to mount in volumes for `docker build` without adding Dockerfile instructions: https://docs.docker.com/build/guide/mounts/#add-a-cache-mount ...this would preclude us from doing something like [using the daemon for build image extension](https://github.com/buildpacks/pack/pull/1791) to improve performance.

text/0000-extension-layer.md
natalieparellano: Are there any implications for build image extension and run image extension running in parallel?

text/0000-extension-layer.md
```diff
@@ -0,0 +1,90 @@
...
+
+The workspace would no longer be the Kaniko context. But the benefit of having it in the first place seems quite limited anyway.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Allow multi-stage Dockerfiles
```
natalieparellano, line 52: This is perhaps orthogonal to this RFC but I do wonder if there is some value in having extensions leave some sort of breadcrumbs for buildpacks about what was added. Maybe we wouldn't make the content of extension "layers" available to buildpacks, but perhaps metadata or env var files?

natalieparellano (commented):
Thank you @c0d1ngm0nk3y for the RFC. I've added a few comments to try to understand it better.

text/0000-extension-layer.md
jabrown85: I would love to see n example extension laid out in the RFC

text/0000-extension-layer.md
jabrown85: I think we should consider combining these - `CNB_OUTPUT_DIR` could have `$CNB_OUTPUT_DIR/build-layer` or similar?

text/0000-extension-layer.md
jabrown85: Is the idea that by using the new ext-layer dir as the context we don't have to deal with volumes in the Dockerfiles?


```
FROM <blah>
# because we downloaded some-expected-file.sh into $CNB_OUTPUT/ext-layer
RUN some-expected-file.sh
```
If the extension needed something from the `/app` directory, would we expect extension authors to copy it to `/ext-layer` then?

Maybe `$CNB_OUTPUT/context` instead of `layer` would make it clearer?

text/0000-extension-layer.md
pbusko: For instance, if we try to extend a distroless image, there's no way to extend it meaningfully with additional packages/files by the means available in the run image itself (sometimes even `coreutils` are not present). The extension could populate the extension layer with necessary files/packages/binaries and manipulate them from the generated Dockerfile in the context of a run image.

jabrown85: Sounds great - lets get that example added into the RFC. An example extension + Dockerfile that makes use of the files in the context. It'll help communicate the idea to a broader group

text/0000-extension-layer.md
pbusko: > I think we should consider combining these - `CNB_OUTPUT_DIR` could have `$CNB_OUTPUT_DIR/build-layer` or similar?

This could be combined with the suggestion from @natalieparellano. We could have `/layers-ext` root directory and the `CNB_OUTPUT_DIR` would point to `/layers-ext/<extension-id>`. Lifecycle would expect run/build Dockerfiles in the root of this directory, as well as `build-layer` folder.

text/0000-extension-layer.md
pbusko: > Is the idea that by using the new ext-layer dir as the context we don't have to deal with volumes in the Dockerfiles?

Exactly.

> If the extension needed something from the /app directory, would we expect extension authors to copy it to /ext-layer then?

Since extensions are not allowed to write to the `/app` directory is there any valid reason to copy files from there? Given the fact that extensions are allowed to read, it should be perfectly fine to copy necessary stuff to the dedicated folder.

text/0000-extension-layer.md
pbusko: This RFC does not implies volumes usage, but rather build context:

- https://github.com/GoogleContainerTools/kaniko#kaniko-build-contexts
- https://docs.docker.com/build/building/context/

text/0000-extension-layer.md
c0d1ngm0nk3y: > This is perhaps orthogonal to this RFC but I do wonder if there is some value in having extensions leave some sort of breadcrumbs for buildpacks about what was added. Maybe we wouldn't make the content of extension "layers" available to buildpacks, but perhaps metadata or env var files?

Do you have a concrete use case in mind? I have a hard time to see where this might be useful.

text/0000-extension-layer.md
natalieparellano: Got it. In that case, if we want to make both the "extension layers" directory and the app directory available in the build context, these would need to be children of a parent directory that would be prepared by the platform, correct? Could you add some illustrative details to the RFC text?

pbusko:
@jabrown85 @natalieparellano the RFC has been updated with the proposed changes.

text/0000-extension-layer.md
natalieparellano: We can table this question for now. It might in fact be useful but is probably orthogonal to this RFC.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+- RFC Pull Request:
+- CNB Pull Request:
+- CNB Issue:
+- Related: [RFC#0105 Support Dockerfiles](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC introduces support for Extension Layers to allow data transfer between the build environment and the Kaniko execution.
```
natalieparellano, line 16:
```suggestion
This RFC introduces support for Extension configurable context to allow data transfer between the build environment and the Kaniko execution.
```

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This change allows extensions to possess their own layers to utilize during the generation/extend process. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
```
natalieparellano, line 21:
```suggestion
This change allows extensions to create their own context for the extend phase during the generation phase. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
```

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+
+This would allow distroless run images to be extended.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive layer. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific layers is ignored at build and launch time, it serves only the extension phase.
```
natalieparellano, line 28:
```suggestion
This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive output directory. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific context is ignored at build and launch time, it serves only the extension phase.
```

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+- New root directory `/layers-ext` is introduced which contains extension layers.
```
natalieparellano, line 33: Is there a better name than `/layers-ext`?

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable layer `/layers-ext/<extension-id>` for each extension which passed detection.
```
natalieparellano, line 34:
```suggestion
- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable directory `/layers-ext/<extension-id>` for each extension which passed detection.
```

natalieparellano (commented):
Thanks for the updates @c0d1ngm0nk3y. The proposed changes make sense and should be relatively easy to implement in the lifecycle. I wonder if there's a better term compared to "layers" so as not to create confusion with respect to buildpack layers. The purpose, lifecycle, etc of these context directories is very different.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+- The `/layers-ext/<extension-id>` is provided to the `./bin/generate` as `<output>` directory.
```
natalieparellano, line 35: Today the spec says "If an image extension is missing /bin/generate, the image extension root /generate directory MUST be treated as a pre-populated <output> directory." This would allow for pre-populated context directories as well. I wonder if there's a case for decoupling the pre-populated context directories from a missing /bin/generate.

pbusko: I would say that the pre-populated context directory as part of the `/generate` directory is a valid use-case. Since no `./bin/generate` executable will be called it would be basically the only way to provide a context.

text/0000-extension-layer.md
pbusko: Something like `/extensions-gen`?

natalieparellano: Or perhaps `/contexts`? With apologies for my bikeshedding... buildpacks generate "layers" and extensions generate "contexts" has a nice parity.

pbusko: > buildpacks generate "layers" and extensions generate "contexts" has a nice parity.

The most important piece of information produced by extensions are the Dockerfiles, so IMHO the folder should have somewhat more general name, which covers contexts, Dockerfiles and additional toml files with build arguments. However I do not see it as a deal breaker, so I'm perfectly fine with `/contexts`

loewenstein:
I wonder what the state of this is? @natalieparellano is there anything missing from your point of view? @pbusko are we on top of all provided review feedback?

text/0000-extension-layer.md
```diff
@@ -0,0 +1,118 @@
...
+[motivation]: #motivation
+
+This change allows extensions to create their own context for the extend phase during the generation phase. Additionally, it ensures that extension output does not inadvertently interfere with other extension or buildpack layers during the build, and it does not unintentionally become part of the final application image.
+
+This would allow distroless run images to be extended.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive output directory. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific context is ignored at build and launch time, it serves only the extension phase.
```
natalieparellano, line 28: Do we need a way to provide different contexts when extending the build or the run image? What about `/context.build` and `/context.run` (if they need to be different) and just `/context` to provide the same for both?

text/0000-extension-layer.md
natalieparellano: > the folder should have somewhat more general name, which covers contexts, Dockerfiles and additional toml files with build arguments

This is a fair point. And in fact, we already have `CNB_GENERATED_DIR` which defaults to `<layers>/generated`. Perhaps we can simply copy any output `./context` directories to `<layers>/generated/<extension-id>/<build|run>/context`.

Currently, the lifecycle creates a temp directory for extension output. But perhaps it would be more efficient to provide `<layers>/generated/<extension-id>` as `CNB_OUTPUT_DIR`. This would require some small changes to the Buildpack API for extensions.

natalieparellano:
> I wonder what the state of this is? @natalieparellano is there anything missing from your point of view? @pbusko are we on top of all provided review feedback?

I left two new comments ([here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400706357) and [here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400695401)) but otherwise I think this is ready. If we don't evolve it any further after this week we can put it up for votes.

pbusko:
> I left two new comments ([here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400706357) and [here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400695401)) but otherwise I think this is ready. If we don't evolve it any further after this week we can put it up for votes.

The comments have been addressed

loewenstein:
> > I wonder what the state of this is? @natalieparellano is there anything missing from your point of view? @pbusko are we on top of all provided review feedback?
> 
> I left two new comments ([here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400706357) and [here](https://github.com/buildpacks/rfcs/pull/301#discussion_r1400695401)) but otherwise I think this is ready. If we don't evolve it any further after this week we can put it up for votes.

@natalieparellano are you putting up for vote?

loewenstein:
@jabrown85 could you help bringing this to a vote and getting it merged. That would be great. I hesitate to immediately start working on spec and implementation pull requests - inviting people to vote might lead to more review comments that lead to changes, but I would really like to see https://github.com/buildpacks/spec/issues/378 closed and [0.13](https://github.com/buildpacks/spec/milestone/27) released as soon as possible.

text/0000-extension-layer.md
```diff
@@ -0,0 +1,128 @@
...
+- The `$CNB_GENERATED_DIR/<extension-id>` is provided to the `./bin/generate` as `<output>` (`$CNB_OUTPUT_DIR`) directory.
+- In addition to the files specified in [RFC#0105](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md), the extension may create the following folders with an arbitrary content:
+    - `<output>/context`
+
+    or the image-specific folders:
+
+    - `<output>/context.run`
+    - `<output>/context.build`
+  
+  If `<output>/context` is provided together any of the image-specific folders the phase must fail.
```
natalieparellano, line 43: Could you clarify this statement? I'm not quite sure what it means

text/0000-extension-layer.md
```diff
@@ -0,0 +1,128 @@
...
+# What it is
+[what-it-is]: #what-it-is
+
+This follows up on RFC-0105 and proposes that during the execution of the extension's `./bin/generate`, an extension is allowed to write arbitrary data to the `context` folder within its exclusive output directory. This data then becomes accessible during the execution of the `extend` phase via Kaniko build context. The content of these extension-specific context is ignored at build and launch time, it serves only the extension phase.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+- Before execution of the `./bin/generate`, the lifecycle will create a distinct writable layer `$CNB_GENERATED_DIR/<extension-id>` for each extension which passed detection.
+- The `$CNB_GENERATED_DIR/<extension-id>` is provided to the `./bin/generate` as `<output>` (`$CNB_OUTPUT_DIR`) directory.
```
natalieparellano, line 34: Currently the extender is looking for Dockerfiles in `<generated>/run/<image extension ID>/Dockerfile` and `<generated>/build/<image extension ID>/Dockerfile` (where `<generated>` defaults to `<layers>/generated`). Should we change the Platform API so that the extender looks for `<generated>/<image extension ID>/run.Dockerfile`, or should the lifecycle re-arrange the files after the generate phase? The former might be simpler in the long run, but platforms such as `pack` look for the presence of files in `<generated>/build` and `<generated>/run` to infer that extension is needed.

text/0000-extension-layer.md
loewenstein, line 43: A missing "with" is the only explanation I could come up with.

```suggestion
  If `<output>/context` is provided together with any of the image-specific folders the phase must fail.
```

text/0000-extension-layer.md
pbusko: I would go with writing to `<generated>/<image extension ID>/run.Dockerfile`  and/or `<generated>/<image extension ID>/build.Dockerfile`, as suggested in this RFC. Platforms such as `pack` might have a fallback mechanism, instead of adding more complex logic on the lifecycle side.

loewenstein:
@natalieparellano @jabrown85 anything blocking a vote, merge and implementation?

text/0000-extension-layer.md
natalieparellano: @pbusko I'm in favor of removing this complexity from the lifecycle. Could we add an `extend` boolean to the `[build-image]` table in analyzed.toml, to match `[run-image]`:

```
[run-image]
  image = "<image>"
  reference = "<image reference>"
  extend = false

[build-image]
  reference = "<image reference>"
  extend = false
```

This will give platforms a simple way to check if extension is needed.

natalieparellano:
@jabrown85 could you give this one a read?

PR #302, opened 2023-10-17
Create 0000-creator-skip-sbom.md

kritkasahni-google:
@natalieparellano fyi ^^

### DIFF
--- a/text/0000-creator-skip-sbom.md
+++ b/text/0000-creator-skip-sbom.md
@@ -1,0 +1,131 @@
+# Meta
+[meta]: #meta
+- Name: Enable CNB_SKIP_SBOM IN /cnb/lifecycle/creator
+- Start Date: (fill in today's date: 2023-10-17)
+- Author(s): kritkasahni-google
+- Status: Draft
+- RFC Pull Request: 
+- CNB Pull Request: 
+- CNB Issue: 
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


text/0000-creator-skip-sbom.md
```diff
@@ -0,0 +1,131 @@
...
+
+Platforms that execute lifecycle today via /cnb/lifecycle/creator are unable to skip restoring SBOM layer from previous app image unless they skip reusing previous app image entirely.
+
+# Prior Art
+[prior-art]: #prior-art
+
+We already support enabling CNB_SKIP_LAYERS in /cnb/lifecycle/analyzer and /cnb/lifecycle/restorer, and CNB_SKIP_RESTORE in /cnb/lifecycle/creator.
+* CNB_SKIP_LAYERS in /cnb/lifecycle/analyzer to skip restoring SBOM from previous app image.
+* CNB_SKIP_LAYERS in /cnb/lifecycle/restorer to skip reusing previous app image layers entirely.
+* CNB_SKIP_RESTORE in /cnb/lifecycle/creator to skips restoring SBOM plus all other layers entirely from previous app image.
```
natalieparellano, line 95: In the future it might be clearer if there were only two inputs - `CNB_SKIP_LAYERS` for buildpack layer directories, and `CNB_SKIP_SBOM` for the SBOM layer. We could deprecate `CNB_SKIP_LAYERS` in the analyzer and `CNB_SKIP_RESTORE` in the creator to keep everything consistent. But, that may be more effort than it's worth.

jabrown85 (approved):

text/0000-creator-skip-sbom.md
jabrown85: Yeah - I think `CNB_SKIP_RESTORE` could be great. I don't want to go wild with options if we think they would likely both be set the same by platforms.

PR #304, opened 2023-11-23
Fix broken link

enrichman:
A link to the `lifecycle descriptor` RFC was using an old branch.

Fixed.


natalieparellano (approved):

natalieparellano:
Aye, I think we need to fix the DCO 😞 https://github.com/buildpacks/rfcs/pull/304/checks?check_run_id=28521018372

enrichman:
Thanks @natalieparellano, not sure which was the problem but I had to resync everything. Probably the DCO was added later and I had very old commits in my fork without the Sign-off-by!

PR #308, opened 2024-02-18
Initial draft, up for feedback

rashadism:
A basic implementation on `pack detect`.
Looking forward on getting feedback and improving on this and coming up with the best version of this, and implenenting this.

### DIFF
--- a/text/0000-my-feature.md
+++ b/text/0000-my-feature.md
@@ -1,0 +1,113 @@
+# Meta
+[meta]: #meta
+- Name: Implementing pack detect command
+- Start Date: 2024-02-15
+- Author(s): @rashadism
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The `pack detect` command is introduced to the Cloud Native Buildpacks ecosystem, providing a way to run only the detect phase of the buildpack lifecycle. This feature enhances the developer experience by allowing them to quickly determine which buildpacks are relevant for their application without progressing through the entire build process. This was partially discussed in [issue #681](https://github.com/buildpacks/pack/issues/681) but the issue was about implementing a `dry-run` flag. With further discussion with @jjbustamante, decided to go forward with this as a new pack pack command rather than a flag.
+
+# Definitions
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
+
+# Motivation
+[motivation]: #motivation
+
+- Simplify and streamline the build process by providing a targeted command for buildpack detection.
+- Reduce build times by skipping unnecessary phases of the buildpack lifecycle.
+- Enable developers to quickly identify which buildpacks are applicable to their application without waiting for the entire build process to complete, or having to `Ctrl+C` midway through.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This provides a high level overview of the feature.
+
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Ideally, the user should run something like `pack detect --path ./path/to/project --builder builder:name` and it should run only the detect binary in the lifecycle and output the logs / output of it. Preferrably, can also output the `group.toml` to a directory specified with `--detect-output-dir`.
+
+The following flags should be supported and they will work more or less like `pack build`.
+
+|Short|Long|type|
+|----|-------------|----|
+|-B |--builder|string|
+|-b |--buildpack|strings|
+|-r |--buildpack-registry|string|
+|-e|--env|stringArray|
+||--env-file|stringArray|
+||--extension|strings|
+|-h|--help|
+|-p|--path|string|
+||--post-buildpack|stringArray|
+||--pre-buildpack|stringArray|
+||--detect-output-dir|string|
+
+# Migration
+[migration]: #migration
+
+This feature does not introduce any breaks to public APIs or compatibility. It provides additional functionality within the existing Cloud Native Buildpacks CLI tooling, enhancing the developer experience without requiring changes to existing workflows or configurations.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+Initially thought of implementing this through something like `pack build --detect`. But after further discussion with @jjbustamante and for the following reasons, decided to do this functionality to a new command.
+- The main use case of `pack build` is to create OCI images, and detect is just a binary in the lifecycle, so it doesn't make much sense to include it in there.
+- To avoid making the mostly used `pack build` command overly complicated.
+
+Also, instead of `pack detect`, something like `pack execute --phase detect` can also be done, where the user can select exclusively what phase they need to run. Can start by implementing just the `detect` phase.
+
+# Prior Art
+[prior-art]: #prior-art
+
+This has been discussed in Issue #681 before, and looked like it was a long awaited feature and currently a few workarounds are being used to get this functionality.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+Fill after initial discussion
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+Since this is a new command, the functionality of this command will have to be amended to the spec / docs.
+
+# History
+[history]: #history
+
+<!--
+## Amended
+### Meta
+[meta-1]: #meta-1
+- Name: (fill in the amendment name: Variable Rename)
+- Start Date: (fill in today's date: YYYY-MM-DD)
+- Author(s): (Github usernames)
+- Amendment Pull Request: (leave blank)
+
+### Summary
+
+A brief description of the changes.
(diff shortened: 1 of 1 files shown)


text/0000-my-feature.md
```diff
@@ -0,0 +1,113 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The `pack detect` command is introduced to the Cloud Native Buildpacks ecosystem, providing a way to run only the detect phase of the buildpack lifecycle. This feature enhances the developer experience by allowing them to quickly determine which buildpacks are relevant for their application without progressing through the entire build process. This was partially discussed in [issue #681](https://github.com/buildpacks/pack/issues/681) but the issue was about implementing a `dry-run` flag. With further discussion with @jjbustamante, decided to go forward with this as a new pack pack command rather than a flag.
```
natalieparellano, line 15: Worth noting, though we care about the outcome of `detect` we'll also need to run `analyze` to get information about the run image that MAY impact the outcome of detect via CNB_TARGET_* env vars.

text/0000-my-feature.md
```diff
@@ -0,0 +1,113 @@
...
+- If applicable, describe the differences between teaching this to existing users and new users.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Ideally, the user should run something like `pack detect --path ./path/to/project --builder builder:name` and it should run only the detect binary in the lifecycle and output the logs / output of it. Preferrably, can also output the `group.toml` to a directory specified with `--detect-output-dir`.
+
+The following flags should be supported and they will work more or less like `pack build`.
+
+|Short|Long|type|
```
natalieparellano, line 47: Probably also needed:

```
  -d, --descriptor string            Path to the project descriptor file
      --docker-host string           Address to docker daemon that will be exposed to the build container.
                                     If not set (or set to empty string) the standard socket location will be used.
                                     Special value 'inherit' may be used in which case DOCKER_HOST environment variable will be used.
                                     This option may set DOCKER_HOST environment variable for the build container if needed.
      --gid int                      Override GID of user's group in the stack's build and run images. The provided value must be a positive number
      --lifecycle-image string       Custom lifecycle image to use for analysis, restore, and export when builder is untrusted.
      --network string               Connect detect and build containers to network
      --pull-policy string           Pull policy to use. Accepted values are always, never, and if-not-present. (default "always")
      --run-image string             Run image (defaults to default stack's run image)
      --uid int                      Override UID of user in the stack's build and run images. The provided value must be a positive number
      --workspace string             Location at which to mount the app dir in the build image
```

* `descriptor` controls things like environment and buildpacks to use
* `docker-host` for accessing the run image
* `lifecycle-image` we need to run the `analyze` step
* `gid` and `uid` could maybe affect how buildpack `./bin/detect` behaves
* `network` and `workspace` for configuring the container environment where buildpacks' `./bin/detect` are run
* `pull-policy` and `run-image` for influencing the run image that actually gets used

natalieparellano (commented):
Thanks for kicking this off @rashadism! I would love to see this happen :)

rashadism:
Thank you @natalieparellano , will update on the stuff you pointed out.

rashadism:
I opened a corresponding [draft PR ](https://github.com/buildpacks/pack/pull/2132). @natalieparellano can u have a look

natalieparellano:
@jjbustamante is going to steward this one :)

rashadism:
> @jjbustamante is going to steward this one :)

nice, looking forward to it (:

text/0000-my-feature.md
```diff
@@ -0,0 +1,123 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The `pack detect` command is introduced to the Cloud Native Buildpacks ecosystem, providing a way to run only the detect phase of the buildpack lifecycle. This feature enhances the developer experience by allowing them to quickly determine which buildpacks are relevant for their application without progressing through the entire build process. This was partially discussed in [issue #681](https://github.com/buildpacks/pack/issues/681) but the issue was about implementing a `dry-run` flag. With further discussion with @jjbustamante, decided to go forward with this as a new pack pack command rather than a flag.
```
dmikusa, line 15: Is the `pack pack` part expected or a typo?

text/0000-my-feature.md
```diff
@@ -0,0 +1,123 @@
...
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
+
+# Motivation
+[motivation]: #motivation
+
+- Simplify and streamline the build process by providing a targeted command for buildpack detection.
+- Reduce build times by skipping unnecessary phases of the buildpack lifecycle.
+- Enable developers to quickly identify which buildpacks are applicable to their application without waiting for the entire build process to complete, or having to `Ctrl+C` midway through.
```
dmikusa, line 27:
```suggestion
- Enable developers to quickly identify which buildpacks are applicable to their application without waiting for the entire build process to complete, or having to `Ctrl+C` midway through.
- Lighter-weight integration testing of the build plan.
```

text/0000-my-feature.md
```diff
@@ -0,0 +1,123 @@
...
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+Initially thought of implementing this through something like `pack build --detect`. But after further discussion with @jjbustamante and for the following reasons, decided to do this functionality to a new command.
+- The main use case of `pack build` is to create OCI images, and detect is just a binary in the lifecycle, so it doesn't make much sense to include it in there.
+- To avoid making the mostly used `pack build` command overly complicated.
+
+Also, instead of `pack detect`, something like `pack execute --phase detect` can also be done, where the user can select exclusively what phase they need to run. Can start by implementing just the `detect` phase.
```
dmikusa, line 88: This might be worth doing. It could be nice for integration testing purposes to be able to run just the build or just the generate phases of a buildpack. Not that we'd include those into this RFC, but this would leave the CLI open for that possibility.

When it comes to integration testing buildpacks with complicated build plans, right now, you have three choices:

1. Don't integration test them.
2. Supply all the buildpacks required to run the buildpack under test. That way all the set up and requirements for this buildpack are complete.

If we could run build or generate specifically, then that would open up the ability to mock out the requirements and run more focused, faster integration tests.

Anyway, that is all just to say I would be in support of this CLI invocation as it leaves open the possibility of running other phases directly.

text/0000-my-feature.md
```diff
@@ -0,0 +1,123 @@
...
+- Define any new terminology.
+- Define the target persona: buildpack author, buildpack user, platform operator, platform implementor, and/or project contributor.
+- Explaining the feature largely in terms of examples.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing users and new users.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Ideally, the user should run something like `pack detect --path ./path/to/project --builder builder:name` and it should run the analyze binary, followed by the detect binary in the lifecycle and output the logs / output of it. Preferrably, can also output the `group.toml` to a directory specified with `--detect-output-dir`. The reason to run the analyze binary is to get information about the run image that may impact the outcome of detect via CNB_TARGET_* environment variables.
```
dmikusa, line 43: Will this go so far as to select a build plan? or will this just output all of the possibilities?

Personally, I'm hoping for the former, that it will attempt to select a plan as well.

text/0000-my-feature.md
```diff
@@ -0,0 +1,123 @@
...
+||--lifecycle-image|string|
+||--network|string|
+|-p|--path|string|
+||--post-buildpack|stringArray|
+||--pre-buildpack|stringArray|
+||--pull-policy|string|
+||--run-image|string|
+||--uid|int|
+||--workspace|string|
+
```
dmikusa, line 69: What would be the plan for the exit code of this command?

I think we should specify this because it's likely that we'll build tooling around this to run tests. We want the exit code to signal errors and for it to be stable.

text/0000-my-feature.md
rashadism: Yes, it looks like the better path to take

text/0000-my-feature.md
rashadism: This would select the build plan, and either write it to disk or pretty print it in the terminal

text/0000-my-feature.md
rashadism: If I understood your question correctly,
Ideally it should be 0 if a build plan selection is successful. In the case of failure, we can come up with some common cases and define it accordingly.

dmikusa: Ok, yes. 0 for success and 1 for failure probably covers most of what is necessary. I'm not sure we need to define a bunch of non-zero exit codes. It does make me wonder what the lifecycle returns 🤔  Maybe we should just make sure the lifecycle exit code is returned from this command? I suspect we'd get good results that way and then this command would be inline with the platform spec.

jjbustamante: I think we also need to add some examples executions of the command with the expected output, that helps a little bit to understand better what is going to do and if we need to pay attention to some details.

text/0000-my-feature.md
rashadism: It was a typo, fixed

natalieparellano:
I think this one might be ready for another round of reviews.

natalieparellano (approved):
This looks good to me. I think we need to fix DCO :)

natalieparellano:
Maintainers, please review! I'll move this one to "voting" on Thursday.

dmikusa (approved):

jabrown85:
@rashadism can I get your acceptance on the DCO? I can mark it passed and merge if you'll reply back with a 👍

jjbustamante (approved):

PR #309, opened 2024-03-06
Add additional metadata to app images

dmikusa:
RFC that proposes to add additional metadata to app container images: ports and health check information.

[Readable](https://github.com/dmikusa/rfcs/blob/app-metadata/text/0000-additional-app-metadata.md)


text/0000-additional-app-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+```
+{
+    ...
+    "Config": {
+        ...
+        "ExposedPorts": {
+            "8080/tcp": {},
+            "9000/udp": {}
+        },
+        "Healthcheck": {
```
jjbustamante, line 89: I am curious about how the OCI image spec deals with [Healthchecj](https://github.com/opencontainers/image-spec/blob/v1.1.0/config.md#properties)

> Healthcheck object, OPTIONAL
> This property is reserved for use, to [maintain compatibility](https://github.com/opencontainers/image-> spec/blob/v1.1.0/media-types.md#compatibility-matrix).

Does **reserved for use** mean, we can use it, just to be compatible with Docker format, but it could change in the future to be in some other place in OCI image spec?

dmikusa: To me, it sounds like they are reserving it to prevent conflicts. It's not part of the OCI spec, but the OCI spec authors seem concerned enough with compatibility that they don't want to overlap with certain popular things outside the spec.

My read is that this is OK to use if the intent is for compatibility with Docker. If we were introducing our own format for the health check data or something that could conflict with Docker, then it would not be OK.

text/0000-additional-app-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+    ...
+}
+```
+
+The `Test` section is an array of the command to run. If `use-shell` is `true`, then the first item in the array is set to `CMD-SHELL`. If `use-shell` is `false` then the first item in the array is set to `CMD`. In addition, values for interval, timeout, start period, and start interval should be converted to nano seconds. This is all to be consistent with the way Docker embeds metadata from a `Dockerfile`, which is a major goal of this proposal.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None beyond the time to implement.
```
natalieparellano, line 110: Is there any concern that this could cause confusion when multiple process types are defined? Do we need to warn / alert the user somehow about this?

That said I think this should be fine to enable. Buildpack authors don't have to support it if they don't want to.

dmikusa: The ports/health check information is specified by the buildpack per process type, so if I have `web` and `scheduler` process types, you can define different ports/health check info for each one. Unfortunately, there's not parity for that in the container image metadata. There's only the capacity to to put metadata for one process type in the container's ports/health check info so we have to pick one of them. 

I think the default process type set by buildpacks seems most reasonable and least likely to be surprising (i.e. best choice if we put anything in this container metadata) because if you're going to `docker run` the image then that's what is run. Given that, I think it stands to reason that the metadata in the container image would reference the thing that is run in this manner.

I'm not sure what opportunities we have to insert some warnings/helpful logging. Where I see this being potentially confusing is that when you run the non-default process type, (I assume, I haven't checked) Docker is still going to see this metadata for the default process type and attempt to use it as ports/health check info (unless the user overrides it). That might be confusing if you don't remember (or know) that the ports/health check metadata is on the container image. Especially the health check. The container would start and could potentially fail if the health check is different between the two process types.

I'm not totally sure what we can do about it, especially at this level. Maybe `pack` could have an opt-out flag for including this metadata (although buildpacks could do that too)? Buildpacks could also log at build time if this metadata is being set. Ideally, there would be some kind of reminder that it's set at runtime but I don't know that is possible and it would be noise a lot of the time, like you'd only want a warning when something other than the default process type is run & there is metadata set on the container image.

dmikusa:
Curious if there are any more thoughts/feedback on this one.

I'd also be interested in helping with the implementation of this. Not my normal area, but it seems like a good one to branch out if I can get some help/points to get started.

samj1912:
I think the general proposal and implementation makes sense to me. I wonder if ports and healthchecks should be scopes inside processes though.

dmikusa:
Sorry, not sure I follow you @samj1912. Do you have an example of what you're thinking?

text/0000-additional-app-metadata.md
AidanDelaney: Should we provide a `healthcheck` binary, similar to `launcher`?  Something that executes the correct heathcheck for the running process but allows `HEALTHCHECK` to be the `healthcheck` for the running process?

text/0000-additional-app-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+[summary]: #summary
+
+When using a `Dockerfile`, it is possible to set additional metadata items like exposed ports and health check information. This metadata can be used by external systems, like container orchestrators, to automatically configure and deploy application images. It is not presently possible to set this metadata using buildpacks.
+
+# Definitions
+[definitions]: #definitions
+
+Dockerfiles settings for the proposed metadata:
+
+- Ports can be set with [`EXPOSE`](https://docs.docker.com/reference/dockerfile/#expose)
```
AidanDelaney, line 22: You motivate supporting `EXPOSE` by talking about orchestration systems that will use the information in  `EXPOSE`.  Could you provide a concrete example?  I'm wondering if `HEALTHCHECK` can be supported without needing to also support `EXPOSE`.

dmikusa: I don't use it myself, but someone I was talking (the spark for this RFC), made it seems like Google Cloud Run does things that way. Docker/Docker Compose also do, to a lesser extent. If you export that metadata, I believe you can `docker run -P` and it just exposes the default ports.

text/0000-additional-app-metadata.md
dmikusa: I like that idea. Is that possible? How would it know the correct process type to run? 

If you put `/cnb/healthcheck` into `HEALTHCHECK`, when the health check is invoked `/cnb/healthcheck` would need to figure out the process type. It could default to the default process type, but I'm not sure how it would know if a different process type was being used. Is that metadata available somewhere?

text/0000-additional-app-metadata.md
durcon: Another example is Prometheus' or Grafana Agent's Docker Service Discovery. The exposed Port will be used to scrape metrics. If you don't expose (or map) your port, the service discovery doesn't scrape the container.

text/0000-additional-app-metadata.md
dmikusa: > Should we provide a healthcheck binary, similar to launcher? Something that executes the correct heathcheck for the running process but allows HEALTHCHECK to be the healthcheck for the running process?

This would actually be really helpful, **IF** we can make this process skip invoking the `exec.d` binaries. It should source the environment, but not run exec.d binaries. This is because some exec.d binaries have side effects when they run, so it's not really safe to execute them *every* time we run the health check, which could be multiple times a minute.

We actually learned this the hard way with the Paketo Health Checker buildpack :( 

Having said that, it also seems like a tremendous amount of additional work, although please correct me if that's an invalid assumption, and I wonder if we couldn't just do something in the launcher that would allow us to launch the health check binary like it normally does, but skip the exec.d binaries.

> If you put /cnb/healthcheck into HEALTHCHECK, when the health check is invoked /cnb/healthcheck would need to figure out the process type. It could default to the default process type, but I'm not sure how it would know if a different process type was being used. Is that metadata available somewhere?

I was thinking about this more and it seems like we'd know this if it were implemented in the launcher, since the launcher knows what process to run. So maybe that's not really a big deal.

@natalieparellano what do you think about the above two points?

text/0000-additional-app-metadata.md
dmikusa: Thanks @durcon, I added that to the motivations list.

tomkennedy513:
I really like this!

PR #311, opened 2024-04-11
Remove Windows Containers support

AidanDelaney:
Remove support for Windows Containers in CNB tooling and documentation, but retain support in the Buildpacks spec.

[readable](https://github.com/buildpacks/rfcs/blob/ff8c235b14be3afa3c3a3215d882b1428d47fbcf/text/0000-remove-windows-containers-support.md)

### DIFF
--- a/text/0000-remove-windows-containers-support.md
+++ b/text/0000-remove-windows-containers-support.md
@@ -1,0 +1,97 @@
+# Meta
+[meta]: #meta
+- Name: Remove Windows Containers Support
+- Start Date: 2024-04-11
+- Author(s): aidan.delaney
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: "N/A"
+
+# Summary
+[summary]: #summary
+
+Retain Windows Containers support in the buildpacks specification, but remove Windows Containers support from `lifecycle`, `pack` CLI, and documentation.
+
+# Definitions
+[definitions]: #definitions
+
+Make a list of the definitions that may be useful for those reviewing. Include phrases and words that buildpack authors or other interested parties may not be familiar with.
+
+* [Windows Containers](https://learn.microsoft.com/en-us/virtualization/windowscontainers/about/): Containers derived from Windows Server Containers.  For example: `mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2022`
+
+# Motivation
+[motivation]: #motivation
+
+- There are no known users of Windows Container support within the buildpacks ecosystem.
+- Supporting Windows Containers is straightforward at a specification level.  However, the CNB project has neither the time nor the expertise to continue to support Windows Containers in our tooling.
+- Removing Windows Container support from CNB tooling will speed up development, simplify testing and impact no existing users.
+- Establish platform parity between `pack` and `kpack`.  `kpack` has never supported building Windows Containers.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- We want to remove Windows Containers related documentation, but clearly state that the specification is platform neutral.
+- Remove Windows Containers support from `pack`, but retain the Windows platform release of the `pack` CLI.
+    * this ensures `pack` users can still invoke `pack` on a Command or Power Shell console
+- Remove Windows Containers support from `lifecycle`.  Lifecycle can remove the Windows platform release as it will then be guaranteed to run on a Linux-based container.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+
+# Migration
+[migration]: #migration
+
+Teams that require Windows Container support will have to
+
+* continue to use the last release of `pack` that supports Windows Containers,
+* fork `pack` and `lifecycle` to maintain Windows Containers support, or
+* adopt alternative tooling that supports Windows Containers.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Windows Containers support is well tested within `pack`.  Removing support for Windows Containers removes a unique selling point of the `pack` CLI and `lifecycle`
+
+# Alternatives
+[alternatives]: #alternatives
+
+- We have discussed platform support with `pack` users.  There are no `pack` users who want to maintain Windows Container support.  Should engineering effort become available to help with development and maintenance, then this proposal can be dropped.
+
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

text/0000-remove-windows-containers-support.md
```diff
@@ -0,0 +1,97 @@
...
+
+* [Windows Containers](https://learn.microsoft.com/en-us/virtualization/windowscontainers/about/): Containers derived from Windows Server Containers.  For example: `mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2022`
+
+# Motivation
+[motivation]: #motivation
+
+- There are no known users of Windows Container support within the buildpacks ecosystem.
+- Supporting Windows Containers is straightforward at a specification level.  However, the CNB project has neither the time nor the expertise to continue to support Windows Containers in our tooling.
+- Removing Windows Container support from CNB tooling will speed up development, simplify testing and impact no existing users.
+- Establish platform parity between `pack` and `kpack`.  `kpack` has never supported building Windows Containers.
```
tomkennedy513, line 30: kpack actually does support building windows containers (https://github.com/buildpacks-community/kpack/pull/546), albeit in a slightly more limited capacity. There is an open discussion on the kpack side to remove windows support (https://github.com/buildpacks-community/kpack/discussions/1366)

text/0000-remove-windows-containers-support.md
```diff
@@ -0,0 +1,97 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+
+- We want to remove Windows Containers related documentation, but clearly state that the specification is platform neutral.
```
hone, line 35: How hard will it be to make a platform neutral specification format, i.e. POSIX vs. windows?

text/0000-remove-windows-containers-support.md
```diff
@@ -0,0 +1,97 @@
...
+* continue to use the last release of `pack` that supports Windows Containers,
+* fork `pack` and `lifecycle` to maintain Windows Containers support, or
+* adopt alternative tooling that supports Windows Containers.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Windows Containers support is well tested within `pack`.  Removing support for Windows Containers removes a unique selling point of the `pack` CLI and `lifecycle`
+
+# Alternatives
```
hone, line 58: This came up in the WG today. Would we want to tie this to a spec change and just remove it all together from the spec?

text/0000-remove-windows-containers-support.md
AidanDelaney: We're hoping to maintain the status-quo in the specification where we currently state things like 

* "MUST NOT be run for Windows builds" (https://github.com/buildpacks/spec/blob/main/buildpack.md#purpose-2)
* "...with the OS path list separator (e.g. : on Linux, ; on Windows)." (https://github.com/buildpacks/spec/blob/main/buildpack.md#targets)
* "For Windows-based images..." (https://github.com/buildpacks/spec/blob/main/buildpack.md#targets-1)

It will be difficult to keep in mind the POSIX/Windows differences when changing the specification.  However, I suspect this will be no-more difficult than currently as we have no known Windows containers users.

edmorley:
Given https://medium.com/buildpacks/deprecation-announcement-windows-container-feature-in-cloud-native-buildpacks-bbb70351343d should this PR be moved out of draft status and merged? :-)

jabrown85:
@AidanDelaney care to undraft this and update anything that you think needs updating? Then we can get it merged.

AidanDelaney:
> @AidanDelaney care to undraft this and update anything that you think needs updating? Then we can get it merged.

We don't yet have consensus on this.  But I'm happy with the approach where the Spec retains Windows support but we drop Windows support from our implementations.

text/0000-remove-windows-containers-support.md
jjbustamante: @AidanDelaney can we update the statement in the RFC to reflect Tom's comment?

text/0000-remove-windows-containers-support.md
```diff
@@ -0,0 +1,97 @@
...
+- Supporting Windows Containers is straightforward at a specification level.  However, the CNB project has neither the time nor the expertise to continue to support Windows Containers in our tooling.
+- Removing Windows Container support from CNB tooling will speed up development, simplify testing and impact no existing users.
+- Establish platform parity between `pack` and `kpack`.  `kpack` has never supported building Windows Containers.
+
+# What it is
+[what-it-is]: #what-it-is
+
+- We want to remove Windows Containers related documentation, but clearly state that the specification is platform neutral.
+- Remove Windows Containers support from `pack`, but retain the Windows platform release of the `pack` CLI.
+    * this ensures `pack` users can still invoke `pack` on a Command or Power Shell console
```
jjbustamante, line 37: On Windows, `pack` users will be able to invoke `pack` on a command or power shell and use `Linux containers on Windows`, right?

text/0000-remove-windows-containers-support.md
```diff
@@ -0,0 +1,97 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+
+# Migration
+[migration]: #migration
+
+Teams that require Windows Container support will have to
+
+* continue to use the last release of `pack` that supports Windows Containers,
```
jjbustamante, line 49: I have a question here: When not using the `creator`,  `pack` tries to download the latest `lifecycle` and then calculate the API version supported. 

This is an screen-shoot from [lifecycle](https://hub.docker.com/r/buildpacksio/lifecycle/tags) dockerhub 

Screenshot from 2024-12-19 08-39-06

Once lifecycle stops supporting Windows, we will not have an image for `windows/amd64`, let's suppose this happens on lifecycle `>= 0.21.x` version. On the other hand, we will keep releasing pack binaries for windows. 

When a `pack` user on Windows runs a build I think we will need to handle the cases where the lifecycle image for  `windows/amd64` is not available and print a warning message instead of a error which I think is gonna be the case.

jjbustamante (approved):
Looks good to me, I left a few comments that I think we should consider if we want to add some notes.

text/0000-remove-windows-containers-support.md
jabrown85: Yes - `pack` will continue to work against Docker Desktop (linux), podman or whatever linux container systems are docker-api compatible.

jjbustamante: Should we add that to the RFC?

PR #313, opened 2024-04-11
Export Run Image Metadata

joeybrown-sf:
[readable](https://github.com/buildpacks/rfcs/blob/ddb182852188465c164d8ed630c6e682e5e19df2/text/0000-export-run-image-metadata.md)

### DIFF
--- a/text/0000-export-run-image-metadata.md
+++ b/text/0000-export-run-image-metadata.md
@@ -99,6 +99,8 @@ This metadata is written to the [`lifecycle.metadata` label](https://github.com/
 
 These values are written to image labels. And they are among values that `pack inspect` returns.
 
+`report.toml` was introduced in [RFC 0040](/text/0040-export-report.md). That RFC includes some context for why `report.toml` exists.
+
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
 
--- a/text/0000-export-run-image-metadata.md
+++ b/text/0000-export-run-image-metadata.md
@@ -12,7 +12,13 @@
 # Summary
 [summary]: #summary
 
-The `report.toml` file created by the lifecycle exporter and rebaser should include `run-image.image`, `run-image.reference`, and `run-image.top-layer`. These values are not necessarily known prior to export or rebase they can be critical to a platform rebase process.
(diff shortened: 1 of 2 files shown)


text/0000-export-run-image-metadata.md
```diff
@@ -0,0 +1,139 @@
...
+This metadata is written to the [`lifecycle.metadata` label](https://github.com/buildpacks/spec/blob/main/platform.md#iobuildpackslifecyclemetadata-json) and it can be accessed by querying a docker daemon or registry. So we will be writing this data to two outputs.
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Do nothing and continue to require platforms to retrieve this information via alternative means--either querying the docker daemon or registry.
+    - Rebase process may remain suboptimal for some platform providers.
+
+2. Write all the metadata labels to `report.toml`.
+    - This could break platform operators that are using the `terminationMessagePath` [pattern](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#pod-v1-core). Because some of the metadata includes unbounded arrays, we could explode this report beyond the max size of 4096 bytes.
```
natalieparellano, line 89: @tomkennedy513 do you have any feedback here?

natalieparellano: FWIW I don't think we're close to hitting the limit yet (though it's been said before...)

chenbh: IIRC kpack doesn't blindly dump the report.toml directly, it parses out [only the stuff it cares about](https://github.com/buildpacks-community/kpack/blob/ae5eb15b8b9632fda50bafc4d2fad0f8c2c8a2a8/pkg/cnb/build_metadata.go#L22-L28) and [writes that to `terminationMessagePath` instead](https://github.com/buildpacks-community/kpack/blob/ae5eb15b8b9632fda50bafc4d2fad0f8c2c8a2a8/pkg/cnb/build_metadata.go#L22-L28).

I don't think we've encountered any issues with length yet, but I image that's because we limit the unbounded array to just the id, version, and homepage metadata of buildpacks.

natalieparellano (approved):

jjbustamante (approved):
💥

jabrown85 (approved):

PR #314, opened 2024-06-26
RFC for OCI Image Annotations on Buildpacks

candrews:
The `buildpack` tool should set OCI annotations on the OCI artifacts it produces providing users of these buildpacks with a consistent, standard mechanism for gathering information about the buildpack, including how to find its documentation and what version it is. The annotation values can be gathered from existing data sources (such as `buildpack.toml`) therefore not requiring any additional effort on the part of users of the `buildpack` tool.

Originally proposed to Paketo at https://github.com/paketo-buildpacks/rfcs/pull/307 where it was pointed out that this change would best be made in the `buildpack` tool, see https://github.com/paketo-buildpacks/rfcs/pull/307#issuecomment-2135394931

### DIFF
--- a/text/0000-oci-image-annotations.md
+++ b/text/0000-oci-image-annotations.md
@@ -45,7 +45,7 @@ The target personas as buildpack users, platform operators, and platform impleme
 # How it Works
 [how-it-works]: #how-it-works
 
-When building the buildpack, the `pack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
+When packaging the buildpack, the `pack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
 
 The other annotation values come from `buildpack.toml` mapped to OCI annotations as follows:
 
--- a/text/0000-oci-image-annotations.md
+++ b/text/0000-oci-image-annotations.md
@@ -12,7 +12,7 @@
 # Summary
 [summary]: #summary
 
-The `buildpack` tool should set OCI annotations on the OCI artifacts it produces providing users of these buildpacks with a consistent, standard mechanism for gathering information about the buildpack, including how to find its documentation and what version it is. The annotation values can be gathered from existing data sources (such as `buildpack.toml`) therefore not requiring any additional effort on the part of users of the `buildpack` tool.
+The `pack` tool should set OCI annotations on the OCI artifacts it produces providing users of these buildpacks with a consistent, standard mechanism for gathering information about the buildpack, including how to find its documentation and what version it is. The annotation values can be gathered from existing data sources (such as `buildpack.toml`) therefore not requiring any additional effort on the part of users of the `pack` tool.
 
 # Definitions
 [definitions]: #definitions
@@ -45,7 +45,7 @@ The target personas as buildpack users, platform operators, and platform impleme
 # How it Works
 [how-it-works]: #how-it-works
 
-When building the buildpack, the `buildpack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
+When building the buildpack, the `pack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
 
 The other annotation values come from `buildpack.toml` mapped to OCI annotations as follows:
 
@@ -66,7 +66,7 @@ The following example values are from [Paketo Buildpack for Java 13.0.1](https:/
 # Migration
 [migration]: #migration
 
-The `buildpack` tool would be modified to set the annotations. Because the tool _should_ set these annotations (not _must_ set), buildpacks created with earlier versions of the tool are still consider to be valid in accordance with the distribution specification. 
+The `pack` tool would be modified to set the annotations. Because the tool _should_ set these annotations (not _must_ set), buildpacks created with earlier versions of the tool are still consider to be valid in accordance with the distribution specification.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-oci-image-annotations.md
+++ b/text/0000-oci-image-annotations.md
@@ -1,0 +1,116 @@
+# Meta
+[meta]: #meta
+- Name: OCI Image Annotations on Buildpacks
+- Start Date: 2024-06-26
+- Author(s): @candrews
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- CNB Pull Request: (leave blank)
+- CNB Issue: (leave blank)
+- Supersedes: (put "N/A" unless this replaces an existing RFC, then link to that RFC)
+
+# Summary
+[summary]: #summary
+
+The `buildpack` tool should set OCI annotations on the OCI artifacts it produces providing users of these buildpacks with a consistent, standard mechanism for gathering information about the buildpack, including how to find its documentation and what version it is. The annotation values can be gathered from existing data sources (such as `buildpack.toml`) therefore not requiring any additional effort on the part of users of the `buildpack` tool.
+
+# Definitions
+[definitions]: #definitions
+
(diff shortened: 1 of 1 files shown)


text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,115 @@
...
+* `org.opencontainers.image.description` (when possible)
+
+The target personas as buildpack users, platform operators, and platform implementors. Any of those groups will be able to more easily understand the origin (source), version, and other information about the buildpack. This information can then be used manually or with the aid of tools to get release notes which aid these personas in making informed decisions.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When building the buildpack, the `buidpack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit ref.
+
+The other annotation values come from `buildpack.toml` mapped to OCI annotations as follows:
```
edmorley, line 51: What happens when an app has more than one buildpack? The proposed schema doesn't support more than one entry.

candrews: The RFC is for the buildpacks themselves - not for the apps built by the buildpacks.

For example, it would apply to the [Paketo Java buildpack, `gcr.io/paketo-buildpacks/java`](https://github.com/paketo-buildpacks/java), not to the application it builds.

edmorley: Ah ok. I think the RFC wording might need adjusting to make that clearer.

For example here:

> The `buildpack` tool should set OCI annotations on the OCI artifacts it produces

"buildpack tool" isn't a phrase that's really used by this project, and read to me as meaning "the buildpack".

So it then read as:

"The buildpack should set OCI annotations on the OCI artifacts it produces"

...and the only OCI artifacts created when the buildpack runs is the app OCI image.

It sounds like what you're really proposing is a change to the buildpack distribution specification:
https://github.com/buildpacks/spec/blob/main/distribution.md

edmorley: > It sounds like what you're really proposing is a change to the buildpack distribution specification:

Combined with proposing that the Pack CLI implement these changes, so that when a buildpack maintainer runs `pack buildpack package` to package their buildpack into the OCI image form of distribution it has the proposed labels set.

candrews: Instead of:
> The `buildpack` tool should set...

I've updated the RFC to say:
> `pack buildpack package` should set

Does that eliminate the confusion?

> It sounds like what you're really proposing is a change to the buildpack distribution specification:
> https://github.com/buildpacks/spec/blob/main/distribution.md

I believe so - is there something I should do in this RFC to indicate as much

edmorley: > Does that eliminate the confusion?

Yeah that does thank you.

> I believe so - is there something I should do in this RFC to indicate as much?

So I think this comes down to whether this RFC is proposing only a change to the Pack CLI's own implementation, or a change to the specification which then the Pack CLI (plus any other implementations) would implement. If the latter, than the spec change could be described under the "spec change" section of the RFC template.

natalieparellano:
Awesome to see this! I believe this supersedes https://github.com/buildpacks/rfcs/pull/196

text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,116 @@
...
+- `org.opencontainers.image.revision`: 09747b1df0a56aea74ce9b01af89df6feb1fc50a
+- `org.opencontainers.image.title`: Paketo Buildpack for Java
+- `org.opencontainers.image.version`: 13.0.1
+- `org.opencontainers.image.url`: https://paketo.io/docs/howto/java
+- `org.opencontainers.image.description`: A Cloud Native Buildpack with an order definition suitable for Java applications
+
+# Migration
+[migration]: #migration
+
+The `buildpack` tool would be modified to set the annotations.
```
jkutner, line 69:
```suggestion
The `buildpack` tool would be modified to set the annotations. Because the tool _should_ set these annotations (not _must_ set), buildpacks created with earlier versions of the tool are still consider to be valid in accordance with the distribution specification. 
```

Just adding this to be clear

candrews:
> Thanks for opening this RFC. May I recommend extending this RFC not just to buildpack but also to builders?

That sounds like a good idea to me!

text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,116 @@
...
+
+The outcome will be that users and tools will be able to gather more information about buildpacks, facilitating use cases such as gathering releases notes and finding documentation.
+
+# What it is
+[what-it-is]: #what-it-is
+
+`pack buildpack package` should set the following OCI annotations on the images it produces:
+
+- `org.opencontainers.image.source`
+- `org.opencontainers.image.revision`
```
jkutner, line 37: I think these may also need to be optional, in the event that the Git or VCS info isn't available.
```suggestion
- `org.opencontainers.image.source` (when possible)
- `org.opencontainers.image.revision` (when possible)
```

text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,116 @@
...
+- `org.opencontainers.image.version`
+- `org.opencontainers.image.url` (when possible)
+- `org.opencontainers.image.description` (when possible)
+
+The target personas as buildpack users, platform operators, and platform implementers. Any of those groups will be able to more easily understand the origin (source), version, and other information about the buildpack. This information can then be used manually or with the aid of tools to get release notes which aid these personas in making informed decisions.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When building the buildpack, the `buildpack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit ref.
```
jkutner, line 48: > git commit ref

Should this be the Git commit sha instead of the ref?

candrews: You are correct - that should be the sha, not the commit ref.

jkutner:
This is looking good. I'm going to bring it up in working group today, but then I think it's ready to move forward. Thanks @candrews

text/0000-oci-image-annotations.md
candrews: I've updated all occurrences of "`buildpack` tool" to be "`pack` tool" for consistency

text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,116 @@
...
+- `org.opencontainers.image.revision`: 09747b1df0a56aea74ce9b01af89df6feb1fc50a
+- `org.opencontainers.image.title`: Paketo Buildpack for Java
+- `org.opencontainers.image.version`: 13.0.1
+- `org.opencontainers.image.url`: https://paketo.io/docs/howto/java
+- `org.opencontainers.image.description`: A Cloud Native Buildpack with an order definition suitable for Java applications
+
+# Migration
+[migration]: #migration
+
+The `pack` tool would be modified to set the annotations. Because the tool _should_ set these annotations (not _must_ set), buildpacks created with earlier versions of the tool are still consider to be valid in accordance with the distribution specification.
```
joshwlewis, line 69:
```suggestion
The `pack` tool would be modified to set the annotations. Because the tool _should_ set these annotations (not _must_ set), buildpacks created with earlier versions of the tool are still considered to be valid in accordance with the distribution specification.
```

text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,116 @@
...
+- `org.opencontainers.image.version`
+- `org.opencontainers.image.url` (when possible)
+- `org.opencontainers.image.description` (when possible)
+
+The target personas as buildpack users, platform operators, and platform implementers. Any of those groups will be able to more easily understand the origin (source), version, and other information about the buildpack. This information can then be used manually or with the aid of tools to get release notes which aid these personas in making informed decisions.
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When building the buildpack, the `pack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
```
joshwlewis, line 48:
```suggestion
When packaging the buildpack, the `pack` tool can get the values for the `org.opencontainers.image.source` and `org.opencontainers.image.revision` annotations from git. `org.opencontainers.image.source` is derived from the git origin and `org.opencontainers.image.revision` is the git commit hash.
```

Since the command is `pack buildpack package`, I think `packaging the buildpack` makes more sense here.

jkutner (approved):

jkutner:
FCP should have ended last Thursday, but I failed to note that FCP had started. I'm going to give this until next week, and then finalize it. Sorry for the delay.

jkutner:
This RFC is accepted. I'm just trying to figure out the merge process (our tooling is broken :( )

PR #323, opened 2025-01-16
ArtifactHub.io Integration

joeybrown-sf:
[readable](https://github.com/joeybrown-sf/rfcs/blob/artifact-hub/text/0000-artifact-hub-integration.md)


text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,166 @@
...
+
+[definitions]: #definitions
+
+### [ArtifactHub](https://github.com/artifacthub/hub)
+
+A CNCF project that is a web-based application that enables finding, installing, and publishing Cloud Native packages.
+
+### [ArtifactHub.io](https://artifacthub.io/)
+
+An instance of the ArtifactHub project that is hosted with CNFC resources and maintained by ArtifactHub project
```
jkutner, line 36:
```suggestion
An instance of the ArtifactHub project that is hosted with CNCF resources and maintained by ArtifactHub project
```

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,166 @@
...
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+We would need to consider changes to the ArtifactHub type schemas when modifying our spec. It is possible that the
+Artifact Hub artifact schemas and API could drift from the official RFC speec and implementation. We would need to
+consider backwards compatibility with the ArtifactHub implementation.
+
+# Alternatives
```
jkutner, line 109: Another alternative: we could also sunset our registry move to ArtifactHub entirely

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,166 @@
...
+
+- [registry-namespaces](https://github.com/buildpacks/registry-namespaces)
+- [registry-api](https://github.com/buildpacks/registry-api)
+- [registry-index](https://github.com/buildpacks/registry-index)
+
+# Unresolved Questions
+
+[unresolved-questions]: #unresolved-questions
+
+Is there a Buildpacks team that owns the Registry Service now? Are they willing to take on the burden of owning this? If
```
jkutner, line 137: Yea, it's owned by [Platform Team](https://github.com/buildpacks/community/blob/main/TEAMS.md#Platform-Team)

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,166 @@
...
+This is the registry used by pack when it is identifying buildpacks referenced with the URN pattern of
+`urn:cnb:builder[:<id>[@<version>]]` or
+`urn:cnb:registry[:<id>[@<version>]]` ([ref](https://buildpacks.io/docs/for-app-developers/how-to/build-inputs/specify-buildpacks/)).
+This will remain in place.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+We would need to consider changes to the ArtifactHub type schemas when modifying our spec. It is possible that the
```
jkutner, line 105: hmm, are you sure we need to include them? or could they just be a supplement. or at least a spec extension?

what does this schema look like?

text/0000-artifact-hub-integration.md
jjbustamante: @jkutner  I am curious about how the Tekton integration ended up published on artifacthub.io  

Screenshot from 2025-02-12 07-41-28

Is this something Javier was doing?

text/0000-artifact-hub-integration.md
joeybrown-sf: I will update

joeybrown-sf:
I have made substantial changes to this RFC after discussing with an ArtifactHub.io maintainer and doing some more research. I am now proposing to decommission [registry.buildpacks.io](https://registry.buildpacks.io/).

text/0000-artifact-hub-integration.md
joeybrown-sf: After some more research, I have actually made this part of the main proposal.

text/0000-artifact-hub-integration.md
joeybrown-sf: Upon further review, I do not think we would need to consider changes to the Artifact Hub schema.

[Here](https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml) is the Artifact Hub schema. If it does not suffice, there are also custom annotations we can provide. Here is an example of [custom annotations](https://artifacthub.io/docs/topics/annotations/keptn/).

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+registry. This could also cause additional toil for buildpack authors if they are burdened with publishing to two
+indexes. Therefore, we should not do this unless we plan on migrating (in at least some capacity) to treat
+ArtifactHub.io as an official buildpacks registry. This work would happen in the Artifact Hub repository. Proposal
+discussion can be found [here](https://github.com/artifacthub/hub/issues/4352).
+
+### Integrate CNB services and tools to use ArtifactHub.io as the official CNB registry
+
+This should happen in phases so that buildpack authors and `pack` users have minimal disruption to established
+workflows. CNB is committed to backwards compatibility, but there are sometimes good reasons to break this backwards
+compatibility. This RFC proposes a slow phase-out and eventual decommission of the buildpacks registry. More details
```
jkutner: I don't think that shutting down the old buildpacks registry necessarily means breaking backwards compat. If ArtifactHub matches the features we provide today, then changing our "backend" means we're still backwards compat.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+# Summary
+
+[summary]: #summary
+
+This RFC proposes that the Buildpacks project integrates and augments the
+bespoke [registry service](https://registry.buildpacks.io/) with [ArtifactHub.io](https://artifacthub.io/) and
+eventually deprecates the buildpack registry.
+
+This RFC proposes that in 12 months, the buildpacks registry index becomes read-only. Buildpack publishers
+will have 12 months to change their publish process so that buildpacks are no longer published to the registry index,
```
jkutner: hmm, will they actually need to change their publish process, or could our tooling (i.e. `pack`) paper over the changes so that they go forward not worrying about the backend change?

I suppose they would have to change their expectations about the website and slack notifications.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+Because CNB already has a buildpack registry, there is a necessary migration story. The following requirements
+outline the necessary work. The term "requirement" is used instead of "step" because some requirements can be done in
+parallel, while some have dependencies on other requirements.
+
+### Requirement 1 - Approval in Artifact Hub for work to proceed
+
+The [issue](https://github.com/artifacthub/hub/issues/4352) must be accepted in Artifact Hub. Subsequent work is
+outlined in the third requirement.
+
+### Requrement 2 - CNB Announcement
```
jkutner: I think this should be step zero (even before the above issue is accepted.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+
+`pack` will continue reading from the CNB registry API.
+
+At this point, publishers can publish to either the buildpack registry index or ArtifactHub.io. By the end of 12 months,
+publishers should only use the ArtifactHub.io method for publishing, because the buildpack registry index will become
+readonly at this point and will not accept new publishes.
+
+### Requirement 6 - Update `pack` to pull from ArtifactHub.io
+
+We will introduce a new flag to `pack` called `registry-schema`. The options for this flag are `cnb` and `artifacthub`.
```
jkutner: instead of `cnb`, i think we should use `legacy` or something that strongly indicates you don't want to favor it. The `cnb` value kinda implies it's the preferred one

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+This flag would be threaded through to all the current `pack` commands that rely on the registry, including all the
+`pack buildpack` commands. If this flag is set to `artifacthub`, `pack` will utilize ArtifactHub.io. For instance, the
+`pack buildpack register` command will call the ArtifactHub.io API.
+
+From month 0-12, the flag will default to `cnb` and users will be warned about the upcoming behavior changes.
```
jkutner: I think we should flip the flag at 6 months, and terminate the registry at 12 months.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,339 @@
...
+After month 18, the registry service will be terminated. At this point, ArtifactHub.io will be the official source for
+buildpacks.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+This is a disruptive change for buildpack publishers. The ownership and publishing model will be different.
+
+This is a destructive change that will affect `pack`. Any versions of `pack` that depend on the buildpack registry will
```
jkutner: this is the biggest risk IMO. we should think about how we might get ahead of this (like releasing something in `pack` ASAP that helps future proof)

jkutner:
My biggest bit of feedback is that I think we should tighten up the timeline for cutting over to the ArtifactHub. But I'm less certain about the shutdown of the old registry (either 12 or 18 months)

text/0000-artifact-hub-integration.md
joeybrown-sf: The publishing process would be different.

Instead of using pack to explicitly update the cnb registry, publishers would add a new directory in their Artifact Hub registry source.

For instance, today if publisher wanted to publish verion 1.30.1 of their buildpack, they would do the following:
1. Push the image to their registry
2. Update the CNB registry index (manually or via `pack`)
3. CNB registry is updated when the indexer eventually scans the registry index.

Meanwhile, if this RFC is adopted, publishers would do the following:
1. Push the image to their registry
2. Create a new directory in their Artifact Hub registry source (a git repo they have configured in the Artifact Hub admin console).
3. Artifact Hub is updated when it's scanner checks the Artifact Hub registry source.
4. CNB registry is updated when the indexer eventually scans Artifact Hub.

text/0000-artifact-hub-integration.md
joeybrown-sf: It would be a bit different because the publishing process for the CNB registry is an explicit operation that happens against the cnb index repository. Future publishing would be implicit in that publishers modify a git repo they own and then the Artifact Hub scanner picks it up.

text/0000-artifact-hub-integration.md
jkutner: ok, so we're saying existing users of Buildpack Registry will need to take a one-time action (step 2)?

joeybrown-sf: yes that is correct. They will need to claim ownership of their artifacts in the ArtifactHub.io management console.

dmikusa: Do you have any idea about how long the delay is for steps 3 & 4? Just wondering if we're talking a couple of minutes or a couple of hours or longer?

text/0000-artifact-hub-integration.md
dmikusa: I don't know what the API for artifact hub is like or if this is in any way feasible, but what about running a proxy for requests from old versions of pack? If it could proxy and transparently map those requests to the new API, then old versions of pack would in theory keep working.

Not sure if there are any other platform implementations that integrated with the buildpacks registry, but that would be a way to keep other non-pack implementations working too.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+break as the various requirements of this RFC are implemented.
+
+ArtifactHub.io will be on the hot path for some `pack` operations.
```
dmikusa: What is the uptime like for this service? Does the CNCF have any guarantees about service availability?

What happens if there is an issue with the service? Do we have someone to page? Or a way to open a ticket and get things addressed?

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+
+ArtifactHub.io will be on the hot path for UI discoverability.
+
+If ArtifactHub.io goes away in the future, CNB will have to adapt. CNB discoverability and tooling could be left in a
```
dmikusa: Does the CNCF have a health report for ArtifactHub? Does it have funding guaranteed for a period of time? Are there other projects from the CNCF where ArtifactHub is on the critical path? 

Maybe we want to consider adding a layer of indirection, that we control, so that pack isn't talking to ArtifactHub directly. That way if it goes away, pack's talking to our layer of indirection, and we can swap out a backend easier.

text/0000-artifact-hub-integration.md
joeybrown-sf: That's a really good thought @dmikusa I will look into this.

text/0000-artifact-hub-integration.md
joeybrown-sf: Great questions. I will try to get some answers.

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+From month 0-6, there will be only additive changes to the registry service. It will maintain full backwards
+compatibility for 6 months.
+
+From month 6-18, the registry service will be read-only. No buildpacks will be able to publish to the registry service
+after 6 months, but it will still serve `GET` requests so that `pack` does not break.
+
+After month 18, the registry service will be terminated. At this point, ArtifactHub.io will be the official source for
+buildpacks.
+
+#### Processes - CNB Registry is terminated
```
edmorley: Just to clarify, by "CNB Registry is terminated" do you mean everything, including the indexer and API/UI components at https://registry.buildpacks.io/ ? Or just the publishing/index management parts? (I can't tell whether the dashboard will still exist and use ArtifactHub as a data source, or ...?)

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+### Requirement 3 - Add support in Artifact Hub for component buildpacks and builders.
+
+The work here is in Artifact Hub. This author's understanding is that work is done mostly by Artifact Hub maintainers,
+but CNB maintainers should be open and available for consulting.
+
+> [!WARNING]
+> We need to ensure the Artifact
+> Hub [package metadata](https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml) suffices
+> for Component Buildpacks and Builders. If not, we need to ensure we create
+> adequate [custom annotations](https://artifacthub.io/docs/topics/annotations/keptn/).
```
edmorley: Does this spec need to cover what custom annotations we should use for the buildpack and builder types on ArtifactHub.io?

eg Should Buildpack API version be a field? Supported Targets? Distro flavour? Or for builders, things like run/build image, Lifecycle version?

If the entirety of https://registry.buildpacks.io/ is going away (including the indexer/API/database parts), then I presume we'll need to make sure that the ArtifactHub types support all fields we need for lookup and searching (including fields we currently exact from the Docker Image in the current indexer).

For example, one of the features we're missing in the current CNB registry is the ability to eg filter by CNBs that support the latest Buildpack API versions. To support this use-case we'd presumably need Buildpack API as a custom annotation etc.

Lastly, if we have custom annotations that contain mirror metadata from the buildpack/builder for convenience/searchability - do we need some way to validate these to make sure they are correct? ie: To prevent a buildpack declaring support for one Buildpack API version or target in its `buildpack.toml` but a different value in its ArtifactHub entry? If we were still using https://github.com/buildpacks/github-actions I'd suggest that as the place to do so - but I guess that action will be sunset too since ArtifactHub will pull directly from people's GitHub repos? Do we need a lint action or similar? Maybe the ArtifactHub maintainers have some suggestions or existing tooling to help with this?

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+# Motivation
+
+[motivation]: #motivation
+
+Component Buildpacks and Builders should be easily discoverable to make them more usable and approachable. Today, the
+[buildpacks registry](https://registry.buildpacks.io/) service is the solution to discoverability. This registry service
+is a very important tool, but there is room for improvement. There are many features that ArtifactHub.io provides that
+the buildpack registry does not provide. Some features include rich descriptions, usage information, activity,
+dependency information, email subscriptions to packages, webhooks, related packages, and, furthermore, it is a service
+that the buildpacks team does not need to maintain.
```
edmorley: Another motivation we could list for sunsetting the old CNB registry is that its architecture has a number of issues, eg the constant rescanning of images resulting in hitting Docker Hub rate limits, xref:
https://cloud-native.slack.com/archives/C032YE21V1T/p1746521791976579

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+    
+    else pack registry-schema=artifacthub
+    activate dev
+    dev->>ahapi: pack build, inspect, etc
+    dev->>app: app built
+    deactivate dev
+    end
+```
+
+### Requirement 7 - Terminate the buildpacks registry service
```
edmorley: Do we need to mention the need to update https://github.com/buildpacks/github-actions with deprecation messages and eventually sunsetting its publishing functionality etc?

text/0000-artifact-hub-integration.md
```diff
@@ -0,0 +1,595 @@
...
+> [!WARNING]
+> We need to ensure the Artifact
+> Hub [package metadata](https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml) suffices
+> for Component Buildpacks and Builders. If not, we need to ensure we create
+> adequate [custom annotations](https://artifacthub.io/docs/topics/annotations/keptn/).
+
+### Requirement 4 - Mirror buildpack artifacts from the buildpack [registry index](https://github.com/buildpacks/registry-index) to ArtifactHub.io
+
+_The goal here is to create a process that will register all the current (and new) buildpacks from the registry index to
+ArtifactHub.io._
```
edmorley: I wonder if we should intentionally skip mirroring some CNBs, such as those that are stale/using EOL Buildpack APIs or were clearly only for testing/demo purposes? I suspect we have quite a lot of inactive/unusable CNBs in the current registry - it seems like now might be a good time to drop some of that clutter?

For example, anything older than Buildpack API 0.7 won't work with recent lifecycle versions:
https://github.com/buildpacks/lifecycle?tab=readme-ov-file#supported-apis

(People would still be free to publish any API 0.6 and older CNBs themselves, but we could skip doing it for them... I kinda prefer the idea of not pre-emptively adding a bunch of stuff that will likely never be used, and instead waiting to hear for a need for it)

Also, I wonder if we should first ask the big publishers (eg Heroku, Paketo, GCP, ...) if they want to publish their CNBs first, before we mass reserve all of their namespaces and make them claim ownership back?

edmorley (commented):
Thank you for putting this together! I'm very excited about the prospect of us having fewer services/components to maintain :-)

text/0000-artifact-hub-integration.md
joeybrown-sf: There is not a published uptime report, nor are there guarantees about service availability. 

There is a #artifact-hub channel in the CNCF Slack and [github issues](https://github.com/artifacthub/hub/issues) that can be used to communicate with maintainers.

Do note that Artifact Hub does not host the packages.

text/0000-artifact-hub-integration.md
joeybrown-sf: There is a CLO report [here](https://clomonitor.io/projects/cncf/artifact-hub) here that is at 99%. I will note, it's in the top 7 of the projects listed here.

text/0000-artifact-hub-integration.md
joeybrown-sf: I am proposing terminating the entire instance.

text/0000-artifact-hub-integration.md
joeybrown-sf: That is a good idea!

text/0000-artifact-hub-integration.md
joeybrown-sf: I will add this.

text/0000-artifact-hub-integration.md
joeybrown-sf: This is a good point. I will add this.

text/0000-artifact-hub-integration.md
joeybrown-sf: Artifacthub is the primary distribution mechanism for helm charts

joeybrown-sf:
Hey folks I have substantially changed the content of this RFC. I resolved a bunch of open comment threads because I wanted to start a fresh dialog without getting distracted by conversations that were no longer necessarily relevant.

I have spiked out an integration and proved the concept. In the RFC, I now have a simple and concrete plan for this proposed integration and migration.

Please check it out and let me know if you have feedback or if I resolved a still-pertinent question of yours.

joeybrown-sf:
@jericop, @dmikusa Hey folks I was hoping yall could take a look at this when you get a moment. I have updated the RFC significantly. I have outlined a concrete path to migrating from publishing buildpack index metadata directly to the CNB Registry.

I'm particularly interested in your feedback about the proposed 6 month timeframe to migrate buildpack metadata from your existing publishing process to the new process. I plan on providing some tooling and a walkthrough of migrating and I am more than happy to contribute efforts and work with yall on the Paketo side.

Let me know if there is more information I can provide!

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+
+# Definitions
+
+[definitions]: #definitions
+
+__[Artifact Hub](https://github.com/artifacthub/hub)__ - CNCF project that is a web-based application that enables finding, installing, and publishing Cloud Native packages.
+
+__[ArtifactHub.io](https://artifacthub.io/)__ - Public good instance of the ArtifactHub project that is hosted with CNCF resources and maintained by ArtifactHub project maintainers.
+
+__[ArtifactHub Kind](https://github.com/artifacthub/hub/blob/master/docs/repositories.md)__ - OCI artifacts that Artifact Hub knows about. Artifact Hub supports over 20 kinds of Cloud Native artifacts, including Argo Templates, Backstage Plugins, Helm Charts, and others. This RFC proposes 2 new Artifacts _Kinds_: __cnb-buildpack__ & __cnb-builder__.
```
dmikusa, line 32: > This RFC proposes 2 new Artifacts _Kinds_: __cnb-buildpack__ & __cnb-builder__.

Is a cnb buildpack type sufficent? We really have two types of buildpacks, component (i.e. I do something) and composite (I bundle together some buildpacks). I don't know what this system is like, but is it flexible enought to have one type that supports both of those needs?

My head goes to wanting to see different things for these. For a component buildpack, I want to see things in the buildpack.toml, like the version and maybe the configuration (if we could do that generically). For the composites, it's about the order groups. What does this composite buildpack pull together? I want to see a list of that. I guess that's why I'm asking, cause in the UI I'd want to see different things based on the type of CNB.

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+# Footnotes
+
+#### <a id="minimal-work"></a> 1. There is minimal work required from CNB team to implement the cnb-buildpack & cnb-builder kinds in Artifact Hub. The following quote is from one of the ArtifactHub maintainers:
+> Please note that we'd rather to take care ourselves of the changes to add support for the new artifact kind to AH once we've agreed on the plan.
+
+[source](https://github.com/artifacthub/hub/issues/4352#issuecomment-3751449070)
+
+#### <a id="artifact-claims"></a> 2. I have personally claimed several organizations and repository names in ArtifactHub.io.
+
+Here is what I have claimed:
```
dmikusa, line 278: So when you say you've claimed these, is that blocking some malicious user from coming along and trying to register/steal them?

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+The process that exports data from ArtifactHub.io to CNB Registry will be able to trust that
+```
+ArtifactHub.io repository name + ArtifactHub.io buildpack name = CNB Registry ID
+```
+
+This trust is possible because I have [claimed several organizations and buildpacks](#artifacthub-claims) on behalf of community contributors in order to maintain trust between ArtifactHub.io and the CNB Registry. This allows us to maintain trust without requiring a mapping of ArtifactHub.io attributes to CNB attributes.
+
+## Pack Changes
+
+Also after the migration, we will remove the experimental buildpack publishing commands from `pack`. If users are on an old version of pack and execute this command, it will open a github issue that will be immediately closed with `Not Planned`, a short explanation, and a link to the docs.
```
dmikusa, line 119: What about the feature in pack where you can reference `--buildpack <buildpack-id>` and it will look it up in the registry? Is this going to change? Will we support pointing to different registries? If you can self-host ArtifactHub, users might want to host their own internally (like for offline/air gapped envs), so being able to point to something specific could be helpful.

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+
+## CNB App Changes
+
+No changes necessary*. The CNB Registry is backwards compatible.
+
+*If you rely on a buildpack that is not migrated to ArtifactHub.io, and you reference that buildpack via a Registry URI, then your build will likely fail in the future.
```
dmikusa, line 125: I'm not sure I follow how apps could break here. I'm definitely concerned with this case as we really want to avoid Paketo users from having builds fail unexpectedly.

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+
+## CNB Platform Changes
+
+No changes necessary*. The CNB Registry is backwards compatible.
+
+*If your developers rely on a buildpack that is not migrated to ArtifactHub.io, and you or they reference that buildpack via a Registry URI, then their builds will likely fail in the future.
```
dmikusa, line 131: I'm not sure I am following what will happen with the existing data that is in the buildpacks registry now. Is that going to be migrated in some way to the new system? Or is that being put onto buildpack authors?

Paketo has a large history of buildpacks. If we want to keep this accessible to through the registry, what would we need to do? This is definitely something we'd want to do.

text/0135-artifact-hub-integration.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Add ArtifactHub.io Banner to registry.buildpacks.io
+
+registry.buildpacks.io will be going away at the end of the migration deadline. Add a banner so people know that
+
+#### Step 4 - Buildpack Author Migration with 6 Month Deadline
+
+The 6 month countdown begins now.
+
+At this point, Buildpack Authors are equipped to migrate to ArtifactHub.io as their buildpack index publishing platform. Authors will have 6 months, during which they can publish indexes to either or both indexes (CNB Registry and ArtifactHub.io), where ArtifactHub.io metadata takes precedence.
```
dmikusa, line 179: Are there any plans to offer tooling for the new API? Right now we have things like the Github Action for publishing to the registry. Will that action be updated to work with the new registry? or are there existing actions for publishing to ArtifactHub?

text/0135-artifact-hub-integration.md
joeybrown-sf: Yes that is correct. organizations and repositories are both globally unique in ArtifactHub. So I created them in Artifacthub.io.

If we can trust the repository names of the ArtifactHub.io cnb-buildpacks are the same as CNB Registry namespaces, then we don't need to maintain a trusted map between ArtifactHub.io cnb-buildpacks and CNB Registry buildpacks.

text/0135-artifact-hub-integration.md
joeybrown-sf: > feature in pack where you can reference --buildpack <buildpack-id> and it will look it up in the registry

No this feature will not change.

This feature depends on the [registry spec](https://github.com/buildpacks/spec/blob/main/extensions/buildpack-registry.md), which defines a CNB registry as a git repo.

Platforms (pack users) will still be able to point to a registry of their choosing. The CNB Registry will continue to operate and be maintained, but it will be pulling data from ArtifactHub.io and will not serve as an index publishing platform.

text/0135-artifact-hub-integration.md
joeybrown-sf: Let's say I published a blog 2 years ago and referenced my buildpack `joeys-buildpacks/dotnet`. There could be users out there that use my buildpack (referenced with the Registry URI). If I migrate my dotnet buildpack to Artifacthub.io and put it in the repository called `joeys-buildpacks`, everything will continue to work.

But if I do not migrate the `joeys-buildpacks/dotnet` index to ArtifactHub.io, it will get deleted from the CNB Registry. Any apps that reference this buildpack via the CNB Registry URI will fail to build because Registry URI resolution will fail.

I also want to make sure Paketo users are not impacted. If the Paketo team migrates all their buildpacks, paketo users should see no impact.

I am happy to help the Paketo maintainers with this work.

text/0135-artifact-hub-integration.md
joeybrown-sf: The existing data will be deleted if it is not migrated to ArtifactHub.io. This RFC puts all the migration work on the Buildpack Authors. The CNB org will not be responsible for migrating any community buildpacks.

If we want to keep this accessible to through the registry, what would we need to do

You would need to create `artifacthub-pkg toml` files for all existing buildpacks. Check out my [proof of concept](https://github.com/joeybrown-sf/h-buildpacks/tree/main/catalog/paketo-buildpacks) repo I exported the go buildpacks for paketo. 

I have an export script I'm working on that will create `artifacthub-pkg.yml` files for existing packages (that's what I used to export the go buildpacks). Then going forward, you would be responsible for creating new ones as new versions are published.

text/0135-artifact-hub-integration.md
joeybrown-sf: I planned on providing a walkthrough for Buildpack authors. I would love to would with you to make sure it fits.

To publish buildpack indexes, you don't work directly with the ArtifactHub.io API. Rather, you set up your Repositories in ArtifactHub.io (one-time operation), then artifacthub.io scans the associated git registry for metadata files it recognizes (`artifacthub-pkg.yml`) and adds that data to its database.

text/0135-artifact-hub-integration.md
dmikusa: Do you have any concerns about trusting what's in ArtifactHub?

Right now, let me see if I understand things right. 
- Anyone can publish to our registry.
- They just need to do so from a Github user and then the buildpack ID namespace must match Github user from the token use to publish, right?

Like I can publish a buildpack under `dmikusa/my-buildpack` if I do so from my Github user dmikusa, but if Github user `bad-actor` tried to publish `dmikusa/my-buildpack` then our automation would reject it, right?

I guess what we'd want to make sure the same is true for ArtifactHub. We wouldn't want someone to pick a user `foo` and then try to publish to `dmikusa/my-buildpack`. It should be restricted some way that only things for the `dmikusa` user in ArtifactHub can publish buildpacks with the namespace `dmikusa/*`.

It sounds like that's the case though. Is that your thought too?

text/0135-artifact-hub-integration.md
dmikusa: Ah, gotcha. So as long as we follow the steps here there will be no breakage for users. I think that works.

Happy to work together on this. Paketo's always willing to be the Guinea pig for things like this 😄

text/0135-artifact-hub-integration.md
dmikusa: Ok, so Paketo would need to make a new repo under our Github org. We'd publish all the ArtifactHub data there. Then periodically, ArtifactHub will pick up the data from our repo (and we register this repo with ArtifactHub).

The format doesn't look too bad. I'm sure that we could write a CI job that does that, but I do feel like that is probably something we'll do and a bunch of other people will do also. It seems like it would be a good target for something to share. Perhaps that's something we could work together on and publish through https://github.com/buildpacks/github-actions?

text/0135-artifact-hub-integration.md
dmikusa: Gotcha. Happy to work together on this, appreciate the help.

text/0135-artifact-hub-integration.md
joeybrown-sf: Correct. Creating an ArtifactHub.io repository and pointing it to a git repository can be done by anyone.

The automation that pulls the cnb-buildpack data from ArtifactHub.io will build the registry ID by using the repository name from ArtifactHub.io as the namespace in the CNB Registry.

Github user `bad-actor `would need would to point the dmikusa ArtifactHub.io repo url to a git repo with an index that points to a malicious image.

Right now, the CNB organization I created in ArtifactHub.io owns the `dmikusa` ArtifactHub.io repository name.
- It happens to be pointed to a fake github url now.
- In fact, it's not a cnb-repository, it's a tekton-pipeline repository (repository names are globablly-unique--not just unique for an artifact type)

Only members of the CNB Organization in ArtifactHub.io would be able to point the `dmikusa` ArtifactHub.io repository to a different git repository that contains the `artifacthub-pkg.yml` files.

joeybrown-sf: > Do you have any concerns about trusting what's in ArtifactHub?

In terms of ArtifactHub.io security itself or malicious actors doing name-based attacks?

For name-based attacks, I think my manual efforts protect the community against name-based attacks.

As far as ArtifactHub.io security itself, I am placing trust in the CNCF review process and trusting those maintainers to keep it secure.

text/0135-artifact-hub-integration.md
joeybrown-sf: Yes you can use a new repo or one you already have. Or you could make a repo for each of your buildpacks, but I think that seems overly tedious and I don't know what the benefit would be to an organization.

I didn't realize we had that github-actions repo existed that's great!

I would love to collaborate on this!

text/0135-artifact-hub-integration.md
joeybrown-sf: Short answer: I think.

I admittedly was playing around with component buildpacks and only glanced at composite buildpacks.

ArtifactHub.io has 2 main UI elements.

The first is a readme that cnb-authors will provide. This readme supports markdown, so you can put whatever markdown you want here.

The second UI element is a widget (that can be embedded externally). It might be really nice to have a standard for sharing composite buildpack hyperlinks in this widget.

This is great feedback. I will mess around with my dev env implementation and give you a more concrete answer this week.

text/0135-artifact-hub-integration.md
jericop: I just read through the proposal and all sounds good to me. Daniel raised the questions I was going to ask so I'm good. I think a little collaboration will go a long way to making this process seamless for paketo users. Great writeup!

text/0135-artifact-hub-integration.md
jkutner: For a composite buildpack, the metadata about the component buildpacks is on the artifact, so it should be posssible, even if all we do is link to them or something. But yea, good thought

jkutner (approved):

text/0135-artifact-hub-integration.md
XavierMP14, line 1: X

PR #329, opened 2025-03-31
RFC: DNS Label Compatible Process Types

joshwlewis:
[Rendered](https://github.com/joshwlewis/rfcs/blob/subdomain-processes/text/0000-dns-label-process-types.md)


text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+- Author(s): @joshwlewis
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request:
+- CNB Pull Request:
+- CNB Issue:
+- Supersedes:
+
+# Summary
+
+Process types in the Cloud Native Buildpacks buildpack specification are less restrictive than the RFC 1123 DNS label definition. Therefore, CNB process types may not be compatible with internet routing or kubernetes resource naming. This RFC proposes to restrict CNB process types to meet RFC 1123 and increase usage compatibility with other CNCF / OSS usage.
```
runesoerensen:
```suggestion
Process type identifiers, as defined in the Cloud Native Buildpacks buildpack specification, are less restrictive than the RFC 1123 DNS label definition. Therefore, CNB process types may not be compatible with internet routing or Kubernetes resource naming. This RFC proposes stricter process type identifier requirements to increase compatibility with other CNCF / OSS projects as well as RFC 1123.
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Definitions
+
+**RFC 1123**: The IETF ["Requirements for Internet Hosts" RFC](https://datatracker.ietf.org/doc/html/rfc1123#section-2.1) describes the syntax for DNS host names, domain names, and labels.
+
+**Kubernetes Resource Names**: Many resource names are restricted to RFC 1123, as outlined [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).
+
+**Process Types**: The name of a process in a Cloud Native Buildpacks result as [described in `launch.toml`](https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml).
```
runesoerensen:
```suggestion
**Process Type**: The `type` identifier for a launch process as [specified in `launch.toml`](https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml).
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Motivation
+
+The images resulting from CNB builds are often deployed to internet connected environments, where it's useful to connect or manage the image lifecycle with a domain name.
```
runesoerensen:
```suggestion
The images resulting from CNB builds are often deployed to internet connected environments, where it's useful to connect or manage the image lifecycle with an internet routable domain name.
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+While DNS Labels as defined in RFC 1123 are more restrictive. Some notable differences:
+
+- Underscores (`_`) are not allowed
+- Periods (`.`) are not allowed
+- Dashes ('-') are not allowed as the first or last character
+- Uppercase and lowercase letters are not differentiated
+- Must be less than 63 characters
+
+The buildpack specification will be modified to match RFC 1123 for DNS Labels.
```
runesoerensen:
```suggestion
The buildpack specification will be modified to require process types compatible with RFC 1123 for DNS Labels, and require lowercase letters for compatibility with Kubernetes.
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# How it Works
+
+The platform specification will be changed to reject process types that do not comply with RFC 1123 DNS Labels. When a process type fails to meet this syntax, appropriate error messaging may be given by the platform, for example:
```
runesoerensen:
```suggestion
The platform specification will be changed to reject process types that do not comply with lowercased RFC 1123 DNS Labels. When a process type fails to meet this syntax, appropriate error messaging may be given by the platform, for example:
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+The "Static_Assets" process type must comply with RFC 1123 DNS label syntax. Underscores are not allowed. Uppercase letters are not allowed."
+```
+
+While uppercase letters are allowed in RFC 1123, uppercase and lowercase are not differentiated -- `API` and `api` are the same. DNS providers and Kubernetes enforce (or coerce) uppercase to lowercase to prevent duplicate entries with different casing. The buildpack spec should also enforce lowercase letters for the same reason.
+
+# Migration
+
+This is a breaking API change, and will require a new Buildpack API version.
+
+This may require buildpacks to change the way process types are defined. For buildpacks using statically chosen process types that do not comply with RFC 1123, they'll need to change the chosen process types. For buildpacks using dynamically assigned process types, they'll need to coerce process types (by substitution and/or truncation) to meet the specification or risk builds failing when the dynamically chosen name does not comply.
```
runesoerensen:
```suggestion
Buildpacks specifying process types that do not comply with the stricter requirements proposed in this RFC will need to change the way process types are defined. Buildpacks defining process types dynamically should implement logic to coerce process types (by substitution, filtering and/or truncation) to meet the specification and avoid build failures.
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Drawbacks
+
+- It's a breaking API change and will cause ecosystem churn.
+
+# Alternatives
+
+## Enforce RFC 1123 host names
+
+Instead of enforcing the DNS label syntax, enforce the host name syntax. This would allow periods (in correct postions) and allow up to 254 characters.
```
runesoerensen: Would perhaps make sense to only allow up to 253 characters (if we want to optimize for Kubernetes compatibility with this alternative approach) https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+- Must be less than 63 characters
+
+The buildpack specification will be modified to match RFC 1123 for DNS Labels.
+
+# How it Works
+
+The platform specification will be changed to reject process types that do not comply with RFC 1123 DNS Labels. When a process type fails to meet this syntax, appropriate error messaging may be given by the platform, for example:
+
+```
+The "Static_Assets" process type must comply with RFC 1123 DNS label syntax. Underscores are not allowed. Uppercase letters are not allowed."
```
runesoerensen:
```suggestion
The "Static_Assets" process type must comply with lowercased RFC 1123 DNS label syntax. Underscores are not allowed. Uppercase letters are not allowed."
```

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+**RFC 1123**: The IETF ["Requirements for Internet Hosts" RFC](https://datatracker.ietf.org/doc/html/rfc1123#section-2.1) describes the syntax for DNS host names, domain names, and labels.
+
+**Kubernetes Resource Names**: Many resource names are restricted to RFC 1123, as outlined [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).
+
+**Process Type**: The `type` identifier for a launch process as [specified in `launch.toml`](https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml).
+
+# Motivation
+
+The images resulting from CNB builds are often deployed to internet connected environments, where it's useful to connect or manage the image lifecycle with an internet routable domain name.
```
dmikusa, line 26: I don't fully follow the motivations here. Could you expand on how this is being used in a way that triggers the problem? For example, what part of K8s is looking at the process types in the image and trying to create resources or map DNS names to the application based on the process types? I'm kind of surprised that K8s would be looking at these because they're buildpack-specific metadata. Is this something K8s is doing automatically? or is this a convention that users follow?

Just trying to wrap my brain around how this is being encountered in the wild.

joshwlewis: You are correct -- k8s doesn't use or look at the CNB process type identifier in any way. This might happen when building a platform running on k8s that supports CNBs, though. For instance, you might want to run a k8s `service` for the `web` process on a pack-built image. You might be tempted to name that service `web` or `web.some-namespace.mycluster.mycompany.com`, so the naming within k8s and the DNS entry within the cluster has some semantic meaning for the application developer. However, buildpacks today can contribute process type identifiers like `Web_API`, which can't be used for k8s naming and/or DNS subdomains. So the platform I'm describing would have to reject, filter, or substitute non-compliant process types identifiers.

dmikusa: Gotcha, sounds reasonable to me. Thanks for explaining.

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,139 @@
...
+The "Static_Assets" process type must comply with lowercased RFC 1123 DNS label syntax. Underscores are not allowed. Uppercase letters are not allowed."
+```
+
+While uppercase letters are allowed in RFC 1123, uppercase and lowercase are not differentiated -- `API` and `api` are the same. DNS providers and Kubernetes enforce (or coerce) uppercase to lowercase to prevent duplicate entries with different casing. The buildpack spec should also enforce lowercase letters for the same reason.
+
+# Migration
+
+This is a breaking API change, and will require a new Buildpack API version.
+
+Buildpacks specifying process types that do not comply with the stricter requirements proposed in this RFC will need to change the way process types are defined. Buildpacks defining process types dynamically should implement logic to coerce process types (by substitution, filtering and/or truncation) to meet the specification and avoid build failures.
```
dmikusa, line 74: Can we implement this in two phases? First phase, buildpacks violating this will result in a WARNING message. Second phase, buildpacks violating this will break.

That or have a flag that you can use to control the outcome. i.e. `--strict-proc-names=warn` or `--strict-proc-names=error`.

Either of these options would make the migration process easier & give buildpack authors that don't closely follow the spec changes an easy want to tell that something is going on.

dmikusa: The flag would be really nice, if you could opt-into that before updating your buildpack APIs. It would make testing and checking for impacted buildpacks easier.

joshwlewis: Good idea. Added.

edmorley:
Do we still want to proceed with this? (I think it's worth doing)

jabrown85:
> Do we still want to proceed with this? (I think it's worth doing)

I'd be good with pushing this forward. It seems like a sensible thing to align naming restrictions against. I put it in ready for review status so I won't forget to bring it up next meeting I attend

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,147 @@
...
+
+```
+The "Static_Assets" process type must comply with lowercased RFC 1123 DNS label syntax. Underscores are not allowed. Uppercase letters are not allowed."
+```
+
+While uppercase letters are allowed in RFC 1123, uppercase and lowercase are not differentiated -- `API` and `api` are the same. DNS providers and Kubernetes enforce (or coerce) uppercase to lowercase to prevent duplicate entries with different casing. The buildpack spec should also enforce lowercase letters for the same reason.
+
+# Migration
+
+This is a breaking API change, and will require a new Buildpack API version.
```
jkutner, line 72: I considered asking if there's a way would could instead define a mapping, or substitution thing, where we convert incompatible process types to compatible ones. But that could mean, in theory, two process types get mapped to the same thing. I think we just want to avoid that as much.

jkutner (approved):

text/0000-dns-label-process-types.md
```diff
@@ -0,0 +1,147 @@
...
+
+Buildpacks specifying process types that do not comply with the stricter requirements proposed in this RFC will need to change the way process types are defined. Buildpacks defining process types dynamically should implement logic to coerce process types (by substitution, filtering and/or truncation) to meet the specification and avoid build failures.
+
+In order to ease the transition, a new flag may be introduced to pack and/or lifecycle to control enforcement of this requirement. For example, a `--strict-process-types` flag could be introduced that might be set to `warn` or `error`. Calls with `pack build foo --strict-process-types=error ...` would emit error messages and reject builds with non-compliant process types while calls with `pack build foo --strict-process-types=warn ...` would emit warning messages while still allowing the build to complete.
```
jkutner, line 76: we could reuse CNB_DEPRECATION_MODE instead of a new --strict-process-types flag? (RFCs 0049 lines 79–95, and 0110 lines 107–110 already define warn/error/silent
  semantics)

PR #338, opened 2026-06-11
RFC: Rust launcher implementation

hone:
[Rendered](https://github.com/hone/cnb-rfcs/blob/launcher-rs/text/0000-rust-launcher-implementation.md)


text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+
+Looking at the recent security advisories in recent years, the CVEs flagged against the Go standard library and compiler fall into a couple of buckets:
+
+- HTTP/2 & Networking Layer (Largest bucket) - the largest volume of high-severity CVEs are flagged against `net/http`, `net/http/httputil`, and `crypto/tls`.
+- Privilege Escalation & Unix - CVE-2023-29403 exhibited a case where `setuid` / `setgid` did not alter the behavior.
+
+While Rust is not immune to CVEs or security patches, but there are fewer because Rust's standard library is much smaller which decreases the attack surface area. It does not include an HTTP server, TLS stack, or a JSON parser. Instead, this is pushed to the crate dependencies in a Rust stack. In addition, Rust has a minimalist runtime that doesn't include a garbage collector or a runtime scheduler. Rust maps tasks to native OS threads which removes this abstraction between the code and the OS, which removes the abstraction between the code and the OS, trading Go's runtime scheduling for a reduced attack surface area.
```
Malax:
```suggestion
While Rust is not immune to CVEs or security patches, but there are fewer because Rust's standard library is much smaller which decreases the attack surface area. It does not include an HTTP server, TLS stack, or a JSON parser. Instead, this is pushed to the crate dependencies in a Rust stack. In addition, Rust has a minimalist runtime that doesn't include a garbage collector or a runtime scheduler. Rust maps tasks to native OS threads which removes this abstraction between the code and the OS, trading Go's runtime scheduling for a reduced attack surface area.
```

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+
+The biggest core Rust vulnerability in recent years was CVE-2024-24576 where the Rust standard library's `std::process::Command` did not properly escape arguments when invoking batch files (`.bat` and `.cmd`). This allowed attackers to inject arbitrary shell commands since Windows passes arguments as a single raw string.
```
Malax:
```suggestion
The biggest core Rust vulnerability in recent years was [CVE-2024-24576](https://nvd.nist.gov/vuln/detail/cve-2024-24576) where the Rust standard library's `std::process::Command` did not properly escape arguments when invoking batch files (`.bat` and `.cmd`). This allowed attackers to inject arbitrary shell commands since Windows passes arguments as a single raw string.
```

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+As it stands, when CVEs are discovered within the Go standard library (such as networking or TLS-related issues), container scanners flag them against any image containing a compiled Go binary, including the CNB `launcher`. For platform operators, this creates significant operational noise. They are forced to choose between silencing security alerts or constantly updating and rebuilding the `launcher` to keep up with the "treadmill" of Go runtime patches as seen in [RFC#336](https://github.com/buildpacks/rfcs/pull/336).
+
+Looking at the recent security advisories in recent years, the CVEs flagged against the Go standard library and compiler fall into a couple of buckets:
+
+- HTTP/2 & Networking Layer (Largest bucket) - the largest volume of high-severity CVEs are flagged against `net/http`, `net/http/httputil`, and `crypto/tls`.
+- Privilege Escalation & Unix - CVE-2023-29403 exhibited a case where `setuid` / `setgid` did not alter the behavior.
```
Malax:
```suggestion
- Privilege Escalation & Unix - [CVE-2023-29403](https://nvd.nist.gov/vuln/detail/cve-2023-29403) exhibited a case where `setuid` / `setgid` did not alter the behavior.
```

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+1. Process Replacement: To seamlessly replace the launcher process with the user's application, the implementation uses `std::os::unix::process::CommandExt::exec()`. This maps directly to the `execve` syscall, completely clearing the launcher from the process tree and memory, analogous to Go's `syscall.Exec`.
+1. Strict Exit Code Mapping: The implementation uses Rust's `Result` type and a centralized `LaunchError` enum to strictly map internal failures (e.g., parsing errors, environment failures) to the exact integer exit codes mandated by the CNB API (like `PlatformApiIncompatible` or `BuildpackApiIncompatible`).
+
+The `launcher` will live in a separate repo from the current `lifecycle` repo and will live at <https://github.com/heroku/launcher-rs>. This provides a few benefits:
```
Malax: Is `heroku` a typo here? Shouldn't this be in the `buildpacks` org instead? If not, why should it live in the Heroku org?

Malax:
FWIW, this is very exciting to me. :)

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+- opens the door for component maintainers for Rust :pray:
+- versioning around the `launcher` can match the stability of the API surface area which is fairly stable
+
+# Migration
+[migration]: #migration
+
+The `launcher` is a critical runtime level component so it's important we ensure stability and verify parity. In order to achieve this there will be a phased rollout:
+1. Smoke Testing - The Rust implementation will run alongside the go one in the `lifecycle` acceptance tests so any drifts can be caught ahead of time.
+1. Deprecation Period - Once we deem the Rust binary is mature and can be a drop in replacement, we will deprecate the Go binary.
+1. Dual Distribution - We will ship both the Go and Rust binaries within `buildpacksio/lifecycle` images simultaneously for this time period also during the testing period. A ~700 KB is a minimal size bump.
```
jabrown85, line 68: Is there a world in which we bump version numbers and leave the go binary behind with no dual shipping? 

Builder authors and Platforms are in charge of lifecycle version already depending on the mode (trusted/untrusted).

Malax: > leave the go binary behind with no dual shipping

You mean that we never dual ship and that in version X the Rust launcher is _the_ launcher?

If so, I guess the intent here is to be able to test the Rust launcher in production without having to resort to weird custom versions/nightlies for the price of ~700KB for everyone - even if they are not interested in the new launcher.

Personally, dual shipping for at least one version (range) seems like a practical approach to make sure we're ready to make the move to the Rust launcher and have non-CNB maintainers be able to test this in their setup without much hassle compared to using a nightly version or something.

Of course, in either case we need to vet this new launcher thoroughly before, but with the dual shipping we can get some real world testing in earlier.

I think there is a world were we don't double ship. A decision/tradeoff has to be made here.

hone:
From CNB WG:

* Update motivation to include smaller dependency surface area vs. Go. The rest of `lifecycle` is a "dev tool" which pulls in heavy dependencies like `moby`. Splitting this makes the SBOM smaller.
* What if we don't do dual distribution and just ship it as part of a 1.0 style release after we feel confident
* add versioning
* @jabrown85 to add more "outside in" `launcher` acceptance tests to ensure byte for byte compatible output messages/errors.
* Port more comprehensive test suite from <https://github/jabrown85/launcher-rs>.

text/0000-rust-launcher-implementation.md
hone: oops, mistake at 2 in the morning.

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+1. Rust as the Default - Once we feel comfortable, the Rust `launcher` will become the default where the Go implementation can be used as a fallback.
+1. Full Replacement - Once the deprecation window has expired, the Go implementation be replaced by the Rust binary.
+
+## Dual Distribution
+
+During the testing period, users and platforms can opt into the Rust launcher through three distinct entrypoints:
+
+- Runtime Opt-In (App Developers) - The existing Go launcher will act as a shim. If a specific environment variable is present at runtime (e.g., `CNB_EXPERIMENTAL_RUST_LAUNCHER=true`), the Go launcher will immediately delegate execution to the Rust binary. This allows testing on any existing application image with both binaries.
+- CLI Opt-In (Pack Users) - `pack` can introduce an experimental flag. When used, the exported OCI image will have its `ENTRYPOINT` permanently set to the Rust binary, bypassing the Go shim entirely.
```
jjbustamante, line 77: @hone / @jabrown85  / @Malax 

When `pack` is running `pack build` we are suppose to include both binaries in the app image, then on runtime if app developer set `CNB_EXPERIMENTAL_RUST_LAUNCHER=true` will redirect to the other binary, but if `pack build --<some-launcher-rust-flag>` is used then `pack` should include the `rust` launcher in the app image only, is that how this CLI Opt-In would work?

hone: I wonder if it makes sense to just support the build path and not try to support the env var/runtime thing by shipping both launchers. While I think the env var makes it easier for app devs, it adds a lot of complexity that I'm not sure is worth it with the shim.

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+1. Strict Exit Code Mapping: The implementation uses Rust's `Result` type and a centralized `LaunchError` enum to strictly map internal failures (e.g., parsing errors, environment failures) to the exact integer exit codes mandated by the CNB API (like `PlatformApiIncompatible` or `BuildpackApiIncompatible`).
+
+The `launcher` will live in a separate repo from the current `lifecycle` repo and will live at <https://github.com/buildpacks/launcher-rs>. This provides a few benefits:
+
+- keep language specific ecosystem tooling separate
+- opens the door for component maintainers for Rust :pray:
+- versioning around the `launcher` can match the stability of the API surface area which is fairly stable
+
+# Migration
+[migration]: #migration
```
jabrown85, line 63: # Migration Strategy: Cliff Release

We propose a cliff migration. The boundary is the lifecycle major version: `0.XX.X` releases keep shipping the existing Go launcher, and `1.XX.X` (along with every future major version) ships the new Rust launcher. There is no gradual swap or per-feature toggle. The version you pick determines which launcher you get.

## Dual release window

For one quarter, the maintainers will dual release. The `0.XX.X` line and the `1.XX.X` line are published in parallel.

The purpose of this window is to absorb misses. If something in the Go launcher did not make its way into the Rust launcher correctly, the quarter gives us time to find it and fix it forward in the Rust implementation while the Go launcher is still available as a safety net.

## What the window buys consumers

**Builder and platform maintainers** can stay on or revert to a `0.XX.X` release if a critical bug in the Rust launcher affects their users. Reverting is a version pin, not a migration, so recovery is fast.

**Consumers of the Go launcher library code** get the same protection. During the quarter, the maintainers will make sure any launcher changes are also landed in the Go library on the `0.XX.X` releases, so code depending on those packages continues to behave consistently while teams migrate.

## After the quarter

Once the quarter ends, the `0.XX.X` line is no longer maintained in parallel. At that point, previously public methods in the Go launcher library _may_ be removed at the maintainers' discretion.

The intent is to keep as much as makes sense. Platform facing surfaces such as the launcher related files, structs, and validation are expected to stay stable so consumers of those pieces are not forced to change. The part that is expected to go away over time is the Go launcher implementation itself.

text/0000-rust-launcher-implementation.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- New Language - A transition to Rust requires contributors who are comfortable with a new language that we don't use for any other project today. This can be made easier through the use of AI/agentic tools to assist in code management and we may be able to attract some Rust contributors like the maintainers of [libcnb.rs](https://github.com/heroku/libcnb.rs).
+- Double the Maintenance - Until the Go implementation is replaced, material changes will need to be handled across both implementations. The biggest thing in our favor here is that the `launcher` interface is stable, so there aren't a lot of planned changes atm.
+
+# Alternatives
+[alternatives]: #alternatives
+
```
dmikusa, line 88: I'm not saying it's the perfect option, but have we considered VEX files as a solution? If the project shipped a VEX file, probably maintained via CI, that said the CVEs didn't impact lifecycle, users could pick that up and augment their scanner reports. 

I'm not sure if that's sufficient for all use cases, but thought I'd throw it out there.

jabrown85: I have not given that any thought. That's probably worth doing for the project regardless of this RFC. Do you have tools you like to generate VEX files?

hone: @dmikusa yeah, I did think of VEX files, but you probably still need to incorporate it into your flow. For a platform with many users, it means pushing that downstream to users of the users of lifecycle.

I agree with Jesse that I think we should probably do that regardless.

dmikusa: My thought was to use https://www.gofaq.org/en/how-to-use-govulncheck-to-find-known-vulnerabilities/ because it can tell if something actually impacts a go binary and try to convert to VEX. Not sure it outputs directly to VEx.

PR #342, opened 2026-08-20
BuildKit Multi-Architecture Builds

jericop:
Proposes adding BuildKit as an execution backend for pack build to enable multi-architecture container image builds with a single invocation.

[Rendered](https://github.com/jericop/cnb-rfcs/blob/buildkit-mutliarch-build/text/0000-buildkit-multiarch-build.md)

### DIFF
--- a/text/0000-buildkit-multiarch-build.md
+++ b/text/0000-buildkit-multiarch-build.md
@@ -824,11 +824,19 @@ Notes on the boundary:
   drawback). The daemon-only output concepts it does NOT implement (`-daemon`/
   docker-socket access, load-to-local-daemon, the launch cache, and local `--layout`/
   OCI export) are simply not part of this backend rather than rejected flags.
-- **Extensions (`--extension`, Dockerfiles / image extension)** are intentionally OUT
-  OF SCOPE for this change, to limit complexity. This backend executes a fixed
-  analyzer → detector → builder → exporter and does not run the extender/kaniko phase,
-  so order-defined extensions are not applied. Supporting them would be separate future
-  work, not part of this proposal.
+- **Extensions (`--extension`, image extension) ARE supported.** When the resolved
+  order includes extensions, the backend runs the generator (the `generate` phase)
+  and applies the emitted build- and run-image Dockerfiles by translating each
+  Dockerfile instruction directly into LLB, so the extend happens **inside the build
+  engine with no separate extender/kaniko phase** (kaniko-free). The build-image
+  Dockerfile extends the environment the buildpacks then detect/build against; the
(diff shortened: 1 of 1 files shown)


text/0000-buildkit-multiarch-build.md
```diff
@@ -0,0 +1,989 @@
...
+3. **Assemble `FROM run-image` natively.** Pack's BuildFunc assembles the image as
+   `FROM <run-image>` + `llb.Copy` of each emitted layer source (buildpack layers
+   from `/layers/<bp>/<layer>`, app from `/workspace`, the launcher file; app slices
+   honored via `IncludePatterns`; process-types via a tiny synthesized tree). The
+   run image is read digest-pinned from the analyzer's `/layers/analyzed.toml` and is
+   never modified.
+4. **BuildKit builds + pushes.** BuildKit snapshots the layer filesystems, computes
+   the layer diffIDs, and pushes one image per platform assembled into a single OCI
+   index — no intermediate tags, and **no layer data ever egresses to the host**.
+5. **Finalize authors the CNB metadata.** Pack calls the lifecycle `phase/finalize`
```
jabrown85: Can we have BuildKit push the produced layers + config as content-addressed blobs, by digest, with no tag and pass that off to lifecycle to do the assembly in finalize? That way we only have one registry operation and lifecycle does it.

text/0000-buildkit-multiarch-build.md
```diff
@@ -0,0 +1,989 @@
...
+# Enable experimental features
+pack config experimental true
+
+# Multi-architecture build with the buildkit backend
+pack build registry.example.com/myapp:latest \
+  --path ./app \
+  --builder jericop/ubuntu-noble-builder:buildkit-native-export \
+  --run-image paketobuildpacks/ubuntu-noble-run:latest \
+  --platforms linux/amd64,linux/arm64 \
+  --build-backend buildkit \
```
jabrown85: I'd like to explore making `lifecycle` be the orchestrator instead of this being in `pack` - let `pack` feed the buildkit info into lifecycle and do this maybe?

```

BUILDKIT_HOST=unix:///run/buildkit/buildkitd.sock
lifecycle creator \
  -build-backend buildkit \        # default "" = classic in-process path (today's behavior)
  -run-image <ref> \               # digest-pinned; read from analyzed.toml if omitted
  -platform linux/amd64,linux/arm64 \   # matrix; default = host arch only
  ...
  myorg/myapp:latest
```

The `emit` export and `finalize` still get built like you have here but instead of `pack` having to call finalize - `lifecycle` calls it in-process. For platforms that aren't `pack` and want to accomplish similar things the primitives are there so they could fan out and fan back in to do the `finalize`

wdyt?

